PR #17, opened 2021-10-07
Added interface and function abi

QuinnLee:
This is the first pass on creating a `Contract`/ `Interface` and support an ABI input.  Right now, the `functions` are supported and generated methods on the `Contract` class return the encoded abi.  In the next interaction, these calls should interact with a contract.

### DIFF
--- a/packages/abi-coder/interface.test.ts
+++ b/packages/abi-coder/interface.test.ts
@@ -44,11 +44,6 @@ describe('Interface', () => {
     expect(
       functionInterface.decodeFunctionData('entry_one', '0x000000000c36cb9c000000000000002a')
     ).to.eql([BigNumber.from(42)]);
-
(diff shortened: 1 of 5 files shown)


packages/abi-coder/interface.test.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  it('raises an error if the arguments do not match the function input types', () => {
+    expect(() => functionInterface.encodeFunctionData('entry_one', [true])).to.throw('Invalid u64');
+    expect(() => functionInterface.encodeFunctionData('entry_one', [11, 11])).to.throw(
+      'Types/values length mismatch'
+    );
+    expect(() =>
+      functionInterface.decodeFunctionData('entry_one', '0x1111111111111111000000000000002a')
+    ).to.throw('data signature does not match function entry_one');
+  });
```
AlicanC, line 99: @QuinnLee, unlike on the others we don't set `functionInterface` to anything here. Is this intended?

PR #80, opened 2021-12-14
Fix calling contracts with structs

AlicanC:


QuinnLee (approved):
lgtm

packages/providers/src/provider.ts
```diff
@@ -258,19 +310,31 @@ export default class Provider {
     request: TransactionRequest;
     wait: () => Promise<TransactionResult & { data: Uint8Array }>;
   }> {
+    const dataArray = arrayify(data);
+    const functionSelector = dataArray.slice(0, 8);
+    const isStructArg = dataArray.slice(8, 16).some((b) => b === 0x01);
```
digorithm, line 315: I'm just curious and a bit confused: how come we're confirming it's a struct arg by checking `b === 0x01`?

digorithm (approved):
Great stuff! 🔥 

Just one clarification comment, but feel free to merge whenever.

packages/providers/src/provider.ts
AlicanC: It's an ABI-encoded boolean and we are checking if it has any bytes matching `0x01` in it to see if we should interpret it as true or false.

I was going to make these lines prettier, but I decided to leave a comment instead (and forgot the comment).

I have identified multiple ways to implement this and each one was uglier than the other. Then I realized we only support exactly one argument and when we add support for 0 and 1+ I will have to go over this again. So I just left it like this and will get back to it when it's clear how non-1 args are going to work.

Issue #39, opened 2021-10-25
Add `signTransaction`


QuinnLee:
Currently it is hard to test/dev against a provider/contract.  We should mock it out

QuinnLee:
current specs are https://github.com/FuelLabs/sway-lib-std/issues/17

PR #96, opened 2022-01-17
feat: add signed transaction (#39)

luizstacio:
Add signed transaction.


adlerjohn:
Why was this closed?

luizstacio:
> Why was this closed?

The signature was using the ETH signing method, I'm working on this with Fuel signing specs.

PR #103, opened 2022-01-21
Add `signMessage`

luizstacio:
## Add `signMessage`;

- Create a JSON file to test wallet signing and other methods.
- New classes under the wallet
  - Signer: Responsible for holder keys and providing signature on data;
  - Hasher: Wrap sha256; this is the first step to a `proposal` to remove the use of `@ethersproject/sha2`;
  - Wallet: Wrap, Signer, and Provider to better UX

## Configurations

- Disable `"import/prefer-default-export"` proposal to avoid typos on imports.
- Add support to JSON import on `packages/wallet`.


packages/wallet/src/wallet.test.json
```diff
@@ -0,0 +1,8 @@
+{
```
QuinnLee, line 1: is this the same fixtures as `fuels-rs` is using too? might be good to sync them

luizstacio: Yes, I got it from `fuels-rs`. I was thinking about creating some way to share via module. Maybe create a repo that is a module for `rust` and `typescript`.

QuinnLee (approved):
approved! but can you let @digorithm take a look before merging?

packages/wallet/src/signer.ts
```diff
@@ -0,0 +1,76 @@
...
+    const keyPair = getCurve().keyFromPrivate(privateKeyBytes, 'hex');
+
+    // @TODO: defineReadOnly these properties
+    // Slice(1) removes the encoding scheme from the public key
+    this.publicKey = hexlify(keyPair.getPublic(false, 'array').slice(1));
+    this.privateKey = hexlify(privateKeyBytes);
+    this.address = hash(this.publicKey);
+  }
+
+  sign(msg: BytesLike) {
```
digorithm, line 42: What do you think about renaming `sign` to `signMessage`? Message signing and transaction signing _might_ be slightly different, requiring two different implementations. At the very least, when signing an actual transaction this function will be taking a whole `Transaction` object, right? So they'll likely differ.

digorithm (changes requested):
It's looking great, awesome job! Just left a few nits and comments.

packages/wallet/src/signer.ts
luizstacio: I was thinking about encoding the transaction on the wallet level and sign using this method. In this way, the Signer will work for both. Maybe I could be missing some points but I can change later on the PR for the `wallet.signTransaction`.

Maybe changing the name of the param would be a good idea.

digorithm: That makes sense!

digorithm (approved):
Looking great! Ship it! :shipit:

PR #107, opened 2022-01-28
Add signTransaction

luizstacio:
# Summary
- Create a new package `@fuels-ts/testcases` with mock transactions data. Also, move the currently used ones to these pack.
- Update to use `canonical signature` on `ellipict.sign`.
- Create `Hasher.hashTransaction` it create an `txid` as specified on [here](FuelLabs/fuel-specs/blob/master/specs/protocol/tx_validity.md#valid-signatures).
- Create `Wallet.signTransaction`.

closes: #39


QuinnLee (approved):

packages/wallet/src/hasher.ts
```diff
@@ -11,6 +16,51 @@ export function hashMessage(msg: string) {
...
+  }
+
+  // Zero out input fields
+  transaction.inputs = transaction.inputs.map((input) => {
+    const inputClone = { ...input };
+
+    switch (inputClone.type) {
+      // Zero out on signing: txoPointer
+      case InputType.Coin: {
+        // inputClone.txoPointer = 0;
```
digorithm, line 40: If I understand correctly if the transaction's input is a coin, shouldn't we set the `txoPointer` to `0` when signing? Why is this commented out?

packages/wallet/src/hasher.ts
```diff
@@ -11,6 +16,51 @@ export function hashMessage(msg: string) {
...
+        return inputClone;
+      }
+      // Zero out on signing: txID, outputIndex, balanceRoot, stateRoot, and txoPointer
+      case InputType.Contract: {
+        // inputClone.txoPointer;
```
digorithm, line 45: Same here, shouldn't we set `txoPointer` to `0`?

packages/wallet/src/hasher.ts
luizstacio: We didn't have the field `txoPointer` yet on the transaction.

digorithm: Oh I see, makes sense then.

PR #110, opened 2022-02-02
feat: zero out output fields on hashTransaction

luizstacio:
# Summary

- Set to zero on `hashTransaction` the defined `outputs`.
- Add `lodash.deepClone` to remove reference from `transactionRequest` param.


AlicanC (approved):
LGTM.

Issue #120, opened 2022-02-09
Wallet Generate


luizstacio:
Add `Wallet.generate()` to create a random `privateKey`

PR #121, opened 2022-02-11
Generate random Wallet

luizstacio:
## Summary

- Adds, Generate random wallet;
- Adds the possibility to pass an extra entropy;
- `Wallet.generate(options)` open for optional configuration like `generate from mnemonic`;
- Lock `@ethersproject/random@5.5.1` to avoid non-intentional updates.

## Refs:
https://github.com/ethereumjs/ethereumjs-wallet/blob/master/src/index.ts
https://github.com/crypto-browserify/randombytes
https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
https://github.com/ethers-io/ethers.js/blob/master/packages/random
https://nodejs.org/api/crypto.html#cryptorandombytessize-callback

## Discussion

The `randomBytes` is critical to have a secure source of entropy. Because of this, I have chosen the package `@ethersproject/random`. It has a similar implementation from other modules/wallets. Where the idea is to; Use [crypto.randomBytes() on `Node.js`](https://nodejs.org/api/crypto.html#cryptorandombytessize-callback) and [crypto.getRandomValues() on the `browser`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues).

closes: #120

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,7 @@ Thumbs.db
 ## Jekyll & gh-pages
 .jekyll-cache
 _site/
+
+
+## Packages lock file
+packages/**/*-lock.json
--- a/package-lock.json
+++ b/package-lock.json
@@ -15077,7 +15077,8 @@
     "node_modules/safe-buffer": {
       "version": "5.1.2",
       "resolved": "https://registry.npmjs.org/safe-buffer/-/safe-buffer-5.1.2.tgz",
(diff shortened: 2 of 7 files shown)


QuinnLee (approved):
lgmt, there are some linting warnings about docs, but no blockers.

luizstacio:
> lgmt, there are some linting warnings about docs, but no blockers.

Thanks for the warning will fix it with on next doc generation.

digorithm (approved):
LGTM!

PR #124, opened 2022-02-14
feat!: Update interface and fragment to ignore gas, amount and color in

QuinnLee:
abi format

### DIFF
--- a/packages/abi-coder/src/fragments/function-fragment.ts
+++ b/packages/abi-coder/src/fragments/function-fragment.ts
@@ -1,3 +1,4 @@
+import type { JsonFragmentType } from '@ethersproject/abi';
 import { FormatTypes, ParamType } from '@ethersproject/abi';
 
 import type { JsonFragment } from './fragment';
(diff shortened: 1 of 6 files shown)


AlicanC (approved):

packages/abi-coder/src/fragments/function-fragment.ts
```diff
@@ -17,16 +18,31 @@ function formatOverride(this: ParamType, format?: string): string {
...
+      strictInputs: FunctionFragment.strictArguments(inputs),
       outputs: outputs.map(ParamType.fromObject),
     };
 
     return new FunctionFragment(params);
   }
 
+  static strictArguments(fragment: readonly JsonFragmentType[]): boolean {
+    return !(
+      fragment.length >= 3 &&
```
adlerjohn, line 35: Note that currently the compiler must emit _exactly_ 4 parameters. It will still work with your code, but if you want to be very specific then this check should be `== 4`. Not a big deal since the compiler will be fixed soon-ish anyways to no longer have this requirement.

digorithm (approved):
:shipit:

PR #142, opened 2022-03-02
chore: Update releaserc and allow dry-runs on ci

QuinnLee:
realized this is easier to `--dry-runs`  and semantic-release + mono repos are bad for my OCD


luizstacio (approved):
This is really good to avoid issues on release. Nice!

AlicanC (approved):

PR #156, opened 2022-03-05
Fix encoding of B256s with leading zero bytes

AlicanC:
Turns out the previous "B256 with all zero bytes" issue we fixed was actually a "B256 with leading zero bytes" issue.

With this PR, the bug is fixed and encoder only accepts exactly 32 bytes. So `0x0`, `0x00` and such are invalid now which should be safer.


digorithm (approved):
Uh, nice catch!

QuinnLee (approved):

PR #168, opened 2022-03-09
Upgrade to Forc v0.6.0

AlicanC:

PR #169, opened 2022-03-09
feat: update abi code to support new syntax

luizstacio:
### Summary

- Add support to new syntax, also `forc v6`
- Add support to no params func `method()`
- Rebuild example contracts binary

closes: #154 #167


services/fuel-core/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/fuellabs/fuel-core:v0.4.1
+FROM ghcr.io/fuellabs/fuel-core
```
adlerjohn, line 1: Shouldn't this be kept pinned? Otherwise a new breaking release of `fuel-core` will break all CI flows here.

adlerjohn (changes requested):

.gitignore
```diff
@@ -134,3 +134,6 @@ packages/**/*-lock.json
 
 # Generated files
 **/__generated__/**
+
+# Forc output folder
+out
```
QuinnLee, line 139: add new line!

packages/providers/src/test-contract/src/main.sw
```diff
@@ -8,29 +8,29 @@ struct TestStruct {
...
       a: !value.a,
       b: value.b + 1,
     }
   }
-  fn barfoo(gas_: u64, amount_: u64, asset_id_: b256, value: u64) -> u64 {
+  fn barfoo(value: u64) -> u64 {
     63
   }
-  fn foobar(gas_: u64, amount_: u64, asset_id_: b256, value: ()) -> u64 {
+  fn foobar(value: ()) -> u64 {
```
mohammadfawaz, line 33: There is no reason to keep the unit type `value: ()` anymore. It can be dropped.

packages/contract/src/test-contract/src/main.sw
```diff
@@ -4,25 +4,25 @@ use std::storage::store;
...
   }
-  fn increment_counter(gas_: u64, amount_: u64, asset_id_: b256, amount: u64) -> u64 {
+  fn increment_counter(amount: u64) -> u64 {
     let value: u64 = get(COUNTER_KEY);
     let value = value + amount;
     store(COUNTER_KEY, value);
     value
   }
-  fn counter(gas_: u64, amount_: u64, asset_id_: b256, value: ()) -> u64 {
+  fn counter(value: ()) -> u64 {
```
mohammadfawaz, line 25: `value: ()` can be dropped.

packages/providers/src/test-contract/src/main.sw
```diff
@@ -8,29 +8,29 @@ struct TestStruct {
...
 }
 
 impl TestContract for Contract {
-  fn foo(gas_: u64, amount_: u64, asset_id_: b256, value: u64) -> u64 {
+  fn foo(value: u64) -> u64 {
     log_u64(3735928559);
     value
   }
-  fn boo(gas_: u64, amount_: u64, asset_id_: b256, value: TestStruct) -> TestStruct {
+  fn boo(value: TestStruct) -> TestStruct {
```
mohammadfawaz, line 22: Technically, we can break this struct into multiple arguments now but I suppose it's good to keep a struct test. That being said, maybe it's a good idea to have another method here that requires more than 1 argument?

packages/providers/src/test-contract/src/main.sw
adlerjohn: We should still have a test for that though since it _is_ a valid parameter type.

luizstacio: It's valid to keep this, for validate empty params?

mohammadfawaz: Yes that makes sense. Maybe we should keep one version with no parameters and one with an empty parameter.

QuinnLee (approved):

luizstacio:
@AlicanC just add some more changes from #168

AlicanC (approved):

packages/contract/src/test-contract/src/main.sw
luizstacio: I kept this one and also add a new method `counternoparams` as this is the contract package. On the `providers` I have removed the empty param.

packages/providers/src/test-contract/src/main.sw
luizstacio: I will not address this in this PR.

AlicanC (approved):

adlerjohn (approved):

PR #170, opened 2022-03-10
feat: call contract method with mutiple params

luizstacio:
### Summary

- Fix encode for call a contract with multiple params;
- Add complementary tests on package `provider`
- Remove `isStruct` from the encoded interface

Contract example;
```
abi TestContract {
  fn foo(value: u64) -> u64;
  fn boo(value: TestStruct) -> TestStruct;
  fn barfoo(value: u64) -> u64;
  fn foobar(value: ()) -> u64;
  fn foobar_no_params() -> u64;
  fn sum(a: u64, b: u64) -> u64;
  fn sum_test(a: u64, test: SumStruct) -> u64;
  fn sum_single(test: SumStruct) -> u64;
  fn sum_multparams(a: u64, b: u64, c: u64, d: u64, e: u64) -> u64;
  fn echo_b256(a: b256) -> b256;
  fn add_ten(param: SingleParamStruct) -> u64;
}
```

PR #173, opened 2022-03-10
chore: organize scripts

AlicanC:
This PR adds a file called `scripts.ts` that contains the two scripts we use: `contractCallScript` and `returnZeroScript`.

I know this conflicts with #170 but it should be a an easy one. If not we can merge #170 and I can rebase this.

### DIFF
--- a/packages/providers/src/index.ts
+++ b/packages/providers/src/index.ts
@@ -1,4 +1,5 @@
 export * from './provider';
 export { default as Provider } from './provider';
 export * from './transaction-request';
+export * from './scripts';
 export * from './util';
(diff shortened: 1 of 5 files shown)


digorithm:
This is great! Should we let https://github.com/FuelLabs/fuels-ts/pull/170 go in first because it came first and solves a blocker @luizstacio was having?

luizstacio:
I also have some ongoing changes on this file to support function without return.

luizstacio:
@AlicanC It's for you to merge and squash, I will rebase my ongoing changes with this PR, after merge.

AlicanC:
@luizstacio rebased!

AlicanC:
This was included in #178

PR #187, opened 2022-03-13
chore: move test to contract folder

luizstacio:
### Summary

- Move contract test to inside contract folder.


AlicanC (approved):
Sure, but why aren't we also moving `call-test-contract.test.ts`?

QuinnLee (dismissed):

luizstacio:
> Sure, but why aren't we also moving `call-test-contract.test.ts`?

I didn't notice that! Thanks for pointing, I'm adding this also

QuinnLee (approved):

Issue #175, opened 2022-03-10
Pass additional `inputs` and `outputs` when executing a contract `method`


luizstacio:
### Use case

contract
```
fn deposit() {
    let asset_id = (msg_asset_id()).into();
    assert(asset_id == ETH_ID || asset_id == TOKEN_ID);

    let sender = msg_sender();
    let total_amount = get::<u64>(key_deposits(sender, asset_id)) + msg_amount();

    store(key, total_amount);
}
```

typescript
```ts
const result = await contract.functions.deposit({
  inputs: [
     {
          type: InputType.Coin,
          assetId: '0x0000000000000000000000000000000000000000000000000000000000000000',
          amount: BigNumber.from(500),
          owner: '0x0000000000000000000000000000000000000000000000000000000000000000',
          witnessIndex: 0,
          maturity: 0,
          predicate: '0x',
          predicateData: '0x',
     },
  ],
  outputs: [
     {
          type: OutputType.Coin,
          amount: BigNumber.from(500),
          assetId: `0x0000000000000000000000000000000000000000000000000000000000000000`,
          to: contractId,
     },
  ]
});
```

luizstacio:
https://github.com/FuelLabs/fuels-rs/issues/159

luizstacio:
Closing this issue in favor of: https://github.com/FuelLabs/fuels-ts/issues/198

PR #188, opened 2022-03-13
feat: add extra transaction request overrides

luizstacio:
### Summary

When calling a method from the contract, sometimes we need to send coins and maybe add some other things like `gasPrice`, `bytePrice`, etc. 

Because of this, I think we shouldn't limit what can be overridden but instead enable to override the `transactionRequest` payload completely.

The suggested implementation will merge the additional `transactionRequest` into the default, overriding the default configs with the new ones, **except for `inputs` and `outputs`** which will be concatenated with the default `inputs and outputs`

Notes: Another change suggest is to change the `Overrides` param to `CallOptions{ transaction }` this would allow adding more options in the future without breaking the Interface if we enable to pass the entire `TranasctionRequest`.

### Discussion;

- Should we allow the entire transaction override?
- Should we concatenate the inputs and outputs?
   - If we do so, should we have another config to override instead of concatenating? like a `transactionRequestOverride` or something like a `transactionRequest` hook from [axios](https://axios-http.com/docs/req_config)? Enabling to override something right before submitting the call.

closes: #175

### DIFF
--- a/packages/contract/src/contract.ts
+++ b/packages/contract/src/contract.ts
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import type { BigNumberish } from '@ethersproject/bignumber';
+import type { BytesLike } from '@ethersproject/bytes';
 import { Logger } from '@ethersproject/logger';
 import type { JsonFragment, FunctionFragment } from '@fuel-ts/abi-coder';
 import { Interface } from '@fuel-ts/abi-coder';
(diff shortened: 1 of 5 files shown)


AlicanC:
I can't comment on these discussion points since I feel like these are uncharted territories and we should first complete some apps like SwaySwap before we can decide.

I say we should just add whatever's necessary for SwaySwap and move on. (Just adding `inputs` and `outputs` to `Overrides`?)

Note: It looks like @digorithm is working on something similar on the Rust side: https://github.com/FuelLabs/fuels-rs/issues/138

packages/contract/src/contract.ts
```diff
@@ -21,6 +26,55 @@ export type Overrides = {
...
+      },
+    ],
+  };
+
+  // If no extra payload is passed return default payload
+  if (!overrides) return defaultPayload;
+
+  const overrideList = Array.isArray(overrides) ? overrides : [overrides];
+
+  return overrideList.reduce<TransactionRequest>(
```
QuinnLee, line 64: 👍

QuinnLee (approved):

luizstacio:
> I can't comment on these discussion points since I feel like these are uncharted territories and we should first complete some apps like SwaySwap before we can decide.
> 
> I say we should just add whatever's necessary for SwaySwap and move on. (Just adding `inputs` and `outputs` to `Overrides`?)
> 
> Note: It looks like @digorithm is working on something similar on the Rust side: [FuelLabs/fuels-rs#138](https://github.com/FuelLabs/fuels-rs/issues/138)

Understood. Yeah, I'm needing this  I think in this case, we should think ahead and improve with an API that better supports future changes without breaking things. This is something that I feel will be needed as I'm doing the swayswap development.  And with my previous exp with libs like `Axios`.

Do you think this makes sense @AlicanC?

AlicanC:
Sure!

PR #189, opened 2022-03-13
package.json and package-lock.json version updated to latest

Nightfurex:


luizstacio (approved):

AlicanC (changes requested):
What's the reasoning behind this change? If there is a reason to bump this version then our release flow should be modified to do this automatically.

adlerjohn:
@AlicanC the _root_ package.json and package-lock.json are not modified by the release flow, hence why it's still at 0.0.1 instead of 0.3.0.

AlicanC:
To my knowledge, the version of the root package does not matter at all. It wouldn't even make sense if we move to versioning packages independently. (#191)

Maybe we should just remove the field? We don't have a version on root `Cargo.toml`s: https://github.com/FuelLabs/fuels-rs/blob/master/Cargo.toml

adlerjohn:
If the root package isn't doing anything, then yeah that version should just be removed or set of 0.0.0 if removing is not possible.

pedronauck:
I modified the CI workflow to always change the root package version to have our current version:

https://github.com/FuelLabs/fuels-ts/blob/master/package.json#L3
https://github.com/FuelLabs/fuels-ts/blob/master/scripts/utils/changePackages.ts#L9-L16

I think that by having the main package with version updated is good since we're using fixed versions and people that don't know about our repo, don't get mess about which version we are.

PR #192, opened 2022-03-19
Convenient Coin APIs

AlicanC:
[feat!: add TransactionRequest classes](https://github.com/FuelLabs/fuels-ts/pull/192/commits/f1eef6625eaa431b4d49ccab3e8cdb234f630b6e)

- Replaces `ScriptTransactionRequest`/`CreateTransactionRequest` types with classes that have methods like `addCoins`, `addContract`, etc. to make things easier.

[feat: add wallet coin APIs](https://github.com/FuelLabs/fuels-ts/pull/192/commits/36efdd07d7368840279c8b4489764d8b0ef338d0)

- Adds methods: `getCoinsToSpend`, `getCoins`, `getBalances` and `transfer`
- Adds `fund(tx)` method which calculates the cost of the tx and adds coins from the wallet to the tx to cover it.

[feat: add CoinQuantity](https://github.com/FuelLabs/fuels-ts/pull/192/commits/c15343bc7ee3d1b3cf2e19778dbe091f88e207dc)

- Adds `CoinQuantity` which is a combination of an amount and an AssetId. This replaces `SpendQueryElement`, and instead of just `{ amount, assetId }` it also accepts `[amount, assetId]`, allowing simpler APIs.

[feat: add interfaces package](https://github.com/FuelLabs/fuels-ts/pull/192/commits/7ca775e60cbb23137f7cd74cfb6f8a9ebbbd9614)

- Adds `@fuel-ts/interfaces`, which is similar to fuel-core-interfaces. Currently contains `AbstractWallet` and `AbstractContract`. This allows working with Wallets and Contracts without importing their `@fuel-ts/xyz` packages and prevents circular deps.

---

A transfer (single-source/multi-destination/multi-asset) after these changes:
```ts
const coins = await sender.getCoinsToSpend([
  [amount.mul(2), assetIdA],
  [amount.mul(2), assetIdB],
]);

const request = new ScriptTransactionRequest({ gasLimit: 1000000 });
request.addCoins(coins);
request.addCoinOutputs(receiverA, [
  [amount, assetIdA],
  [amount, assetIdB],
]);
request.addCoinOutputs(receiverB, [
  [amount, assetIdA],
  [amount, assetIdB],
]);

const response = await provider.sendTransaction(request);
```

Before:
```ts
const coins = await provider.getCoinsToSpend(sender, [
      { assetId: assetIdA, amount: amount.mul(2) },
      { assetId: assetIdB, amount: amount.mul(2) },
    ]);

await provider.sendTransaction({
  type: TransactionType.Script,
  gasPrice: BigNumber.from(0),
  gasLimit: BigNumber.from(1000000),
  bytePrice: BigNumber.from(0),
  script: returnZeroScript.bytes,
  inputs: coins.map((coin) => ({
    type: InputType.Coin,
    ...coin,
    witnessIndex: 0,
  })),
  outputs: [
    {
      type: OutputType.Coin,
      to: receiverA.address,
      assetId: assetIdA,
      amount,
    },
    {
      type: OutputType.Coin,
      to: receiverA.address,
      assetId: assetIdB,
      amount,
    },
    {
      type: OutputType.Coin,
      to: receiverB.address,
      assetId: assetIdA,
      amount,
    },
    {
      type: OutputType.Coin,
      to: receiverB.address,
      assetId: assetIdB,
      amount,
    },
    { type: OutputType.Change, assetId: assetIdA, to: sender },
    { type: OutputType.Change, assetId: assetIdB, to: sender },
  ],
  witnesses: ['0x'],
});
```

### DIFF
--- a/packages/contract/src/call-test-contract/call-test-contract.test.ts
+++ b/packages/contract/src/call-test-contract/call-test-contract.test.ts
@@ -16,7 +16,7 @@ const setup = async (abi: ReadonlyArray<JsonFragment> | Interface = abiJSON) => 
 
   // Create wallet
   const wallet = Wallet.generate({ provider });
-  await seedWallet(wallet, [{ assetId: NativeAssetId, amount: 1 }]);
+  await seedWallet(wallet, [[1, NativeAssetId]]);
 
   // Deploy contract
   const bytecode = readFileSync(join(__dirname, './out/debug/call-test.bin'));
--- a/package-lock.json
+++ b/package-lock.json
@@ -1982,6 +1982,10 @@
       "resolved": "packages/hdwallet",
       "link": true
     },
+    "node_modules/@fuel-ts/interfaces": {
+      "resolved": "packages/interfaces",
+      "link": true
+    },
     "node_modules/@fuel-ts/merkle": {
       "resolved": "packages/merkle",
       "link": true
@@ -25658,7 +25662,7 @@
     },
     "packages/abi-coder": {
       "name": "@fuel-ts/abi-coder",
-      "version": "0.3.0",
+      "version": "0.4.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@ethersproject/abi": "^5.4.1",
@@ -25671,12 +25675,12 @@
     },
     "packages/constants": {
       "name": "@fuel-ts/constants",
-      "version": "0.3.0",
+      "version": "0.4.0",
       "license": "Apache-2.0"
     },
     "packages/contract": {
       "name": "@fuel-ts/contract",
-      "version": "0.3.0",
+      "version": "0.4.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@ethersproject/bignumber": "^5.5.0",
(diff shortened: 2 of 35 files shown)


AlicanC:
[Support submitting multiple transactions at once #194](https://github.com/FuelLabs/fuels-ts/issues/194) was going to be resolved in this PR, but I've found a better way to do it which is blocked by a fuel-core issue so I left it to not block this.

luizstacio:
This type of interface looks practical but same time not clear... I would suggest keeping an object even if you need to repeat the object keys. Or if we keep it this way maybe would be clear to be `[coinId, amount]`.
```
const coins = await sender.getCoinsToSpend([
  [amount.mul(2), assetIdA],
  [amount.mul(2), assetIdB],
]);
```

packages/contract/src/call-test-contract/call-test-contract.test.ts
```diff
@@ -16,7 +16,7 @@ const setup = async (abi: ReadonlyArray<JsonFragment> | Interface = abiJSON) =>
 
   // Create wallet
   const wallet = Wallet.generate({ provider });
-  await seedWallet(wallet, [{ assetId: NativeAssetId, amount: 1 }]);
+  await seedWallet(wallet, [[1, NativeAssetId]]);
```
luizstacio, line 19: This type of interface looks practical but same time not clear... I would suggest keeping an object even if you need to repeat the object keys. Or if we keep it this way maybe would be clear to be `[coinId, amount]`.
```
const coins = await sender.getCoinsToSpend([
  [amount.mul(2), assetIdA],
  [amount.mul(2), assetIdB],
]);
```

luizstacio (changes requested):

packages/contract/src/call-test-contract/call-test-contract.test.ts
AlicanC: I chose that order because:
- We already had `transfer` implemented in fuels-rs and it was AssetId-last. (`dest,amount,asset_id` and not `dest,asset_id,amount`)
- "Amount then unit"  feels natural:
  `2 stacks of Saronite Ore` -> `[2 * STACK_SIZE, SARONITE_ORE]`
  `100 dollars and 52 euros` -> `[[100, USDC], [52, EURS]]`

luizstacio: I think makes sense to keep the consistency between the `rs and ts`.  Even if I think the array API is a bit strange... @AlicanC thanks

luizstacio (approved):

Issue #146, opened 2022-03-03
Add Support to mnemonic and HDWallets on the Wallet package


luizstacio:
- [x] `Wallet.fromSeed`
- [x] `Wallet.fromExtendedKey`
- [x] `Wallet.fromMnemonic(mnemonic[,path="m/44’/60’/0’/0/0" [,wordlist]]) -> Wallet`
- [ ] `Wallet.derivePath`
- [ ] `Wallet.deriveIndex`

PR #196, opened 2022-03-24
feat: add hdwallet and mnemonic features to wallet

luizstacio:
### Summary

- Add `fromSeed, fromMnemonic, fromExtendedKey` methods to Wallet. The designed solution didn't keep seed, mnemonic phrase, or other related info on the wallet instance, removing possible leaking of it on the wallet level. Because of that `derivePath and deriveIndex` methods can't be implemented on the wallet level. At this time the idea is to bring a new High Order Component something like `Wallet Manager`.
- Add `extendPublicKey` on `@fuels-ts/signer`, converting a compact publicKey to an extended publicKey.
- Add the `connect` method to `@fuels-ts/wallet`, connect to a different provider without the need to recreate the wallet.

closes: #146

### DIFF
--- a/packages/wallet/src/wallet.ts
+++ b/packages/wallet/src/wallet.ts
@@ -2,8 +2,10 @@ import type { BigNumberish } from '@ethersproject/bignumber';
 import type { BytesLike } from '@ethersproject/bytes';
 import { NativeAssetId } from '@fuel-ts/constants';
 import { hashMessage, hashTransaction } from '@fuel-ts/hasher';
+import { HDWallet } from '@fuel-ts/hdwallet';
 import { AbstractWallet } from '@fuel-ts/interfaces';
-import { Provider, ScriptTransactionRequest, transactionRequestify } from '@fuel-ts/providers';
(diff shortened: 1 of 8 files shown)


QuinnLee (approved):

packages/wallet/src/wallet.ts
```diff
@@ -196,4 +210,37 @@ export default class Wallet extends AbstractWallet {
...
     return new Wallet(privateKey, generateOptions?.provider);
   }
+
+  /**
+   * Create wallet from a seed
+   */
+  static fromSeed(seed: string, path?: string): Wallet {
+    const hdWallet = HDWallet.fromSeed(seed);
+    const childWallet = hdWallet.derivePath(path || Wallet.defaultPath);
+    const wallet = new Wallet(<string>childWallet.privateKey);
```
QuinnLee, line 220: stylistically i think we can just return here

luizstacio: I made the change makes sense

QuinnLee (approved):

PR #197, opened 2022-03-25
fix!: Update to use new ABI format and only allow json framgents

QuinnLee:
Migrate to use the new array format.

One thing I noticed is that the ABI coder allowed strings and all the tests were testing this.  Currently, we only allow JSON ABI formats.  I update all the tests and refactored the code to do this.

### DIFF
--- a/packages/abi-coder/src/interface.test.ts
+++ b/packages/abi-coder/src/interface.test.ts
@@ -131,20 +131,20 @@ describe('Interface', () => {
         inputs: [
           {
             name: 'arg',
-            type: 'u16[3]',
+            type: '[u16; 3]',
           },
         ],
         name: 'takes_array',
         outputs: [
           {
             name: '',
-            type: 'u16[2]',
+            type: '[u16; 2]',
           },
         ],
       },
     ]);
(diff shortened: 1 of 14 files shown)


QuinnLee:
I wasnt sure the best way to do the tests.  With the json ABI format, we return an array w/ keys lookup.
e.g
```
['foo', bar: 'foo']
```

packages/abi-coder/src/interface.test.ts
```diff
@@ -131,20 +131,20 @@ describe('Interface', () => {
...
             name: '',
-            type: 'u16[2]',
+            type: '[u16; 2]',
           },
         ],
       },
     ]);
     expect(functionInterface.encodeFunctionData('takes_array', [[1, 2, 3]])).toEqual(
-      '0x00000000f0b878640000000000000001000000000000000100000000000000020000000000000003'
+      '0x00000000058734b90000000000000001000000000000000100000000000000020000000000000003'
```
luizstacio, line 147: We also have tests to call the function with this type of input and output?

luizstacio (approved):

packages/abi-coder/src/interface.test.ts
QuinnLee: will create this as a follow up

Issue #198, opened 2022-03-26
Enable forward custom amount and asset on contract calls


luizstacio:
On Sway is possible to access amount and asset using `msg_amount()`, `msg_asset_id()`.

Contract;
```
fn deposit() {
    let asset_id = (msg_asset_id()).into();
    assert(asset_id == ETH_ID || asset_id == TOKEN_ID);

    let sender = msg_sender();
    let total_amount = get::<u64>(key_deposits(sender, asset_id)) + msg_amount();

    store(key, total_amount);
}
```

Use case;
```
const result = await contract.functions.deposit({
  amount: 1_000_000,
  assetId: '0x0000000000000000000000000000000000000000000000000000000000000000',
});
```

PR #199, opened 2022-03-26
feat: forward amount and assetId on contract call

luizstacio:
[chore: update lib-core and lib-std](https://github.com/FuelLabs/fuels-ts/commit/a6dfafe450af7dc16eb33e1259f6a46e4c6ad28a)

 - Update std libs with latest changes on `msg_asset_id()` and `msg_amount()`
 
 [feat: forward asset and amount on contract call](https://github.com/FuelLabs/fuels-ts/commit/b6b64d640318e5389bd5b9faba5b4e0c08adc724)
 
 - Enable forward custom asset and amount on contract call

On typescript;
```
const result = await contract.functions.deposit({
  amount: 1_000_000,
  assetId: '0x0000000000000000000000000000000000000000000000000000000000000000',
});
```

On Sway these values can be retrieved like this;
```
  fn return_context_amount() -> u64 {
    msg_amount()
  }
  fn return_context_asset() -> b256 {
    (msg_asset_id()).into()
  }
```

closes: #198


packages/contract/src/scripts.ts
```diff
@@ -9,35 +12,56 @@ import { Script, ReceiptType } from '@fuel-ts/providers';
...
+    // Load the amount into 0x12
+    Opcode::ADDI(0x12, REG_ZERO, data_offset + 32),
+    // Load word into 0x12
+    Opcode::LW(0x12, 0x12, 0),
+    // Load the asset id to use to 0x13.
+    Opcode::ADDI(0x13, REG_ZERO, data_offset),
+    // Call the transfer contract.
+    Opcode::CALL(0x10, 0x12, 0x13, REG_CGAS),
+    Opcode::RET(REG_ONE),
   */
```
AlicanC, line 32: Is this correct? A comment says `Load gas forward to 0x11`, but I don't see `0x11` being used.

AlicanC (commented):
LGTM w/ 1 comment

packages/contract/src/scripts.ts
luizstacio: This is a good point, during don't correctly the answer. cc @digorithm.

digorithm: Oh, I forgot to remove that comment on the Rust SDK I guess? We're still not forwarding gas. Once we do we can use 0x11 (or any unreserved) for that.

AlicanC (approved):

PR #200, opened 2022-03-27
fix: set json fragment name optional

luizstacio:
### Summary

- When declaring a JSONFrament `input/outputs` name is not required.

#### Example

This works;
```ts
    const contract = await setup([
      {
        type: 'function',
        name: 'return_context_asset',
        outputs: [
          {
            name: 'arg0',
            type: 'b256',
          },
        ],
      },
    ]);
```

This should work also;
```ts
    const contract = await setup([
      {
        type: 'function',
        name: 'return_context_asset',
        outputs: [
          {
            type: 'b256',
          },
        ],
      },
    ]);
```

### DIFF
--- a/packages/abi-coder/src/abi-coder.ts
+++ b/packages/abi-coder/src/abi-coder.ts
@@ -28,20 +28,22 @@ export default class AbiCoder {
   }
 
   getCoder(param: JsonFragmentType): Coder {
+    const name = param.name || '';
+
     switch (param.type) {
       case 'u8':
       case 'u16':
       case 'u32':
       case 'u64':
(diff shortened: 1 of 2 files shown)


luizstacio:
@QuinnLee, this was implemented here: https://github.com/FuelLabs/fuels-ts/pull/197/files#diff-620d54c98858ccefe59323ac69ab85f1a37cd15f6d11b5aaad6d8c34504a7a4fR4 any solid reason for name to be required?

QuinnLee (approved):

AlicanC (approved):

PR #218, opened 2022-04-15
chore: forc with latest version

QuinnLee:
updating this test


luizstacio (approved):

adlerjohn:
Would it be better to generate these things automatically in CI instead of checking them in?

AlicanC:
@adlerjohn there wasn't a good way to source forc during development and on CI when we started. If we can get forc in CI then we can gitignore these.

AlicanC (approved):

Issue #229, opened 2022-04-26
Add coins to transaction, avoiding input duplication


luizstacio:
The SDK is allowing inserting duplicated inputs/coins into a request.

1. When forwarding `asset amount`, the SDK should replace the current input with the new one with enough amount to pay `amount + gas`, removing the need to add coins to the request manually.
2. When the forwarded `assetId` is not present on the `inputs`, we should also add a new input.
3. The SDK should check to duplicate inputs before sending the request?

### Use case
```ts
await contract.functions.<method>({
        assetId: '0x0000....0000',
        amount: 1_000_000,
        // The request on transformRequest has already been full filled with
        // Coins to cover the gas cost but it will not be enough to cover gas + amount forwarded
        transformRequest: async (request) => {
          // Query amount of coins to cover the transaction cost + gas
          const coins = await wallet.getCoinsToSpend([[1_500_000, '0x0000....0000']]);
          // Add the coins to the transaction
          request.addCoins(coins);
          // Now request has duplicated inputs
          return request;
        },
});
```

### Temporary solution
```ts
await contract.functions.<method>({
        assetId: '0x0000....0000',
        amount: 1_000_000,
        transformRequest: async (request) => {
          // Remove all previous inputs with the same coin
          request.inputs = request.inputs.filter(i => {
            return !(i.type === InputType.Coin && i.assetId === coinFrom.assetId);
          });
          const coins = await wallet.getCoinsToSpend([[1_500_000, '0x0000....0000']]);
          request.addCoins(coins);
          // Now request has no duplicated inputs
          return request;
        },
});
```

PR #232, opened 2022-04-29
feat: replace `amount`/`assetId` overrides with `forward`

AlicanC:
This PR replaces the current way of forwarding coins to a contract call from the `amount`/`assetId` override combo to a single `forward` property.

Another change is that when you call a contract using `forward`, the SDK will automatically get coins from the wallet so:
- You don't have to call `getCoinsToSpend` and manually insert coins.
- Both this new functionality and the current txn funding mechanism uses a single `getCoinsToSpend` call so https://github.com/FuelLabs/fuels-ts/issues/229 is fixed as a side effect.

```ts
// Before
await contract.functions.<method>({
        assetId: '0x0000....0000',
        amount: 1_000_000,
        transformRequest: async (request) => {
          // Remove all previous inputs with the same coin
          request.inputs = request.inputs.filter(i => {
            return !(i.type === InputType.Coin && i.assetId === coinFrom.assetId);
          });
          const coins = await wallet.getCoinsToSpend([[1_500_000, '0x0000....0000']]);
          request.addCoins(coins);
          // Now request has no duplicated inputs
          return request;
        },
});

// After
await contract.functions.<method>({
        forward: [1_000_000, '0x0000....0000'],
});
```

### DIFF
--- a/packages/contract/src/call-test-contract/call-test-contract.test.ts
+++ b/packages/contract/src/call-test-contract/call-test-contract.test.ts
@@ -163,8 +163,7 @@ describe('TestContractTwo', () => {
       },
     ]);
     const result = await contract.functions.return_context_amount({
-      amount: 1_000_000,
-      asset_id: NativeAssetId,
+      forward: [1_000_000, NativeAssetId],
     });
     expect(BigNumber.from(result).toNumber()).toBe(1_000_000);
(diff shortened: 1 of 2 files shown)


luizstacio (approved):
LGTM

PR #238, opened 2022-05-03
refactor: add turborepo, pnpm and tsup

### LINKED ISSUE #236: chore: migrate away from `lerna`
https://github.com/lerna/lerna/issues/2703 :( 

We should get an alternative

pedronauck:
This PR introduces the replacement of Lerna and NPM (closes #236) with [Turborepo](https://turborepo.org), [PNPM](https://pnpm.io) and adds [tsup](https://tsup.egoist.sh) for build. Also, this pull request adds some general improvements listed bellow:

## Tasks

- [x] Add Turborepo as monorepo manager
- [x] Change to PNPM instead of NPM
- [x] Use imports directly with [publishConfig](https://pnpm.io/package_json#publishconfig) setting *(don't need `build --watch` tasks)*
- [x] Use [tsup](https://tsup.egoist.sh/) for build using `esm | cjs | iife` formats
- [x] Change Github actions to use PNPM
- [x] Use [npm-run-all](https://www.npmjs.com/package/npm-run-all) in order to run docker services and test in serial
- [x] Add docker services commands inside a `Makefile` (529a8d51)
- [x] Tests Github Actions

👉🏻 And after some changes I struggle with other issues (that's already solved too):

- [x] Create an `index.ts` entry point for `wordlists` and `merkle-shared` packages in order to follow monorepo "pattern"
- [x] Exclude `node_modules` and `dist` files on `tsconfig.base.json`

## 🤩 Wins

Now we have some new benefits because of the new toolchain:

✅ A very good cache system on the build pipeline [(thanks to Turborepo)](https://turborepo.org/docs/core-concepts/caching)
✅ So much faster package installations (thanks to PNPM)
✅ Consume packages internally without watch builds
✅ Better `package.json` definitions with specific builds for CommonJS and ESM modules (thanks to tsup)
✅ Since we're using `tsup` and `pnpm` we don't need `references` and `composite: true` on `tsconfig` files anymore

## Development Breaking changes 

- The `run-services` command was renamed to `services:run`
- We have now a `ci:test` command that runs `services:run`, `test` and `services:clean`

## 🔗 References

→ Why use PNPM? https://pnpm.io/motivation
→ Why use Turborepo? https://turborepo.org/docs#why-turborepo


pedronauck:
**→ About the Github Actions:** I'm using [Act](https://github.com/nektos/act) locally in order to run the actions, but can't run yet the publishing-related actions. The build action is running, but it's getting stuck in the test step (because I don't have a `secret.GITHUB_TOKEN` configured locally)

**→ About publishing:** One possible way we have to test publishing using another tools without messing around with our versions is making something like that: https://developer.entando.com/v6.1/tutorials/ecr/how-to-create-local-npm-registry.html#start-local-nexus-as-a-docker-container

I'll try here locally!

pedronauck:
Since we already with a very big pul request, I'll let this pull request using Lerna just for publishing and create another one in other to replace this part of the system on it.

QuinnLee:
How would we publish on npm with the new setup?

.github/workflows/sematic-release.yaml
```diff
@@ -17,19 +17,37 @@ jobs:
     steps:
```
QuinnLee, line 17: I think we can remove this.  We couldn't get Sematic-release + mono repo to run on CI, it would error out 

https://github.com/pmowrer/semantic-release-monorepo

AlicanC:
I won't be able to fully review until tomorrow but I have some qs:

1- Does pnpm support something like Yarn's "Offline Cache"? https://yarnpkg.com/features/offline-cache

I think it would be great if we could commit all of our deps to the repo (like Yarn does in [their repo](https://github.com/yarnpkg/berry/tree/master/.yarn/cache)) so we avoid left-pad-like issues and also easily track dep changes.

---

2- How hard would it be to move to Yarn 2/3 if we decided to in the future? How hard are we depending on pnpm?

.github/workflows/sematic-release.yaml
pedronauck: Remove the entire Github Action @QuinnLee ?

pedronauck:
> I won't be able to fully review until tomorrow but I have some qs:
> 
> 1- Does pnpm support something like Yarn's "Offline Cache"? https://yarnpkg.com/features/offline-cache
> 
> I think it would be great if we could commit all of our deps to the repo (like Yarn does in [their repo](https://github.com/yarnpkg/berry/tree/master/.yarn/cache)) so we avoid left-pad-like issues and also easily track dep changes.
> 
> 2- How hard would it be to move to Yarn 2/3 if we decided to in the future? How hard are we depending on pnpm?

1. About offline support @AlicanC, PNPM has already https://pnpm.io/cli/install#--offline

2. The only thing we're depending on PNPM direct import, to do that we're using a `main` field on `package.json` pointing directly to the `index.ts` file and adding a `publishConfig` appointing to the compiled files (as you can [see here](https://github.com/FuelLabs/fuels-ts/pull/238/commits/d5a7021df755722e94b1d7d6f2795e0e197c8fca#diff-ae8b97dabdf446b7f229ac907def90a770ef59693a48fee298fdd926ab353d3dR7)). I think that don't have anything anymore coupling us with it. So, a change in the future for any other package manager should be easy.

pedronauck:
> How would we publish on npm with the new setup?
> 
> I've been doing
> 
> ```shell
> lerna version --conventional-commits --no-push
> npx lerna publish from-git
> ```
> 
> manually for versioned releases

Yes @QuinnLee, I don't change anything related to `lerna publish` mainly because the pull request is already too big, I'm working on another one just to see about this. So, we keep the same flow for publishing using Lerna.

For now, I'm checking [Rush](https://rushjs.io/) for publishing, and in fact, what I'm seeing is a lot of very good features. Rush is made my Microsoft, so there's a big company behind, less chances to die in the future like Lerna haha.

.github/workflows/sematic-release.yaml
QuinnLee: yes!

luizstacio (approved):
Nices change!
Testing locally everything is working as expected also!

AlicanC:
How can I work on SwaySwap and fuels-ts at the same time using this setup? Meaning:
- I have two VSCode windows open: fuels-ts and swayswap/client
- I add a new API on fuels-ts
- I save
- I go to swayswap window and just use the API

I used to do this by doing `yarn lerna exec -- yarn link` on `fuels-ts` and `yarn link fuels` on `swayswap`, but I can't install with Yarn since we changed to `workspace:*` versions that Yarn doesn't seem to support.

AlicanC:
I also used to run `npx tsc --build --watch` at `fuels-ts` root to watch all packages and build them on changes. This doesn't seem to work now. (Probably because we removed the references setup.) How can I achieve the same?

.github/workflows/publish-npm-master.yaml
```diff
@@ -18,25 +18,48 @@ jobs:
...
+      - name: Setup PNPM
+        uses: pnpm/action-setup@v2.1.0
+        with:
+          version: 7.0.0
+          run_install: true
+
+      - name: Run ESLint and Build
+        run: |
+          pnpm build
+          pnpm lint
```
camsjams, line 52: `lint` should be before `build` right? 

If lint fails it should not build invalid code.

PR #239, opened 2022-05-04
feat: add contract slots on contract deployment

luizstacio:
### Summary
- Remove `std-lib` and `std-core`. To use the inherit libs from `forc`. 
- Add `storage slots` (initial storage values), on `deployContract`.
- Add tests to `storage slots` on `storage-contract-test`.
- Test support for `b256`  storage slot. To test it we added a `asm` code on contract.`b256` are WIP on `storage::get` from `sway`.

### Notes
- Only values with `32 bytes` are supported on storage slots. To make a better UX, I have made a `hex pad right`, enabling to pass values that are not `b256` like `u64`. Ex. a value `0x1000000000000001` will be sent as `0x1000000000000001000000000000000000000000000000000000000000000000`.

closes: #126


packages/providers/src/transaction-request/storage-slot.ts
```diff
@@ -1,14 +1,37 @@
-import { hexlify } from '@ethersproject/bytes';
+import type { BytesLike } from '@ethersproject/bytes';
```
QuinnLee, line 1: 👍

QuinnLee (approved):
👍  Looks like most of the PR is `.sw` files

AlicanC (approved):

PR #242, opened 2022-05-05
fix: github action for publish on master and replaces Lerna publishing

pedronauck:
## ✅ Tasks 

- [x] This pull request intends to fix the Github Action that publishes a new version on each push on `master`, broken earlier because we change to Turborepo and PNPM.
- [x] Also creates a new script `pnpm release:stable` that enables to use `lerna version` just to create changelogs based on conventional commits, but use pnpm to publish packages

## 📑 Description

I tried to fix just what was broken in the last pull request without changing so much our release flow. Mainly because after talking with @QuinnLee, I said that maybe [changesets](https://github.com/changesets/changesets) could be the best replacement for Lerna (since Rush does so much more things than we want). But changesets will have some implications on our development flow.

Because of that, I left Lerna just generating our changelogs based on conventional commits, and the rest it's been done by [some scripts](https://github.com/FuelLabs/fuels-ts/tree/fix/publish-master/scripts) involving PNPM and Turborepo.

## 👨🏻‍💻 Dev Breaking Changes

So, after this pull request the only change is that now for releases we need to run:

```sh
$ pnpm release:stable
```

This script will replace all `workspace:*` dependencies and turn Lerna able to read `package.json` on each package. Then will run `npx lerna version` with some arguments, generating changelogs and all git stuff are doing using shell commands.


luizstacio:
@pedronauck how is the flow to work locally with `npm link`?

scripts/release-stable.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (response.confirm) {
+    sh.exec(`pnpm publish -r --tag=latest --no-git-checks --force`);
+    sh.exec('git add ./CHANGELOG.md');
+    sh.exec('git add ./package.json');
+    sh.exec('git add ./lerna.json');
+    sh.exec('git add ./packages/**/CHANGELOG.md');
+    sh.exec('git add ./packages/**/package.json');
+    sh.exec(`git commit -m "bump to version ${version}"`);
+    sh.exec(`git tag -a ${version} -m "bump to version ${version}"`);
+    sh.exec('git push origin master --tags');
```
luizstacio, line 45: I think it should create a PR

scripts/release-stable.ts
```diff
@@ -0,0 +1,47 @@
...
+
+(async () => {
+  const mainPkgJSON = await fs.readJSON(resolveDir('./package.json'));
+
+  /**
+   * Change all package.json with the version from main package.json
+   * Run lerna to generate changelogs and bump version based on conventional commits
+   */
+  await changeAllPkgJSON(mainPkgJSON.version);
+  const args = `--conventional-commits --no-push --no-git-tag-version --no-commit-hooks --no-ignore-changes --yes`;
```
luizstacio, line 17: This sounds a bit problematic because if anything breaks during the running of this, we will need to checkout or do another manual reversion of the files.

The probability looks low, and also if running on CI it only will need to be re-runned.

scripts/release-stable.ts
luizstacio: Create a new branch with the version name and then push and open the PR. something like this.

luizstacio (changes requested):

QuinnLee (approved):

scripts/release-stable.ts
pedronauck: But if it will open a pull request, when will occur the `publish` execution for npm? 🧐 In fact, this command is just a small replacement of what @QuinnLee was already using with `lerna` commands globally, but now integrated with PNPM workspaces.

Because there's no sense create a pull request to be reviewed because if we're publishing a stable version, we already had reviews and just need to publish. This will be executed on `master` branch, just need a commit and tag in order to represent that at that point had a new version.

scripts/release-stable.ts
pedronauck: Cool, did some kinda a `trycatch` making a `git checkout .` if some error occurs. But, because of that, the git commands are just executing after all release/changelogs/version happens.

pedronauck:
Hey @luizstacio @QuinnLee , I found an alternative to generate changelogs instead of Lerna: [git-chglog](https://github.com/git-chglog/git-chglog)

I implemented it [in our scripts](https://github.com/FuelLabs/fuels-ts/pull/242/commits/f4ef88cb8f058a7f363b3996d53b9bfff4b3283d), so our workflow keeps the same, but we don't have Lerna anymore 🔥

luizstacio (approved):

PR #246, opened 2022-05-05
fix: change build tasks in order to use pnpm link correctly

pedronauck:
## Description 

This pull request intends to change our environment in order to be able to use `pnpm link` correctly.

After talking with @luizstacio we tried to keep the direct import in all `package.json` by using the `publishConfig`, but we notice that if you link some package to the global pnpm store, it'll keep the same `package.json` that uses the `main` field pointing to the Typescript file. This is bad if you trying to use the linked package with some bundler (vite, react-scripts, etc) because the bundler won't handle loaders to bundle outside the package directory, and by consuming `index.ts` file this is needed.

So, we change to instead of consuming direct inside our monorepo, we need to build/watch and have all `package.json` field pointing to the same publish config. It's an annoying trade-off, but by now is the only way to work.

I'll open an issue on PNPM repository in order to check if is possible to run this scenário.

## Main Changes

- Now for work inside our monorepo, we need to run in the root directory:

```sh
$ pnpm build:watch
```

This is specified in our README as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,7 @@ const contract = MyContract__factory.connect(contractId, provider);
 git clone git@github.com:FuelLabs/fuels-ts.git
 cd fuels-ts
 pnpm install
+pnpm build
 ```
 
 ### Testing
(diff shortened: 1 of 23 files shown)


luizstacio:
This enables us to use `pnpm link` that is similar to `npm link` used today.

luizstacio (approved):

PR #257, opened 2022-05-09
Upgrade to Forc v0.12.1 and some chores

AlicanC:
[chore!: drop long format function names](https://github.com/FuelLabs/fuels-ts/commit/607cadb08bb6e09454c0ea8def4a02aebbf4c4f1)

When we encounter the ABI definition of, for example, `fn foo(bar: u64)`, we generated two methods: one named `foo` and another named `foo(u64)`. AFAIK, unlike in the Ethereum ecosystem, this isn't useful for us since we don't have overloads and referring with just the name is okay. So I removed the long format fns.

[chore: add JsonAbi type for convenience](https://github.com/FuelLabs/fuels-ts/commit/c4486cdb55289ad5ba0b0a966bbd6ce8932d9b61)

We used `ReadonlyArray<JsonFragment>` to refer to an object in Fuel JSON ABI Format. I created an alias named `JsonAbi` to make it more clear and also moved all JsonAbi related types to their own file and also added some commentary.

[feat!: upgrade to Forc v0.12.1](https://github.com/FuelLabs/fuels-ts/commit/55c959c4882f0d46e747f9a33f9462abc20f7b03)

Just a regular upgrade. `example-contract` was erroring so I removed in-repo stdlib projects. (We needed this anyways.)


packages/abi-coder/src/json-abi.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Types for Fuel JSON ABI Format as defined on:
+ * https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md#json-abi-format
+ */
+
+export interface JsonAbiFragmentType {
```
QuinnLee, line 6: 👍

packages/contract/src/contract.ts
```diff
@@ -1,8 +1,9 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import type { BigNumberish } from '@ethersproject/bignumber';
 import { Logger } from '@ethersproject/logger';
-import type { JsonFragment, FunctionFragment } from '@fuel-ts/abi-coder';
+import type { FunctionFragment } from '@fuel-ts/abi-coder';
```
QuinnLee, line 4: CI is fialing here

luizstacio:
Looks really better now!

packages/contract/src/contract.ts
AlicanC: Whoops. Seems like ESLint autofix on save didn't work.

PR #266, opened 2022-05-12
feat: move from BigNumber to BigInt

AlicanC:
This PR:
- Adds a new package called `@fuel-ts/math` which has some utilities for working with bigints. This is a copy from Ethers v6 and we can move to their package when that's out of beta.
- Removes any dependency to `@ethersproject/bignumber` and uses native bigints instead.
- Fixes the test for UtxoIdCoder, which was a copy of WitnessCoder that tested Witness :D
- Upgrades to Jest 28. (27 had issues with bigints) #267

Notes:
- In the `transactions` package, instead of changing all BigNumbers to BigInts, I changed the ones lte u32 to just `number`. They are mostly lengths and indexes and those are natively represtented as numbers and not bigints in JS so this makes our APIs more convenient.


packages/abi-coder/src/coders/number.ts
```diff
@@ -1,16 +1,20 @@
...
+type NumberCoderType = 'u8' | 'u16' | 'u32' | 'u64';
+
+type ToDecodedType<TType extends NumberCoderType> = TType extends 'u64' ? bigint : number;
+
+export default class NumberCoder<TType extends NumberCoderType> extends Coder {
   // This is to align the bits to the total bytes
   // See https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md#unsigned-integers
   length: number;
-  baseType: string;
+  baseType: TType;
```
luizstacio, line 14: Maybe rename it to `BaseType`?

luizstacio (approved):
Just one comment

QuinnLee (approved):
👍

camsjams:
Should we add this to the typedoc list?

AlicanC:
> @AlicanC Should we add this to the typedoc list?

Yeah, nice catch.

PR #271, opened 2022-05-13
Improve Script Error Reporting

AlicanC:
[feat: improve TransactionResponse](https://github.com/FuelLabs/fuels-ts/pull/271/commits/782c4f3105a33935c47f24547f67aea3f26bb168)

- Extracts `TransactionResponse` from `provider.ts` into its own file.
- Adds `TransactionResponse.waitForResult()` which doesn't throw on failure so you can do your own error handling.

[feat: improve Script error handling](https://github.com/FuelLabs/fuels-ts/pull/271/commits/fb0f3c6b1ce89cbb98ad262641085875d84b7d88)

- Uses the new `waitForResult` to do better error reporting

Before:
image
After:
image

### DIFF
--- a/packages/providers/src/transaction-response/transaction-response.ts
+++ b/packages/providers/src/transaction-response/transaction-response.ts
@@ -1,0 +1,132 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { arrayify } from '@ethersproject/bytes';
+import type {
+  ReceiptCall,
+  ReceiptLog,
+  ReceiptLogData,
+  ReceiptPanic,
+  ReceiptReturn,
+  ReceiptReturnData,
+  ReceiptRevert,
+  ReceiptTransfer,
+  ReceiptTransferOut,
+  ReceiptScriptResult,
+} from '@fuel-ts/transactions';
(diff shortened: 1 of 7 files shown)


packages/providers/src/transaction-response/transaction-response.ts
```diff
@@ -0,0 +1,132 @@
...
+    return transaction;
+  }
+
+  /** Waits for transaction to succeed or fail and returns the result */
+  async waitForResult(): Promise<TransactionResult<any>> {
+    const transaction = await this.#fetch();
+
+    switch (transaction.status?.type) {
+      case 'SubmittedStatus': {
+        // TODO: Implement polling or GQL subscription
```
luizstacio, line 97: Can you create an issue for this?

luizstacio (changes requested):

QuinnLee:
cc @camsjams

packages/providers/src/transaction-response/transaction-response.ts
AlicanC: We don't need to track this separately since it will be a mandatory change when we will eventually need to upgrade to a fuel-core version that doesn't immediately produce blocks.

luizstacio (approved):

Issue #216, opened 2022-04-13
Multiple contract calls in a single transaction


luizstacio:
if it's possible to write a script that can call multiple contract/contracts methods/calls on a single TX. Would be amazing if the SDK provide a simple API to build these actions in a bundle TX.

# Use case

On swayswap to create liquidity between two pairs, you need to follow steps 1, 2, and 3 in sequence. This executes 3 transactions but would be nicer to build all 3 method calls, combine and send in a single TX.

1. Deposit coin A
2. Deposit coin B
3. Create liquidity

<img width="500" alt="Screen Shot 2022-04-13 at 6 03 06 PM" src="https://user-images.githubusercontent.com/3941923/163222523-bba9e1b6-a0f9-4bd8-a81f-fcae6ca7da74.png">

AlicanC:
Update:

I finally got Multicall script to work today, but a lot needs to happen before we can merge it:
- It currently works with Forc@master. Forc with the features we need must be stable.
- The `@fuel-ts/contract` API still only supports single calls. (@camsjams' #303 has `prepareCall` that will be useful for this.)
- The script works with the heap so it needs `std::alloc` and `std::ptr`: https://github.com/FuelLabs/sway/pull/1627
- The script uses generics so it needs this fixed: https://github.com/FuelLabs/sway/issues/1628 (PR: https://github.com/FuelLabs/sway/pull/1671)

Some good-to-haves would be:
- https://github.com/FuelLabs/sway/issues/1077 (PR: https://github.com/FuelLabs/sway/pull/1118)
- https://github.com/FuelLabs/sway/issues/1188
- https://github.com/FuelLabs/sway/issues/1585

PR #274, opened 2022-05-14
Chores

AlicanC:
[fix: repo scripts](https://github.com/FuelLabs/fuels-ts/pull/274/commits/531f733ad8a95888bd63aa5537352ec1697c7bba)

**Days since someone got rekt by not having `set -euo pipefail`: 0**

We didn't have `set -euo pipefail` in our .sh files so we weren't noticing one of the contracts was failing to build. (The reason was the forgotten `lib-std` and `lib-core` dependencies.) I've fixed the issue and also added `set -euo pipefail` to all of our .sh files. 

[fix: coder type](https://github.com/FuelLabs/fuels-ts/pull/274/commits/9e5deafd710fe9d40a712cc85fc5b1d0fa381cca)

`maturity` was previously typed `u32` and was later changed to `u64`. We forgot some and didn't notice because we encode them the same.

[chore: replace hardcoded script with actual Sway code](https://github.com/FuelLabs/fuels-ts/pull/274/commits/0041076f41535c59b71cbca9b1561e7c76b97d5a)

This one replaces our hardcoded contract call script with an actual Sway script. It's a hacky script but we can improve it as Sway allows us to.

This should also get us closer to https://github.com/FuelLabs/fuels-ts/issues/216


QuinnLee (approved):

packages/transactions/src/coders/transaction.ts
```diff
@@ -63,7 +63,7 @@ export class TransactionScriptCoder extends Coder {
     parts.push(new NumberCoder('gasPrice', 'u64').encode(value.gasPrice));
     parts.push(new NumberCoder('gasLimit', 'u64').encode(value.gasLimit));
     parts.push(new NumberCoder('bytePrice', 'u64').encode(value.bytePrice));
-    parts.push(new NumberCoder('maturity', 'u32').encode(value.maturity));
+    parts.push(new NumberCoder('maturity', 'u64').encode(value.maturity));
```
luizstacio, line 66: I think we should keep it as spec even if the decoder results in the same 8 bytes. https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/tx_format.md

luizstacio (changes requested):

packages/transactions/src/coders/transaction.ts
AlicanC: Hmm. I thought we changed maturity to u64. There are other places where we type maturity u64 so I've opened an issue to fix this completely: https://github.com/FuelLabs/fuels-ts/issues/275

Let's keep this change though so at least all maturities are u64 instead of only one being u32.

AlicanC: Actually, this fixes a bug and we need it because the JS type is `bigint` and this returns `number`.

luizstacio: cc: @adlerjohn

luizstacio (approved):

packages/transactions/src/coders/transaction.ts
adlerjohn: https://github.com/FuelLabs/fuels-ts/issues/275#issuecomment-1127022049

PR #276, opened 2022-05-16
Uses native crypto APIs

### LINKED ISSUE #264: Remove AES-js and use native crypto libs on browser and node
Per #203 there is a todo within [@fuel-ts/keystore](https://github.com/FuelLabs/fuels-ts/blob/master/packages/keystore/src/keystore.ts#L5) to swap out aes-js with native solutions.

camsjams:
Fixes #264

Uses native `crypto` module in Node and native `crypto`/`SubtleCrypto` in web browser. The advantage here aside from performance would be using native tools built into environment with more strict security and maintenance routines.

The `encrypt` and `decrypt` functions in `keystore` are now asynchronous (they were not before).

Noting that [wallet-manager](https://github.com/FuelLabs/fuels-ts/blob/master/packages/wallet-manager/src/wallet-manager.ts#L2) already used these functions as `Promises`.

### DIFF
--- a/packages/keystore/src/universal-crypto.ts
+++ b/packages/keystore/src/universal-crypto.ts
@@ -1,33 +1,33 @@
-import type {createCipheriv, createDecipheriv} from 'crypto'
+import type { createCipheriv, createDecipheriv } from 'crypto';
 
 type UniversalCrypto = {
-    getRandomValues: (length: number) => Uint8Array;
-    randomBytes: (length: number) => Uint8Array;
-    subtle: SubtleCrypto;
-    createCipheriv: typeof createCipheriv;
-    createDecipheriv: typeof createDecipheriv;
-}
+  getRandomValues: (length: number) => Uint8Array;
+  randomBytes: (length: number) => Uint8Array;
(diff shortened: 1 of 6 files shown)


packages/keystore/src/universal-crypto.ts
```diff
@@ -0,0 +1,34 @@
+import type { createCipheriv, createDecipheriv } from 'crypto';
```
luizstacio, line 1: This is working on a web app mainly with typescript also?

luizstacio (changes requested):
LGTM. One question only

packages/keystore/src/universal-crypto.ts
luizstacio: This is a nice question, I thought the same thing when I was asking haha...

We should open an issue to have `SDK tests on browser compatibility` in the future.

luizstacio: https://github.com/FuelLabs/fuels-ts/issues/284

luizstacio (approved):

Issue #268, opened 2022-05-12
Improve AbiCoder


AlicanC:
- Add a way to check if a type is a reference type

luizstacio:
This is also related to #231 also this comment: https://github.com/FuelLabs/fuels-ts/issues/231#issuecomment-1127644950

We can completely remove the check the type and just [check if args has more them 8 bytes.](https://github.com/FuelLabs/fuels-ts/blob/59cbeff8331f3c3a4e0373e67816fc143d69c2e5/packages/contract/src/scripts.ts#L27)

luizstacio:
> Am I misunderstanding the table here? [FuelLabs/sway#1368](https://github.com/FuelLabs/sway/issues/1368)
> 
> It looks like `const isRef = (bytes) => bytes.length > 8;` wouldn't return the correct result.

I think the table has changed from the type it was like what I said. If wall struct should be pointer we really need to check the type

Issue #231, opened 2022-04-28
Fix the handling of the remaining Sway types in and out of a contract method


digorithm:
Same as https://github.com/FuelLabs/fuels-rs/issues/239. 

Many changes in the way types are handled. We should investigate what's the current state on `fuels-ts` in relation to this and implement the appropriate fixes.

camsjams:
From what I can tell, based on my short time with this project, there are no updates needed.

If anyone is available to point me in the right direction that would be great.

PR #277, opened 2022-05-16
Improve contract call script

AlicanC:
[fix: contract call naming and values](https://github.com/FuelLabs/fuels-ts/pull/277/commits/fd1ed8d6f3f5d151ea15aeb7bb55cd36a7e8a375)

Fixed some misunderstandings in the contract calling code:

We used a var named `isStruct` to indicate how we should encode the arg. This isn't actually about the arg being a struct or not, but about if it's a reference type or not. Renamed it to `isReferenceType`.

While the end result was the same, `getScriptDataOffset()` was actually returning the offset for the call data and not the script data. Fixed this.

[chore: improve contract call tests](https://github.com/FuelLabs/fuels-ts/pull/277/commits/406bc872be665098930976a8de69926e2887bf4a)

Added a few more test cases to make testing the contract call script easier.

[feat: add @fuel-ts/script package](https://github.com/FuelLabs/fuels-ts/pull/277/commits/255f9d3e62fc3a5ee19bb00d90f9187f2b2ecd94)

❗**PR UPDATE**❗   Added a new package called `@fuel-ts/script`. Basically like `@fuel-ts/contract` but for scripts. It isn't packed with features or anything but it has a nice clean test that helps a lot when you just need to debug a script call and not a whole contract call.

[chore: improve contract call script](https://github.com/FuelLabs/fuels-ts/pull/277/commits/ecfc5e6b8e9d05d702aaead77156efe8b98021b3)

Typed script_data so the compiler can calculate memory addresses instead of us doing it manually. ❗**PR UPDATE**❗   Also removed the patching we did to get the script length in favor of `std::tx::tx_script_length()`. It's bugged so we need to insert `asm(r1: 0x0000000000000000000000000000000000000000000000000000000000000000) {}` into code, but that's a lot cleaner.


packages/providers/src/script.ts
```diff
@@ -103,17 +103,16 @@ export class Script<TData = void, TResult = void> {
   }
 
   getScriptDataOffset() {
-    return (
-      VM_TX_MEMORY + TRANSACTION_SCRIPT_FIXED_SIZE + ASSET_ID_LEN + AMOUNT_LEN + this.bytes.length
-    );
+    return VM_TX_MEMORY + TRANSACTION_SCRIPT_FIXED_SIZE + this.bytes.length;
```
luizstacio, line 106: `ASSET_ID_LEN + AMOUNT_LEN` is always required on the contract call offset I think, we should keep both here.

As `getArgOffset()` its ysed for struct inputs and `ASSET_ID_LEN + AMOUNT_LEN` are required on all calls.

luizstacio (changes requested):

luizstacio:
About `isStructType` maybe we can work to remove it already by checking the byte size direct on the encoding method. Ref.: https://github.com/FuelLabs/fuels-ts/issues/231

packages/providers/src/script.ts
luizstacio: Makes sense.

I think some thoughts on this;
- Maybe we should create another method that gives the `offset` where the data should be placed.
- With `lw` we are able to pass values bigger than `u64` to `OP:CALL` and this offsets the data to go after `amount`
   - When forwarding `asset_id` we also add this to the offset, even when the params of the `call` are not of the type struct

luizstacio (changes requested):

packages/contract/src/contract-call-script/src/main.sw
```diff
@@ -21,14 +19,33 @@ fn get_script_data_offset() -> u64 {
...
+    contract_id: ContractId,
+    fn_selector: [u8; 8],
+    fn_arg_or_ptr: u64,
+}
 
-    asm(r1: call_data_offset, r2: amount_offset, r3: asset_id_offset) {
-        lw r2 r2 i0;
+fn call_contract(call_data: CallData, amount: u64, asset_id: b256) {
+    asm(r1: call_data, r2: amount, r3: asset_id) {
         call r1 r2 r3 cgas;
```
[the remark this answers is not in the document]
luizstacio: @adlerjohn thanks for pointing this out. I miss informed here; Without `lw` the size was fixed to `u16`. And when passing a value up to `u64`, the value was parsed incorrectly.

This was fixed here:  https://github.com/FuelLabs/fuels-rs/pull/175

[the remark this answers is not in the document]
luizstacio: Knowing that all values from `u8` to `u64` are encoded into `8 bytes`. I can see that the `lw` is not required, and maybe this could be an issue in another place where we `decode` the `amount`? Makes sense?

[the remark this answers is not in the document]
luizstacio: @adlerjohn @AlicanC as we have tests to return the amount from context and this test is now working with values like `1_000_000` without the `lw` we can safely remove it.

luizstacio (approved):

packages/contract/src/contract-call-script/src/main.sw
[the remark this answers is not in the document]
luizstacio: Also open a issue to remove it on `fuels-rs`: https://github.com/FuelLabs/fuels-rs/issues/285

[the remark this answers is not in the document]
adlerjohn: Not sure I follow. `lw` is required for integers that don't fit into 12 bits, no?

packages/contract/src/contract-call-script/src/main.sw
```diff
@@ -21,14 +19,33 @@ fn get_script_data_offset() -> u64 {
...
+    let script_data_ptr = get_script_data_offset();
+    let script_data = asm(r1: script_data_ptr) {
+        r1: T
+    };
+    script_data
+}
+
+struct CallData {
+    contract_id: ContractId,
+    fn_selector: [u8; 8],
```
adlerjohn, line 32: Not sure this is correct. Each element here is one word, so this takes up 8 words. There's no compacting supported yet.

adlerjohn, line 32: This doesn't look correct. Each `u8` in the array takes up one word. So in total this takes up 8 words.

adlerjohn (changes requested):

packages/contract/src/contract-call-script/src/main.sw
AlicanC: Yeah, just experienced this. Didn't break anything on this PR but when I tried logd-ing based on `size_of<[u8; 8]>()` it was logging too much.

adlerjohn: Suggest just using a `u64` here, or perhaps a `u32`.

AlicanC:
Updated the PR, please see the OP.

packages/abi-coder/src/interface.ts
```diff
@@ -118,9 +118,11 @@ export default class Interface {
       return selector;
     }
 
-    const isStruct = !(inputs.length === 1 && inputs[0].type === 'u64');
+    const isReferenceType = !(inputs.length === 1 && inputs[0].type === 'u64');
```
adlerjohn, line 121: Should this not also check for `u8`, `u16`, `u32`, and `bool`, or are those all folded into `u64` earlier?

packages/contract/scripts/build-contract-call-script.sh
```diff
@@ -2,5 +2,9 @@
 
 set -euo pipefail
 
-forc build -p src/contract-call-script
+SCRIPT_DIR="src/contract-call-script"
+BIN_DIR="$SCRIPT_DIR/out/debug"
+
+forc build -p $SCRIPT_DIR --print-finalized-asm
 pnpm exec ts-node scripts/process-contract-call-script.ts
+# forc parse-bytecode "$BIN_DIR/contract-call-script.bin" > "$BIN_DIR/contract-call-script.txt"
```
adlerjohn, line 10: Remove debugging statement.

packages/contract/src/contract-call-script/src/main.sw
```diff
@@ -1,34 +1,36 @@
 script;
 
-use std::context::registers::*;
+use std::contract_id::ContractId;
 
-const CONTRACT_ID_LEN = 32;
-const WORD_SIZE = 8;
+fn get_script_data<T>() -> T {
```
adlerjohn, line 5: This is actually quite an interesting workaround for not having script `main()` parameters. 👍

packages/contract/src/contract-call-script/src/main.sw
```diff
@@ -1,34 +1,36 @@
...
+    // This line fixes a bug
```
adlerjohn, line 6: Can you clarify in this comment what bug? Is there an issue tracking it?

packages/providers/src/transaction-request/transaction-request.ts
```diff
@@ -35,6 +33,17 @@ import { witnessify } from './witness';
...
 export { TransactionType };
 
+// We can't import this from `@fuel-ts/script` because it causes
+// cyclic dependency errors so we duplicate it here.
+export const returnZeroScript: AbstractScript<void> = {
+  /*
+    Opcode::RET(REG_ZERO)
+    Opcode::NOOP
+  */
+  bytes: arrayify('0x24000000'),
```
adlerjohn, line 43: Hmmm. The encoding of instructions may (will) change in the future. Is there some way to make this more maintainable? Doesn't have to be in this PR.

Aside: `0x0000` isn't a `NOOP` IIRC, but rather an invalid instruction. Doesn't actually matter since it's unreachable.

packages/script/scripts/build-call-test-script.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -euo pipefail
+
+SCRIPT_PATH="src/call-test-script"
+BIN_DIR="$SCRIPT_PATH/out/debug"
+
+forc build -p $SCRIPT_PATH --print-finalized-asm
+# forc parse-bytecode "$BIN_DIR/call-test-script.bin" > "$BIN_DIR/call-test-script.txt"
```
adlerjohn, line 9: Remove debugging statement.

packages/script/src/call-test-script/Forc.lock
```diff
@@ -0,0 +1,12 @@
+[[package]]
```
adlerjohn, line 1: You probably don't want to commit the lockfile here, otherwise the stdlib version won't get bumped automatically when you bump the version of `forc` in CI.

packages/abi-coder/src/interface.ts
AlicanC: This is one of the things that will be covered in https://github.com/FuelLabs/fuels-ts/issues/268. I will add a TODO here.

packages/contract/scripts/build-contract-call-script.sh
AlicanC: I didn't forget this, but left it for the next dev to use. I like leaving useful stuff like this around.

Maybe I should put this behind a `FUELS_DEBUG` envvar to make that obvious?

packages/contract/src/contract-call-script/src/main.sw
AlicanC: Yeah, I love it! Hopefully will be in std soon: https://github.com/FuelLabs/sway/issues/1062

BUT, being able to have `main()` args and also forc generating JSON ABIs from scripts would be great.

packages/contract/src/contract-call-script/src/main.sw
AlicanC: I don't even know what the bug is :D 

First started realizing the bug here: https://fuellabs.slack.com/archives/C01KBNGSWGJ/p1652788684067789
And reduced it to "some kind of a padding issue" here: https://fuellabs.slack.com/archives/C01KBNGSWGJ/p1652799760675829

packages/providers/src/transaction-request/transaction-request.ts
AlicanC: Yeah this is why we are moving from these to the new script package and actually writing Sway. This is the only instance that was left.

Will leave a TODO for the issue: https://github.com/FuelLabs/fuels-ts/issues/281

AlicanC:
Changes since the last review round: https://github.com/FuelLabs/fuels-ts/pull/277/files/ecfc5e6b8e9d05d702aaead77156efe8b98021b3..ae5ce6069144ae37385c9898182ea7de9d6fc6c1

adlerjohn (approved):

PR #288, opened 2022-05-20
feat!: add Multicall

AlicanC:
This PR adds a new script called `Multicall` and exposes two new APIs that use it, `Contract.dryRunMulticall` and `Contract.submitMulticall`, for making multiple contract calls in a single transaction.

```ts
const [resultA, resultB] = await contract.submitMulticall([
  contract.prepareCall.foo(1337),
  contract.prepareCall.boo({ a: true, b: 1337 }),
]);
```


luizstacio:
@AlicanC would be great to add `await contract.simulateMulticall([`

luizstacio (approved):

PR #289, opened 2022-05-20
ci: add action for validate pull request

pedronauck:
## Problem

Since we are generating changelogs based on conventional commits, when we are squashing pull requests without the right title we are lost a lot of information when generating our changelogs.

## Solution

This pull request just adds a Github action that validates the title according to conventional commits and also checks lint and tests.


AlicanC (commented):
LGTM

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,50 @@
+on:
+  pull_request:
+    types: [opened, synchronize, ready_for_review, edited, closed]
```
luizstacio, line 3: I think we can remove closed here and add `reopened`
```
[opened, reopened , synchronize, ready_for_review, edited]
```

luizstacio, line 3: I think we can remove closed here and add `reopened`
```
[opened, reopened , synchronize, ready_for_review, edited]
```

luizstacio (changes requested):

QuinnLee (approved):

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,50 @@
...
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Lint & Test
+        run: |
+          pnpm build
+          pnpm lint
+          pnpm build
+          pnpm ci:test-coverage
```
adlerjohn, line 50: Why not simply add these to `ci.yml`? It looks like most of these two worksflow are the same.

pedronauck: I split it up because of the pull request title validation, thinking this shouldn't run on push master, but is it possible to put it all together? if is I think there's no problem put all together

.github/workflows/pr.yaml
pedronauck: done ✅

.github/workflows/pr.yaml
adlerjohn: Yes, you can do it together. Just add an `if` clause to the `Pull request title check` job that checks if it's a PR.

```yml
if github.event_name == 'pull_request'
```

should suffice.

adlerjohn: Alternatively, keep it as a separate workflow that only runs on PRs!

adlerjohn (changes requested):
https://github.com/FuelLabs/fuels-ts/pull/289#discussion_r880006067

adlerjohn (approved):
utACK

PR #294, opened 2022-05-26
chore: fix generated file and class names

### LINKED ISSUE #292: Fix TypeChain generated Class names
Before: 
`ExchangeContractAbi__factory

After:
`Exchange_contractAbi__factory`

Following up from changes introduced in #283

camsjams:
Fixes #292 

Previous version of `normalizeName` from TypeChain: https://github.com/dethcrypto/TypeChain/blob/2fcc74371ed0c2c587fb87e9ce0faa935e22942a/packages/typechain/src/parser/normalizeName.ts

Recently updated version of `normalizeName` from TypeChain: https://github.com/dethcrypto/TypeChain/blob/3a8a99ad4641af8a4dae498566ccd837e15d7034/packages/typechain/src/parser/normalizeName.ts

Luckily there was a copy already inside the fuels-ts repo

### DIFF
--- a/packages/typechain-target-fuels/src/parser/parseSvmTypes.test.ts
+++ b/packages/typechain-target-fuels/src/parser/parseSvmTypes.test.ts
@@ -1,4 +1,4 @@
-import { parseSvmType } from './parseSvmTypes';
+import { parseSvmType, normalizeName } from './parseSvmTypes';
 
 describe('parseSvmTypes', () => {
   it('it maps a raw type to type', () => {
@@ -49,5 +49,19 @@ describe('parseSvmTypes', () => {
         { name: 'bar', type: { type: 'u8', bits: 8, originalType: 'u8' } },
       ],
     });
+  });
+});
+
+describe('normalizeName', () => {
+  it('it should make nice names', () => {
+    expect(normalizeName('DsToken')).toEqual('DsToken');
+    expect(normalizeName('test')).toEqual('Test');
+    expect(normalizeName('ds-token')).toEqual('DsToken');
+    expect(normalizeName('ds_token')).toEqual('DsToken');
+    expect(normalizeName('Aaa_bbb_CCDD-EEE')).toEqual('AaaBbbCCDDEEE');
+    expect(normalizeName('ds token')).toEqual('DsToken');
+    expect(normalizeName('name.abi')).toEqual('NameAbi');
+    expect(normalizeName('1234name.abi')).toEqual('NameAbi');
+    expect(normalizeName('ERC20.abi')).toEqual('ERC20Abi');
   });
 });


packages/typechain-target-fuels/src/index.ts
```diff
@@ -3,12 +3,14 @@ import { readFileSync } from 'fs';
...
 import type { CodegenConfig, Config, FileDescription } from 'typechain';
-import { getFilename, normalizeName, TypeChainTarget } from 'typechain';
+import { getFilename, TypeChainTarget } from 'typechain';
+
 
 import { codegenAbstractContractFactory, codegenContractTypings } from './codegen';
 import { FACTORY_POSTFIX } from './common';
 import { extractAbi, extractDocumentation, parse } from './parser/abiParser';
 import type { Contract, RawAbiDefinition } from './parser/abiParser';
+import { normalizeName } from './parser/parseSvmTypes';
```
luizstacio, line 13: Running it locally here; Looks like it solved the class names but not file names, resulting in issues with imports.

luizstacio (changes requested):

packages/typechain-target-fuels/src/index.ts
QuinnLee: Yeah, I think the issue is https://github.com/FuelLabs/fuels-ts/blob/master/packages/typechain-target-fuels/src/parser/abiParser.ts#L191

QuinnLee (approved):

luizstacio:
> @luizstacio are your requested changes looking good?

I'm testing it using swayswap and because the name of the file has the underscore, looks like the `normalizeName` didn't try to improve the name and just use it. 

I just write a custom `normalizeName` and tested locally this one should work better for us;
```ts
function normalizeName(rawName: string): string {
  const transformations: ((s: string) => string)[] = [
    (s) => s.replace(/_/g, '-'), // replace "_" to '-'
    (s) => s.replace(/\s+/g, '-'), // spaces to - so later we can automatically convert them
    (s) => s.replace(/\./g, '-'), // replace "."
    (s) => s.replace(/-[a-z]/g, (match) => match.slice(-1).toUpperCase()), // delete '-' and capitalize the letter after them
    (s) => s.replace(/-/g, ''), // delete any '-' left
    (s) => s.replace(/^\d+/, ''), // removes leading digits
    (s) => s.replace(/[a-z]/, (match) => match.slice(-1).toUpperCase()), // upper case first letter
  ];

  const finalName = transformations.reduce((s, t) => t(s), rawName);

  if (finalName === '') {
    throw new Error(`Can't guess class name, please rename file: ${rawName}`);
  }

  return finalName;
}
```

luizstacio (approved):
Nice!

PR #298, opened 2022-05-27
feat: update bytes length to u64

luizstacio:


packages/transactions/src/coders/witness.ts
```diff
@@ -4,8 +4,8 @@ import { Coder, NumberCoder } from '@fuel-ts/abi-coder';
 import { ByteArrayCoder } from './byte-array';
 
 export type Witness = {
-  /** Length of witness data, in bytes (u16) */
-  dataLength: number;
+  /** Length of witness data, in bytes (u64) */
```
QuinnLee, line 7: I think `u32` is probably large enough

adlerjohn: This change is related to https://github.com/FuelLabs/fuel-specs/issues/298 and https://github.com/FuelLabs/fuel-tx/pull/130#discussion_r881818372

AlicanC: Note:

With `u32`s we will be off-spec but still behave on-spec and solve the immediate issue at SwaySwap and won't need to move to `bigint`s.

Also, even though these are `u64`s, their actual values will be in the `u32` range. I think we should have a follow-up issue to implement something so we can make these `u64`s but still have them be `number`s in JS (and not `bigint`s).

QuinnLee (approved):

PR #299, opened 2022-05-27
chore: update sway param calling

### LINKED ISSUE #231: Fix the handling of the remaining Sway types in and out of a contract method

camsjams:
Fixes #231 

Some of these param types are not working fully, and I found that passing by value for `str[1] - str[8]` was not working as expected (If I removed the padding for the pointer and passed by value, I'd get a `panic` receipt).

Here are the current state of the tests I've written, which may be the best outcome of what I was doing, unless these are duplicated elsewhere:
image

The code in [fuels-ts](https://github.com/FuelLabs/fuels-ts/blob/master/packages/abi-coder/src/json-abi.ts#L34-L47) (thanks @AlicanC :+1: ) mostly mirrors whats happening in [fuels-rs](https://github.com/FuelLabs/fuels-rs/blob/master/packages/fuels-contract/src/contract.rs#L368-L382) post https://github.com/FuelLabs/fuels-rs/pull/261

### DIFF
--- a/packages/coverage-contract/src/coverage-contract.test.ts
+++ b/packages/coverage-contract/src/coverage-contract.test.ts
@@ -68,7 +68,7 @@ describe('CoverageContract', () => {
     expect(result).toBe('fu');
   });
 
-  it.only('should test str[3] variable type', async () => {
+  it('should test str[3] variable type', async () => {
     const result = await contractInstance.functions.echo_str_3('fue');
     expect(result).toBe('fue');
   });


packages/coverage-contract/src/main.sw
```diff
@@ -0,0 +1,108 @@
...
+    fn echo_str_2(input: str[2]) -> str[2];
+    fn echo_str_3(input: str[3]) -> str[3];
+    fn echo_str_8(input: str[8]) -> str[8];
+    fn echo_str_9(input: str[9]) -> str[9];
+    fn echo_tuple_u8(input: (u8, u8)) -> (u8, u8);
+    fn echo_tuple_u64(input: (u64, u64)) -> (u64, u64);
+    fn echo_tuple_mixed(input: (bool, u64)) -> (bool, u64);
+    fn echo_array_u8(input: [u8; 2]) -> [u8; 2];
+    fn echo_array_u64(input: [u64; 5]) -> [u64; 5];
+    fn echo_array_bool(input: [bool; 2]) -> [bool; 2];
```
QuinnLee, line 37: Is an array of tuples, an array of strings, or nested arrays out of scope?

packages/coverage-contract/src/coverage-contract-types/factories/CoverageContractAbi__factory.ts
```diff
@@ -0,0 +1,451 @@
...
+      },
+    ],
+  },
+  {
+    type: "function",
+    inputs: [
+      {
+        name: "input",
+        type: "(u8, u8)",
+        components: null,
```
QuinnLee, line 216: Tuples have components now.  I think this needs to be regenerated

packages/coverage-contract/src/coverage-contract-types/factories/CoverageContractAbi__factory.ts
```diff
@@ -0,0 +1,451 @@
...
+      },
+    ],
+  },
+  {
+    type: "function",
+    inputs: [
+      {
+        name: "input",
+        type: "[u8; 2]",
+        components: null,
```
QuinnLee, line 270: same comment about tuples - I think arrays have components now and this should be regenerated.

PR #308, opened 2022-05-31
fix: use `isArray` in reduce

QuinnLee:


matt-user (approved):
Thanks, I will give this a shot and report back.

PR #310, opened 2022-05-31
fix: Use internal random bytes function to avoid browser ESM file

### LINKED ISSUE #309: Fix WARNING: Missing strong random number source
This looks to have been introduced with the updated build process, as the dist files generated by wallet are optimized for the browser explicitly.

The best fix would be to use the internal universal crypto -> random bytes: https://github.com/FuelLabs/fuels-ts/blob/master/packages/keystore/src/universal-crypto.ts

camsjams:
Fixes #309


luizstacio (changes requested):
I think we should export `Keystore` on `fuels` package. To be able to use randomBytes from `fuels`

luizstacio (approved):
LGTM

PR #324, opened 2022-06-09
fix: add angular-changelog and check for fuels-ts package.json

QuinnLee:
I was getting 
Screen Shot 2022-06-09 at 7 03 45 PM

adding the angular changelog seems to fix it

Also, update the package json update script to ignore directories like `forc`

### DIFF
--- a/scripts/utils/changePackages.ts
+++ b/scripts/utils/changePackages.ts
@@ -21,7 +21,9 @@ export async function changeAllPkgJSON(version: string) {
   const spinner = ora('Bumping versions').start();
 
   for (const dir of dirs) {
-    changePkgJsonVersion(`./packages/${dir}`, version);
+    if (dir.includes('fuels-ts')) {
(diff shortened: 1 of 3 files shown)


scripts/utils/changePackages.ts
```diff
@@ -21,7 +21,9 @@ export async function changeAllPkgJSON(version: string) {
   const spinner = ora('Bumping versions').start();
 
   for (const dir of dirs) {
-    changePkgJsonVersion(`./packages/${dir}`, version);
+    if (dir.includes('fuels-ts')) {
```
luizstacio, line 24: Why if requires this fuels-ts?

luizstacio (changes requested):

luizstacio (approved):

scripts/utils/changePackages.ts
QuinnLee: this is to avoid updating the `forc-bin` package.json file

PR #327, opened 2022-06-13
chore: update forc and refactor

QuinnLee:
Updating `forc`

Can someone confirm `force_transfer` -> `force_transfer_to_contract` is correct?

### DIFF
--- a/packages/contract/src/token-test-contract/token_contract/src/main.sw
+++ b/packages/contract/src/token-test-contract/token_contract/src/main.sw
@@ -13,7 +13,7 @@ impl Token for Contract {
     }
 
     fn force_transfer_coins(coins: u64, asset_id: ContractId, target: ContractId) {
-        force_transfer(coins, asset_id, target);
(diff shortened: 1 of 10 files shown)


packages/contract/src/token-test-contract/token_contract/src/main.sw
```diff
@@ -13,7 +13,7 @@ impl Token for Contract {
     }
 
     fn force_transfer_coins(coins: u64, asset_id: ContractId, target: ContractId) {
-        force_transfer(coins, asset_id, target);
+        force_transfer_to_contract(coins, asset_id, target);
     }
```
QuinnLee, line 18: Is this correct?

luizstacio: cc: @mohammadfawaz

mohammadfawaz: That is correct.

PR #331, opened 2022-06-14
feat: add predicate support

### LINKED ISSUE #304: Support for predicates
Tracking https://github.com/FuelLabs/fuels-rs/issues/341

The Fuel VM allows for a Coin or Message input to be locked behind a predicate. To spend one of these inputs, the predicate is provided as a field of the input. If hash(predicate) == Input.owner && eval(predicate) , the UTXO can be spent. Note that in this case, no signature is required as the predicate alone is sufficient to spend the UTXO.

The SDK will therefore need to be able to:

attach a compiled predicate to an Input.
attach predicate data to an input. (This is analagous to providing scriptdata to a script - the data are parameters to the main function)
send a transaction without the need for any wallet or signer
set an output's owner to the hash of a compiled predicate

camsjams:
Closes #304 

Will be enhanced once https://github.com/FuelLabs/sway/issues/1981 is completed

PR for Sway is ready for review: https://github.com/FuelLabs/sway/pull/1984


packages/predicate/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@fuel-ts/predicate",
+  "version": "0.6.0",
```
QuinnLee, line 3:
```suggestion
  "version": "0.7.0",
```

packages/predicate/src/test-predicate-basic/Forc.lock
```diff
@@ -0,0 +1,12 @@
+[[package]]
+name = 'core'
+dependencies = []
+
+[[package]]
+name = 'std'
+source = 'git+https://github.com/fuellabs/sway?tag=v0.14.5#60c626cf486d5c53c44d84db1ec81cb90174cad3'
```
adlerjohn, line 7: Not a fan of how this uses a hard-coded and old version of `forc`. But this is a repo-wide issue I think?

adlerjohn (commented):
As a general framework it looks fine. Will have to be replaced by the stdlib version once that's available as you noted.

packages/predicate/src/test-predicate-basic/Forc.lock
adlerjohn: It is what it is

adlerjohn (commented):
CI failing

adlerjohn:
In that case I would hold off on merging this PR until it's unblocked by stdlib inclusion of getting predicate data correctly. Don't want to have CI failing in `master`.

QuinnLee:
do the changes https://github.com/FuelLabs/fuels-ts/pull/344 conflict

luizstacio (changes requested):
Are the contracts present on the predicate only for tests? If yes, I think we should avoid committing the binaries and lock files, just the minimal files requested, as it should be built again on `build-time`

AlicanC (changes requested):
Requested changes are here: https://github.com/FuelLabs/fuels-ts/pull/349

luizstacio (approved):

AlicanC (approved):

PR #335, opened 2022-06-16
chore: forc update 0.16

QuinnLee:
Updating forc to `0.16`

### DIFF
--- a/packages/contract/src/storage-test-contract/src/main.sw
+++ b/packages/contract/src/storage-test-contract/src/main.sw
@@ -4,13 +4,18 @@ use std::storage::store;
 use std::storage::get;
 
 abi StorageTestContract {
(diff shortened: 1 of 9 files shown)


packages/contract/src/storage-test-contract/src/main.sw
```diff
@@ -4,13 +4,19 @@ use std::storage::store;
 use std::storage::get;
 
 abi StorageTestContract {
+  #[storage(write)]
   fn initialize_counter(value: u64) -> u64;
+  #[storage(write)]
+  #[storage(read)]
```
mohammadfawaz, line 10: These two annotations can be merged into `#[storage(read, write)]`

AlicanC (approved):

PR #336, opened 2022-06-17
fix: EnumCoder not padding its value

AlicanC:
EnumCoder needed to left-pad its value to the max length among all possible values, but that wasn't the case.

The actual changes are in [enum.ts](https://github.com/FuelLabs/fuels-ts/pull/336/files#diff-5cc0231e0e475c5030836a1e71284195e87dafe4e2d6875a707c27d61f276647) and the rest is for length calculation.

### DIFF
--- a/packages/transactions/src/coders/byte-array.ts
+++ b/packages/transactions/src/coders/byte-array.ts
@@ -2,8 +2,6 @@ import type { BytesLike } from '@ethersproject/bytes';
 import { arrayify, concat, hexlify } from '@ethersproject/bytes';
 import { Coder } from '@fuel-ts/abi-coder';
 
-const padToBytes = 8;
-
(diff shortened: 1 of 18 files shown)


luizstacio (dismissed):

packages/transactions/src/coders/byte-array.ts
```diff
@@ -25,9 +30,8 @@ export class ByteArrayCoder extends Coder<BytesLike, string> {
     const data = arrayify(value);
     parts.push(data);
     // Write padding
-    const pad = padToBytes - (this.length % padToBytes);
-    if (pad % padToBytes) {
-      parts.push(new Uint8Array(pad).fill(0));
```
QuinnLee, line 30: `padToBytes` on line 5 can be remove too

QuinnLee (approved):
👍🏼

PR #346, opened 2022-06-19
fix: enable prettier errors via eslint

helciofranco:
## Summary

- This PR adds a configuration necessary to enable prettier errors via eslint.
- Enable `eslint-plugin-prettier` (runs prettier as an ESLint rule).
- Keep `eslint-config-prettier` enabled (that only turns rules off).
- Fix `extends` array order, because it should always be the last on `extends` array (so it gets the chance to override other configs).
- Add `.eslintrc.js` to include files by `tsconfig.eslint.json` ([see reference](https://github.com/typescript-eslint/typescript-eslint/blob/b229ee4ace1417d2d4f7d6cafa039f073e845c3a/tsconfig.eslint.json#L10) from the official repo)
- Remove an unused import (from `packages/contract/src/util.ts`).

## Before (using only the `eslint-config-prettier`) ⚠️ 
before

## Now ✅
now

## Reference
- [eslint-config-prettier - Installation](https://github.com/prettier/eslint-config-prettier#installation)
- [eslint-plugin-prettier - Recommended Configuration](https://github.com/prettier/eslint-plugin-prettier#recommended-configuration)
- [tsconfig.eslint.json](https://github.com/typescript-eslint/typescript-eslint/blob/b229ee4ace1417d2d4f7d6cafa039f073e845c3a/tsconfig.eslint.json#L10)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,9 +11,9 @@ module.exports = {
     'airbnb-base',
     'airbnb-typescript/base',
     'plugin:@typescript-eslint/recommended',
-    'prettier',
     'plugin:eslint-comments/recommended',
     'plugin:@typescript-eslint/recommended',
+    'plugin:prettier/recommended',
   ],
   settings: {
     jsdoc: {
--- a/packages/contract/src/util.ts
+++ b/packages/contract/src/util.ts
@@ -1,7 +1,6 @@
 import type { BytesLike } from '@ethersproject/bytes';
 import { hexlify, arrayify, concat } from '@ethersproject/bytes';
 import { sha256 } from '@ethersproject/sha2';
-import { ZeroBytes32 } from '@fuel-ts/constants';
 import { calcRoot } from '@fuel-ts/merkle';
 
 const getContractRoot = (bytecode: Uint8Array): string => {
--- a/tsconfig.eslint.json
+++ b/tsconfig.eslint.json
@@ -5,5 +5,6 @@
     "types": ["@types/node"],
     "noEmit": true,
     "allowJs": true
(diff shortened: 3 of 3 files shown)


AlicanC:
Hey, thanks for the PR!

This behavior is intentional. We don't need to see Prettier errors in the editor since they are auto-fixed on save anyways. We avoid cluttering this way and offer a cleaner DX. This could be avoided in your PR by disabling the `prettier/prettier` rule in `.vscode/settings.json` but showing these errors seems to be the point of your PR so I vote to close.

helciofranco:
> Hey, thanks for the PR!
> 
> This behavior is intentional. We don't need to see Prettier errors in the editor since they are auto-fixed on save anyways. We avoid cluttering this way and offer a cleaner DX. This could be avoided in your PR by disabling the `prettier/prettier` rule in `.vscode/settings.json` but showing these errors seems to be the point of your PR so I vote to close.

They are auto-fixed if you use VSCode, but not if you use other editors. `.vscode/settings.json` is VSCode-only. 

If necessary keep it off, `eslint-plugin-prettier` can be removed from the the repo. It’s not used currently.

AlicanC:
> They are auto-fixed if you use VSCode, but not if you use other editors.

My assumption was that our VSC config was so simple that non-VSC devs could easily replicate it.

I do understand the issue though. With the change I've mentioned this PR would provide the same DX for VSC devs, while delivering better DX for non-VSC devs. (Then they could disable `prettier/prettier` and turn on auto-fix if they like.)

I've assigned myself too and will take a deeper look into this.

PR #348, opened 2022-06-20
feat!: make prepareCall return ContractCall

AlicanC:
This PR makes some necessary changes for https://github.com/FuelLabs/fuels-ts/pull/288. While it was possible to write cleaner code, I thought it would be wiser to not commit to any abstractions until at least #288 and #315 land so we have a clearer path.

The idea here is to formalize the concept of a contract call in a struct called `ContractCall`, and with #288 be able to execute multiples of these in a single tx.

---

[fix: remove types for unused props](https://github.com/FuelLabs/fuels-ts/pull/348/commits/6317ff96d2e853a63343bdf9c432e0b0c97d6b07)

We actually don't put these props on Interfaces.

[feat!: make prepareCall return ContractCall](https://github.com/FuelLabs/fuels-ts/pull/348/commits/b5b5d5d1c03c755c1e3f53ef46c3a183211cd329)

Now `prepareCall` returns a new struct named `ContractCall`. There are also other refactors around this.

### DIFF
--- a/packages/contract/src/contract-factory.test.ts
+++ b/packages/contract/src/contract-factory.test.ts
@@ -70,7 +70,6 @@ describe('Contract Factory', () => {
     );
 
     const dryRunResult = await contact.dryRunResult.increment_counter(1);
-    expect(dryRunResult.blockId).toBeFalsy();
     expect(dryRunResult).toEqual(
       expect.objectContaining({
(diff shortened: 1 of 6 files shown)


luizstacio (approved):
LGTM

PR #356, opened 2022-06-24
chore: allow tests to run in parallel

AlicanC:


luizstacio:
@AlicanC @camsjams was talking about other approaches also;

- Run a node for each switch of tests: @AlicanC said it was very power consumption and not really a good solution
- Have a queue to ask to manage the seeder was not try

I think for now this solution looks good

PR #358, opened 2022-06-25
chore: reorg CI config

AlicanC:
Currently we have a single action that both runs the tests and checks the PR title. They have to share their run triggers so tests rerun when the PR title changes for example. This PR splits those into two so that doesn't happen.

This PR also stops running tests when master is merged or before publishing. We don't need this since we can "require branches to be up to date before merging which ensures pull requests have been tested with the latest code".

### DIFF
--- a/.github/workflows/pr-test.yaml
+++ b/.github/workflows/pr-test.yaml
@@ -2,8 +2,6 @@ name: "Test PR"
 
 on:
   pull_request:
-    branches:
-      - "**" # matches every branch
 
 jobs:
   cancel-previous-run:
--- a/.github/workflows/pr-test.yaml
+++ b/.github/workflows/pr-test.yaml
@@ -1,11 +1,10 @@
-on:
-  push:
-    branches:
-      - master
-  pull_request:
-    types: [opened, reopened, synchronize, ready_for_review, edited]
+name: "Test PR"
 
-name: "Validate PR"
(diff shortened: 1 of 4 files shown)


AlicanC:
@adlerjohn could you please verify if the "Require branches to be up to date before merging." setting is enabled or enable it when this PR is otherwise approved?

.github/workflows/pr-test.yaml
```diff
@@ -1,11 +1,10 @@
+name: "Test PR"
+
 on:
-  push:
-    branches:
-      - master
```
adlerjohn, line 4: I would keep this for the sole reason that even with requiring branches to be up to date, things can slip through. CI minutes are free, so it's better to have this check than not.

.github/workflows/pr-test.yaml
```diff
@@ -1,11 +1,10 @@
...
   pull_request:
-    types: [opened, reopened, synchronize, ready_for_review, edited]
+    branches:
+      - "**" # matches every branch
```
adlerjohn, line 6: You don't need these lines. Without a branch specifier, it'll run on PRs against all branches.

adlerjohn:
The "Require branches to be up to date before merging" setting is now enabled.

adlerjohn (approved):

PR #361, opened 2022-06-27
chore: use `concurrency` to cancel previous CI runs

AlicanC:
From this SO page, it seems like the way we cancel previous runs is outdated and GitHub's native `concurrency` setting is the current practice: https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre

The current GH action we use haven't been touched for two years and the alternatives looked similar.

### DIFF
--- a/.github/workflows/pr-lint.yaml
+++ b/.github/workflows/pr-lint.yaml
@@ -7,6 +7,10 @@ on:
       - edited
       - synchronize
 
+concurrency:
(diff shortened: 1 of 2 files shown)


pedronauck (approved):

adlerjohn (approved):

adlerjohn:
@AlicanC can you also do this for our other repos? Good find.

AlicanC:
@adlerjohn, yup. Was planning to do that after seeing it work well here.

Issue #333, opened 2022-06-16
Improve Contract API methods


luizstacio:
Today the approach used on the contract API uses `namespaces` and `overrides` params to deliver multiple utilities like;
- Submit a contract call
- Simulate a contract call
- Run a `dry-run` call to a contract
- Build a transaction
- Override `gasPrice`, `bytePrice`, `gasLimite`, `forward coins`, etc...

The current namespaces are the following:
```ts
dryRun!: { [key: string]: any };
dryRunResult!: { [key: string]: any };
submit!: { [key: string]: any };
submitResult!: { [key: string]: any };
prepareCall!: { [key: string]: any };
simulate!: { [key: string]: any };
simulateResult!: { [key: string]: any };
```

The purpose is to move from the namespace approach to a chain o commands more near to the `fuels-rs API` achieving;
- Smaller learn curve between SDKs
- Better readability
- Better reusability of code on the app sides

### Examples:

Moving from:
```
await contract.submit.deposit({
  forward: [fromAmount, coinFrom.assetId],
});
```
To:
```
await contract
    .deposit()
    .forward([fromAmount, coinFrom.assetId])
    .submit();
```
or more similar from the `fuels-rs` would be;
```ts
await contract
    .deposit()
    .callParams([fromAmount, coinFrom.assetId])
    .call();
```

This makes it easier to read as the chain command is auto-explaining and also will provide good auto-complete using typescript.

One example of reusability is;
```
const request = contract
    .deposit()
    .forward([fromAmount, coinFrom.assetId]);
const result = request.simulte();

if (<something>) {
   const response = request.submit();
   return response;
}
```
Also instead of using `submitResult` to get the receipts would be possible to do something like;
```
contract
    .deposit()
    .forward([fromAmount, coinFrom.assetId])
    .submit()
    .receipts();
```

luizstacio:
Also thinking about multi-call, would be great to have something, like;

```ts
contract
    .deposit()
    .forward([fromAmount, coinFrom.assetId])
    .deposit()
    .forward([fromAmount, coinFrom.assetId])
    .add_liqudity()
    .forward([fromAmount, coinFrom.assetId])
    .submit()
    .receipts();
```

luizstacio:
With new changes on the contract with multi-call this got a bit more priority on my opinion, I will start a draft

PR #370, opened 2022-06-30
refactor: update contract api

luizstacio:
- [x] Changes on contract API #333 
- [x] Update typechain to new contract API
- [x] Fix tests
- [x] Update docs
- [ ] Improve typescript infer for ABI JSON (Didn't found a way to infer types from the JSON)


packages/contract/src/contract-new.ts
```diff
@@ -0,0 +1,273 @@
...
+    } else {
+      this.provider = null;
+      this.wallet = null;
+    }
+
+    Object.keys(this.interface.functions).forEach((name) => {
+      const fragment = this.interface.getFunction(name);
+      Object.defineProperty(this.functions, fragment.name, {
+        value: this.buildFunction(fragment),
+        writable: false,
```
QuinnLee, line 265: Not a blocker - Feels weird, i think we should have getter function to get `functions` and maybe make `this.functions`  private. But I'm not too opinionated on this

luizstacio: Didn't find a nice way to implement the methods with types directly in the class. Mainly to support types...

Would be possible to do it by add `[key: string]: any` but will lose the power to use the contract without generated types like:
```ts
// With this type you can inform the method expects a bigint and return a bigint.
const {value} = contract.functions.method<[bigint], bigint>(1).call();
```

QuinnLee (approved):

packages/contract/src/contracts/functions/base-invocation-scope.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  /**
+   * Submits a transaction to the blockchain.
+   *
+   * This is a final action and will spend the coins and change the state of the contract.
+   * It also means that invalid transactions will throw an error, and consume gas. To avoid this
+   * running invalid tx and consuming gas try to `simulate` first when possible.
+   */
+  async call<T = TReturn>(options?: CallOptions): Promise<FunctionInvocationResult<T>> {
+    assert(this.contract.wallet, 'Wallet is required!');
```
AlicanC, line 152: This naming here is odd. Assertions are used to check invariants, and are preferably removed from production code using something like [babel-plugin-unassert](https://github.com/unassert-js/babel-plugin-unassert).

An assertion would make sense here if, for example, it was impossible for a user to call `call` without a `this.contract.wallet`, but TypeScript didn't understand this so we still had to check to make TS happy. We don't have such a case here since it's very possible for a user to create a wallet-less Contract and call `call`. We should just do regular error handling here.

https://en.wikipedia.org/wiki/Assertion_(software_development)#Comparison_with_error_handling

packages/contract/src/contracts/functions/invocation-scope.ts
```diff
@@ -0,0 +1,56 @@
...
+    this.callParameters = callParams;
+
+    if (callParams?.forward) {
+      this.forward = coinQuantityfy(callParams.forward);
+      // Update transaction script with new forward params
+      this.setArguments(...this.args);
+    }
+
+    return this;
+  }
```
AlicanC, line 55: This should also handle `gasLimit` if we are to keep `forward`. (See my other comment on `forward`.)

AlicanC (commented):
Left some comments.

packages/contract/src/types.ts
```diff
@@ -0,0 +1,43 @@
...
+  fundTransaction: boolean;
+}>;
+
+export type CallConfig<T = unknown> = {
+  func: FunctionFragment;
+  contract: Contract;
+  callParameters?: CallParams;
+  txParameters?: TxParams;
+  forward?: CoinQuantity;
+  args: T;
```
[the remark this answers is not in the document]
luizstacio: - The forward in this case in the values passed on `call parameters` already converted to `CoinQuantity`, and `gasLimit` as its passed without any data transformation, we wire from `txParameters`.
- I think to build the call you don't need just a matter to save the inputted value.
- I'm not sure if I understand. But the spec is more relevant to the OP call, which is built on when calling: `this.transactionRequest.setScript`. line: `58`

packages/contract/src/contracts/functions/invocation-results.ts
```diff
@@ -0,0 +1,101 @@
...
+    this.gasUsed = getGasUsage(callResult);
+  }
+
+  protected getDecodedValue(callResult: CallResult) {
+    const encodedResults = contractCallScript.decodeCallResult(callResult);
+    const returnValues = encodedResults.map((encodedResult, i) => {
+      const { contract, func } = this.functionScopes[i].getCallConfig();
+      return contract.interface.decodeFunctionResult(func, encodedResult)?.[0];
+    });
+    return (this.isMultiCall ? returnValues : returnValues?.[0]) as T;
```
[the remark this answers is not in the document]
luizstacio: No, because if you mount a multi-call with a single call method, which can be useful in cases where you dynamically mount the number of calls. In this case, the SDK user would need to also check if the value is an array or not.

Knowing that is a multi-call or single makes it easier to have consistency. Also for the user is a transparent process. as the multi-call and single-call upper classes set this.

packages/contract/src/contracts/functions/base-invocation-scope.ts
luizstacio: This assertion is a custom method from `./utils` but maybe is a good idea to change the name.

packages/contract/src/contracts/functions/invocation-scope.ts
luizstacio: Makes sense, as the gas limit would be strict to the call method, would not make sense to the user to call `txParams`.
https://github.com/FuelLabs/fuels-ts/issues/379

packages/contract/src/contracts/functions/base-invocation-scope.ts
luizstacio: Also on the other point, it's not possible to `submit a valid tx` without a wallet instance, as we need it to add funds and pay the gas fee. The only time is possible to submit a tx without a wallet is with `dryRun` in this case we have other methods such as `dryRun` and `get`.

AlicanC: By possibility I specifically mean whether it's possible for the assertion in question to fail.

For example:
```ts
  async call<T = TReturn>(options?: CallOptions): Promise<FunctionInvocationResult<T>> {
    // We do regular error handling here because the user can create a Contract
    // without a wallet and call `call`.
    // In production, we can't remove this check. We need to properly notify the user
    // about what was wrong.
    if (!this.contract.wallet) {
      throw new Error('Wallet is required!');
    }
    
    return this.#call(options);
  }

  async #call<T = TReturn>(options?: CallOptions): Promise<FunctionInvocationResult<T>> {
    // We do an assertion here because even though TypeScript thinks `this.contract.wallet`
    // can be undefined, it is impossible for this to happen.
    // We call this method from only one place and we make this check already.
    // In production, we can safely remove this check.
    assert(this.contract.wallet, 'Expected a wallet.');

    await this.prepareTransaction(options);
    const response = await this.contract.wallet.sendTransaction(this.transactionRequest);

    return FunctionInvocationResult.build<T>(
      this.functionInvocationScopes,
      response,
      this.isMultiCall
    );
  }
  ```

PR #394, opened 2022-07-12
fix: adjust release to avoid npm auth issue

camsjams:


pedronauck (approved):

pedronauck:
I had the same problem on `fuel-ui` @camsjams 😅

PR #400, opened 2022-07-13
feat: enable users to pay for transactions

luizstacio:
This PR enables users to estimate the gas cost of a transaction and add this as gasLimit. Limiting the amount that can be expended on a single transaction.

This PR includes the following methods;
- `@fuel-ts/transactions`
  - `byteSize()` return the byte size of the transaction;
  - `chargeableByteSize()` return byte size of the tx - byte size of witness;
  - Update: `calculateFee()` to return a Also the `assetId`;
  - Update `transaction-reponse` add `gasUsed` on the returned response;
 - `@fuel-ts/provider`
   - `getTransactionCost(transaction)` return a object with important information about the cost of tx like;
      - minGasPrice, minBytePrice, gasPrice, bytePrice, byteSize, gasUsed, fee;
   - `getNodeInfo()` return the information about the node like `minGasPrice, minBytePrice`;
   - Update: `sendTransaction` before submit we call `getTransactionCost` and check if gasPrice, bytePrice and gasLimit. Meets the required amounts of tx minGasPrice, minBytePrice and gasUsed. **This prevents tx from failure after submitting and expending coins**. 
- `@fuel-ts/contract`
  - `getTransactionCost()` return the transaction cost of the invocation scope.
  
 ## Use case example
 
 ```ts
// Users will mount a transaction
// that can be single or multi-call
const scope = await contract.functions
      .return_context_amount()
      .callParams({
        forward: [1_000_000, NativeAssetId],
      });

// User can estimate and show on screen how much
// the transaction will cost
const {
        bytePrice,
        gasPrice,
        gasUsed: gasLimit,
        fee
} = await scope.getTransactionCost();
console.log('Network fee in NativeCoins', fee);

// User can set the gasLimit, bytePrice and gasPrice
// with the estimated cost
const { value } = await scope
      .txParams({
        bytePrice,
        gasPrice,
        gasLimit,
      })
      .call<bigint>();
 ```
 
 This approach was discussed here: https://github.com/FuelLabs/fuels-ts/discussions/403


packages/providers/src/transaction-request/transaction-request.ts
```diff
@@ -298,11 +311,33 @@ abstract class BaseTransactionRequest implements BaseTransactionRequestLike {
...
+  }
+
+  /**
+   * Return the minimum amount in native coins required to create
+   * a transaction.
+   *
+   * Note: this is required even if the gasPrice and bytePrice
+   * are set to zero.
+   */
+  calculateFee(): CoinQuantity {
```
QuinnLee, line 330: 👍🏼

packages/providers/src/util.ts
```diff
@@ -1,7 +1,16 @@
...
-import { InputType, OutputType, TransactionType, TransactionCoder } from '@fuel-ts/transactions';
+import {
+  ReceiptType,
+  InputType,
+  OutputType,
+  TransactionType,
+  TransactionCoder,
+} from '@fuel-ts/transactions';
+
+import type Provider from './provider';
```
QuinnLee, line 12: probably can remove

packages/math/src/math.ts
```diff
@@ -0,0 +1,9 @@
+import { toNumber } from './convert';
```
QuinnLee, line 1: should we flush out other math functions we may need? like `div`, `sub`, `add` and etc?

QuinnLee (commented):
lgtm! just minor stuff

packages/math/src/math.ts
luizstacio: I add multiply and max

QuinnLee (dismissed):

QuinnLee (approved):

packages/contract/src/__test__/contract.test.ts
```diff
@@ -308,4 +309,127 @@ describe('Contract', () => {
...
+        bytePrice: 1,
+      });
+    // Get transaction cost using bytePrice and gasPrice from
+    // invocation scope
+    const transactionCost = await invocationScope.getTransactionCost();
+
+    expect(transactionCost.bytePrice).toBe(1n);
+    expect(transactionCost.gasPrice).toBe(1n);
+    expect(transactionCost.fee).toBeGreaterThanOrEqual(2n);
+    expect(transactionCost.gasUsed).toBeGreaterThan(1000n);
```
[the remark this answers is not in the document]
luizstacio: Really nice never use like this.

QuinnLee (approved):

PR #410, opened 2022-07-15
fix: run forc-bin build before other packages

### LINKED ISSUE #408: `pnpm build` in CI is failing

0.7.0

### Steps to Reproduce

- Open a PR

You'll see that sometimes the PR validation steps fail, but sometimes not.
There's an intermittent error in CI build process. 

After upgrading `forc` to 0.18.1, it's happening more frequently.

### Expected Behavior

pnpm build in CI should work

### Actual Behavior

pnpm build in CI is failing intermittently.

LuizAsFight:
- fix `pnpm build` broken when running first time after download repo in linux environment (including CI)
- run `forc-bin` build before other packages

Fix #408

### DIFF
--- a/packages/forc-bin/package.json
+++ b/packages/forc-bin/package.json
@@ -20,6 +20,7 @@
   ],
   "license": "Apache-2.0",
   "scripts": {
+    "preinstall": "pnpm setup",
     "install": "./dist/install.js",
     "setup": "tsup --dts",
     "update": "./dist/update.js && ./dist/install.js"
--- a/packages/forc-bin/package.json
+++ b/packages/forc-bin/package.json
@@ -20,9 +20,8 @@
   ],
   "license": "Apache-2.0",
   "scripts": {
-    "preinstall": "pnpm setup",
+    "preinstall": "tsup --dts",
     "install": "./dist/install.js",
-    "setup": "tsup --dts",
     "update": "./dist/update.js && ./dist/install.js"
   },
   "bin": {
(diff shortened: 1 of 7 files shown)


packages/forc-bin/package.json
```diff
@@ -22,6 +22,7 @@
   "scripts": {
     "preinstall": "pnpm build",
     "install": "./dist/install.js",
+    "setup": "pnpm build",
```
luizstacio, line 25: here we have two scripts equal, and we are calling both on. We should have only one...
```
"prebuild": {
      "dependsOn": ["^prebuild", "setup"]
    },
```

luizstacio (changes requested):

turbo.json
```diff
@@ -1,7 +1,10 @@
 {
   "$schema": "https://turborepo.org/schema.json",
   "pipeline": {
-    "prebuild": {},
+    "preinstall": {},
+    "prebuild": {
+      "dependsOn": ["^prebuild", "preinstall"]
+    },
     "build": {
       "dependsOn": ["^build", "prebuild"],
```
luizstacio, line 9: we should add preinstall here?

luizstacio (changes requested):

turbo.json
LuizAsFight: yes, some prebuild commands require forc-bin to be ready already

luizstacio (approved):

PR #418, opened 2022-07-25
feat(ABI): add support to generic types 

### LINKED ISSUE #402: Add support for generic structs/enums from the JSON ABI
### Motivation

The JSON ABI now includes a field `typeArguments` alongside with `components` for supporting dynamic structs/enums. 

We should implement these changes on the `ABI parser` and also on `typechain` if needed.

Refs:
https://github.com/FuelLabs/fuel-specs/issues/381
https://github.com/FuelLabs/sway/issues/1993
It was released in sway `0.18.0`
https://github.com/FuelLabs/sway/releases/tag/v0.18.0
https://github.com/FuelLabs/sway/pull/2218

### LINKED ISSUE #433: Support "arrays of Structs" in contract method arguments in ABI

0.8

### Steps to Reproduce

When trying to get sighash for a method like this:
```
abi MyContract {
  fn generic_type_function(
    arg1: [MyStruct; 3],
  )
}
```

### Expected Behavior

result should be like: `a[s(...);3]`

### Actual Behavior

result is: `a[struct MyStruct;3]`

LuizAsFight:
- Created our own `ParamType` as fuel abi is getting more and more different from ETH abi. -> Inspired by `ethersproject/abi/src.ts/fragments.ts`. Closes #402 
- Add support to Arrays of Structs in contract methods arguments from ABI. Closes #433 

We're blocked in typechain, it will be handled in #430


packages/contract/src/__test__/generic-types-contract/generic-types-contract.test.ts
```diff
@@ -0,0 +1,30 @@
...
+      .generic_type_function({
+        bim: b256,
+      })
+      .call();
+
+    expect(value).toEqual(b256);
+  });
+});
+
+// correct ->     complex_function(s<a[b256;3],u8>(a[b256;3],e<u64>(u64,bool)))
```
luizstacio, line 29: Can we add more details to this comment?
It's hard to understand the propose of the comment

luizstacio (changes requested):

packages/contract/src/__test__/call-test-contract/call-test-contract.test.ts
```diff
@@ -275,4 +275,22 @@ describe('CallTestContract', () => {
     // tx execution context
     await expectContractCall();
   });
+
+  // TODO: We should support array arguments, then unskip this test
```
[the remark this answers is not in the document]
QuinnLee: I'll open an issue.  I have a branch with some fixes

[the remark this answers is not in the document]
QuinnLee: https://github.com/FuelLabs/fuels-ts/issues/420

packages/abi-coder/src/fragments/param-type.ts
```diff
@@ -0,0 +1,175 @@
...
+      indexed: value.indexed !== undefined ? Boolean(value.indexed) : undefined,
+      components: value.components ? value.components.map(ParamType.fromObject) : undefined,
+      typeArguments: value.typeArguments
+        ? value.typeArguments.map(ParamType.fromObject)
+        : undefined,
+    });
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  static isParamType(value?: any): value is ParamType {
```
luizstacio, line 172: Can we change from `any` to `?: ParamType`? or `value?:  { isParamType: boolean }`?

packages/abi-coder/src/fragments/param-type.ts
```diff
@@ -0,0 +1,175 @@
...
+    Object.freeze(this);
+  }
+
+  getSignaturePrefix(): string {
+    if (this.type) {
+      const arrayMatch = arrayRegEx.exec(this.type)?.groups;
+      const structMatch = structRegEx.exec(this.type)?.groups;
+      const enumMatch = enumRegEx.exec(this.type)?.groups;
+
+      // TODO: return 'a' after this PR is merged
```
luizstacio, line 111: Just a reminder to add more details in the comments. To reduce required context knowledge.

packages/abi-coder/src/fragments/param-type.ts
```diff
@@ -0,0 +1,175 @@
+import { Logger } from '@ethersproject/logger';
+import { defineReadOnly } from '@ethersproject/properties';
+
+import { arrayRegEx, enumRegEx, structRegEx } from '../abi-coder';
+
+const version = 'abi/5.6.4';
```
luizstacio, line 6: The version here is the version of the current package; in this case, `v0.10.0` you could query from the `package.json`.

packages/abi-coder/src/fragments/param-type.ts
```diff
@@ -0,0 +1,175 @@
...
+  readonly baseType?: string;
+  readonly indexed?: boolean;
+  readonly components?: Array<ParamType>;
+  readonly typeArguments?: Array<ParamType>;
+  readonly arrayLength?: number;
+  readonly arrayChildren?: ParamType;
+  readonly isParamType?: boolean;
+
+  constructor(constructorGuard: typeof CONSTRUCTOR_GUARD, params: ParamTypeProps) {
+    if (constructorGuard !== CONSTRUCTOR_GUARD) {
```
luizstacio, line 72: This is a very used approach on the ETH lib, we normally don't implement it, why do you think is a good choice to add this?

packages/abi-coder/src/fragments/param-type.ts
```diff
@@ -0,0 +1,175 @@
...
+  readonly baseType?: string;
+
+  // Indexable Paramters ONLY (otherwise null)
+  readonly indexed?: boolean;
+
+  // Tuples ONLY: (otherwise null)
+  //  - sub-components
+  readonly components?: Array<ParamType>;
+
+  // Tuples ONLY: (otherwise null)
```
luizstacio, line 47: I think the comment is not up to date with `typeArguments`.

packages/abi-coder/src/fragments/param-type.ts
```diff
@@ -0,0 +1,175 @@
...
+
+export interface ParamTypeProps {
+  // The local name of the parameter (of null if unbound)
+  readonly name?: string;
+
+  // The fully qualified type (e.g. "address", "tuple(address)", "uint256[3][]"
+  readonly type: string;
+
+  // The base type (e.g. "address", "tuple", "array")
+  readonly baseType?: string;
```
luizstacio, line 38: I think we have the opportunity now to transform BaseType into an Enum. Maybe we should open a issue for this.

packages/abi-coder/src/fragments/param-type.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  // Indexable Paramters ONLY (otherwise null)
+  readonly indexed?: boolean;
```
luizstacio, line 41: What is an indexed parameter? Do we support it?

packages/abi-coder/src/fragments/param-type.ts
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    if (this.baseType === 'array' && this.arrayChildren) {
+      return `${this.arrayChildren.getSighashSignature()}[${
+        this.arrayLength && this.arrayLength >= 0 ? String(this.arrayLength) : ''
+      }]`;
+    }
+
+    if (this.baseType === 'tuple' && this.components) {
+      const typeArgumentsSignature = this.typeArguments
```
luizstacio, line 135: For safer check we can change this to `Array.isArray(this.typeArguments)`.

packages/abi-coder/src/fragments/param-type.ts
```diff
@@ -0,0 +1,175 @@
...
+    this.isParamType = true;
+
+    Object.freeze(this);
+  }
+
+  getSignaturePrefix(): string {
+    if (this.type) {
+      const arrayMatch = arrayRegEx.exec(this.type)?.groups;
+      const structMatch = structRegEx.exec(this.type)?.groups;
+      const enumMatch = enumRegEx.exec(this.type)?.groups;
```
luizstacio, line 109: Can you just use `.test`? As we just want to see if it returns true.

packages/abi-coder/src/fragments/param-type.ts
```diff
@@ -0,0 +1,175 @@
...
+  constructor(constructorGuard: typeof CONSTRUCTOR_GUARD, params: ParamTypeProps) {
+    if (constructorGuard !== CONSTRUCTOR_GUARD) {
+      abiLogger.throwError('use fromString', Logger.errors.UNSUPPORTED_OPERATION, {
+        operation: 'new ParamType()',
+      });
+    }
+    populate(this, params);
+
+    const match = this.type.match(paramTypeArray);
+    if (match && match[1]) {
```
luizstacio, line 80: Maybe invert the if and inside the else use destruct on the match to add more sense?

```ts
const [type, size] = match;
```

luizstacio (changes requested):

packages/abi-coder/src/fragments/param-type.ts
LuizAsFight: I used as I see other places in the repo use this same logger pattern.
`const logger = new Logger(process.env.BUILD_VERSION || '~');`,

packages/abi-coder/src/fragments/param-type.ts
LuizAsFight: This approach is good for safety of classes. As it's "impossible" to have same reference of  `CONSTRUCTOR_GUARD` object outside of this file, creating this class only works using `ParamType.fromObject` method.

This blocks to extend ParamType and try to do something like
```
class HackyParamType extends ParamType {
  constructor(params) {
    super(params);

    // any hacky code
  }
}
```

Combining this with `Object.freeze(this);` in the end of constructor, makes the class `immutable and unextendable`.

LuizAsFight: for this class, I will remove the safety `CONSTRUCTOR_GUARD` part, as it just exports methods signatures. But should consider this approach to other more criticals classes that handle things like prices, bytePrices, gasLimit etc..

packages/abi-coder/src/interface.test.ts
```diff
@@ -256,8 +257,52 @@ describe('Interface', () => {
...
+              },
+            ],
+          },
+        ],
+        outputs: [{ name: 'ret', type: 'u64' }],
+      },
+    ]);
+
+    expect(functionInterface.getFunction('entry_one').getInputsSignature()).toEqual(
+      'entry_one(u64,s<b256,bool>(bool,u64))'
```
QuinnLee, line 305: 👍🏼

packages/abi-coder/src/fragments/param-type.ts
LuizAsFight: This indexed is something that came from ETH. We don't have it in our ABI structure. Removing it.

Ref: https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md

packages/abi-coder/src/fragments/param-type.ts
LuizAsFight: as mentioned in prev comment, refactored this to use `stringRegEx` instead

packages/abi-coder/src/interface.test.ts
LuizAsFight: nice that we have your PR with array/enum already merged. Fixes are included here, plus some more tests specifics to array/enum in dynamic typing structure.

Plus fix of #433  (which I didn't notice before when reviewing your PR)

packages/abi-coder/src/interface.test.ts
```diff
@@ -1,6 +1,7 @@
 import { hexlify } from '@ethersproject/bytes';
 
 import FunctionFragment from './fragments/function-fragment';
+import { ParamType } from './fragments/param-type';
```
luizstacio, line 4: I think this can be removed

luizstacio (changes requested):

luizstacio (approved):

PR #437, opened 2022-08-04
fix: fuel-ts packages linking to node_modules

### LINKED ISSUE #286: Improve `@fuel-ts` packages references inside `fuels-ts`
When clicking on `from "@fuel-ts/<package>"`. The link opens the `node_modules` folder. But it should open the package folder.

PR #238 used the typescript directly, which was good, but `pnpm link` was broken.
PR #246 fixed the issue with `pnpm link`, but the correct track of the package was lost opening the node_modules.

We should have both
- Good linking
- `pnpm link` working

LuizAsFight:
- Fixed linking packages to inside `node_modules` folder 
- Remove old Lerna config

Closes #286

### DIFF
--- a/.github/actions/ci-setup/action.yaml
+++ b/.github/actions/ci-setup/action.yaml
@@ -5,7 +5,7 @@ inputs:
     default: 16.15
   pnpm-version:
     description: "PNPM version"
-    default: 7.5.1
+    default: 7.8.0
 runs:
   using: "composite"
   steps:
--- a/.changeset/khaki-fishes-allow.md
+++ b/.changeset/khaki-fishes-allow.md
@@ -1,0 +1,5 @@
+---
+"forc-bin": patch
+---
+
+Update forc version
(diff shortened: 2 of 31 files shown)


QuinnLee (approved):

.github/actions/ci-setup/action.yaml
```diff
@@ -5,7 +5,7 @@ inputs:
     default: 16.15
   pnpm-version:
     description: "PNPM version"
-    default: 7.5.1
+    default: 7.8.0
```
pedronauck, line 8: You could use `default: latest` instead maybe, no?!

LuizAsFight: I think it's better to keep specific version because
- setups should be as stable as possible
- we're using pnpm version in cache key for CI 
```
key: ${{ runner.os }}-node-${{ inputs.node-version }}-pnpm-${{ inputs.pnpm-version }}-lock-${{ hashFiles('**/pnpm-lock.yaml') }}
```

We don't wanna unexpected problems out from nothing, sometimes spending time to debug and understand that the problems was with `pnpm` version specifically. 
From times to times we should upgrade it to easier identify breaking changes if there are any

PR #441, opened 2022-08-05
feat!: bech32

camsjams:
I refactored this a couple of times after looking through the fuels-rs version of the `Bech32` implementation.

### What first?

I think it make sense to first look over these two files before diving into all of the changes:
Address: https://github.com/FuelLabs/fuels-ts/blob/feature/bech32/packages/address/src/address.ts
Address Utils: https://github.com/FuelLabs/fuels-ts/blob/feature/bech32/packages/address/src/utils.ts

What is nice about the Rust version is that the `Bech32` value is a container rather than just a string. This was what I eventually realized made the most sense as this allows a little more clarity of what kind of `Address` you are handling.

There are still some places where an address is an ambiguous `string` or `BytesLike` which I opted to leave alone for now, but open to also refactoring them as part of this issue.

### DIFF
--- a/packages/interfaces/src/index.ts
+++ b/packages/interfaces/src/index.ts
@@ -26,3 +26,9 @@ export abstract class AbstractWallet {
 export type AddressLike = AbstractAddress | AbstractWallet;
 
 export type ContractIdLike = AbstractAddress | AbstractContract;
+
+export abstract class AbstractPredicate {
+  abstract bytes: Uint8Array;
+  abstract address: AbstractAddress;
+  abstract types?: ReadonlyArray<any>;
+}
--- a/packages/predicate/package.json
+++ b/packages/predicate/package.json
@@ -25,14 +25,14 @@
   "dependencies": {
     "@fuel-ts/abi-coder": "workspace:*",
     "@fuel-ts/address": "workspace:*",
-    "@fuel-ts/constants": "workspace:*",
     "@fuel-ts/contract": "workspace:*",
     "@fuel-ts/interfaces": "workspace:*",
-    "@fuel-ts/keystore": "workspace:*",
-    "@fuel-ts/math": "workspace:*",
-    "@fuel-ts/providers": "workspace:*",
     "@fuel-ts/wallet": "workspace:*",
-    "@fuel-ts/transactions": "workspace:*",
     "@ethersproject/bytes": "^5.5.0"
+  },
+  "devDependencies": {
+    "@fuel-ts/constants": "workspace:*",
+    "@fuel-ts/math": "workspace:*",
+    "@fuel-ts/providers": "workspace:*"
   }
 }
--- a/packages/predicate/src/predicate.test.ts
+++ b/packages/predicate/src/predicate.test.ts
@@ -1,5 +1,6 @@
 import { Address } from '@fuel-ts/address';
 import { NativeAssetId } from '@fuel-ts/constants';
+import type { AbstractAddress } from '@fuel-ts/interfaces';
 import type { BigNumberish } from '@fuel-ts/math';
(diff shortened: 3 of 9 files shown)


packages/address/src/address.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (!isBech32(this.bech32Address)) {
+      logger.throwArgumentError('Invalid Bech32 Address', 'address', address);
+    }
+  }
+
+  get address(): Bech32Address {
+    return this.bech32Address;
+  }
+
+  get b256Address(): B256Address {
```
luizstacio, line 34: I think it would be good to export a `toB256 ` and `toBytes` instead of setting getting properties.

packages/address/src/address.ts
```diff
@@ -0,0 +1,71 @@
...
+    return toB256(this.bech32Address);
+  }
+
+  get byteAddress(): Uint8Array {
```
luizstacio, line 38: It would be great adding some doc comments.

packages/address/src/address.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  equals(other: Address): boolean {
+    return this.bech32Address === other.bech32Address;
+  }
+}
+
+/**
+ * Takes a Public Key, hashes it, and creates an Address wrapper
+ */
+export function fromPublicKey(publicKey: string): Address {
```
luizstacio, line 54: I think we could implement these functions inside the Address class as static members.

packages/address/src/utils.ts
```diff
@@ -0,0 +1,87 @@
...
+ * Decodes a Bech32 address string into Decoded
+ */
+export function fromBech32(address: Bech32Address): Decoded {
+  return bech32m.decode(address);
+}
+
+/**
+ * Converts a B256 address string into Bech32
+ */
+export function toBech32(address: B256Address): Bech32Address {
```
luizstacio, line 31: Why enforce the type B256Address, Bech32Address? I think just string would make easear.

packages/contract/src/contracts/contract.ts
```diff
@@ -10,19 +12,23 @@ import { FunctionInvocationScope } from './functions/invocation-scope';
...
 
   constructor(
-    id: string,
+    id: string | AbstractAddress,
     abi: JsonAbi | Interface,
     walletOrProvider: Wallet | Provider | null = null
   ) {
     this.interface = abi instanceof Interface ? abi : new Interface(abi);
-    this.id = id;
+    if (id instanceof AbstractAddress) {
```
luizstacio, line 27: Maybe it's easier to check if it is a type of string.

packages/contract/src/contracts/contract.ts
```diff
@@ -10,19 +12,23 @@ import { FunctionInvocationScope } from './functions/invocation-scope';
...
+      this.id = id;
+    } else {
+      this.id = isBech32(id) ? new Address(id) : fromB256(id);
```
luizstacio, line 30: I think this can be something very common maybe we should abstract it inside the Address class.

```
Address.fromAny(string); // Something like this hehe
```

luizstacio (changes requested):

luizstacio (changes requested):

packages/address/src/utils.ts
luizstacio: Makes sense I think this would help to enforce it: https://github.com/FuelLabs/fuels-ts/pull/441#discussion_r948255727

luizstacio (dismissed):

luizstacio (approved):

QuinnLee (approved):

PR #443, opened 2022-08-06
chore: test return data larger than a `WORD`, test empty `enum` properties

### LINKED ISSUE #145: Handle return data larger than a `WORD`
Initially discovered here: https://github.com/FuelLabs/fuels-rs/pull/112 

> When the returned data from a contract call is larger than a word,
it isn't stored in Return.val, but in ReturnData.data.
> 
> Before the changes in here, we were pulling the data only from
> Return.val, which fortunately handled most cases. Until now,
> with some methods returning, for instance, a b256, which would cause
> the data to not be in Return.val as we expected, leading to errors.

The same issue should be happening in `fuels-ts` as well.

### LINKED ISSUE #234: feat: support "empty" enums
```
enum SmallEnum {
    v1: (),
    v2: (),
}
```

tracking https://github.com/FuelLabs/fuels-rs/issues/231

camsjams:
Fixes #145 
Fixes #234

### DIFF
--- a/.changeset/khaki-fishes-allow.md
+++ b/.changeset/khaki-fishes-allow.md
@@ -1,5 +1,4 @@
 ---
-"forc-bin": patch
 ---
 
 Update forc version
--- a/.changeset/wild-ants-tie.md
+++ b/.changeset/wild-ants-tie.md
@@ -1,0 +1,4 @@
+---
+---
+
+Create one single aggregated release for fuels repository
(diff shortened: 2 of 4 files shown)


packages/contract/src/__test__/coverage-contract/src/main.sw
```diff
@@ -0,0 +1,172 @@
+contract;
```
luizstacio, line 1: 💚💚💚

luizstacio (approved):

QuinnLee (approved):

PR #446, opened 2022-08-08
fix: update required coins on single call

luizstacio:


QuinnLee (approved):

QuinnLee:
do we need a follow up for multi-calls?

luizstacio:
> do we need a follow up for multi-calls?

No, multi-call was working fine, issue was just on single.

LuizAsFight (approved):

PR #471, opened 2022-08-26
feat!: add simpler Option style

### LINKED ISSUE #470: Add support to Option<T>
## Description

Today if a contract uses the sway object `Option<T>` as a return or argument of a function, we parse it as a custom Enum from the ABI, requiring the user to take action and pass an object `{ None: [] }` or `{ Some: <value> }` to match the custom enum.

The ideal implementation of `Option<T>` on typescript would be to abstract the `enum Option` and;
- In case of argument: simplify the argument to be the valid value (`T`) or optional in this case null or undefined.
- On the return: we should return the value `T` or null/undefined.

Todos;
- [ ] Implement an abstraction for the `Option<T>` by identifying it from the json ABI.
- [ ] Add the abstraction type inside the `typechain` to match the implementation.

### Contract

```
contract;

use std::{option::Option};

abi MyContract {
  fn test_option(param: Option<u32>) -> Option<u32>;
}

  fn test_option(param: Option<u32>) -> Option<u32> {
    param
  }
}
```

### Current way on Typescript

```ts
// When calling the contract, we require the user to 
// create an object { None: [] } for no params and { Some: 12 }
// when it will provide the value.
const { value } = await contract.functions
  .test_option({
    None: [],
  })
  .call();
// value will be {  None: [] }
const { value } = await contract.functions
  .test_option({
    Some: 12,
  })
  .call();
// value will be {  Some: 12 }
```

### Desired implementation on Typescript

```ts
// When calling the contract that has a params Option<T>, 
// we can make the param optional and create the underlining abstraction
// to create the correct params object { None: [] } or { Some: 12 }
const { value } = await contract.functions
  .test_option()
  .call();
// value will be null
const { value } = await contract.functions
  .test_option(12)
  .call();
// value will be 12
```

camsjams:
Closes #470

- Adds Option support
- Adds tests for Option
- Updates Typechain types to use new Option style

### DIFF
--- a/.changeset/tidy-parrots-care.md
+++ b/.changeset/tidy-parrots-care.md
@@ -1,0 +1,6 @@
+---
+"@fuel-ts/providers": minor
+"@fuel-ts/wallet": minor
+---
+
+Add `excludeId` to getCoinsToSpend
--- a/packages/providers/src/operations.graphql
+++ b/packages/providers/src/operations.graphql
@@ -208,8 +208,14 @@ query getCoinsToSpend(
   $owner: Address!
   $spendQuery: [SpendQueryElementInput!]!
   $maxInputs: Int
+  $excludedIds: [UtxoId!]
 ) {
-  coinsToSpend(owner: $owner, spendQuery: $spendQuery, maxInputs: $maxInputs) {
+  coinsToSpend(
+    owner: $owner
+    spendQuery: $spendQuery
+    maxInputs: $maxInputs
+    excludedIds: $excludedIds
+  ) {
     ...coinFragment
   }
 }
--- a/packages/providers/src/provider.ts
+++ b/packages/providers/src/provider.ts
@@ -371,7 +371,9 @@ export default class Provider {
     /** The quantitites to get */
     quantities: CoinQuantityLike[],
     /** Maximum number of coins to return */
-    maxInputs?: number
+    maxInputs?: number,
+    /** IDs of coins to exclude */
+    excludedIds?: BytesLike[]
   ): Promise<Coin[]> {
     const result = await this.operations.getCoinsToSpend({
       owner: owner.toB256(),
@@ -380,6 +382,7 @@ export default class Provider {
         amount: quantity.amount.toString(),
       })),
       maxInputs,
+      excludedIds: excludedIds?.map((id) => hexlify(id)),
     });
 
     const coins = result.coinsToSpend;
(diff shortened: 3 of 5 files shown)


packages/contract/src/__test__/coverage-contract/coverage-contract.test.ts
```diff
@@ -175,4 +177,24 @@ describe('Coverage Contract', () => {
...
+  it('should test Option<u32> extraction', async () => {
+    const INPUT_SOME = 123;
+    const { value: Some } = await contractInstance.functions
+      .echo_option_extract_u32(INPUT_SOME)
+      .call();
+    expect(Some).toStrictEqual(INPUT_SOME);
+
+    const INPUT_NONE = undefined;
+    const { value: None } = await contractInstance.functions
+      .echo_option_extract_u32(INPUT_NONE)
```
luizstacio, line 196: On my opinion it should be possible to execute this `echo_option_extract_u32()`

luizstacio (changes requested):
I think we should also include typechain on this change.

packages/abi-coder/src/abi-coder.ts
```diff
@@ -82,7 +83,9 @@ export default class AbiCoder {
         obj[component.name] = this.getCoder(component);
         return obj;
       }, {});
-      return new EnumCoder(enumMatch.name, coders);
+      return param.type === 'enum Option'
```
QuinnLee, line 86: I was trying to think of a better way to de-nest this but couldn't.  My gut tells me we should refactor `getCoder` but probably not a blocker for this pr

packages/abi-coder/src/abi-coder.ts
```diff
@@ -82,7 +84,9 @@ export default class AbiCoder {
         obj[component.name] = this.getCoder(component);
         return obj;
       }, {});
-      return new EnumCoder(enumMatch.name, coders);
+      return param.type === OPTION_CODER_TYPE
+        ? new OptionCoder(enumMatch.name, coders)
+        : new EnumCoder(enumMatch.name, coders);
```
luizstacio, line 89:
```suggestion
     const isOptionEnum = param.type === OPTION_CODER_TYPE;
     if (isOptionEnum) {
         return new OptionCoder(enumMatch.name, coders);
     }
     return new EnumCoder(enumMatch.name, coders);
```

luizstacio (changes requested):

luizstacio (approved):
Just one code comment.

luizstacio (approved):

QuinnLee (approved):

PR #226, opened 2022-04-20
feat: add `excludedIds` param to getCoinsToSpend

AlicanC:
The feature was added to fuel-core a while ago but was missing fuels-ts implementation.

PR #472, opened 2022-08-29
feat: Add ExcludeId to filter out utxo_ids on coinsToSpend

QuinnLee:
Closes out https://github.com/FuelLabs/fuels-ts/pull/226

## TLDR

Users can filter out utxo's from `getCoinsToSpend`

### DIFF
--- a/packages/wallet/src/transfer.test.ts
+++ b/packages/wallet/src/transfer.test.ts
@@ -1,6 +1,5 @@
 import { NativeAssetId } from '@fuel-ts/constants';
 import { Provider, ScriptTransactionRequest } from '@fuel-ts/providers';
-import { assert } from 'console';
 
 import { generateTestWallet } from './test-utils';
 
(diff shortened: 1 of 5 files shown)


packages/wallet/src/transfer.test.ts
```diff
@@ -1,5 +1,6 @@
 import { NativeAssetId } from '@fuel-ts/constants';
 import { Provider, ScriptTransactionRequest } from '@fuel-ts/providers';
+import { assert } from 'console';
```
luizstacio, line 3: I think we can remove this here.

luizstacio (changes requested):

packages/wallet/src/transfer.test.ts
QuinnLee: done!

luizstacio (approved):

PR #479, opened 2022-09-02
chore: improve coverage step on github action

pedronauck:
This action updates the comment on PR with the coverage instead of creating a new one on each run, also it has a better visual detail about the coverage.


packages/providers/src/transaction-request/transaction-request.ts
```diff
@@ -10,10 +10,8 @@ import type {
   AbstractScript,
 } from '@fuel-ts/interfaces';
 import type { BigNumberish } from '@fuel-ts/math';
-import { multiply } from '@fuel-ts/math';
```
QuinnLee, line 13: 👍🏼

QuinnLee (approved):

luizstacio (approved):

LuizAsFight (approved):

Issue #480, opened 2022-09-02
Simplify type generation CLI


luizstacio:
Today users need to install `fuelchain` and `typechain-target-fuels` to generate Types from a `ABI JSON`. The idea is to implement a single CLI tool for type generation and remove the need to install `typechain-target-fuels`.

Issue #430, opened 2022-07-29
Implement generic on typechain


LuizAsFight:
### Motivation

After `0.18.1`, sway is supporting `typeArguments` in ABI.

But `type` field inside `components` doesn't tell what type is being set (as mentioned in [sway/2386](https://github.com/FuelLabs/sway/issues/2386)). Then we don't know what dynamic type is being set for each argument. Instead we just have the final type.

After this issue is merged [sway/2386](https://github.com/FuelLabs/sway/issues/2386), we should implement dynamic typing in typechain

This is also blocked by the implementation on the ABI: #431

luizstacio:
With the new flat ABI, it's not possible to use the default implementation of the [typechain](https://www.npmjs.com/package/typechain); the issue is related to the way the typechain checks if an abi file is empty or not ([see file line in typechain](https://github.com/dethcrypto/TypeChain/blob/92faf28a91fa17a6cb9c34bd29ea9b2cae5d57b3/packages/typechain/src/typechain/io.ts#L42)). 

I have implemented locally a solution for typechain to support a external `extractAbi` function witch is provided on the typechain target. I'm going to open a PR on typechain.

luizstacio:
> With the new flat ABI, it's not possible to use the default implementation of the [typechain](https://www.npmjs.com/package/typechain); When using it, the typechain throws `Error occured: Not a valid ABI`. This issue is related to the way the typechain checks if an abi file is empty or not ([see file line in typechain](https://github.com/dethcrypto/TypeChain/blob/92faf28a91fa17a6cb9c34bd29ea9b2cae5d57b3/packages/typechain/src/typechain/io.ts#L42)).
> 
> I have implemented locally a solution for typechain to support a external `extractAbi` function witch is provided on the typechain target and added the implementation for the flat ABI JSON on the `typechain-target-fuels`. I'm going to open a PR on typechain to see if is possible to get this new feature.

https://github.com/dethcrypto/TypeChain/pull/757

PR #481, opened 2022-09-02
feat: change from typechain to fuelchain

luizstacio:
This PR includes a fork from the original TypeChain. Including it under the name `FuelChain`.

The original TypeChain CLI, doesn't enable support for custom ABI JSON formats. This started to be an issue with the new FLAT ABI format. We have opened one PR in the TypeChain original project, but we didn't get comments on it. 

The idea is first to implement this fork and, in the near future, migrate to a single CLI command that doesn't require the installation of `fuelchain` and `typechain-target-fuels`. https://github.com/FuelLabs/fuels-ts/issues/480

More info can be found on:
- The original issue: https://github.com/FuelLabs/fuels-ts/issues/430
- The PR open on TypeChain: https://github.com/dethcrypto/TypeChain/pull/757


packages/fuelchain/src/parser/parseEvmType.ts
```diff
@@ -0,0 +1,193 @@
+/* eslint-disable consistent-return */
+/* eslint-disable no-case-declarations */
+/* eslint-disable default-case */
+/* eslint-disable @typescript-eslint/no-use-before-define */
+/* eslint-disable no-plusplus */
+/* eslint-disable radix */
```
LuizAsFight, line 6: if we're ok in getting more and more different from `typechain` code, would be nice to fix this anti-patterns pointed by eslint

packages/fuelchain/src/utils/modules.ts
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable import/no-dynamic-require */
+/* eslint-disable consistent-return */
+/* eslint-disable global-require */
```
LuizAsFight, line 3: if we're ok in getting more and more different from `typechain` code, would be nice to fix this anti-patterns pointed by eslint

packages/fuelchain/src/utils/logger.ts
```diff
@@ -0,0 +1,29 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
LuizAsFight, line 1: if we're ok in getting more and more different from `typechain` code, would be nice to fix this anti-patterns pointed by eslint

packages/fuelchain/src/parser/parseEvmType.ts
LuizAsFight: there's more along other files, should fix there too if possible

packages/fuelchain/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "fuelchain",
+  "version": "0.13.0",
+  "description": "🔌 FuelChain is fork from TypeChain",
```
LuizAsFight, line 4: missing `a`:

"FuelChain is `a` fork from TypeChain"

LuizAsFight (changes requested):
if possible remove lint warnings and `eslint-disable` comments to improve code quality and remove anti-patterns

QuinnLee (approved):

packages/fuelchain/src/utils/modules.ts
luizstacio: The idea is to change the least amount of code possible. In the near future, we can also address and improve the code on the feature to improve CLI. This could also result in a completely different code, as we can use the opportunity to remove `typechain-target-*` and remove dynamic codes. https://github.com/FuelLabs/fuels-ts/issues/480

QuinnLee (approved):

packages/fuelchain/src/utils/modules.ts
LuizAsFight: ok makes sense, we can work in refactor it in a different issue. if we're stable at this point I don't see reasons to keep code similar

LuizAsFight (approved):
LG(reat)TM

PR #494, opened 2022-09-07
fix: Decode transaction-request with offsets

QuinnLee:
## TLDR

Update the decoding logic for transactions


packages/transactions/src/coders/transaction.test.ts
```diff
@@ -15,27 +16,41 @@ describe('TransactionCoder', () => {
...
     expect(encoded).toEqual(
-      '0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000d5579c46dfcc7f18207013e65b44e4cb4e2c2298f4ac457ba8f82743f31e930b'
+      '0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000d5579c46dfcc7f18207013e65b44e4cb4e2c2298f4ac457ba8f82743f31e930b0000000000000001d5579c46dfcc7f18207013e65b44e4cb4e2c2298f4ac457ba8f82743f31e930b0000000000000000d5579c46dfcc7f18207013e65b44e4cb4e2c2298f4ac457ba8f82743f31e930bd5579c46dfcc7f18207013e65b44e4cb4e2c2298f4ac457ba8f82743f31e930b00000000000000000000000000000000d5579c46dfcc7f18207013e65b44e4cb4e2c2298f4ac457ba8f82743f31e930b'
     );
 
     const [decoded, offset] = new TransactionCoder().decode(arrayify(encoded), 0);
 
     expect(offset).toEqual((encoded.length - 2) / 2);
-    expect(JSON.stringify(decoded)).toEqual(JSON.stringify(transaction));
+    expect(JSON.parse(JSON.stringify(decoded))).toMatchObject(
```
QuinnLee, line 51: the stringified json order was different, this is done to ignore the key-value ordering

packages/transactions/src/coders/transaction.test.ts
```diff
@@ -15,27 +16,41 @@ describe('TransactionCoder', () => {
       maturity: 0,
       scriptLength: 0,
       scriptDataLength: 0,
-      inputsCount: 0,
+      inputsCount: 1,
```
luizstacio, line 19: Only to enforce better testing can we add a different value for each field of the tx on this test?

luizstacio (changes requested):

packages/transactions/src/coders/transaction.test.ts
QuinnLee: sure, let me add it.  I'll also add additional tests to test input, output and witnesses in isolation

QuinnLee: done! also updated the `transaction.create` tests

packages/transactions/src/coders/transaction.test.ts
```diff
@@ -38,7 +40,78 @@ describe('TransactionCoder', () => {
     expect(JSON.stringify(decoded)).toEqual(JSON.stringify(transaction));
   });
 
-  it('Can encode TransactionCreate', () => {
+  it('Can encode TransactionScript with inputs, outputs and witnesses', () => {
```
luizstacio, line 43: I think we can add some data on all these 0 values as this will not validate but only parse is good for testing.

gasPrice, gas limit, maturity, lengths etc...

luizstacio (changes requested):

packages/transactions/src/coders/transaction.ts
```diff
@@ -239,16 +239,16 @@ export class TransactionCreateCoder extends Coder<TransactionCreate, Transaction
...
     const bytecodeLength = decoded;
+    [decoded, o] = new NumberCoder('u8').decode(data, o);
     const bytecodeWitnessIndex = decoded;
+    [decoded, o] = new NumberCoder('u16').decode(data, o);
     const storageSlotsCount = decoded;
+    [decoded, o] = new NumberCoder('u8').decode(data, o);
     const inputsCount = decoded;
+    [decoded, o] = new NumberCoder('u8').decode(data, o);
     const outputsCount = decoded;
+    [decoded, o] = new NumberCoder('u8').decode(data, o);
```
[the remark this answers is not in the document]
QuinnLee: yeah it was a bug

packages/transactions/src/coders/transaction.test.ts
QuinnLee: Sounds good, I'll update the transactions.  What about witnesses, input, and outputs?  I think it's a bit overkill for these because they have separate coders, but I can update these here or as a follow-up issue.

luizstacio: Maybe random values would be better than = values for each field. But it already looks great.

If we could also generate some random witnesses data would be nice also.

luizstacio (approved):

Issue #499, opened 2022-09-09
Race condition bug on "my-fueled-app" example


itspauloroberto:
### What version of fuels-ts are you using?

Latest (v0.13.0)

### Steps to Reproduce

1. Run the "my-fueled-app" sample application according to the Quickstart guidelines walkthrough.
2. Click the "Increment" button several times, and the faster you can. (You can also use a script to click on it very fast to make it easier to reproduce)
3. Look into the console of the fuel-core API running, and you can see errors mentioning: "Transaction id was already used".

### Expected Behavior

The expected behavior is to use the cleanup function to prevent newer promises from being fired before older ones get fulfilled.

Or it can also be a loading state, or even to keep track of the last used transaction id, but I do feel that's too much for a simple task.

### Actual Behavior

Basically, the errors are popping up on the console of the fuel-core API running. And this is because of the promise race condition bug which occurs because the old promise was fulfilled before the newer promise. The application is not handling that edge case because there is no cleanup being used on hooks or there is no loading state or something like this.

digorithm:
@luizstacio what's the current status on this?

PR #500, opened 2022-09-09
fix: solve race condition bug on sample application

itspauloroberto:
This Pull Request closes issue #499 

### What I've done
- Added a new state variable to keep track locally of the counter before actually increasing the real counter.
- Added a new hook to update the real counter when the local counter was updated, and only if there is no pending update ongoing to avoid race conditions (using a useEffect hook cleanup function).
- Changed the displayed counter to the real counter value.

### Why i have done this way
I feel that this is the most elegant and quick solution for the problem since it prevents additional simultaneous requests only by using a cleanup function which is a useEffect's hook feature ready to use, and this is enough to prevent a race condition bug.

I know that there are other ways to accomplish that, like, for instance, keeping track of transaction ID or creating a loading state on the button, but taking into consideration the user experience and the effort necessary to make those solutions i felt that mine solution is more appropriated.


docs/examples/my-fuel-dapp/frontend/src/App.tsx
```diff
@@ -24,30 +24,45 @@ const contract = ContractAbi__factory.connect(CONTRACT_ID, wallet);
...
+        // Creates a transactions to call the increment function passing the amount
+        // we want to increment, because it creates a TX and updates the contract state
+        // this requires the wallet to have enough coins to cover the costs and also
+        // to sign the Transaction
+        const { value } = await contract.functions.increment(1).call();
+        setBalance(Number(value));
+      }
+    }
+    fetch();
+    return () => {
```
luizstacio, line 53: I think it would be great to add a loading state to let the user know tx has been executed instead of just blocking the button. And change button text to `Incrementing...`

luizstacio (changes requested):

docs/examples/my-fuel-dapp/frontend/src/App.tsx
itspauloroberto: I do agree with you. But, the blocking only occurs if the user clicks the button repeatedly at a very fast speed, so it's clearly not even noticeable. To make a loading state here would add some "latency" to the operation that is meant to be quick. But I will try doing it and see how the experience feels.

docs/examples/my-fuel-dapp/frontend/src/App.tsx
```diff
@@ -24,30 +24,45 @@ const contract = ContractAbi__factory.connect(CONTRACT_ID, wallet);
...
+  }, [counter]);
+
   async function increment() {
-    // Creates a transactions to call the increment function passing the amount
-    // we want to increment, because it creates a TX and updates the contract state
-    // this requires the wallet to have enough coins to cover the costs and also
-    // to sign the Transaction
-    const { value } = await contract.functions.increment(1).call();
-    setCounter(Number(value));
+    setCounter(balance + 1);
```
LuizAsFight, line 59: I think we should have a loading state also, then won't need to have balance at all.

1- you can create loading: `const [loading, setLoading] = useState(false); `
2- this `increment` function can be created with useCallback, and listen to `loading`
3- in the start of `increment` function: `if (loading) return`
4- before you execute contract call: `await setLoading(true)`
4- after you execute contract call `await setLoading(false)`

then will never be called twice

LuizAsFight: Other strategy would be to `debounce` or `throttle` the click function

LuizAsFight (changes requested):

docs/examples/my-fuel-dapp/frontend/src/App.tsx
itspauloroberto: Thanks! @LuizAsFight that's way more clever rather than the cleanup hook. I've even switched the code back to what it was and simply added a loading state and that's all.

docs/examples/my-fuel-dapp/frontend/src/App.tsx
itspauloroberto: I've put the loading state, but the `Incrementing...` looked **very weird** because the operation is almost instant. So what I did was updating disabled attribute on the button according to loading state and it feels better now.

luizstacio:
As the example app, it's been removed (#679). We can close this issue. Thanks, @itspauloroberto, for the contribution.

PR #511, opened 2022-09-15
chore: update forc 0.24.3

QuinnLee:
Updating `forc`!

### DIFF
--- a/.changeset/neat-singers-bathe.md
+++ b/.changeset/neat-singers-bathe.md
@@ -1,0 +1,8 @@
+---
+"@fuel-ts/contract": minor
+"@fuel-ts/example-contract": minor
+"@fuel-ts/predicate": minor
+"@fuel-ts/script": minor
+---
+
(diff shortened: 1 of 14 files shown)


.changeset/neat-singers-bathe.md
```diff
@@ -0,0 +1,9 @@
+---
+"@fuel-ts/contract": minor
+"@fuel-ts/example-contract": minor
+"forc-bin": minor
```
luizstacio, line 4: We need to remove `forc-bin` from here otherwise, the CI will change the version and break.

Can you try to find a why for changesets to ignore this package? Maybe we can open another issue.

luizstacio (changes requested):

.changeset/neat-singers-bathe.md
QuinnLee: done!

luizstacio (approved):

PR #515, opened 2022-09-22
feat: add message support

### LINKED ISSUE #504: Add message support
Add support for InputMessage and OutputMessage. As development of the bridge progresses, we will need support for being able to spend and create messages. Refer to the currently implemented spec [here](https://github.com/FuelLabs/fuel-specs/tree/8bb43ae61aab69e72275c004792aef34efed1d9b), and you can find the parallel rust sdk issue [here](https://github.com/FuelLabs/fuels-rs/issues/530).

Here is a non exhaustive list of what needs to be done:
- [ ] Add `InputMessage` to inputs.ts in transactions (then also under provider)
- [ ] `OutputWithdrawal` needs to be updated to `OutputMessage` in outputs.ts (then also under provider)
- [ ] Add `getMessages` to the provider (similar to `getCoins`)
- [ ] Add `addMessage` and `addMessages` to transaction-request.ts (similar to `addCoin` and `addCoins`)
   - [ ] `getCoinInputWitnessIndexByOwner` should be modified to support the `owner` field on input messages
   - [ ] can still try to auto add a change output but for the base asset

### LINKED ISSUE #513: Use new Message specs once fuel-core update comes out
### Motivation

The `owner` field was removed from Messages as it became redundant, as part of https://github.com/FuelLabs/fuel-core/issues/568.

Once fuel-core is released with a version tag using this update, the Message support in the SDK will also need to be updated:
- Encoding of message inputs
- Hashing of message id
- Initial state JSON

### Usage example

_No response_

### Possible implementations

_No response_

camsjams:
Closes #504 and

Closes #513 (just used newer Docker image to cover for this)


packages/contract/src/__test__/coverage-contract/coverage-contract.test.ts
```diff
@@ -1,8 +1,11 @@
+import { arrayify, hexlify } from '@ethersproject/bytes';
+import { ArrayCoder, NumberCoder } from '@fuel-ts/abi-coder';
 import { NativeAssetId } from '@fuel-ts/constants';
 import type { BN } from '@fuel-ts/math';
 import { bn, toHex } from '@fuel-ts/math';
-import { Provider, LogReader } from '@fuel-ts/providers';
-import { TestUtils } from '@fuel-ts/wallet';
+import { Provider, LogReader, ScriptTransactionRequest } from '@fuel-ts/providers';
+import type { Message } from '@fuel-ts/providers/src/message';
```
luizstacio, line 7: Need export on the index.ts from providers

packages/contract/src/__test__/coverage-contract/coverage-contract.test.ts
```diff
@@ -1,8 +1,11 @@
+import { arrayify, hexlify } from '@ethersproject/bytes';
```
luizstacio, line 1: We should clean up this non used variables

packages/transactions/src/coders/input.ts
```diff
@@ -191,7 +192,129 @@ export class InputContractCoder extends Coder<InputContract, InputContract> {
...
+    parts.push(new NumberCoder('u16').encode(value.predicateLength));
+    parts.push(new NumberCoder('u16').encode(value.predicateDataLength));
+    parts.push(encodedData);
+    parts.push(new ByteArrayCoder(value.predicateLength).encode(value.predicate));
+    parts.push(new ByteArrayCoder(value.predicateDataLength).encode(value.predicateData));
+
+    return concat(parts);
+  }
+
+  decode(data: Uint8Array, offset: number): [InputMessage, number] {
```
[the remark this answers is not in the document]
luizstacio: We should add this to the transaction-request decode process

services/fuel-core/Dockerfile
```diff
@@ -21,7 +22,6 @@ CMD exec ./fuel-core run \
     --ip ${IP} \
     --port ${PORT} \
     --db-path ${DB_PATH} \
-    --utxo-validation \
     --min-gas-price ${MIN_GAS_PRICE} \
```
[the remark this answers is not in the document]
luizstacio: I think if it works without could be an issue once we have some predicate tests that would require this to be enabled. @Voxelot do you have some idea why this could be happening?

packages/providers/src/provider.ts
```diff
@@ -507,6 +509,40 @@ export default class Provider {
...
+      first: 10,
+      ...paginationArgs,
+      owner: address.toB256(),
+    });
+
+    const messages = result.messages.edges!.map((edge) => edge!.node!);
+
+    return messages.map((message) => {
+      const dataLength = message.data.length;
+      const [data] = new ArrayCoder(new NumberCoder('u8'), dataLength / WORD_SIZE).decode(
```
luizstacio, line 531: Why do we need to divide data length by the `WORD_SIZE).decode(...`. Looking at the specs, it looks like data length

luizstacio: Can we use the message decoder here?

packages/transactions/src/coders/input.test.ts
```diff
@@ -62,4 +62,49 @@ describe('InputCoder', () => {
...
     expect(decoded).toEqual(input);
   });
+
+  it('Can encode Message Id', () => {
+    const input: InputMessage = {
+      type: InputType.Message,
+      amount: bn(1000),
+      sender: '0xf1e92c42b90934aa6372e30bc568a326f6e66a1a0288595e6e3fbd392a4f3e6e',
+      recipient: '0xef86afa9696cf0dc6385e2c407a6e159a1103cefb7e2ae0636fb33d3cb2a9e4a',
+      data: [12, 13, 14],
```
luizstacio, line 72: Data is always an array of bytes, or we should also accept a string here?

luizstacio (changes requested):

packages/contract/src/__test__/coverage-contract/coverage-contract.test.ts
```diff
@@ -1,8 +1,9 @@
 import { NativeAssetId } from '@fuel-ts/constants';
 import type { BN } from '@fuel-ts/math';
 import { bn, toHex } from '@fuel-ts/math';
-import { Provider } from '@fuel-ts/providers';
-import { TestUtils } from '@fuel-ts/wallet';
+import { Provider, ScriptTransactionRequest } from '@fuel-ts/providers';
+import type { Message } from '@fuel-ts/providers/src/message';
```
luizstacio, line 5:
```suggestion
import type { Message } from '@fuel-ts/providers';
```

packages/wallet/src/wallet.ts
```diff
@@ -19,6 +19,7 @@ import type {
   BuildPredicateOptions,
   TransactionResult,
 } from '@fuel-ts/providers';
+import type { Message } from '@fuel-ts/providers/src/message';
```
luizstacio, line 22:
```suggestion
import type { Message } from '@fuel-ts/providers';
```

luizstacio (changes requested):

luizstacio (approved):

Voxelot:
> I think if it works without could be an issue once we have some predicate tests that would require this to be enabled. @Voxelot do you have some idea why this could be happening?

This could be related to https://github.com/FuelLabs/fuel-tx/pull/186

QuinnLee (approved):

PR #517, opened 2022-09-23
feat: Parse Logs and LogData

QuinnLee:


packages/contract/src/__test__/coverage-contract/coverage-contract.test.ts
```diff
@@ -234,20 +234,23 @@ describe('Coverage Contract', () => {
   });
 
   it('should test u8 vector input', async () => {
-    const { value, transactionResult } = await contractInstance.functions
```
QuinnLee, line 237: Update the test to read the logs

packages/abi-coder/src/interface.ts
```diff
@@ -32,9 +38,17 @@ export default class Interface {
...
   readonly abiCoder: AbiCoder;
+  readonly abi: ABI | null;
+  readonly types: ReadonlyArray<JsonFlatAbiFragmentType>;
+  readonly loggedTypes: ReadonlyArray<JsonAbiLogFragment>;
 
   constructor(jsonAbi: JsonAbi | JsonFlatAbi) {
+    this.abi = isFlatJsonAbi(jsonAbi) ? new ABI(jsonAbi) : null;
     this.fragments = coerceFragments(ABI.unflatten(jsonAbi));
+
+    this.types = this.abi ? this.abi.types : [];
```
QuinnLee, line 49: this isn't used, but my ocd kicked in and I added it.  It can be removed

packages/abi-coder/src/json-abi.ts
```diff
@@ -58,10 +73,16 @@ export type JsonAbi = ReadonlyArray<JsonAbiFragment> | JsonFlatAbi;
...
+  readonly loggedTypes: ReadonlyArray<JsonFlatAbiFragmentLoggedType>;
 
   constructor(jsonAbi: JsonFlatAbi) {
     this.types = jsonAbi.types;
     this.functions = jsonAbi.functions;
+    this.loggedTypes = jsonAbi.loggedTypes;
+  }
+
+  parseLoggedType(loggedType: JsonFlatAbiFragmentLoggedType): JsonAbiFragmentType {
+    return ParamType.fromObject(this.parseInput(loggedType.loggedType) as JsonFragmentType);
```
QuinnLee, line 85: It was much easier to reuse the current logic - but this should be cleaned up asap

luizstacio (approved):

packages/abi-coder/src/json-abi.ts
```diff
@@ -111,11 +132,19 @@ export class ABI {
...
-    return abi.unflatten();
+
+    return jsonAbi;
+  }
+
+  unflattenLoggedTypes(): ReadonlyArray<JsonAbiLogFragment> {
+    return this.loggedTypes.map((loggedType) => ({
+      ...loggedType,
+      abiFragmentType: [this.parseLoggedType(loggedType)],
+    }));
```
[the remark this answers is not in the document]
QuinnLee: This is correct.  My understanding is that we `unflatten` it to make it work with the code. We really need to refactor the how the coders and ABI interact because this is where the tech debt is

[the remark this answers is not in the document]
luizstacio: The idea behind keeping the nonthe  flat abi, was to keep possible partial definition of functions using objects. But once logs and other things are going to require types. We should remove it.

PR #518, opened 2022-09-23
feat: improve math with formats/parsing unit related functions

### LINKED ISSUE #165: Create a units utility to work with decimals
During the development of a DApp that manipulates the numbers returned from contracts, we normally have to use a lot of utilities to handle;
- Parsing `bn` to `number`;
- Parse `string` to `number`;
- Parse `bn` to `number` formatted with decimals etc;
- Divide `bn` and more.

This is a reference of the utilities used on swayswap: https://github.com/FuelLabs/swayswap/blob/master/packages/app/src/systems/Core/utils/math.ts

LuizAsFight:
## Improvements to math package

- now `bn` accepts undefined, and use default value `0 (zero)`
- Added helper functions to our math `bn`
  - `format` - format units and precision. Returns a formatted value from any of `BnInput`
  - `formatUnits` - format units. Returns a formatted value from any of `BnInput`
  - `parseUnits` - create a BN instance parsing to unit value.
  - `toFixed` - formats a string to precision. don't consider units
- added functional shortcuts to new helpers as well

#### Usage (all parameters are optional):
```
bn('1001000000200000000').format({
  minPrecision: 3,
  precision: 8,
  units: 9
});

bn('1000000000').formatUnits(7);

bn.parseUnits('100100.00002');
bn.parseUnits('100100.00002', 5);
```

Closes #165


packages/math/src/bn.test.ts
```diff
@@ -280,4 +280,152 @@ describe('Math - Convert', () => {
     expect(bn(maxBytes).toHex(4)).toEqual(maxHex);
     expect(() => bn(over).toHex(4)).toThrow();
   });
+
+  it('should accept possibly undefined to craete bn, and set to 0 when undefined', () => {
```
luizstacio, line 284:
```suggestion
  it('should create bn with number or undefined', () => {
```

luizstacio: maybe add some tests for creating with string also?

packages/math/src/bn.test.ts
```diff
@@ -280,4 +280,152 @@ describe('Math - Convert', () => {
...
+    expect(bn('100000020000').formatUnits(7)).toEqual('10000.0020000');
+    expect(bn('100100000020000').formatUnits(7)).toEqual('10010000.0020000');
+  });
+
+  it('should format with default configs', () => {
+    expect(bn('1000000000').format()).toEqual('1');
+    expect(bn('2').format()).toEqual('0.000000002');
+    expect(bn('22000').format()).toEqual('0.00002');
+    expect(bn('100000020000').format()).toEqual('100');
+    expect(bn('100100000020000').format()).toEqual('100,100');
```
luizstacio, line 315: I think we should keep at least a `.0` not having. `.` and having `,` could lead to the miss understanding of the amount being a decimal.

packages/math/src/bn.ts
```diff
@@ -47,8 +51,8 @@ export class BN extends BnJs implements BNInputOverrides, BNHiddenTypes, BNHelpe
       super(value.substring(2), base || 'hex', endian);
       return;
     }
-
-    super(value, base, endian);
+    const safeValue = value == null ? 0 : value;
```
luizstacio, line 54:
```suggestion
    const defaultValue = value == null ? 0 : value;
```

packages/math/src/bn.test.ts
```diff
@@ -280,4 +280,152 @@ describe('Math - Convert', () => {
...
+      })
+    ).toEqual('100,100.0');
+    expect(
+      bn('100100000020000').format({
+        minPrecision: 2,
+        units: 8,
+      })
+    ).toEqual('1,001,000.00');
+
+    expect(
```
luizstacio, line 411: Maybe it is missing some tests with just `units` and just `minPrecision` with a big number

packages/math/src/decimal.ts
```diff
@@ -0,0 +1,35 @@
+import { DEFAULT_MIN_PRECISION, DEFAULT_PRECISION } from './constants';
+import type { ToFixedConfig } from './types';
+
+// export function toFixed(value?: string | number, precision: number = DEFAULT_PRECISION) {
+export function toFixed(value?: string | number, options?: ToFixedConfig) {
+  const { precision = DEFAULT_PRECISION, minPrecision = DEFAULT_MIN_PRECISION } = options || {};
+
+  // todo: minPrecision higher than precision, should override or respect precision?
```
luizstacio, line 8: respect the precision... maybe throw a error

packages/math/src/decimal.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  const [valueUnits = '0', valueDecimals = '0'] = String(value || '0.0').split('.');
+  const groupRegex = new RegExp(`(\\d)(?=(\\d{${precision}})+\\b)`, 'g');
+  const units = valueUnits.replace(groupRegex, '$1,');
+  let decimals = valueDecimals.slice(0, precision);
+
+  // increase precision if decimal formatted is zero, but has more numbers out of precision
+  // 0.000014 and precision 3 will be = 0.00001 instead of 0.000
+  // todo: should create a prop for this??
```
luizstacio, line 17: I think we should change the default behavior of `toFixed` to just cut on the unit amount and not do the search for the first non-zero. And pass the other logic inside the `format` method.

packages/math/src/bn.test.ts
LuizAsFight: this test is only check undefined compatibility. we have a lot of other tests creating from strings already

packages/math/src/bn.test.ts
```diff
@@ -280,4 +280,187 @@ describe('Math - Convert', () => {
...
+        precision: 8,
+      })
+    ).toEqual('1,001,000,000.20000');
+  });
+
+  it('should parse to bn unit from decimal/inputs/string values', () => {
+    expect(bn.parseUnits('1').toHex()).toEqual(bn('1000000000').toHex());
+    expect(bn.parseUnits('0.000000002').toHex()).toEqual(bn(2).toHex());
+    expect(bn.parseUnits('0.00002').toHex()).toEqual(bn('20000').toHex());
+    expect(bn.parseUnits('100.00002').toHex()).toEqual(bn('100000020000').toHex());
```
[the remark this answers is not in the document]
luizstacio: We have a default amount of units that is 9, then `0.00002` would be `0.000020000`and the final number is `20000`.

packages/math/src/bn.ts
```diff
@@ -47,8 +51,8 @@ export class BN extends BnJs implements BNInputOverrides, BNHiddenTypes, BNHelpe
       super(value.substring(2), base || 'hex', endian);
       return;
     }
-
-    super(value, base, endian);
+    const defaultValue = value == null ? 0 : value;
```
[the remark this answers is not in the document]
luizstacio: I think here we need to test as `==` in order to catch undefined cases

packages/math/src/bn.test.ts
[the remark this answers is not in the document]
LuizAsFight: it's quite confusing but life saver for front-end

luizstacio (approved):

PR #522, opened 2022-09-26
feat: improve Option codegen and types

### LINKED ISSUE #519: Fuelchain does not generate the correct type for OptionOutput<>

0.16.0

### Steps to Reproduce

I have a repro [here](https://github.com/matt-user/option-repro). 
Simply create a contract that has a function which returns an Option.  And then generate the frontend types for the contract.

### Expected Behavior

The generated type should contain information about the contained type such as struct fields etc. OptionOutput<bool | null>

### Actual Behavior

The type generated is OptionOutput<[]>/

camsjams:
Fixes #519


matt-user (approved):
lgtm

luizstacio (approved):

PR #523, opened 2022-09-27
chore: now `bn` accepts `null`

LuizAsFight:


packages/math/src/bn.ts
```diff
@@ -251,7 +251,7 @@ export class BN extends BnJs implements BNInputOverrides, BNHiddenTypes, BNHelpe
 }
 
 // functional shortcut to create BN
-export const bn = (value?: BNInput, base?: number | 'hex', endian?: BnJs.Endianness) =>
+export const bn = (value?: BNInput | null, base?: number | 'hex', endian?: BnJs.Endianness) =>
```
luizstacio, line 254: Should we change the type BNInput?

luizstacio (approved):

packages/math/src/bn.ts
LuizAsFight: I think no. all other methods use BNInput, but shouldn't accept a null/undefined ad values.

for example:
`bn(10).mul()` should not work. I need to explicitly say the value I want to multiply to `mul` method, which has `BNInput` as typing

PR #524, opened 2022-09-28
feat: Add support for Vector as an output

### LINKED ISSUE #487: Add `Vec<T>` Support as an `Output`
### Description
Iterate through the structure and log its contents, then parse those receipts

```rust
fn echo_u8_vector(input: Vec<u8>) -> Vec<u8> {
    let mut i = 0;
    while i < input.len() {
      log(input.get(i).unwrap());
      i += 1;
    }
    
    input
}
```

- The vector length will be retrieved from the returned `Vec` `struct` property `len` (`input` in above example)
- In above code, the vector data will be inside the log as items with type `ReceiptTypeReturnData`
- See [`LogReader`](https://github.com/FuelLabs/fuels-ts/blob/56b635db91cdb3d4d4f3a44d2ebab425b5b596f1/packages/providers/src/LogReader.ts), perhaps we can add there or reuse some of the code

Details: https://github.com/FuelLabs/fuels-rs/issues/353

See also #434 and #497 

### Initial checklist
- [ ] Call a contract fn and receive a Vec<T> as return data in logs
- [ ] TBD Generate the correct `Type` using `Typegen` for the `Vec<T>` Output

See also #434

camsjams:
Closes #487

### DIFF
--- a/.changeset/tidy-keys-fly.md
+++ b/.changeset/tidy-keys-fly.md
@@ -1,0 +1,6 @@
+---
+"@fuel-ts/abi-coder": patch
+"@fuel-ts/contract": patch
+---
+
+Add support to get outputs of type Vector


packages/abi-coder/src/coders/vec.ts
```diff
@@ -16,16 +16,24 @@ export default class VecCoder<TCoder extends Coder> extends Coder<
...
     super('struct', `struct Vec`, 0);
     this.coder = coder;
+    this.logReceiptData = [];
   }
 
   static getBaseOffset(): number {
     return VEC_PROPERTY_SPACE * WORD_SIZE;
   }
 
+  set logs(logs: Array<any>) {
```
QuinnLee, line 31: Should we cast logs as `Array<uknown>` instead of `Array<any>`?

packages/abi-coder/src/abi-coder.ts
```diff
@@ -139,7 +139,11 @@ export default class AbiCoder {
     return concat([results, concat(vectorData)]);
   }
 
-  decode(types: ReadonlyArray<JsonAbiFragmentType>, data: BytesLike): DecodedValue[] | undefined {
+  decode(
```
luizstacio, line 142: It looks a bit weird to pass the logs all the way to the Vec decode method. I was exploring ideas in my mind. To keep this in a more high-level place like the interface and just return a value to know where to place the Vec output parsed from the logs and avoid the throw not being implemented. This would isolate this temp workaround. But don't know if you already tough about this

luizstacio: This would remove the logs here and in the vec decoder

packages/contract/src/__test__/coverage-contract/coverage-contract.test.ts
```diff
@@ -460,4 +460,67 @@ describe('Coverage Contract', () => {
...
+      },
+    ];
+    const EXPECTED = INPUT.map((input) => ({ ...input, bar: bn(input.bar).toHex() }));
+    const { value } = await contractInstance.functions.echo_complex_struct_vector(INPUT).call();
+    const RESULT = value.map((result: ComplexStruct) => ({
+      ...result,
+      bar: bn(result.bar).toHex(),
+    }));
+    expect(RESULT).toStrictEqual(EXPECTED);
+  });
```
luizstacio, line 525: We should add a test using two methods that return Vec in a multicall. The inputs would need to be different in order to test if it will correctly get each returned log.

```ts
contractInstance.multiCall([
    contractInstance.functions.echo_complex_struct_vector(INPUT),
    contractInstance.functions.echo_struct_vector(INPUT_2),
])
```

luizstacio (changes requested):

PR #526, opened 2022-10-04
chore: relocate contract related testing

### LINKED ISSUE #492: Organize contract tests in a better way
Today we have many contracts inside the `packages/contract` that not only test the contract calls but also test parsing. We should move this to a better structure. Maybe a new package `packages/contract-tests`, that we shouldn't release. The idea would also be created in such a way we could use contracts from other repos like https://github.com/FuelLabs/fuels-rs/tree/master/packages/fuels/tests/test_projects. This is an opportunity to also unify in a single repo this test project.

camsjams:
Closes #492


QuinnLee (dismissed):

packages/fuel-gauge/src/utils.ts
```diff
@@ -1,12 +1,8 @@
 import type { BytesLike } from '@ethersproject/bytes';
```
luizstacio, line 1: We can import `BytesLike` from `fuels` and remove depedency for `@ethersproject/bytes`

packages/fuel-gauge/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "author": "Fuel Labs <contact@fuel.sh> (https://fuel.network/)",
+  "scripts": {
+    "prebuild": "pnpm build-projects",
+    "build-projects": "./scripts/build.sh"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@ethersproject/bytes": "^5.5.0",
+    "@ethersproject/logger": "^5.5.0",
+    "@ethersproject/sha2": "^5.6.0",
```
luizstacio, line 14: Can we remove this dependency? By importing from `fuels`

luizstacio (changes requested):
Why the name gauge?

QuinnLee (approved):

luizstacio (approved):

PR #531, opened 2022-10-10
feat: add valueOf to bn and address

QuinnLee:
## TLDR

`valueOf` for bn and addresses


packages/math/src/bn.ts
```diff
@@ -91,6 +91,10 @@ export class BN extends BnJs implements BNInputOverrides, BNHiddenTypes, BNHelpe
     return this.toString(16);
   }
 
+  valueOf(): string {
+    return this.toString();
+  }
+
```
QuinnLee, line 97: not sure to hex this.  `bn` doesnt have value of and Bignumber does [this](https://mikemcl.github.io/bignumber.js/#valueOf)

luizstacio (approved):

Issue #473, opened 2022-08-30
Run forc fmt against repo, perhaps enforce as a linter?


camsjams:
### Motivation

Now that forc fmt works, let's use it!

PR #535, opened 2022-10-11
feat: add forc fmt check

camsjams:
Closes https://github.com/FuelLabs/fuels-ts/issues/473


QuinnLee (approved):
looks like most of the changes are from a formatter

.github/workflows/test.yaml
```diff
@@ -20,6 +20,13 @@ jobs:
       - name: CI Setup
         uses: ./.github/actions/ci-setup
 
+      - name: Forc Format
```
luizstacio, line 23: We can run just `forc fmt --check` here

luizstacio (changes requested):

luizstacio (approved):

LuizAsFight (approved):

PR #543, opened 2022-10-18
chore: upgrading all dependencies to latest version

### LINKED ISSUE #542: Upgrade `ts` and dependencies
### Motivation

We are using `^4.5.2`, the latest version is `4.8.x` https://www.npmjs.com/package/typescript. We should upgrade it and any other dependency 

### Usage example

_No response_

### Possible implementations

_No response_

arboleya:
# Summary

 - Closes #542 

# TODO

 - [x] Upgrade all npm libraries in all<sup>(1)</sup> packages
 - [x] Fix call inside `/scripts/forc-check.sh`
 - [x] Decouple `package` and `forc` versions in `/packages/forc-bin`
   - _With this, there's no need to ignore `forc-bin` in changesets_

# Notes
<sup>(1)</sup> _The `node-fetch` package will remain outdated because recent versions got published as an ESM module _only_. Before using it, we'd need to [change some configs](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#can-i-import-esm-packages-in-my-typescript-project)._


.changeset/smart-jeans-tickle.md
```diff
@@ -0,0 +1,24 @@
+---
+"@fuel-ts/abi-coder": patch
+"@fuel-ts/address": patch
+"@fuel-ts/contract": patch
+"@fuel-ts/example-contract": patch
+"forc-bin": patch
```
luizstacio, line 6: Remove forc bin from the changesets. It will lead to a issue when the release is created.   Can you add on the changesets ignore also?

luizstacio (changes requested):

.changeset/smart-jeans-tickle.md
arboleya: It seems that by ignoring the `forc-bin` package triggers a cascade event in which we need to ignore several other packages that depend on it, namely:

 - `@fuel-ts/contract`
 - `@fuel-ts/example-contract"`
 - `@fuel-ts/predicate"`
 - `@fuel-ts/script"`
 - `forc-bin`
 - `fuel-gauge`
 - `fuels`
 - `typechain-target-fuels`

arboleya: @luizstacio Also, could you please expand on what issue it will lead to when creating the release?

luizstacio: forc-bin package uses the version of the package to download the forc binary, in this way adding here on the changesets will try to download the wrong version.

luizstacio: > It seems that ignoring the `forc-bin` package triggers a cascade event in which we need to ignore several other packages that depend on it, namely:
> 
> * `@fuel-ts/contract`
> * `@fuel-ts/example-contract`
> * `@fuel-ts/predicate`
> * `@fuel-ts/script`
> * `fuels`
> * `typechain-target-fuels`

Do you know why? And if is possible to fix this?

luizstacio, line 6: Need to remove forc-bin from here

luizstacio (changes requested):

.changeset/smart-jeans-tickle.md
arboleya: I circumvented this and documented it on the PR description.

Please, have a look and let me know your thoughts about it.

luizstacio (approved):

luizstacio (approved):

QuinnLee (approved):

packages/forc-bin/src/shared.ts
```diff
@@ -22,15 +22,18 @@ export const getPkgPlatform = () => {
...
-  return version;
+  const { forcVersion } = JSON.parse(pkgJson).config;
+  return forcVersion;
 };
+
 export const setCurrentVersion = async (version: string) => {
   const pkgJson = await fs.readFile(pkgJsonPath, 'utf8');
-  const wrap = (v: string) => `"version": "${v}",`;
+  const { forcVersion } = JSON.parse(pkgJson).config;
+  const wrap = (v: string) => `"forcVersion": "${v}",`;
```
AlicanC, line 35: This change here breaks this script since the field `forcVersion` doesn't have a comma in the end.

arboleya: Nice catch! I'll review and fix this.

PR #551, opened 2022-10-27
feat: export vault

luizstacio:
Add a new method on the wallet-manager to export the vault.

### DIFF
--- a/packages/wallet-manager/src/types.ts
+++ b/packages/wallet-manager/src/types.ts
@@ -29,17 +29,17 @@ export interface WalletManagerState {
   vaults: VaultsState;
 }
 
-export abstract class Vault<TOptions = unknown> {
+export abstract class Vault<TOptions = { secret?: string }> {
   static readonly type: string;
 
   // The variables are not used on the abstract class
   // Did not look enough to find a way to not need to use comments.
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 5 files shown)


QuinnLee (approved):

packages/wallet-manager/src/types.ts
```diff
@@ -29,17 +29,17 @@ export interface WalletManagerState {
   vaults: VaultsState;
 }
 
-export abstract class Vault<TOptions = unknown> {
+export abstract class Vault<TOptions = { secret?: string }> {
   static readonly type: string;
 
   // The variables are not used on the abstract class
   // Did not look enough to find a way to not need to use comments.
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
arboleya, line 37: Isn't the rule of thumb for explicitly unused variables to precede them with a `_`?

By adding this to the `.eslintrc.js` file:

```ts
'@typescript-eslint/no-unused-vars': [
  'warn',
  {
    argsIgnorePattern: '^_',
    varsIgnorePattern: '^_',
  },
],
```

We can simply rename the parameter to `_options`.

```ts
constructor(_options: TOptions) {
  throw new Error('Not implemented');
}
```

The same goes for the parameter `address` in the other methods.

arboleya (approved):
LGTM.

PR #552, opened 2022-10-27
refactor: change sign methods to be async

luizstacio:
This PR changes the sign methods to be async in this way is possible to implement Signer classes that require an async process like Wallet Extension or Hardware Wallets.

### DIFF
--- a/packages/wallet/src/wallet.ts
+++ b/packages/wallet/src/wallet.ts
@@ -75,9 +75,9 @@ export default class Wallet extends AbstractWallet {
    * Sign message with wallet instance privateKey
    *
    * @param message - Message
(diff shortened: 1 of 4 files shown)


QuinnLee (approved):

packages/wallet/src/wallet.ts
```diff
@@ -87,20 +87,20 @@ export default class Wallet extends AbstractWallet {
    * @param transactionRequestLike - TransactionRequestLike
    * @returns string - Signature a ECDSA 64 bytes
```
arboleya, line 88: The `@return` doc for the `signTransaction` method should contain a `Promise<>`:
 - `@returns Promise<string> - Signature a ECDSA 64 bytes`

PR #557, opened 2022-10-28
fix: change order to avoid text busy error

### LINKED ISSUE #546: CI tests fail intermittently
The CI test runner process seems flaky and sometimes fails with something like the error below.

It seems to originate from the `fuel-gauge` package:
- https://github.com/FuelLabs/fuels-ts/actions/runs/3313772573/jobs/5472200619#step:5:2626

error

camsjams:
fixes #546

Modifies flow of CI to avoid error `ETXTBSY: text file is busy`.


arboleya:
Not sure if I got this right, but isn't the `fuels-gauge` package only for testing purposes?

If so, perhaps its build step could live in a `pretest` script, and leave the `build` alone.

Today when we build the project, it mandatorily builds everything inside `fuels-gauge`.

This seems unnecessary, thus making the process take longer.

QuinnLee (dismissed):

arboleya (approved):

Dhaiwat10 (approved):

PR #564, opened 2022-10-30
feat: add convenience function to wallet for easy withdrawal to base chain

pixelcircuits:
Adds a convenience function for withdraw base asset to the base chain. This will be useful for generating test output message for future issues to test with


packages/wallet/src/transfer.test.ts
```diff
@@ -114,4 +116,25 @@ describe('Wallet', () => {
...
+    const tx = await sender.withdraw(recipient, 10);
+    const result = await tx.wait();
+
+    const messageOutReceipt = <TransactionResultMessageOutReceipt>result.receipts[0];
+    expect(result.transactionId).toEqual(messageOutReceipt.sender);
+    expect(recipient.toHexString()).toEqual(messageOutReceipt.recipient);
+    expect(amount.toString()).toEqual(messageOutReceipt.amount.toString());
+
+    const senderBalances = await sender.getBalances();
+    expect(senderBalances).toEqual([{ assetId: NativeAssetId, amount: bn(90) }]);
```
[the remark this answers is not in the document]
pixelcircuits: The 'recipient' is an Ethereum address. Their balance won't be updated on Ethereum until `relayMessageFromFuelBlock` on the Ethereum side MessagePortal contract is called. Full end-to-end testing of this is currently being implemented in [this repo](https://github.com/FuelLabs/bridge-integration-tests)

digorithm (changes requested):
Looks really nice. What do you think about renaming it from `withdraw()` to `transfer()` so we can keep the terminology/API between the Rust SDK and the TS SDK similar? [The equivalent of `withdraw()` in the Rust SDK is `transfer()`.](https://fuellabs.github.io/fuels-rs/v0.27.0/wallets/transferring-assets.html?highlight=transfer#transferring-assets)

digorithm:
> > Looks really nice. What do you think about renaming it from `withdraw()` to `transfer()` so we can keep the terminology/API between the Rust SDK and the TS SDK similar? [The equivalent of `withdraw()` in the Rust SDK is `transfer()`.](https://fuellabs.github.io/fuels-rs/v0.27.0/wallets/transferring-assets.html?highlight=transfer#transferring-assets)
> 
> This new feature is not about transferring coins within the Fuel chain. It is for withdrawing the base asset (ETH) back to Ethereum. This has not yet been implemented in the rust SDK. It's popping up in the typescript SDK first since that's what's being used for the full Ethereum+Fuel end-to-end tests

Oh, I see; it makes sense not to rename it to `transfer()`. But, `withdraw()` on its own doesn't tell me the wallet is withdrawing to the base chain. Maybe the function name could be more explicit about it?

pixelcircuits:
I'm open to suggestions for a different name other than `withdraw`. What does everyone else think? I used the term `withdraw` because I wanted to avoid using the term `ETH` or `L1`. I guess in the modular stack thesis the correct nomenclature would be `settlement layer`. So maybe something like `withdrawToSettlementLayer`? or `withdrawToBaseLayer`? @adlerjohn @Voxelot

digorithm:
> I'm open to suggestions for a different name other than `withdraw`. What does everyone else think? I used the term `withdraw` because I wanted to avoid using the term `ETH` or `L1`. I guess in the modular stack thesis the correct nomenclature would be `settlement layer`. So maybe something like `withdrawToSettlementLayer`? or `withdrawToBaseLayer`? @adlerjohn @Voxelot

Yeah, we shouldn't be mentioning ETH because the base layer is "pluggable". I like both `withdrawToSettlementLayer` and `withdrawToBaseLayer` 🤔 Maybe `BaseLayer` is more generic/broad here and thus would be better?

pixelcircuits:
ok, renamed `withdraw` to `withdrawToBaseLayer` to make things more clear.

digorithm (approved):
LGTM -- great stuff!

adlerjohn:
any of the above suggestions seem reasonable to me

packages/transactions/src/coders/receipt.ts
```diff
@@ -682,14 +682,22 @@ export class ReceiptMessageOutCoder extends Coder<ReceiptMessageOut, ReceiptMess
     value.data.forEach((d) => dataInNumbers.push(d));
     const parts: Uint8Array[] = [];
 
+    // encode data padded to nearest word (8 bytes)
```
luizstacio, line 685: I think here you can use the `ByteArrayCoder `. Like this? https://github.com/FuelLabs/fuels-ts/blob/db8cc6b49616199368463ecd69aae6b3ca0b65d0/packages/transactions/src/coders/input.ts#L73

luizstacio (changes requested):

packages/transactions/src/coders/receipt.ts
pixelcircuits: Yep, thanks!

luizstacio (approved):

PR #576, opened 2022-11-04
feat: add fuelBlockSpend to getMessages query

pixelcircuits:


packages/fuel-gauge/src/coverage-contract.test.ts
```diff
@@ -448,6 +456,13 @@ describe('Coverage Contract', () => {
     // sort by nonce, messages are not guaranteed in order
     receiverMessages.sort((a, b) => a.nonce.toNumber() - b.nonce.toNumber());
 
-    expect(receiverMessages).toStrictEqual(messages);
+    expect(receiverMessages.length).toEqual(messages.length);
+    for (let i = 0; i < receiverMessages.length; i += 1) {
+      expect(receiverMessages[i].amount).toEqual(messages[i].amount);
+      expect(receiverMessages[i].sender).toEqual(messages[i].sender);
+      expect(receiverMessages[i].recipient).toEqual(messages[i].recipient);
+      expect(receiverMessages[i].nonce).toEqual(messages[i].nonce);
```
luizstacio, line 464: Should we add the `fuelBlockSpend` on the checks also?

luizstacio (changes requested):

packages/fuel-gauge/src/coverage-contract.test.ts
pixelcircuits: I don't think we can. This value can change depending on the state of the fuel-core service running. It's the block number that the message was spent in

luizstacio: Just check if the `fuelBlockSpend` is present?

luizstacio (approved):

PR #583, opened 2022-11-07
feat: add contract ID from PanicReceipts 

QuinnLee:
## TLDR

Add missing contractId's from panic receipts

## Follow-up
Figure out if the underlying transactionRequest on the scope should be updated too


packages/fuel-gauge/src/contract.test.ts
```diff
@@ -132,21 +132,10 @@ describe('Contract', () => {
...
-    expect(scope.transactionRequest.getContractInputs()).toEqual([
-      { contractId: contract.id.toB256(), type: 1, txPointer },
-      { contractId: otherContract.id.toB256(), type: 1, txPointer },
-    ]);
-
-    expect(scope.transactionRequest.getContractOutputs()).toEqual([
-      { type: 1, inputIndex: 0 },
-      { type: 1, inputIndex: 1 },
-    ]);
+    const scope = contract.functions.call_external_foo(1336, otherContract.id);
```
QuinnLee, line 135: I think this is sufficient for tests following the pattern [here](https://github.com/FuelLabs/fuels-rs/pull/628/files#diff-99029e6dd80e10a946c69a9ea3b16c35afb83813b457d3bdbddc4905e4c1aeea).

The only issue is that the underlying transaction request doesn't get mutated if there are updates on to it so the `getContract`  assertion is no longer needed.  I'm going to create follow-up issue to see if we should update the `scope.transactionRequest`

packages/providers/src/util.ts
```diff
@@ -41,11 +42,34 @@ export function sleep(time: number = 1000) {
...
+): receipt is ReceiptRevert =>
+  receipt.type === ReceiptType.Revert &&
+  receipt.val.toString('hex') === FAILED_TRANSFER_TO_ADDRESS_SIGNAL;
+
+const getReceiptsWithMissingContractIds = (
+  receipt: TransactionResultReceipt
+): receipt is ReceiptPanic =>
+  receipt.type === ReceiptType.Panic &&
+  receipt.contractId !== '0x0000000000000000000000000000000000000000000000000000000000000000';
+
```
arboleya, line 56: Nice use of type helpers. ⚡ 

Since `getReceiptsWithMissingOutputVariables` and `getReceiptsWithMissingContractIds` don't _actually_ get a list of receipts, I suggest renaming them using question-like nomenclature.

One literal example would be:

- `doesReceiptHaveMissingOutputVariables`
- `doesReceiptHaveMissingContractIds`

packages/providers/src/provider.ts
```diff
@@ -316,9 +315,23 @@ export default class Provider {
...
+      }
+
       transactionRequest.addVariableOutputs(missingOutputVariableCount);
-    } while (tries > MAX_RETRIES || missingOutputVariableCount > 0);
+
+      missingOutputContractIds.forEach(({ contractId }) =>
+        transactionRequest.addContract(Address.fromString(contractId))
+      );
+      tries += 1;
+    } while (tries < MAX_RETRIES);
```
arboleya, line 334: What is the reasoning behind `MAX_RETRIES`, and why is it `10` and not `20` or `30` or other value?

arboleya (changes requested):

packages/providers/src/util.ts
QuinnLee: got it! this makes sense.  Will update!

packages/providers/src/provider.ts
QuinnLee: TBH, I just used the same number as the Rust SDK set.  Not sure what the motivation for it is too. cc @digorithm - should we set a max retry to a different number

arboleya (approved):

PR #604, opened 2022-11-16
feat: add new transaction of Mint type

### LINKED ISSUE #602: SDK failing on parsing transaction in decoding inputs part

0.21.0

### Steps to Reproduce

get a transaction from production and try to decode

### Expected Behavior

Should decode 

### Actual Behavior

Error on decoding

### LINKED ISSUE #603: Add decode transaction of Mint type
### Motivation

There's a new transaction type that we don't support yet.
We should support encoding/decoding this type of transaction too

[fuel-specs/transaction](https://fuellabs.github.io/fuel-specs/master/protocol/tx_format/transaction.html#transactionmint)

### Usage example

_No response_

### Possible implementations

_No response_

LuizAsFight:
Add new Mint transaction
- decoders/encoders
- fix TransactionResponse to work with few props (no receipts, gas, etc..)
- specify better typing in TransactionRequest, as we don't need/should to use Mint transactions in there
- add better typing for Transaction, to improve usage when consuming TransactionResponse / TransactionResult in frontend side
  - now a `Transaction` will have all possible parameters as optional.
  - `Transaction<TransactionType.Script>` will have the exact params for Script transaction only
  - `Transaction<TransactionType.Create>` will have the exact params for Create transaction only
  - `Transaction<TransactionType.Mint>` will have the exact params for Mint transaction only

Closes #603 
Closes #602

### DIFF
--- a/packages/transactions/src/coders/transaction.ts
+++ b/packages/transactions/src/coders/transaction.ts
@@ -11,12 +11,15 @@ import type { Output } from './output';
 import { OutputCoder } from './output';
 import { StorageSlotCoder } from './storage-slot';
 import type { StorageSlot } from './storage-slot';
+import type { TxPointer } from './tx-pointer';
+import { TxPointerCoder } from './tx-pointer';
 import type { Witness } from './witness';
 import { WitnessCoder } from './witness';
 
 export enum TransactionType /* u8 */ {
   Script = 0,
   Create = 1,
+  Mint = 2,
 }
 
 export type TransactionScript = {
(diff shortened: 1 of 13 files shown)


packages/transactions/src/coders/transaction.ts
```diff
@@ -291,7 +294,62 @@ export class TransactionCreateCoder extends Coder<TransactionCreate, Transaction
   }
 }
 
-export type Transaction = TransactionScript | TransactionCreate;
```
QuinnLee, line 294: Could you elaborate on why this was removed? I think we can do  `TransactionScript | TransactionCreate | TransactionMint`, but I may be missing something

LuizAsFight: it wasn't removed, it's just ahead in the code

LuizAsFight: `TransactionScript | TransactionCreate | TransactionMint` was the initial idea, but then it breaks a lot of transactions because of small number of common properties.

then I did a refactor on typing stuff for `Transaction` which now accepts a generic informing transaction type

QuinnLee (approved):

packages/transactions/src/coders/transaction.ts
LuizAsFight: @QuinnLee you can check the changed tests/files to understand better how it's working now

luizstacio (approved):

PR #613, opened 2022-11-21
fix: remove addCoins, addCoin and Run `pnpm run docs`

### LINKED ISSUE #590: Remove old functions
Remove `addCoin` and `addCoins`, they are no longer in use after #588 changes

QuinnLee:
## TLDR

Removed `addCoins` and `addCoin` and `pnpm run docs` 

There is a lot of changes on docs

Closes #590


packages/providers/src/transaction-request/transaction-request.ts
```diff
@@ -315,50 +315,6 @@ abstract class BaseTransactionRequest implements BaseTransactionRequestLike {
     resources.forEach((resource) => this.addResource(resource));
   }
 
-  /**
```
luizstacio, line 318: I think we can remove this import from Coin also.

luizstacio (changes requested):

luizstacio (approved):

arboleya (approved):

PR #620, opened 2022-11-25
feat: add `buildBlockExplorerUrl` helper function

### LINKED ISSUE #600: add helper to build block explorer link
### Motivation

In many projects, users will need to build a link to block explorer, to show transactions or for any other reason.

we should add a helper method to allow it.

- use a default Block Explorer URL (our latest one -> https://fuellabs.github.io/block-explorer-v2/)
- allow the user to configure other blockExplorerUrl

### Usage example

_No response_

### Possible implementations

```
export function getBlockExplorerLink({
  path,
  providerUrl,
}: {
  path: string;
  providerUrl: string;
  blockExplorerUrl: string;
}) {
  const blockUrl = blockExplorerUrl || BLOCK_EXPLORER_URL;
  return `${blockUrl}${path}?providerUrl=${encodeURIComponent(
    providerUrl
  )}`;
}
```

Dhaiwat10:
## Summary

This PR adds a helper function called `buildBlockExplorerUrl` to the `providers` package.

## API
```ts
const url = buildBlockExplorerUrl({ path: '/address/0x123'}); // https://fuellabs.github.io/block-explorer-v2/address/0x123

// or you can pass in a custom block explorer URL
const url = buildBlockExplorerUrl({ blockExplorerUrl: 'https://explorer.fuel.sh', path: '/address/0x123' }); // https://explorer.fuel.sh/address/0x123
```
___

## Other details / questions

- I have added a `patch` version bump to the `providers` package from the pattern I could see in previous PRs. Let me know if this needs to change!
- I noticed that we don't have any tests written for the other utility functions in the file. I'm guessing it's because they are pretty straightforward? Regardless, I have added two simple tests for the function I wrote. Please let me know if it is extraneous and if I should remove the test if it's not adding a lot of value.
- Re. [this comment in the issue](https://github.com/FuelLabs/fuels-ts/issues/600#issuecomment-1326733111), I haven't created a new `utils` folder just yet as I feel like we should tackle that in a separate issue since it would involve migrating a lot if not all of the helper functions into a single folder. I have kept my newly created function in the `util.ts` file inside the `providers` package for now.
- @LuizAsFight I am a little confused about your mention of a `providerUrl` param where you described the possible implementation for this function in the issue [here](https://github.com/FuelLabs/fuels-ts/issues/600#issue-1448734722). I looked at some block explorer links on the Fuel block explorer (eg. https://fuellabs.github.io/block-explorer-v2/transaction/0xe8d60fb06ecba0d914a6082b0a94324a5e6b2fba59e29d0eeac3cafcaf798cec) and didn't spot any occurrences of `providerUrl` in the URLs. Can you please explain why we need to append a `providerUrl` to the generated URL? Thanks!

Closes #600


packages/providers/src/util.test.ts
```diff
@@ -0,0 +1,16 @@
...
+  test('Builds block explorer links using buildBlockExplorerUrl', () => {
+    const result = buildBlockExplorerUrl({ path: '/transaction/0x123' });
+    expect(result).toEqual(`${DEFAULT_BLOCK_EXPLORER_URL}/transaction/0x123`);
+
+    const result2 = buildBlockExplorerUrl({
+      blockExplorerUrl: 'https://explorer.fuel.sh',
+      path: '/transaction/0x123',
+    });
+    expect(result2).toEqual('https://explorer.fuel.sh/transaction/0x123');
+  });
```
arboleya, line 15: The tests should address all possible cases concerning leading/trailing slashes on `path` and `explorerUrl`.

Perhaps iterating over an array with `from`/`expectations` pairs can reduce repetition.

arboleya (changes requested):

Dhaiwat10:
@arboleya thanks so much for the review! I have made the changes. Let me know if the PR looks good now. And should I squash the three commits together?

arboleya (approved):
Good job! 🙌 

You don't need to change anything at this point, but here's another take on it just for reference. Feel free to ignore it unless it makes sense to you.

> **Note:** pseudo code written inside the comments section.

<details>
  <summary>Code</summary>

```ts
import { buildBlockExplorerUrl } from './util';

describe('Providers utils', () => {
  const DEFAULT_BLOCK_EXPLORER_URL = 'https://fuellabs.github.io/block-explorer-v2';

  test('Builds block explorer URL in every possible way', () => {
    const trimSlashes = /^\/|\/$/gm;
    const basePath = 'transaction/0x123';
    const baseBlockExplorer = 'https://explorer.fuel.sh';

    // Here, we assemble and iterate over all possible combinations of
    // `blockExplorerUrl` and `path` by adding leading/trailing slashes to them
    const pathUrls = [basePath, `/${basePath}`, `${basePath}/`, `/${basePath}/`]
    const blockExplorerUrls = [undefined, baseBlockExplorer, `${baseBlockExplorer}/`]

    pathUrls.forEach((path) => {
      blockExplorerUrls.forEach((blockExplorerUrl) => {
        // First, we clean paths manually
        const cleanPath = path.replace(trimSlashes, '');
        const cleanBlockExplorerUrl = blockExplorerUrl?.replace(trimSlashes, '');
        
        // Then we compare them with the ones returned by our method
        const expected = cleanBlockExplorerUrl || `${DEFAULT_BLOCK_EXPLORER_URL}/${cleanPath}`
        expect(buildBlockExplorerUrl({ path, blockExplorerUrl })).toBe(expected);
      });
    });
  });
});
```
</details>

arboleya:
> And should I squash the three commits together?

Yes, all merges are squashed.

Dhaiwat10:
@arboleya thanks, that code looks much better. I have pushed a refactor!

QuinnLee (approved):

LuizAsFight (changes requested):
Looking good !

- explanation for `providerUrl` question, is in the requested change
- nice one in creating the tests, no harm at all we should all be doing it more. If you notice we also have a lot of other files untested, but many of them are also indirectly tested through other processes (executing contracts, scripts, etc...)
- about adding `txId` and `address` as params, also feel like a good idea, but something to think:
  - would need to address conflict between inputs `path`, `txId` and `address`. (for example cannot go to transaction and address page at same time)
  - would add the responsibility of maintaining the route always correct in the sdk side
  - my conclusion is: I like the idea but we can skip for now in this pr, as it's almost ready to go. Then create other issue focused on adding these helper params. Can you create this issue @Dhaiwat10 ?
  
  Thanks !

Dhaiwat10:
@LuizAsFight sounds good! Added the `providerUrl` param. Will create the issue too 🤝

packages/providers/src/util.ts
```diff
@@ -95,3 +95,30 @@ export const calculateTransactionFee = ({
...
+  const explorerUrl = blockExplorerUrl || DEFAULT_BLOCK_EXPLORER_URL;
+
+  // Remove leading and trailing slashes from the path and block explorer url respectively, if present
+  const trimSlashes = /^\/|\/$/gm;
+  const cleanPath = path.replace(trimSlashes, '');
+  const cleanBlockExplorerUrl = explorerUrl.replace(trimSlashes, '');
+  const cleanProviderUrl = providerUrl?.replace(trimSlashes, '');
+
+  return `${cleanBlockExplorerUrl}/${cleanPath}${
+    cleanProviderUrl ? `?providerUrl=${encodeURIComponent(cleanProviderUrl)}` : ''
```
luizstacio, line 122:
```suggestion
  const cleanBlockExplorerUrl = explorerUrl.replace(trimSlashes, '');
  const cleanProviderUrl = providerUrl?.replace(trimSlashes, '');
  const encodedProviderUrl = encodeURIComponent(cleanProviderUrl);

  return `${cleanBlockExplorerUrl}/${cleanPath}${
    cleanProviderUrl ? `?providerUrl=${encodedProviderUrl}` : ''
```

Dhaiwat10: Thanks @luizstacio 🤝

luizstacio (dismissed):

packages/providers/src/util.test.ts
```diff
@@ -0,0 +1,39 @@
...
+        // Then we compare them with the ones returned by our method
+        providerUrls.forEach((providerUrl) => {
+          const cleanProviderUrl = providerUrl?.replace(trimSlashes, '');
+          const encodedProviderUrl = cleanProviderUrl
+            ? encodeURIComponent(cleanProviderUrl)
+            : undefined;
+          const expected = `${cleanBlockExplorerUrl}/${cleanPath}${
+            cleanProviderUrl ? `?providerUrl=${encodedProviderUrl}` : ''
+          }`;
+          expect(buildBlockExplorerUrl({ path, blockExplorerUrl, providerUrl })).toBe(expected);
```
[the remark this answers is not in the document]
Dhaiwat10: Thank you so much for the review and feedback @camsjams. That makes a lot of sense. I just pushed some changes. Can you have a look and let me know if this looks better?

packages/providers/src/util.ts
```diff
@@ -95,3 +95,31 @@ export const calculateTransactionFee = ({
...
+ * Builds a block explorer url based on and the given path, block explorer URL and provider URL
+ */
+export const buildBlockExplorerUrl = ({
+  blockExplorerUrl,
+  path,
+  providerUrl,
+}: {
+  blockExplorerUrl?: string;
+  path: string;
+  providerUrl?: string;
```
LuizAsFight, line 111: we should support receiving both `blockExplorerUrl` and `providerUrl` without the protocol.
if we detect that the user didn't provide protocol (`://`) in the string, we auto prefix the input with `https://`

to increase DX / flexibility, if the user informs `blockExplorerUrl: 'explorer.fuel.sh'`, it should just work

LuizAsFight (changes requested):
sorry for last minute review @Dhaiwat10 , I figured something out and requested a change. I'm also ok if we create a new issue to do this later, but I think it's not a big deal and would be nice to support from start

packages/providers/src/util.ts
Dhaiwat10: @LuizAsFight gotcha, adding this now

LuizAsFight (approved):

Issue #582, opened 2022-11-07
TransactionOutputCoinAssetIdNotFound


pixelcircuits:
### What version of fuels-ts are you using?

0.20.0

### Steps to Reproduce

I believe the issue occurs when `resourceToSpend` returns input messages. To trigger this, add the follow message to the `chainConfig.json` file for the fuel core service that's ran before testing:
```
 {
   "sender": "0xc43454aa38dd91f88109a4b7aef5efb96ce34e3f24992fe0f81d233ca686f80f",
   "recipient": "0x09c0b2d1a486c439a87bcba6b46a7a1a23f3897cc83a94521a96da5c23bc58db",
   "nonce": "0x1000",
   "amount": "0x010000000000",
   "data": "0x00",
   "da_height": "0x00"
 }
```

The message recipient is the same address that's used for a majority of the tests.

### Expected Behavior

The tests should run just fine, with input messages being used to cover gas costs just like a coin input.

### Actual Behavior

The tests fail with the error `TransactionOutputCoinAssetIdNotFound`

QuinnLee:
Im having a hard time replicating this with the lastest version locally and in CI.  I ran it here via CI https://github.com/FuelLabs/fuels-ts/pull/621/files

pixelcircuits:
I just updated my workspace and I can't replicate this issue anymore. It must have been fixed a few weeks ago

PR #621, opened 2022-11-28
fix: add message to chainConfig

QuinnLee:


luizstacio (approved):

arboleya:
@QuinnLee If I got this right, this is not a fix for #582 yet, and if that's the case, perhaps it'd be helpful to mark the PR as a draft to avoid approvals ahead of time.

PR #624, opened 2022-11-28
chore: update default config values for `bn.format`

### LINKED ISSUE #595: Change default values for `bn.format`
### Motivation

Current default parameters for `bn.format` are:
```
{
  precision: 3,
  minPrecision: 1
}
```

let's say the user has `0.002313` ETH, format will show: `0.002`. Abbreviated.

Which looks good. But normally while using the SDK, the user will need the right value, instead of an abbreviated one.

We should change default parameters to:
```
{
  precision: DECIMAL_UNITS, // will be 9
  minPrecision: 3
}
```

### Usage example

_No response_

### Possible implementations

_No response_

Dhaiwat10:
## Summary
This PR updates the default config values for the `bn.format` function as requested in #595.

```ts
// constants.ts
DEFAULT_PRECISION = 9 // previously 3
DEFAULT_MIN_PRECISION = 3 // previously 1
```

Closes #595

### DIFF
--- a/.changeset/unlucky-roses-share.md
+++ b/.changeset/unlucky-roses-share.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/math": patch
+---
+
+The default config values for `bn.format` have been changed.
--- a/packages/math/src/bn.test.ts
+++ b/packages/math/src/bn.test.ts
@@ -312,11 +312,11 @@ describe('Math - BN', () => {
   });
 
   it('should format with default configs', () => {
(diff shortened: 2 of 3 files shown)


arboleya:
Perhaps it's not a _breaking_ change per se, but it could _break_ something that's using it.

I'd suggest a minor bump.

arboleya (dismissed):

luizstacio (approved):

LuizAsFight (approved):
lgtm

Issue #634, opened 2022-11-30
Bug: log data decode issue when using custom errors


camsjams:
### What version of fuels-ts are you using?

v0.21.2

### Steps to Reproduce

1. Create custom errors
2. Trigger a potential error or log
3. Decode log results in a types/value length mismatch

### Expected Behavior

No error should be thrown. We've also improved logs to better understand these issues in the future: https://github.com/FuelLabs/fuels-ts/pull/632

### Actual Behavior

![image](https://user-images.githubusercontent.com/833485/204873859-08fae00e-e567-4ad3-99a2-0fdfb936472e.png)

arboleya:
It looks like we have someone with this problem on Discord _(see the thread [here](https://discord.com/channels/732892373507375164/1034937869736878160/threads/1060499318613745734))_.

himanshudevrani96:
Thanks I am looking for the response

danielbate:
This was resolved in #640, we have an example test case logs a custom error using an enum, very similar to the above replication.

PR #640, opened 2022-12-02
feat: add decoding revert logs

camsjams:
relates to #634

### DIFF
--- a/.changeset/unlucky-roses-share.md
+++ b/.changeset/unlucky-roses-share.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/math": minor
+---
+
+The default config values for `bn.format` have been changed.
--- a/packages/math/src/bn.test.ts
+++ b/packages/math/src/bn.test.ts
@@ -312,11 +312,11 @@ describe('Math - BN', () => {
   });
 
   it('should format with default configs', () => {
-    expect(bn('1000000000').format()).toEqual('1.0');
+    expect(bn('1000000000').format()).toEqual('1.000');
     expect(bn('2').format()).toEqual('0.000000002');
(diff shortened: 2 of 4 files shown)


packages/script/src/errors.ts
```diff
@@ -16,7 +11,8 @@ const printLineWithId = (id: string, line: string) =>
   `${id === ZeroBytes32 ? 'script' : id}: ${line}`;
 
 export class ScriptResultDecoderError extends Error {
-  constructor(result: TransactionResult<'failure'>, message: string) {
+  logs: any[];
```
arboleya, line 14: Could this be anything other than `any`?

Also, should we reconsider our lint configs in this regard?

packages/script/src/errors.ts
```diff
@@ -28,21 +24,7 @@ export class ScriptResultDecoderError extends Error {
...
-            printLineWithId(
-              id,
-              `${
-                type === ReceiptType.LogData ? (r as TransactionResultLogDataReceipt).data : r.val0
-              }`
-            )
-          )
-          .join('\n')}`
-      : null;
+    const logsText = logs.length ? `Logs:\n${JSON.stringify(logs, null, 2)}` : null;
```
arboleya, line 27: Is there any other place/test validating this change?

I presume [this test](https://github.com/FuelLabs/fuels-ts/pull/640/files#diff-39d1aa6104e1987397125435435ff592c6b7057ff4b64ea578709fa37610716d
) `log` deconstructed variable, but I wonder if we didn't have anything in place for this case before that, as nothing seems to have been broken.

arboleya (approved):
Advanced stuff!

I'm approving with some questions and minor considerations.

PR #642, opened 2022-12-05
feat: accept custom `Provider` in wallet generation functions

### LINKED ISSUE #636: No way to pass provider to certain wallet functions

0.21.1

### Steps to Reproduce

Try to pass a provider url to `fromMnemonic` `fromSeed` etc..

### Expected Behavior

The wallet should have its provider set to the passed in url/provider.

### Actual Behavior

If you pass in a provider/url it is not set on the wallet object

Dhaiwat10:
## Summary
This PR adds a new parameter named `provider` to the following wallet generation functions: [`fromSeed`](https://github.com/FuelLabs/fuels-ts/blob/master/packages/wallet/src/wallets.ts#L49), [`fromMnemonic`](https://github.com/FuelLabs/fuels-ts/blob/master/packages/wallet/src/wallets.ts#L59), and [`fromExtendedKey`](https://github.com/FuelLabs/fuels-ts/blob/master/packages/wallet/src/wallets.ts#L70).

## Other notes
To add coverage for this new parameter `provider` in our tests, I wasn't sure if I should create a separate test (or tests) just to test this param or if I should pass in a custom provider to the existing tests for these wallet generation functions. For now, I have gone ahead with the latter.

Closes #636


packages/wallet/src/wallets.ts
```diff
@@ -46,30 +47,35 @@ export class WalletUnlocked extends BaseWalletUnlocked {
...
+    path?: string,
+    passphrase?: BytesLike,
+    provider?: Provider
+  ): WalletUnlocked {
     const seed = Mnemonic.mnemonicToSeed(mnemonic, passphrase);
     const hdWallet = HDWallet.fromSeed(seed);
     const childWallet = hdWallet.derivePath(path || WalletUnlocked.defaultPath);
 
-    return new WalletUnlocked(<string>childWallet.privateKey);
+    return new WalletUnlocked(<string>childWallet.privateKey, provider);
```
arboleya, line 70: Although you didn't do this, I think we can use the contextual opportunity to add a _[new]_ test case to ensure line `68` is not missed — something like:

 - Create wallet fromMnemonic with default path

arboleya (changes requested):
I just added a minor observation but otherwise, LGTM.

packages/wallet/src/wallets.ts
Dhaiwat10: Done!

Dhaiwat10: Hmm interesting the test doesn't pass. Need to look into this

Dhaiwat10: My bad, it obviously will not pass because the code is checking for `account_1`'s public key

QuinnLee (approved):

arboleya (approved):

PR #645, opened 2022-12-06
feat: add `isB256` helper to Address

### LINKED ISSUE #626: add helper method `isB256` to address module
### Motivation

Project currently have a [isBech32](https://github.com/FuelLabs/fuels-ts/blob/563ecc5dcce054619e56ca04c8c9e2514dd40e98/packages/address/src/utils.ts#L41) method in `@fuel-ts/address` module util methods

would be nice to also have a `isB256` method

### Usage example

in frontend, users would be able to do things like
```
  const txId = '0x7d032e7a0953b271d6052ef2ddd20ed6381e659237b13e9544e29a7cf922ddd7';
  if (!isBech32(txId) && !isB256(txId)) {
    throw new Error('Invalid tx ID');
  }

  const wallet = Wallet.fromAddress(txId);

  // code goes on... 
```

### Possible implementations

```
export const isB256 = (val: string) => {
  return /(0x)?[0-9a-f]{64}$/i.test(val);
}
```

camsjams:
Closes #626

### DIFF
--- a/.changeset/empty-dots-cover.md
+++ b/.changeset/empty-dots-cover.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/wallet": minor
+---
+
+Accept custom provider in wallet generation functions like `fromSeed`
--- a/.changeset/plenty-pumas-heal.md
+++ b/.changeset/plenty-pumas-heal.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/providers": patch
+---
+
+Add buildBlockExplorerUrl utility function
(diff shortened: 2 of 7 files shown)


packages/address/src/address.test.ts
```diff
@@ -51,6 +51,48 @@ describe('Address utils', () => {
...
+    const result = utils.isB256(ADDRESS_B256);
+
+    expect(result).toBeTruthy();
+  });
+
+  test('isB256 (b256)', async () => {
+    const result = utils.isB256(ADDRESS_B256);
+
+    expect(result).toBeTruthy();
+  });
```
arboleya, line 64: Duplicated test case.

```suggestion
```

arboleya (changes requested):

LuizAsFight (approved):

arboleya (approved):

PR #646, opened 2022-12-06
feat: adjust publish docs

camsjams:
I moved the generation of docs and publishing to occur in the `next` push, which will update docs at a more optimal time.


arboleya (approved):
Could you please elaborate a bit more on why this is better?

Just trying to get familiar with these actions.

Dhaiwat10 (approved):
Same as Anderson, would love to understand why this is better than the current setup! Still familiarising myself with our publishing workflow 😅

PR #664, opened 2022-12-14
feat: add tests for auth functions

camsjams:
Using same test as https://github.com/FuelLabs/fuels-rs/blob/master/packages/fuels/tests/contracts/auth_testing_contract/src/main.sw

### DIFF
--- a/packages/fuel-gauge/src/auth-testing.test.ts
+++ b/packages/fuel-gauge/src/auth-testing.test.ts
@@ -1,8 +1,8 @@
-import type { Contract } from "fuels";
+import type { Contract } from 'fuels';
 
-import { getSetupContract, getWallet } from "./utils";
+import { getSetupContract, getWallet } from './utils';
 
-const setupContract = getSetupContract("auth_testing_contract");
+const setupContract = getSetupContract('auth_testing_contract');
 
 let contractInstance: Contract;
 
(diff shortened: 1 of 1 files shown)


packages/fuel-gauge/src/auth-testing.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  it('can check_msg_sender [with correct id, using get]', async () => {
+    const wallet = getWallet();
+    try {
+      await contractInstance.functions.check_msg_sender({ value: wallet.address.toB256() }).get();
+
+      throw new Error('it should have thrown');
+    } catch (error) {
+      expect(error).toBeTruthy();
+    }
```
arboleya, line 38: Not sure what's going on here but:
 1. Throwing inside the `try` block looks like a mistake
 2. All assertions must be outside `try/catch` blocks to ensure they always run

However, I think we could `expect(...).toThrow()` in this case:

```suggestion
    const wallet = getWallet();
    expect(() => {
      await contractInstance.functions.check_msg_sender({ value: wallet.address.toB256() }).get();
    }).toThrow();
```

packages/fuel-gauge/src/auth-testing.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    try {
+      await contractInstance.functions
+        .check_msg_sender({ value: wallet.address.toB256().replace('x', 'y') })
+        .call();
+
+      throw new Error('it should have thrown');
+    } catch (error) {
+      expect(error).toBeTruthy();
+    }
```
arboleya, line 52: Here as well, shouldn't it be like this?

```suggestion
    const wallet = getWallet();
    expect(() => {
      await contractInstance.functions
          .check_msg_sender({ value: wallet.address.toB256().replace('x', 'y') })
          .call();
    }).toThrow()
```

arboleya (changes requested):

Dhaiwat10 (approved):
@camsjams if you don't mind, I was wondering what exactly is the purpose of the `fuel-gauge` package and the tests that we are adding in this PR? Are these tests supposed to test if the SDK can correctly handle some auth-related Sway standard library and its responses? Just trying to gather some context here. 😅

arboleya (approved):

Dhaiwat10 (approved):

PR #678, opened 2022-12-16
feat: Add new helper to take unknown string and turn into address

### LINKED ISSUE #628: make Address input more flexible
### Motivation

`Address` is such a powerful tool, allowing us to easily manipulate and convert addresses.

one thing we could do to make it better, is to improve DX on creation.
currently to create `Address`, we need to know what type of data we have in hands, and use the correspondent `static  function`:
```
  let address: Address;

  if (!txId) {
    address = Address.fromRandom()
  }

  if (isBech32(txId)) {
    address = new Address(txId);
  }

  if (isB256(txId)) {
    address = Address.fromB256(txId);
  }

  if (isPublicKey(txId)) {
    address = Address.fromPublicKey(txId);
  }
```

 #626 is as a pre for this issue

### Usage example

```
    address = new Address(txId);
```

only this code would:
- validate any type of input (of course only valid ones)
- create an `Address` from the dynamic input

allowing us to do also other things like 
```
  const b256 = new Address(txId).toB256();
  const bech32 = new Address(txId).toBech32();

  // this code create validated and safe outputs for address strings
```

### Possible implementations

something like the logic mentioned in `Motivation` section, could be used in `Address.constructor`

camsjams:
Closes #628

### DIFF
--- a/.changeset/red-rockets-glare.md
+++ b/.changeset/red-rockets-glare.md
@@ -1,0 +1,5 @@
+---
+"fuels": patch
+---
+
+update docs
--- a/docs/_guide/local-docs.md
+++ b/docs/_guide/local-docs.md
@@ -1,0 +1,49 @@
+There are two parts to the fuels-ts repository documentation
+
+## Part One: Typedoc
+[typedoc](https://typedoc.org/) gathers all types, functions, classes, etc., from our source code. It will read through the `packages` folder, collect all that information, and convert it into a markdown file, modifying the contents of [`docs/packages`](https://github.com/FuelLabs/fuels-ts/tree/master/docs/packages).
+
+
+Furthermore, I've enhanced typedoc to also build out the [Guide documentation](https://fuellabs.github.io/fuels-ts/guide/) with a custom plugin called  `typedoc-plugin-guide-builder` that's in our [repo here](https://github.com/FuelLabs/fuels-ts/tree/master/scripts/typedoc-plugin-guide-builder) (We should move this to a standalone plugin later on). This portion of the typedoc process takes source files from [`docs/_guide`](https://github.com/FuelLabs/fuels-ts/tree/master/docs/_guide), runs them through the plugins and renders code snippets pulled from source code.  
+
+#### How does the `typedoc-plugin-guide-builder` plugin work? 
+See how this markdown document refers to [Sway](https://github.com/FuelLabs/fuels-ts/blob/master/docs/_guide/testing/testing-with-jest.md?plain=1#L6) and [TypeScript](https://github.com/FuelLabs/fuels-ts/blob/master/docs/_guide/testing/testing-with-jest.md?plain=1#L9) code (also note the language hints like `[@code:rust]`, and then see how [reference in the TypeScript code](https://github.com/FuelLabs/fuels-ts/blob/master/packages/example-contract/src/example-contract.test.ts#L1) is written
+
+It writes the final results to [`docs/guide`](https://github.com/FuelLabs/fuels-ts/tree/master/docs/guide).
+
+### Intermission: Repo
+Once typedoc is done generating the markdown docs, the data inside `docs/packages` and `docs/guide` are highly useful. These are complete docs that are checked into the repo and can be found alongside the code for TS-SDK consumers. **One could view and use them just fine inside GitHub file browser or locally within their own filesystem.**
+
+## Part Two: Jekyll
+[jekyll](https://jekyllrb.com/) does the work of rendering the docs in markdown into something the browser understands: HTML. There are currently some [open issues to possibly switch off Jekyll](https://github.com/FuelLabs/fuels-ts/issues/457). We shall see, but [GitHub pages uses Jekyll](https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll) so its the easy choice for now.
+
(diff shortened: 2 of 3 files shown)


packages/address/src/address.ts
```diff
@@ -131,4 +133,32 @@ export default class Address extends AbstractAddress {
   static fromAddressOrString(address: string | AbstractAddress): AbstractAddress {
     return typeof address === 'string' ? this.fromString(address) : address;
   }
+
+  /**
+   * Takes an optional string and returns back an Address. If `address` is nilsy, return back random Address
+   *
+   * @param addressId - Can be a string containing Bech32, B256, or Public Key, or null/undefined
+   * @throws {Error}
+   * thrown if the input string is not nilsy and cannot be resolved to a valid address format
```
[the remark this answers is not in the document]
LuizAsFight: yeah I agree, otherwise user would think the address was right

packages/address/src/address.ts
```diff
@@ -131,4 +133,32 @@ export default class Address extends AbstractAddress {
...
+      return new Address(addressId as Bech32Address);
+    }
+
+    if (isB256(addressId)) {
+      return Address.fromB256(addressId);
+    }
+
+    throw new Error('Unknown address format: only Bech32, B256, or Public Key (512) supported');
+  }
 }
```
LuizAsFight, line 164: Your logic is correct I think, it's good that we're flexible enough to accept empty input, but also make sure user won't use a wrong addresses

One thing I would change is:
The user shouldn't really need to dig into specific static methods (`fromDynamicInput`, `fromPublicKey`, etc..) to create an address.

instead of having static methods for doing it, I think `Address` constructor can centralize handling its inputs.

all the user would need is:

```
const address = new Address(dynamicInput);
```


if done as so, all this entrance static methods will become useless and can be deprecated.

LuizAsFight (changes requested):

packages/address/src/address.ts
arboleya: Agree on everything; I think we can leave it as is.

luizstacio: I think for the dynamic input, we should not have the random method. This can be the source of issues, as the user didn't define anything. It should throw an error in this case. Other than that, it looks good.
```
if (!addressId) {
      return Address.fromRandom();
}
```

packages/address/src/address.ts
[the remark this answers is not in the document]
digorithm: Failing silently here would be catastrophic, haha. Good call @camsjams

packages/address/src/address.ts
digorithm: We can aid this discussion by ensuring everyone's on the same page regarding a mantra: explicit is better than implicit. This philosophy is really good for building secure software.

With that in mind and contrasting what's being discussed here, I agree with @camsjams's point of having `new Address(userInput)` blow up if it's malformed, but if the user wants and knows that it can be malformed, using another, lower-level construct for that end. This is a solid approach that prevents any surprises (surprises are bad here!)

That said, I also agree with @luizstacio; `Address.fromDynamicInput()` (with no param) returning a random address is a relatively surprising behavior; as a user, I would immediately look for something named "random" like `Address.fromRandom()` if I was trying to create a random address. Allowing `Address.fromDynamicInput()` to return a random could cause more trouble than otherwise.

arboleya (approved):

digorithm (approved):
LGTM!

packages/address/src/address.ts
LuizAsFight: I disagree.

you can check whole math/BN package, where we accept anything possible as input.
we use it a lot in front end and it's a really good DX... flexible to use, always work

```
bn('100');
bn(100);
bn('0x64');
bn(byteArray);
bn(buffer);
// same result for all prev examples...


// them you can do operations also with any input
bn(100).add('0x64').sub('50').add(bytesArray).format({precision: 3});
```

now imagine if for all of prev situations, I needed to:
```
bn.fromString('100');
bn.fromNumber(100);
bn.fromHex('0x64');
bn.fromByteArray(byteArray);
bn.fromBuffer(buffer)

// then for operations??
bn.fromString('100').addHex('0x64').subString('50');
```

to me DX and flexibility is key. if it's a valid input we should just keep it simple and accept. I see no reasons to block `new Address(userInput)` 
it's like saying:
"I know your input is valid, but please go and lose your time searching for the correct method".
Having multiple methods just makes the user think and search on things he doesn't really need to. 

I'm ok with blocking invalid/empty number, making it explicit for this case `Address.fromRandom()`



this is my opinion but of course feel free to dismiss my review and merge anyway if rest of the team wants on the other way

luizstacio: I think bn makes sense because is fell on being a number that start with 0 and not random number what will be like Number() instead of returning 0 returning any random number between 0 and Infinity.

luizstacio (approved):

packages/address/src/address.ts
LuizAsFight: I'm don't agree with empty validation as well, but I understand and am ok... what I'm not ok is with constructor accepting only `bech32`... why is that?

if we know it's a vlaid `bech32` or `b256` or `publicKey`, constructor should just accept

digorithm: Context is king. Within the `bn` example you (@LuizAsFight) gave, it does make sense to be a _lot_ more flexible. 

But this is a matter of security and safety; DX does _not_ take precedence over that. Explicitness is unarguably safer than implicitness; easy to back this up; tons of security research is out there. 

Having to explicitly and manually find the right `.from*()` method forces the user to think about how they're handling _public keys_. It goes without saying how sensitive this is.

If the user wants something dynamic, they must _opt-in_ by explicitly and manually picking `.fromDynamicInput()`. 

This removal of implicitness while manipulating _public keys_ is essential and security best practice.

LuizAsFight: awesome thanks for context @digorithm , in terms of getting better security for critical operations, no harm on being more strict.

LuizAsFight (approved):

PR #679, opened 2022-12-16
docs: deprecate quickstart in favor of new quickstart in fuel book

camiinthisthang:
We thank @luizstacio for this quickstart which at one point was the only quickstart available for Fuel.

This is being deprecated in favor of the quickstart in the Fuel Book, which is the most up to date quickstart. This quickstart has caused problems for devs since it isn't maintained. This PR keeps the quickstart page and directs people to the quickstart in the Fuel Book.


digorithm (dismissed):
Having a `Quickstart` section to mention it was deprecated is less than ideal. I suggest —_removing_ the section and adding a link to the new quickstart in the home section.

camiinthisthang:
Made those changes, thank you @digorithm and @camsjams!

Dhaiwat10 (approved):

PR #681, opened 2022-12-16
fix: corrected byteslike comparison with owner

pixelcircuits:
Simple fix for an issue I ran into when trying to manually build transactions


digorithm (approved):
Oh, nice catch!

luizstacio (approved):

PR #682, opened 2022-12-16
fix: change check by instanceof to presence of provider property on contract creation.

luizstacio:
Currently, when trying to create a `Contract` instance using the `WalletLocked` provided by the `Wallet Extension`, is not possible. The reason for it is that the Contract creation checks if the `Wallet` is an instance of `BaseWalletLocked` which will never be true as they are two different bundles. To fix I change the current approach from `instance of` to `property in` in this case checking by the property `provider`.


Dhaiwat10 (dismissed):

arboleya (changes requested):
LGTM, but can we have a few simple unit tests that validate this change?

LuizAsFight (approved):

arboleya (approved):

PR #692, opened 2022-12-20
fix: add missing npm readme for the `fuels` package

### LINKED ISSUE #665: The fuels package has an empty readme on NPM
We should make sure that we are pushing a readme file to the build artifacts that we publish to NPM.

https://www.npmjs.com/package/fuels

Dhaiwat10:
### Summary
Since the `fuels` package does not have a readme in its source code inside the `packages` folder, unlike all the other packages it has an empty readme on npm. This PR adds a prepublish script that copies over the readme at the root of the monorepo to the `fuels` package's bundle before it gets published.

Closes #665


arboleya (dismissed):
Good one! Elegant and straightforward solution.

packages/fuels/package.json
```diff
@@ -55,6 +55,7 @@
   },
   "scripts": {
     "build": "tsup --dts",
-    "build:watch": "tsup --dts --watch"
+    "build:watch": "tsup --dts --watch",
+    "prepublishOnly": "cp ../../README.md ./dist/README.md"
```
luizstacio, line 59: To include files on the dist packages, just list the files on the prop inside the package.json

```json
"files": [
    "dist",
     "./README.md"
  ],
```

luizstacio (changes requested):

packages/fuels/package.json
luizstacio: Also, the readme should not be inside the dist folder after the build but in the root folder. To test it, you can run `pnpm pack`

arboleya: @luizstacio Good catch, but this will only be used at release time, and the `README` file itself is not on the `packages/fuels` directory.

@Dhaiwat10 One simple option you can try is to symlink the `/README.md` into `packages/fuels/README.md`.

> _I've suggested the use of symlinks [here](https://github.com/FuelLabs/fuels-ts/pull/694#issuecomment-1359741111) as well._

arboleya (approved):

packages/fuels/package.json
Dhaiwat10: Can we resolve this conversation since we are using a symlink?

arboleya: Fine by me. Unless @luizstacio thinks otherwise.

luizstacio (approved):

PR #704, opened 2022-12-22
feat: enable update url from provider

luizstacio:
Currently is not possible to update the provider URL instance. The issue with only enabling the replacement of the URL by creating a new provider adds an issue when we want to have a global provider configuration and make sure when changing the URL, all the references to the current Provider will be updated, as currently is needed to create an instance, all the references to the old one will be pointing to the old provider instance.


packages/providers/src/provider.ts
```diff
@@ -206,8 +206,22 @@ export default class Provider {
     /** GraphQL endpoint of the Fuel node */
     public url: string
   ) {
+    this.operations = this.createOperations(url);
```
arboleya, line 209: DRY variable assignment via method reuse.

```suggestion
    this.connect(url)
```

packages/providers/src/provider.ts
```diff
@@ -206,8 +206,22 @@ export default class Provider {
...
     const gqlClient = new GraphQLClient(url);
-    this.operations = getOperationsSdk(gqlClient);
+    return getOperationsSdk(gqlClient);
+  }
+
+  /**
+   * Connect provider to a different Fuel node url
+   */
+  connect(url: string) {
+    this.operations = this.createOperations(url);
```
arboleya, line 224: Connect methods are usually _async_, validate the URL, try to _connect_ somewhere, and throw an error in case of failure. This case doesn't seem to be actually _connecting_, so I'd suggest either:

 1. Change the method name _(`setProviderUrl`?)_
 2. Make the method `async` and run a try/catch validation flow, i.e.:

```suggestion
  async connect(url: string) {
    this.operations = this.createOperations(url);
    try {
      await provider.getNodeInfo(); // or something better
    } catch (error) {
       // handle error, re-throw if necessary
    }
```

packages/providers/src/provider.test.ts
```diff
@@ -152,4 +152,14 @@ describe('Provider', () => {
...
   });
+
+  it('can change the provider url of the curernt instance', async () => {
+    const provider = new Provider('http://127.0.0.1:4000/graphql');
+
+    provider.connect('http://127.0.0.1:8080/graphql');
+
+    expect(async () => {
+      await provider.getNodeInfo();
+    }).rejects.toThrow();
```
arboleya, line 163: Calling a second method _(`getNodeInfo`)_ to validate the first one _(`connect`)_ is not optimal. Only one unit should be considered per test, using mocks if necessary, and the assertion should focus on the `operations` property — the only thing the `connect` method changes.

> Also, try to always error-match using `.toThrow(/my-expected-error/)` to avoid _false-positives_ — when _any_ error can make a test pass, bugs can be silenced.

packages/wallet/src/base-locked-wallet.ts
```diff
@@ -60,7 +60,11 @@ export class BaseWalletLocked extends AbstractWallet {
     if (!provider) {
       throw new Error('Provider is required');
     } else if (typeof provider === 'string') {
-      this.provider = new Provider(provider);
+      if (this.provider) {
+        this.provider.connect(provider);
+      } else {
+        this.provider = new Provider(provider);
+      }
```
arboleya, line 67: I think it'd be good to have a simple test validating all these if/else possibilities.

arboleya (changes requested):

packages/providers/src/provider.ts
luizstacio: I think this is not valid, as the creation of the provider will not be handle by developers making than to add a try catch around the creation of the wallet.

packages/providers/src/provider.ts
luizstacio: This will trigger the TypeScript check as the operations are not assigned on the constructor;

Screenshot 2022-12-22 at 10 27 57

packages/providers/src/provider.ts
luizstacio: Connect is not an async method, as it didn't and should not request data. The application layer is responsible for handling the async operations and errors on bad URL etc.

arboleya (approved):

LuizAsFight (approved):

PR #709, opened 2022-12-26
chore: splitting lib and utils

arboleya:
This PR segregates the exported **library** and _test utilities_.

This ensures that such utilities are not present in the final build of apps using the Fuels SDK.


Dhaiwat10 (commented):
The changes look good but looks like the CI test run is failing.

arboleya:
Yes, I found an unexpected problem when using different builds for `src` and `test`, because then there will be more than one `BaseWalletLocked` reference, and this conditional will malfunction:

https://github.com/FuelLabs/fuels-ts/blob/4c8b0d65eb446d7ae1dbe1f90cf40263b7c6b962/packages/contract/src/contracts/contract-factory.ts#L44-L54

This happens because the imports will be bundled _with_ the `test-utils.js` file and differ from the ones exported with the library at `index.js`:

https://github.com/FuelLabs/fuels-ts/blob/cde2a5d2d73e1f54764e4ba7ca45e6e8d2364011/packages/wallet/src/test-utils.ts#L5-L6

The same would not happen if we'd replace them with the following:

```ts
import { Wallet, WalletUnlocked } from '@fuel-ts/wallet'; 
```

But the latter can't be used because the `test-utils.ts` file is already inside the `@fuel-ts/wallet` package. Packages can't have self-references, I guess. A new `test-utils` package would potentially solve this since all references would point to the `@fuel-ts/wallet` package, like: `import ... from '@fuel-ts/wallet'`.

Initially, I didn't want to go this way (_of creating a new general test utility package_) because I feel that these utilities are better placed inside the packages they belong to. Now I'll have to reevaluate the possibilities.

arboleya:
FWIW: I'm marking the PR as a draft again until I finish this.

arboleya:
> The changes look good but looks like the CI test run is failing.

Fixed via https://github.com/FuelLabs/fuels-ts/pull/709/commits/fb0d58ba473a8aabb0e2c356f7b40004a3b990e7 and subsequent commits.

Dhaiwat10 (approved):

Dhaiwat10 (approved):

Dhaiwat10 (approved):

arboleya:
Regarding the `#context` doc tags like [this one](https://github.com/FuelLabs/fuels-ts/pull/709/files#diff-071995a9d63e58b0b20537c46f7eeb1b1ffe3aa3ab575eb4e818f77158a2ceaaR10-R12), my approach was to duplicate the `#context` lines with the relevant new imports, which now come from the individual packages and not the `fuels` umbrella. I'm just not sure what exactly this `#context` is and if there could be two subsequent lines.

<details>
  <summary>See snippet example</summary>

context

</details>

Dhaiwat10 (approved):

PR #711, opened 2022-12-28
fix: typegen globals expansion

### LINKED ISSUE #710: Typegen missing multiple expanded path globals
Let's say I have two _(or more)_ ABIs inside the `./abis` dir and run this:

```console
$ fuels-typegen -i ./abis/*-abi.json -o ./contracts
```

I'd expect the path global to be expanded and types to be generated for the **_two_** ABI JSON.

However, _only the first_ expanded path is considered, and _the rest_ is completely **ignored**.

arboleya:
- Close #710

### DIFF
--- a/.changeset/sixty-bugs-ring.md
+++ b/.changeset/sixty-bugs-ring.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/abi-typegen": patch
+---
+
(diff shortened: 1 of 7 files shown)


Dhaiwat10 (approved):
LGTM!

PR #712, opened 2022-12-28
feat: Add `getBalance` to contracts and `getContractBalance` to providers

### LINKED ISSUE #696: Add get contract balance to provider
Today if a developer wants to know the balance of a contract, he can write a graphql query but it would be great to include it inside the SDK.

Docs ref: https://graphql-docs.fuel.network/docs/recipes/#get-an-asset-balance-of-a-contract

Dhaiwat10:
- This PR adds two new methods: `getBalance` for contracts, and `getContractBalance` for providers
- Closes #696


arboleya:
> [...] until I figure out how to add tests for both of these new functions

I'm not sure if isolating all e2e's apart from their source code is the best approach, but in the _current_ setting, I believe these tests should go inside the `package/fuel-gauge` — a package solely dedicated to e2e stuff. I assume you'd need to deploy a contract and transfer funds to it first.

You might have to add things to the `chainConfig` as you did; my only concern is how we can keep this file documented since it is tough to read, and it can become increasingly hard to keep track of which test is using what. Currently, the `chainConfig` has many accounts with balances, aiming at the possibility of running tests in parallel, which we do not currently do. You can find some docs about this `initial_state` config [here](https://github.com/FuelLabs/fuels-ts/blob/master/jest.env.ts).

Although I'm trying to give you some thoughts on this matter, please know that I'm unfamiliar with these e2e tests, and I'm looking forward to seeing how you'll address this.

You might get better guidance from @camsjams on this one.

Dhaiwat10:
@arboleya Agreed. That's the approach I am going with now. I wrote an e2e test inside the `fuel-gauge` package and came across an issue (possibly) with our GraphQL API. [Slack convo](https://fuellabs.slack.com/archives/C03FMADHHGB/p1672657909553509)

packages/fuel-gauge/src/token-test-contract.test.ts
```diff
@@ -1,6 +1,6 @@
 import { generateTestWallet } from '@fuel-ts/wallet/test-utils';
 import { readFileSync } from 'fs';
-import { NativeAssetId, toHex, Provider, Wallet, ContractFactory } from 'fuels';
+import { NativeAssetId, toHex, Provider, Wallet, ContractFactory, Address, bn } from 'fuels';
```
arboleya, line 3: It looks like the `Address` import can be removed.

packages/contract/src/contracts/contract.ts
```diff
@@ -62,4 +63,15 @@ export default class Contract implements AbstractContract {
   multiCall(calls: Array<FunctionInvocationScope>) {
     return new MultiCallInvocationScope(this, calls);
   }
+
+  /**
+   * Get the balance for a given assset ID for this contract
+   */
+  getBalance(assetId: BytesLike) {
+    if (!this.provider) {
```
arboleya, line 71: Do you know if there's a case where a contract can be used without a provider?

arboleya (dismissed):
Looking good! 👌 

I'm glad you finally managed to finish it!

I'm approving it as is, just leaving minor observations and questions.

packages/contract/src/contracts/contract.ts
Dhaiwat10: Not sure if it can be _used_ without a provider but I think it can technically be _initialized_ if we don't pass in a provider + we pass in a wallet that does not contain a `provider`. See [this code](https://github.com/FuelLabs/fuels-ts/blob/master/packages/contract/src/contracts/contract.ts#L44-L47)

luizstacio: I think it is not possible to not have a provider in this case, as the constructor assigned the provider from the wallet or the provider it self.

arboleya: > **Note** Just to make it clear, this point I'm touching on might be out of the scope of this PR, but if we agree on changing the params signature, it could become a new issue.

Dhaiwat10: @luizstacio thoughts on this? I can file a separate issue for this if we want to make this change, and we can go ahead with this PR as it is

luizstacio: We should remove the null, as the Contract will not work without a provider or wallet.

packages/providers/src/provider.ts
```diff
@@ -603,6 +603,22 @@ export default class Provider {
...
+    /** The contract ID to get the balance for */
+    contractId: AbstractAddress,
+    /** The asset ID of coins to get */
+    assetId: BytesLike
+  ): Promise<BN> {
+    const { contractBalance } = await this.operations.getContractBalance({
+      contract: contractId.toB256(),
+      asset: hexlify(assetId),
+    });
+    return bn(contractBalance.amount, 10);
```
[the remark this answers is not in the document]
Dhaiwat10: > defaults to bn(amount, hex). No need for the 10

Do `bn(amount, 'hex')` and `bn(amount, 10)` evaluate to the same result? Shouldn't `bn(amount, 16)` evaluate to the same as `'hex'`?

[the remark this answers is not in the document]
luizstacio: Yeah, this is the question if we check fi the value was correct if it's than we have some issues on the file because some of the bn instances are not passing 10.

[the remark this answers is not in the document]
luizstacio: Ok I check the value from the playground it returns a base 10 not a hex:

```
{
  "data": {
    "contractBalance": {
      "amount": "1182963097192509693",
      "assetId": "0x6dfc0f524d3006e103fc4bd733c4f012b5a089fd09361d44833cdf653e9afe5b"
    }
  }
}
```

PR #717, opened 2023-01-03
docs: Update read-only contract call section

### LINKED ISSUE #677: Update read-only sections in the docs

latest

### Steps to Reproduce

The current text under the read-only section is wrong and incomplete. It's using the Rust's SDK terminology (`simulate` instead of `dryRun`, etc.). https://fuellabs.github.io/fuels-ts/guide/contracts/read-only-calls.html. 

This should be updated to reflect the correct API and include more information about when to use `.get` vs `.call`.

### Expected Behavior

NA

### Actual Behavior

NA

Dhaiwat10:
Closes #677

### DIFF
--- a/docs/_guide/contracts/read-only-calls.md
+++ b/docs/_guide/contracts/read-only-calls.md
@@ -2,16 +2,18 @@
 
 # Read-only calls
 
-Sometimes you want to call a contract method that doesn't change the state of the blockchain. For instance, a method that only reads a value from storage and returns it.
+Sometimtes you might want to call a contract method that does not change the state of the blockchain. For instance, a method that only _reads_ a value from storage and returns it without making any changes to storage.
 
-In this case, there's no need to generate an actual blockchain transaction; you only want to read a value quickly.
+In this case, there is no need to create an actual blockchain transaction; you only want to read a value quickly.
 
-You can do this with the SDK. Instead of `.call()`ing the method, use `.simulate()`:
+You can do this with the SDK. Instead of `.call()`ing the method, use `.get()`:
 
-```typescript
-await contract.functions.foo(1336).simulate();
-```
+[@code:typescript](./packages/fuel-gauge/src/contract.test.ts#typedoc:Contract-read-only-call)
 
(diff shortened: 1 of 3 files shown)


docs/_guide/contracts/read-only-calls.md
```diff
@@ -2,16 +2,18 @@
...
-You can do this with the SDK. Instead of `.call()`ing the method, use `.simulate()`:
+You can do this with the SDK. Instead of `.call()`ing the method, use `.get()`:
 
-```typescript
-await contract.functions.foo(1336).simulate();
-```
+[@code:typescript](./packages/fuel-gauge/src/contract.test.ts#typedoc:Contract-read-only-call)
 
-Note that if you use `.simulate()` on a method that _does_ change the state of the blockchain, it won't work properly; it will just `dry-run` it.
+Note that `get()` uses a method called `dryRun()` under the hood, and does not fund the transaction with any coins. If you want to test a read-only call and fund it with coins, use `dryRun()` directly.
```
arboleya, line 13: What's the use case for funding a read-only call? _(using `dryRun` directly)_

Dhaiwat10: @arboleya I'm not sure to be honest but I just tried to follow [this piece of documentation](https://github.com/FuelLabs/fuels-ts/blob/master/packages/contract/src/contracts/functions/base-invocation-scope.ts#L266).

Dhaiwat10: If it's causing more confusion than adding value, we can remove that line

digorithm: The phrasing is confusing, but the use case for running a `dryRun()` with funds attached is valid.

I'd omit that `get()` uses `dryRun()` underneath entirely. `get()` doesn't need funds because it's a true read-only call. 

`dryRun()` might need fund because, although it won't cause any effect on-chain, it could be a test to a transaction that would. 

```suggestion
`get()` doesn't take funding, as it is a read-only call that doesn't alter the chain state. 

If you want to dry run a transaction call that takes funding without altering the chain state, use `dryRun()`. 
```

digorithm:
Gotta sync with master before merging though!

arboleya (approved):

PR #720, opened 2023-01-08
feat: Parse Logs and LogData from all contracts used

### LINKED ISSUE #713: Problem parsing `log()` from external calls
> This problem first came from [this Discord thread](https://discord.com/channels/732892373507375164/1034937869736878160/1052227259785805915)

Given this interaction flow:
 - `User` -> `Contract1` -> `Contract2`

Let's imagine the following:

1. User call `Contract1.method_x`
2. `Contract1.method_x` have a single call to  `Contract2.method_y`
3. `Contract2.method_y` do a simple `log()` call

The flow below will return the `log()` _(from `Contract2`)_ to the user, but it seems like the SDK tries to parse it using the `Contract1` ABI definitions, which seems to be causing problems.

I've put up a simple repro on this [branch](https://github.com/FuelLabs/fuels-ts/tree/aa/chore/external-log-repro) _(diff [here](https://github.com/FuelLabs/fuels-ts/compare/master..aa/chore/external-log-repro))_, and a simple screenshot below for more reference.

<details>
  <summary>Show screenshot</summary>

repro

</details>

camsjams:
Closes #713 

The main idea here is that the receipts data from a transaction has `receipt.id` that are for each contract Id involved, so for every contract that has a log, the log data will show up in the receipts using the associated contract id.

This means we keep track of all attached contracts and then the log data can be parsed correctly using the matching contract's log types.

### DIFF
--- a/.changeset/four-gorillas-admire.md
+++ b/.changeset/four-gorillas-admire.md
@@ -1,0 +1,8 @@
+---
+"@fuel-ts/example-contract": patch
+"fuels": patch
+"@fuel-ts/script": patch
+"@fuel-ts/wallet": patch
+---
+
+Segregating exported library and test utilities
--- a/docs/_guide/testing/index.md
+++ b/docs/_guide/testing/index.md
@@ -11,3 +11,16 @@ In order to test your Sway and TS-SDK applications, you can test your code in a 
 ### Testing with TS-SDK
 
 To test your Sway applications using the TS-SDK, you can pick whatever testing library or framework you feel comfortable with. There isn't any specific testing framework needed, it is entirely up to the user. That being said, the TS-SDK uses [Jest](https://jestjs.io/) for its tests.
+
+### Wallet Test Utilities
+
+You'll often want to create one or more test wallets when testing your contracts.
+
+For this, you can find two simple utilities on the wallet package:
+
+- [`@fuel-ts/wallet`](https://github.com/FuelLabs/fuels-ts/tree/master/packages/wallet#test-utilities)
+
+See also:
+
+1. [Setting up test wallets](../wallets/test-wallets.md)
+2. [Testing with Jest](./testing-with-jest.md)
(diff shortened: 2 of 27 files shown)


packages/abi-coder/src/interface.ts
```diff
@@ -42,13 +42,15 @@ export default class Interface {
   readonly abi: ABI | null;
   readonly types: ReadonlyArray<JsonFlatAbiFragmentType>;
   readonly loggedTypes: ReadonlyArray<JsonAbiLogFragment>;
+  private logTypeHash: { [id: string]: ReadonlyArray<JsonAbiLogFragment> };
```
arboleya, line 45: This variable name could be more self-explanatory.

Here's a simple suggestion to increase clarity.

```ts
/*
  Same as the `loggedTypes` above, but dedicated to external contracts
  added via `<base-invocation-scope.ts>.addContracts()` method. This is
  used to decode logs from contracts other than the main contract
  we're interacting with.
*/
private externalLoggedTypes: { [contractId: string]: ReadonlyArray<JsonAbiLogFragment> };
```

packages/abi-coder/src/interface.ts
```diff
@@ -173,4 +177,8 @@ export default class Interface {
 
     return this.abiCoder.encode(fragment.outputs, values);
   }
+
+  updateLoggedTypes(id: string, loggedTypes: JsonAbiLogFragment[]) {
+    this.logTypeHash[id] = loggedTypes;
+  }
```
arboleya, line 183: This method name suggests that it will update the `loggedTypes` variable, which can be confusing (_when looked at from the outside_) since it modifies the `logTypeHash` instead.

Following my first suggestion of renaming the `logTypeHash` variable to `externalLoggedTypes`, this could follow the same rationale and be something like this: 

```ts
  /*
    Called from `<base-invocation-scope.ts>.addContracts()` method,
    used for saving the `loggedTypes` ABI definitions referring to
    external contracts being added, needed for proper log decoding.
  */
  addExternalLoggedTypes(contractId: string, loggedTypes: JsonAbiLogFragment[]) {
    this.externalLoggedTypes[contractId] = loggedTypes;
  }
```

arboleya (changes requested):

digorithm (approved):
Great work!

PR #721, opened 2023-01-09
test: add test for browser compatibility

### LINKED ISSUE #284: Create test to `browser compatibility`

See also:

Dhaiwat10:
- Related to (but does not close) #284 
- This PR adds one unit test to the `fuels` package that scans the browser bundles of all the child packages exported by the umbrella `fuels` package for any references to Node.js built-in modules that may cause the package to break in a browser environment.
- We have discussed that we should try running all of our tests in a browser environment on top of our current setup which is a Node.js environment, but I don't think we have yet decided on how exactly to go about that. I think this test might be an alternative approach to that same problem, or an added layer of safety in terms of browser compatibility - depends on how we look at it. 

Keeping this a draft for now. Just an idea.


packages/fuels/src/index.test.ts
```diff
@@ -1,5 +1,40 @@
...
+    const moduleName = importMatch[1];
+    listOfModules.add(moduleName);
+  }
+
+  const exportRegex = /export\s+(?:.+\s+from\s+)?["'](.+)["'];?/g;
+  let exportMatch;
+  while ((exportMatch = exportRegex.exec(formattedBundleContent))) {
+    const moduleName = exportMatch[1];
+    listOfModules.add(moduleName);
+  }
```
Dhaiwat10, line 33: The approach that I have gone with is scanning the bundled files for `require` and `import` statements to check for any references to Node.js built-in modules that are only available in a Node environment and not in a browser. I am not sure if this is the absolute best way to go about this, but I have checked that it _does_ get the job done.

arboleya: The rationale is valid, but it'd be hard to have certainty about the complete required stack, since we don't control all the libraries in use. For example, we can import an NPM package that uses some native module, which might be hard or impossible to detect, especially if we consider multiple nested dependencies.

In the end, I believe a declarative approach will suit us better.

One inspiration is how [jest-runner-groups](https://github.com/eugene-manuilov/jest-runner-groups) does it using manual tags.

Dhaiwat10:
Closing this in favor of #728

Issue #615, opened 2022-11-22
Typegen - Add support for Sway Predicates


arboleya:
Also - would we be able to simplify creating JSON abi for Predicates? See this [test file](https://github.com/FuelLabs/fuels-ts/blob/master/packages/fuel-gauge/src/predicate.test.ts#L63-L92). Predicates cannot product ABI from Sway, so developers need to _manually_ type up an ABI JSON structure in order to pass params into Predicates correctly and safely. Its pretty harsh

_Originally posted by @camsjams in https://github.com/FuelLabs/fuels-ts/pull/570#pullrequestreview-1189400462_

arboleya:
@camsjams Do you know why Predicates cannot produce ABI from Sway files?

If we can't start to form a `.sw` file, I can't see how we could skip assembling the JSON by hand.

Ideas?

arboleya:
It seems like we do have ABI generation for Predicates:
 - https://github.com/FuelLabs/sway/issues/2581

Issue #330, opened 2022-06-14
Typegen - Add support for Sway Scripts


AlicanC:
We now generate ABI JSONs for Sway scripts: https://github.com/FuelLabs/sway/pull/1838

We can use those to generate code and types like we do for contracts in `typechain-target-fuels`.

PR #729, opened 2023-01-13
refactor: reenabling Handlebars (`.hbs`) support

arboleya:
As I've started approaching #330 and #615, having a transpilation step for handling `.hbs` became a huge PITA.

This PR revives my initial approach to handling `.hbs` directly, fixing the little intricacies that didn't work before.


Dhaiwat10 (dismissed):
LGTM but I will add that I'm still not very confident about my understanding of the source code of the `abi-typegen` package so I might have missed something 😅

arboleya:
@Dhaiwat10 Thanks for reviewing. The changes here are mainly for configurations and special loaders for `ts-node`, `jest`, and `tsup`, which we need to handle `.hbs` templates directly, as text. I previously tried that but struggled with the `jest` configs. This PR is the conclusion of the first attempt.

luizstacio (approved):

luizstacio (approved):

digorithm (approved):

PR #733, opened 2023-01-16
feat: add helper params to `buildBlockExplorerUrl`

### LINKED ISSUE #622: Add helper params like `txId` and `address` to `buildBlockExplorerUrl`
### Motivation

The [`buildBlockExplorerUrl`](https://github.com/FuelLabs/fuels-ts/blob/dp/issue-600/packages/providers/src/util.ts#L104) helper function requires you to pass in the `path` that you want to append to the base block explorer URL. This is what the current usage looks like:

```ts
const url = buildBlockExplorerUrl({ path: '/address/0x123' });

// or
const url = buildBlockExplorerUrl({ path: '/transaction/0x123456' });
```

Maybe we can improve the DX by adding helper params like `txId`, `address`, `blockNumber`, etc.

### Usage example

This is what the usage could look like:

```ts
const url = buildBlockExplorerUrl({ address: '0x123' });

const url = buildBlockExplorerUrl({ transaction: '0x123456' });
```

### Possible implementations

Implementing these params will require us to assume that all the block explorers that we are building URLs for follow the same conventions for displaying addresses, transactions, etc.

i.e.
`<explorer base url>/address/<address>`
`<explorer base url>/transaction/<transaction id>`

Dhaiwat10:
## Summary
This PR adds three new helper params to the util function `buildBlockExplorerUrl` to make it easier for people to build explorer links.

Closes #622


packages/providers/src/util.ts
```diff
@@ -98,23 +98,87 @@ export const calculateTransactionFee = ({
...
+    },
+    {
+      key: 'txId',
+      value: txId,
+    },
+    {
+      key: 'blockNumber',
+      value: blockNumber,
+    },
+  ];
```
arboleya, line 148: Extra line break.

```suggestion
  ];

```

packages/providers/src/util.ts
```diff
@@ -98,23 +98,87 @@ export const calculateTransactionFee = ({
...
+      key: param.key,
+      value: param.value,
+    }));
+  if (definedValues.length > 1) {
+    throw new Error(
+      `Only one of the following can be passed in to buildBlockExplorerUrl: ${customInputParams
+        .map((param) => param.key)
+        .join(', ')}`
+    );
+  }
```
arboleya, line 161: Extra line break.

```suggestion
  }

```

packages/providers/src/util.ts
```diff
@@ -98,23 +98,87 @@ export const calculateTransactionFee = ({
...
+        .join(', ')}, path`
+    );
+  }
+  if (path && definedValues.length > 0) {
+    throw new Error(
+      `You cannot pass in a path to buildBlockExplorerUrl along with any of the following: ${customInputParams
+        .map((param) => param.key)
+        .join(', ')}`
+    );
+  }
```
arboleya, line 175: Adding fn blocks inside backtick strings can often become hard to read/maintain. Perhaps setting these values on a variable before throwing the error might be cleaner:

```suggestion
    const inputKeys = customInputParams.map(({ key }) => key).join(', ')
    throw new Error(
      `You cannot pass in a path to buildBlockExplorerUrl along with any of the following: ${inputKeys}`
    );
  }
```

packages/providers/src/util.ts
```diff
@@ -98,23 +98,87 @@ export const calculateTransactionFee = ({
...
+  if (definedValues.length === 0 && !path) {
+    throw new Error(
+      `One of the following must be passed in to buildBlockExplorerUrl: ${customInputParams
+        .map((param) => param.key)
+        .join(', ')}, path`
+    );
+  }
```
arboleya, line 168: Extra line break.

```suggestion
  }

```

packages/providers/src/util.ts
```diff
@@ -98,23 +98,87 @@ export const calculateTransactionFee = ({
...
+      key: 'txId',
+      value: txId,
+    },
+    {
+      key: 'blockNumber',
+      value: blockNumber,
+    },
+  ];
+  const definedValues = customInputParams
+    .filter((param) => param.value)
```
arboleya, line 150: Here I assume you want to filter all items with a non-null `value`. For this purpose, I think we could do it like:

```suggestion
    .filter((param) => !!param.value)
```

packages/providers/src/util.ts
```diff
@@ -98,23 +98,87 @@ export const calculateTransactionFee = ({
...
+    .map((param) => ({
+      key: param.key,
+      value: param.value,
+    }));
```
arboleya, line 154:
```suggestion
    .map(({ key, value) => ({ key, value }));

```

packages/providers/src/util.test.ts
```diff
@@ -71,4 +71,45 @@ describe('Providers utils', () => {
...
+    // passing in multiple helper params should throw
+    expect(() =>
+      buildBlockExplorerUrl({
+        blockExplorerUrl: 'explorer.fuel.sh',
+        providerUrl: 'rpc.fuel.sh',
+        address: '0x123',
+        txId: '0x123',
+        blockNumber: 123,
+      })
+    ).toThrow();
```
arboleya, line 85: Please, try always to match something about the thrown Error message to ensure that the error being thrown is the expected one and not another (_sometimes, even unrelated_) error. This can be useful against false positives.

You can use strings or regexes for this.

```suggestion
    ).toThrow(/My pretty regex here/);
```

arboleya (changes requested):
Looking good overall. I added some _(optional)_ aesthetic suggestions and a note about error matching in the tests.

packages/providers/src/util.ts
```diff
@@ -98,23 +98,87 @@ export const calculateTransactionFee = ({
...
+    case 'address':
+      return `address/${value}`;
+    case 'txId':
+      return `transaction/${value}`;
+    case 'blockNumber':
+      return `block/${value}`;
+    default:
+      return `/${key}/${value}`;
+  }
+};
```
LuizAsFight, line 112: can we avoid conditional doing on this style?


```
const pathMap = {
    'address': `address`,
    'txId': `transaction`,
    'blockNumber': `block`
  };
  const path = pathMap[key] || key;
  return `${path}/${value}`;
```

LuizAsFight (changes requested):

Dhaiwat10:
Thanks for the suggestions @arboleya @LuizAsFight, super helpful!

arboleya (approved):

LuizAsFight (approved):

PR #745, opened 2023-01-20
feat!: add main arg support to Scripts

### LINKED ISSUE #698: Add support for `Script` `main()` arguments
Currently, you can pass the script data to a `Script`, but if the script you're integrating with takes `main()` arguments, I believe there's no interface to handle it, [like in the Rust SDK](https://fuellabs.github.io/fuels-rs/v0.33.0/getting-started/running-scripts.html).

This support should be added.

camsjams:
closes #698 

Follow-up in here for additional testing: #757


packages/abi-coder/src/interface.ts
```diff
@@ -150,6 +150,29 @@ export default class Interface {
     return concat([selector, new BooleanCoder().encode(isRef), args]);
   }
 
+  encodeMainFunctionData(
```
arboleya, line 153: These two methods are almost identical; do we need them both?

 - `encodeMainFunctionData`
 - `encodeFunctionData`

I see the little differences in the code, but I need help understanding them.

I guess the questions are:
 1. Are `contract` functions different from `script` functions?
 2. Could something else be changed _(in the `ScriptFactory` or elsewhere)_ to avoid duplicating this method?

packages/script/src/script-factory.ts
```diff
@@ -0,0 +1,106 @@
...
+  wallet: BaseWalletLocked;
+
+  constructor(bytecode: BytesLike, abi: JsonAbi | Interface, wallet: BaseWalletLocked) {
+    this.bytecode = bytecode;
+    this.wallet = wallet;
+    if (abi instanceof Interface) {
+      this.interface = abi;
+    } else {
+      this.interface = new Interface(abi);
+    }
```
arboleya, line 38: Overloading the `abi` parameter seems unnecessary; we could accept `JsonAbi` only.

```suggestion
  constructor(bytecode: BytesLike, abi: JsonAbi, wallet: BaseWalletLocked) {
    this.bytecode = bytecode;
    this.wallet = wallet;
    this.interface = new Interface(abi);
```

packages/script/src/script-factory.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  async callScript(args: Data): Promise<{
+    transactionResult: TransactionResult<any>;
+    response: TransactionResponse;
+    value: any;
+    logs: any[];
+  }> {
+    if (!this.wallet) {
+      return logger.throwArgumentError('Cannot call without wallet', 'wallet', this.wallet);
+    }
```
arboleya, line 91: I wonder if it'd make sense to receive the `wallet` as a parameter on the `call()` method instead.

docs/_guide/scripts/calling-a-script.md
```diff
@@ -4,4 +4,16 @@
...
+
+You can access this low-level helper using `import { callScript } from 'fuels';`.
+
+## Calling a script with a `main` function that takes arguments
+
+Suppose your Sway script `main` function is written using the arguments passed to the `main` function like so:
+
+[@code:rust](./packages/fuel-gauge/test-projects/script-main-args/src/main.sw#typedoc:script-with-main-args)
+
+You can still hand code out a solution wrapper using `callScript` utility to call your script with data. However, if you prefer to use the ABI generated from your script, you can use the `ScriptFactory` helper:
```
arboleya, line 17: I wonder if having multiple ways of calling `Script`'s can cause more confusion than clarity.

I think the approach in `ScriptFactory` is superior and should be the only one.

packages/fuel-gauge/src/script-main-args.test.ts
```diff
@@ -0,0 +1,82 @@
...
+};
+
+describe('Script Coverage', () => {
+  it('can call script and use main arguments', async () => {
+    const wallet = await setup();
+    // #region typedoc:script-call-factory
+    const scriptInstance = new ScriptFactory<BigNumberish>(scriptBin, scriptAbi).script;
+    const foo = 33;
+
+    const { value, logs } = await wallet.callScript<BigNumberish[], BigNumberish>(scriptInstance, [
```
luizstacio, line 35: I think this types should follow the types of [InvokeFunction](https://github.com/FuelLabs/fuels-ts/blob/cef77013e14461b07a525213046859ba1013dd29/packages/contract/src/types.ts#L34) it will help too keep consistency and I think it what you are looking for.

```suggestion
    const { value, logs } = await wallet.callScript<BigNumberish[], BigNumberish>(scriptInstance, [
```

packages/script/src/script-factory.ts
```diff
@@ -0,0 +1,145 @@
...
+    const response = await this.provider.sendTransaction(request);
+    const transactionResult = await response.waitForResult();
+    const result = this.script.decodeCallResult(transactionResult);
+
+    return {
+      transactionResult,
+      response,
+      value: result.value,
+      logs: result.logs,
+    };
```
arboleya, line 143: @camsjams Quick questions:

1. Do the `Contract` calls also expose internal req/res stuff?
2. Why are they helpful in the `Script` context?

Besides, I usually like wrapping non-essential output data in a `meta` property to indicate that they are secondary.

See what you think.

```suggestion
    const request = await this.buildScriptTransaction<TInput>(data, options);
    const response = await this.provider.sendTransaction(request);
    const result = await response.waitForResult();
    const meta = { request, response, result }
    
    const { value, logs } = this.script.decodeCallResult(transactionResult);

    return { value, logs, meta }
```

packages/script/src/script-factory.ts
```diff
@@ -0,0 +1,145 @@
...
+    options?: BuildScriptOptions
+  ): Promise<
+    {
+      transactionResult: TransactionResult<'success' | 'failure'>;
+      response: TransactionResponse;
+    } & Result<TOutput>
+  > {
+    if (!this.provider) {
+      throw new Error('Provider is required');
+    }
```
arboleya, line 131: This condition seems unnecessary since now the constructor requires a `walletOrProvider` to be non-null.

```suggestion
```

arboleya (changes requested):
Looking good in general; I've just left some final comments.

----

At last, I noticed you went with an `options` parameter in the `ScriptFactory.call()` method instead of the chained-methods approach used in `Contract`'s, i.e.:
 - `contract.txParms(..).call(...)`

Without full knowledge of the discrepancies in play, I feel it would be best to have a similar API for interacting with `Script`, `Predicate`, and `Contract` — using chained methods if we'd mimic the current `Contract` implementation.

I can file a new issue for evaluating this option if you agree.

What do you think?

> **Note** _I believe @luizstacio also mentioned something related to the pattern defined by `InvokeFunction`, which could help with standardization._

arboleya:
@camsjams Using similar APIs could be more intuitive. However, I still need to get more familiar with everything, and your reasoning about the differences makes sense to me.

packages/script/src/script-factory.ts
```diff
@@ -0,0 +1,141 @@
...
+      this.provider = walletOrProvider.provider;
+      this.wallet = walletOrProvider;
+    } else {
+      this.provider = walletOrProvider;
+      this.wallet = null;
+    }
+
+    this.script = new Script(
+      bytecode,
+      (args: InputValue<void>[]) =>
```
luizstacio, line 50: Extract this to another function in the class would help with code readability if this is intended to be use only internal we can make it private

luizstacio: Maybe create a `buildScript`?

luizstacio (changes requested):

luizstacio (approved):

digorithm:
> > @arboleya
> > At last, I noticed you went with an `options` parameter in the `ScriptFactory.call()` method instead of the chained-methods approach used in `Contract`'s, i.e.:
> > 
> > * `contract.txParms(..).call(...)`
> 
> This is a good point, although the reason I excluded this is that the chainability is more useful in Contracts where you can change the:
> 
> * required coins
> * add contracts
> * add call/ add calls
> 
> However, the `call` function within https://github.com/FuelLabs/fuels-ts/blob/master/packages/contract/src/contracts/functions/base-invocation-scope.ts#L210 still has the options as a param there, so my goal was to at least match that basic feature.
> 
> Contracts have a lot more features since there are multiple functions, multicalls, and more going on.
> 
> But yes, I will refactor to use the `InvokeFunction` and `FunctionInvocationScope` patterns more closely. I think the main concern I had was that those are defined in `Contract` which depends on `Script` so it involves some deeper refactoring.
> 
> @arboleya @digorithm @Dhaiwat10 @luizstacio please let me know if you have any questions. Otherwise I am going to update per my last comment here.

I understand that the fact that `Script`s are simpler, you went with a simpler configuration approach (optional parameter). 

But, I believe API consistency should take precedence here. I can definitely see people complaining because, for contracts you'd do `contract.txParms(...).call(...)` and then, when running a script, the _same_ configuration (tx params) would now be `script.call([params...], {...})`. I'm sure it's worth delaying a bit and putting some effort to keep the API consistent. Note that this consistency would even span across _languages_, as the Rust SDK is also like that (https://fuellabs.github.io/fuels-rs/v0.36.0/getting-started/running-scripts.html?highlight=script#running-scripts-with-transaction-parameters). That'd be such a nice detail 😃

packages/script/src/functions/invocation-scope.ts
```diff
@@ -0,0 +1,204 @@
...
+              receipt: scriptResult.returnReceipt,
+            }
+          );
+        }
+
+        let value;
+        if (scriptResult.returnReceipt.type === ReceiptType.Return) {
+          value = scriptResult.returnReceipt.val;
+        }
+        if (scriptResult.returnReceipt.type === ReceiptType.ReturnData) {
```
luizstacio, line 69: Can we use a switch instead of if's here?

luizstacio (changes requested):
-  Small comment to improvement the code
- Add a comment to remove the warning of any
- I think we have room to create a common abstract code between the script and contract and reduce the amount of "duplication" don't know if needs to be now.

packages/providers/src/provider.ts
```diff
@@ -190,6 +190,10 @@ export type BuildPredicateOptions = {
   fundTransaction?: boolean;
 } & Pick<TransactionRequestLike, 'gasLimit' | 'gasPrice' | 'maturity'>;
 
+export type BuildScriptOptions = {
+  fundTransaction?: boolean;
+} & Pick<TransactionRequestLike, 'gasLimit' | 'gasPrice' | 'maturity'>;
+
```
arboleya, line 196: @camsjams It looks like `BuildScriptOptions` is not used anywhere. 

Or is it?

arboleya (changes requested):
Left a minor question/request about `BuildScriptOption`.

----

At this point, this PR has grown considerably since the first iteration, and I had a hard time understanding all the interconnected parts. I must admit I still didn't.

However, I feel that we started duplicating logic, as @luizstacio pointed out, especially on the `invocation-{scope,results}` side of things. On the one hand, I fear that if we push this PR too far ahead, we might suffer from the delay; but I also fear that duplication might be harder to purge as time passes.

By only reading the code, my instinct tells me that the `contract` stuff might derive from `script`, as `script` seems simpler. But again, it's hard for me to estimate the effort required here, and I don't know if it should be done on this PR or in the future.

Please, feel free to do as you see fit, but I'd appreciate it if you could take a few mins to elaborate on your reasons and share your thoughts so we can all be on the same page.

Torres-ssf (approved):

packages/script/src/script-factory.ts
```diff
@@ -0,0 +1,23 @@
...
+    this.bytecode = bytecode;
+    this.interface = new Interface(abi);
+
+    this.account = account;
+  }
+
+  prepareScript(): Script<TInput, TOutput> {
+    return new Script<TInput, TOutput>(this.bytecode, this.interface, this.account);
+  }
+}
```
luizstacio, line 23: I think we could remove ScriptFactory and use directly `new Script` without the need first to do `new ScriptFactory` and then do `prepareScript()`.

luizstacio (approved):

Torres-ssf (approved):

packages/program/src/script-request.ts
```diff
@@ -0,0 +1,183 @@
...
+    | TransactionResultRevertReceipt;
+  callResult: CallResult;
+};
+
+function callResultToScriptResult(callResult: CallResult): ScriptResult {
+  const receipts = [...callResult.receipts];
+
+  // Every script call ends with two specific receipts
+  // Here we check them so `this.scriptResultDecoder` doesn't have to
+  const scriptResultReceipt = receipts.pop();
```
digorithm, line 46: I know this is not coming from this PR, but this refactor just brought this to my attention now: we're hard-coding the assumptions that the receipts are ordered. If that changes in the future, it'll trigger some gnarly errors. 

I'm sure it's fine for now, but we should, maybe, refactor this not to assume order and filter the receipts we're looking for in the `receipts`. 

It shouldn't be a gating thing; we can open an issue for this, incur the tech debt now, and address this later.

digorithm (approved):
Good stuff -- looks pretty clean to me.

packages/program/src/script-request.ts
arboleya: Issue created:
 - https://github.com/FuelLabs/fuels-ts/issues/802

arboleya (approved):
The addition of the `program` package couldn't feel more appropriate. 😌

Dhaiwat10 (approved):
Brilliant!

PR #746, opened 2023-01-20
feat: add main arg support to Predicates 

### LINKED ISSUE #700: Add support for `Predicate` `main()` arguments
Currently, you can pass the predicate data to a `Predicate`, but if the predicate you're integrating with takes `main()` arguments, I believe there's no interface to handle it, like in the Rust SDK.

This support should be added.

camsjams:
closes #700

### DIFF
--- a/packages/providers/src/provider.ts
+++ b/packages/providers/src/provider.ts
@@ -604,6 +604,22 @@ export default class Provider {
   }
 
   /**
+   * Returns the balance for the given contract for the given asset ID
+   */
+  async getContractBalance(
+    /** The contract ID to get the balance for */
+    contractId: AbstractAddress,
+    /** The asset ID of coins to get */
+    assetId: BytesLike
+  ): Promise<BN> {
+    const { contractBalance } = await this.operations.getContractBalance({
+      contract: contractId.toB256(),
+      asset: hexlify(assetId),
+    });
+    return bn(contractBalance.amount, 10);
+  }
+
+  /**
    * Returns the balance for the given owner for the given asset ID
    */
   async getBalance(
--- a/.changeset/shy-moles-pay.md
+++ b/.changeset/shy-moles-pay.md
@@ -1,0 +1,6 @@
+---
+"@fuel-ts/contract": minor
+"@fuel-ts/providers": minor
+---
+
+Add `getBalance` method for contracts and `getContractBalance` method for providers
--- a/packages/contract/src/contracts/contract.ts
+++ b/packages/contract/src/contracts/contract.ts
@@ -1,3 +1,4 @@
+import type { BytesLike } from '@ethersproject/bytes';
 import type { FunctionFragment, JsonAbi, JsonFlatAbi } from '@fuel-ts/abi-coder';
 import { Interface } from '@fuel-ts/abi-coder';
 import { Address } from '@fuel-ts/address';
@@ -61,5 +62,16 @@ export default class Contract implements AbstractContract {
 
   multiCall(calls: Array<FunctionInvocationScope>) {
     return new MultiCallInvocationScope(this, calls);
+  }
+
+  /**
+   * Get the balance for a given assset ID for this contract
+   */
+  getBalance(assetId: BytesLike) {
+    if (!this.provider) {
+      throw new Error('Contract instance has no provider.');
+    }
+
+    return this.provider.getContractBalance(this.id, assetId);
   }
 }
--- a/packages/fuel-gauge/src/token-test-contract.test.ts
+++ b/packages/fuel-gauge/src/token-test-contract.test.ts
@@ -1,6 +1,6 @@
 import { generateTestWallet } from '@fuel-ts/wallet/test-utils';
 import { readFileSync } from 'fs';
-import { NativeAssetId, toHex, Provider, Wallet, ContractFactory } from 'fuels';
+import { NativeAssetId, toHex, Provider, Wallet, ContractFactory, bn } from 'fuels';
 import type { BN } from 'fuels';
 import { join } from 'path';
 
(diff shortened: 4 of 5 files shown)


Dhaiwat10 (approved):
🫡

packages/predicate/src/predicate.ts
```diff
@@ -1,20 +1,41 @@
...
+      if (mainFunction !== undefined) {
+        this.types = mainFunction.inputs;
+      } else {
+        logger.throwArgumentError(
+          'Cannot use ABI without "main" function',
+          'Function fragments',
+          abiInterface.fragments
+        );
+      }
+    }
```
arboleya, line 39: If we'd stop overloading the `types` parameter by accepting only `JsonAbi`:

```suggestion
    const abiInterface = new Interface(types);
    const mainFunction = abiInterface.fragments.find(({ name }) => name === 'main');
    
    if (mainFunction !== undefined) {
      this.types = mainFunction.inputs;
    } else {
      logger.throwArgumentError(
        'Cannot use ABI without "main" function',
        'Function fragments',
        abiInterface.fragments,
      );
    }
```

arboleya (changes requested):
Looking good overall. 👌 

I still have specific concerns with parameter type _overloading_ and the trade-off between:
 - `flexibility` _(not always essential)_
 - `simplicity` _(mostly desirable)_

My suggestions are based on the latter. You're free to dismiss them.

I'd love to understand the reasoning behind the decision.

Thanks.

docs/_guide/predicates/index.md
```diff
@@ -13,3 +13,11 @@ Similar to contracts, once you've written a predicate in Sway and compiled it wi
...
 
 In the next section, we show how to interact with a predicate and explore an example where specific signatures are needed to spend the predicate funds.
+
+## Calling a predicate with a `main` function that takes arguments
+
+Suppose your Sway predicate `main` function is written using the arguments passed to the `main` function like so:
+
+[@code:rust](./packages/fuel-gauge/test-projects/predicate-main-args-struct/src/main.sw#typedoc:Predicate-main-args)
+
+You can still create a `Predicate` in the same way as above, but you can pass in a `JsonAbi | Interface` using the JSON generated from `forc`.
```
arboleya, line 23: Now it's `JsonAbi`, the mention of `Interface` can be removed.
```suggestion
You can still create a `Predicate` in the same way as above, but you can pass in a `JsonAbi` using the JSON generated from `forc`.
```

packages/fuel-gauge/src/predicate.test.ts
```diff
@@ -1,11 +1,13 @@
 import { generateTestWallet } from '@fuel-ts/wallet/test-utils';
 import { readFileSync } from 'fs';
-import { Address, NativeAssetId, bn, toHex, toNumber, Provider, Predicate } from 'fuels';
+import { Address, NativeAssetId, bn, toHex, toNumber, Provider, Predicate, Interface } from 'fuels';
```
arboleya, line 3: `Interface` isn't used and can be removed.

```suggestion
import { Address, NativeAssetId, bn, toHex, toNumber, Provider, Predicate } from 'fuels';
```

arboleya (changes requested):

packages/predicate/src/predicate.ts
```diff
@@ -1,20 +1,39 @@
...
 
-  constructor(bytes: BytesLike, types?: ReadonlyArray<JsonAbiFragmentType>) {
+  constructor(bytes: BytesLike, types?: JsonAbi) {
     super();
     this.bytes = arrayify(bytes);
     this.address = Address.fromB256(ContractUtils.getContractRoot(this.bytes));
-    this.types = types;
+
+    if (types) {
+      const abiInterface = new Interface(types as JsonAbi);
```
luizstacio, line 26: It is missing the assignment to variable `this.interface`.
```suggestion
           this. interface = new Interface(types as JsonAbi);
```

packages/fuel-gauge/src/predicate.test.ts
```diff
@@ -352,4 +441,56 @@ describe('Predicate', () => {
...
+    const predicate = new Predicate(testPredicateStruct, predicateMainArgsStructAbi);
+
+    const initialPredicateBalance = await setupPredicate(wallet, amountToPredicate, predicate);
+    const initialReceiverBalance = await wallet.provider.getBalance(receiverAddress, NativeAssetId);
+
+    const validation: Validation = {
+      has_account: true,
+      total_complete: 100,
+    };
+    await wallet.provider.submitSpendPredicate(
```
luizstacio, line 458: We should add a type optional argument here to inform the Validation types. It will help to generate better types after.

luizstacio (changes requested):

LuizAsFight (approved):

packages/providers/src/provider.ts
```diff
@@ -752,7 +752,7 @@ export default class Provider {
     let encoded: undefined | Uint8Array;
     if (predicateData && predicate.types) {
       const abiCoder = new AbiCoder();
-      encoded = abiCoder.encode(predicate.types, predicateData);
+      encoded = abiCoder.encode(predicate.types, predicateData as InputValue[]);
```
arboleya, line 755: Shouldn't this be `InputValue<T>[]`? _(see method signature on line `735`)._

Does it make any difference?

```suggestion
      encoded = abiCoder.encode(predicate.types, predicateData as InputValue<T>[]);
```

arboleya (approved):
Approving as is, but I've left one last comment about Generics.

luizstacio (approved):

packages/providers/src/provider.ts
arboleya: Oh, my bad. You are right.

It looks like `values: InputValue[]` is expected.


https://github.com/FuelLabs/fuels-ts/blob/f521146c328a7fb2c98679ec3f0c9aa6df2f684f/packages/abi-coder/src/abi-coder.ts#L122

PR #758, opened 2023-02-03
fix: allow to use buildBlockExplorer without any param

### LINKED ISSUE #752: allow to use `buildBlockExplorer` without any param
### Motivation

Currently when using `buildBlockExplorer`, the user needs to inform some params, or the function will throw an error [here](https://github.com/FuelLabs/fuels-ts/blob/cef77013e14461b07a525213046859ba1013dd29/packages/providers/src/util.ts#L175)

then in front-end I'm needing to workaround, passing a empty space in path
```
    buildBlockExplorerUrl({
      providerUrl: currentNetworkUrl,
      path: ' ',
    }),
``` 

In my case I'm passing a custom `providerUrl` , but even only `buildBlockExplorerUrl()` should just work only the correct block-explorer URL

### Usage example

_No response_

### Possible implementations

_No response_

Dhaiwat10:
Closes #752

### DIFF
--- a/.changeset/wise-squids-relate.md
+++ b/.changeset/wise-squids-relate.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/providers": patch
+---
+
+Make `buildBlockExplorer` more flexible. It won't throw when not passed anything now.
--- a/packages/providers/src/util.test.ts
+++ b/packages/providers/src/util.test.ts
@@ -30,13 +30,21 @@ const testBlockExplorerUrlWithInputs = ({
 
 describe('Providers utils', () => {
   test('buildBlockExplorerUrl - empty/undefined inputs', () => {
-    const url = buildBlockExplorerUrl({
-      blockExplorerUrl: undefined,
-      path: '/transaction/0x123',
-      providerUrl: undefined,
-    });
-
-    expect(url).toEqual(`${DEFAULT_BLOCK_EXPLORER_URL}/transaction/0x123`);
+    expect(buildBlockExplorerUrl({})).toEqual(`${DEFAULT_BLOCK_EXPLORER_URL}/`);
+    expect(
+      buildBlockExplorerUrl({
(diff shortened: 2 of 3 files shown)


arboleya (approved):
Good one. 👍 

This utility has grown in complexity since the beginning, reinforcing the potential need to split the `util.ts` file into multiple smaller ones, as we initially discussed.

The vast array of options made the code more challenging to read, something that could also be improved with a more granular split using multiple functions.

Not pertinent to this PR, but we could consider refactoring it in the future.

> _This [PR](https://github.com/FuelLabs/fuels-ts/pull/766) does a similar split_.

LuizAsFight (changes requested):
Looking good, test coverage also ok 💯 


I would suggest a small change only:


RN to use it without params, we need to inform empty object., right? `buildBlockExplorer({})`.

I think we should support `buildBlockExplorer()`


it can be easily achieved by changing entrance of `buildBlockExplorer` function to something like:
```
export const buildBlockExplorerUrl = (options: {
  blockExplorerUrl?: string;
  path?: string;
  providerUrl?: string;
  address?: string;
  txId?: string;
  blockNumber?: number;
} = {}) => {
  const {
    blockExplorerUrl,
    path,
    providerUrl,
    address,
    txId,
    blockNumber,
  } = options;
  const explorerUrl = blockExplorerUrl || DEFAULT_BLOCK_EXPLORER_URL;
```

LuizAsFight:
> 

also, add this empty situation on test

Dhaiwat10:
Ah, it's that flaky test (`Transaction id already in use`) failing again.

arboleya (approved):

LuizAsFight (approved):

arboleya:
@Dhaiwat10 I wonder why this wasn't auto-merged yet.

@camsjams Maybe you need to re-approve it since you're the one who requested changes? 🤔

PR #760, opened 2023-02-03
test: Increase coverage on Contract Class

### LINKED ISSUE #759: Increase test coverage of Contract Class
Per #753 to increase overall test coverage and quality.

Missing test coverage of `Contract` class in `package:contract` of `getBalance` and `multiCall` methods.

danielbate:
Adding coverage of `getBalance` and `multiCall` methods.

Closes #759


danielbate:
@Dhaiwat10 thankyou for running the workflow. Looks to be a flaky test? Please could you re-run? I can look into it.

arboleya:
Updating the PR with the `master` branch seems to have broken a test.

danielbate:
@arboleya @Torres-ssf I have removed the problem test case that had broke after merging latest master, as I believe it is now an invalid case.

However, I think that the check for a falsy provider is now redundant given the recent change to make `walletOrProvider` a required param for the `Contract` constructor.

https://github.com/FuelLabs/fuels-ts/blob/9eaad60e1aceffc31882175e9e81ad202bfe41f9/packages/contract/src/contracts/contract.ts#L67-L76

Torres-ssf:
@danielbate Yes. It looks like this validation is no longer required.

Torres-ssf (approved):

danielbate:
@arboleya @arboleya I have raised #769 to handle the refactoring of that function as it seems out of scope of this work.

packages/contract/src/contracts/contract.test.ts
```diff
@@ -36,13 +37,36 @@ describe('Contract', () => {
...
+      'Contract instance has no provider.'
+    );
+  });
+
+  test('Contract instance can multi call functions', async () => {
+    const wallet = Wallet.generate();
+    const contract = new Contract(CONTRACT_ID, ABI, wallet);
+    const calls = [contract.functions.foo(123)];
+    expect(contract.multiCall(calls)).toBeInstanceOf(MultiCallInvocationScope);
+  });
```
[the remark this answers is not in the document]
danielbate: Thankyou for your input @camsjams 😃 Yes this is a very basic validation check for module level coverage, it would be good to understand the test plan for the project, not sure if that is documented anywhere? In some places, especially within fuel-gauge, it seems to cover more feature level testing of a range of packages/functionality. And then some packages do seem to utilise more unit level testing. 

I am quite pro utilising both, rather than just feature, especially in the case of SDKs where modules themselves could be further extrapolated and packages then become untested in isolation.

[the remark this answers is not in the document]
arboleya: @danielbate You got it right; the `fuel-gauge` package currently groups all our e2e tests interacting with the Fuel-Core node, while most unit tests are kept side-by-side with their source counterpart.

packages/program/src/contract.test.ts
```diff
@@ -36,13 +38,49 @@ describe('Contract', () => {
     expect(contract.account).toBe(null);
   });
 
+  test('Contract instance with provider can get balance', async () => {
+    const provider = new Provider('http://localhost:4000/graphql');
+    const contract = new Contract(CONTRACT_ID, ABI, provider);
+    expect(await contract.getBalance([1])).toStrictEqual(bn(0));
```
Dhaiwat10, line 44: I feel like we can slightly improve these assertions. These assertions are checking if the `getBalance` function returns `0` for a non-existent asset ID IIUC. We should instead be checking that the function returns the _correct_ balance for an actual asset ID with a non-zero balance. Thoughts?

arboleya: Makes sense.

danielbate: Thank you @Dhaiwat10 I will make that change. Currently in a position where I am still getting a feel for the amount of testing required at this code level, a lot of what I am putting through at the moment are validation like tests.

I'm also seeing a lot of places where we are not mocking classes that are not under test within the scope of the unit test. In this test, the `Provider` would be an example of that. I think we should be mocking this so we are not dependent on the behaviour or state of other dependencies for our unit test. I'd be quite keen to refactor this throughout the SDK. What are your thoughts on this?

arboleya: @danielbate I agree. However, adjusting it for the whole repo _(at once)_ can become quite an endless task. Doing it in a few places first might give an idea of how difficult it will be, and then doing it in smaller amounts _(per package or test file)_ would make it easier to review.

Also, since many things are interconnected, I believe we will need to share mock helpers between packages to ease the pain of constructing mocks from scratch for every test. I did similar setups in the past; we can talk about it later if you'd like.

danielbate: @arboleya sounds good and that would be great, thankyou!

danielbate:
Converting to draft as I'm taking a bit of a deeper look into mocking some of those other classes. Bit of a headache with the amount of class inheritance, static methods and readonly properties so not looking like the cleanest solution but will push a WIP version a bit later on. Based on conversations, I think isolating the service under test will be worthwhile so going to use this scope as a proving ground.

PR #763, opened 2023-02-06
feat: methods without #[payable] annotation cannot accept coins

### LINKED ISSUE #675: Methods without #[payable] annotation cannot accept coins
### Motivation

Once https://github.com/FuelLabs/sway/pull/3450 is merged the SDK also needs to forbid accepting coins in non-payable functions. The current plan is to introduce function attributes into the JSON ABI schema to mark payable functions as such. I'm happy to take any comments/suggestions on this.

Spec issue: https://github.com/FuelLabs/fuel-specs/issues/446
Compiler change: https://github.com/FuelLabs/sway/pull/3450
fuels-rs issue: https://github.com/FuelLabs/fuels-rs/issues/742

### Usage example

_No response_

### Possible implementations

_No response_

Dhaiwat10:
## Summary
- This PR adds a new field called `attributes` to the `FunctionFragment`s generated by `abi-coder`.
- The `callParams` function now reads this `attributes` field to scan for a `payable` attribute before allowing forwarding of any coins or assets.
- Closes #675


arboleya:
@Dhaiwat10 Great work! It looks practically finished.

Just summarizing what we spoke about, I think the requirements are:

 - `abi-typegen` Update `JSON ABI`-related interfaces (_the rest can probably be rolled back_)
 - `abi-coder` Update `JSON ABI`-related interfaces and function attributes' parser
 - `contract` The crux of the validation, which seems to be spot on already

As for the [_`Text file busy`_ error](https://github.com/FuelLabs/fuels-ts/actions/runs/4112113498/jobs/7096642709#step:5:1493), I remember we had a [fix](https://github.com/FuelLabs/fuels-ts/pull/557) for it a while back.

Maybe @camsjams has some insight about it.

Dhaiwat10:
@arboleya I have rolled back _everything_ that I changed inside of `abi-typegen` because 1. the interfaces were simple enough so we can always put them back in if or when we need them later, and 2. we don't need anything related to attributes on our types right now. Keeping the code that we would not be using right now felt a little unnecessary and could lead to confusions in the future. Let me know if this sounds good!

arboleya:
Yes, that's fine.

At some point, I'll have to sync the ABI `coder` and `typegen`.

They should share common interfaces, especially the ones describing the JSON ABI.

Dhaiwat10:
@arboleya this PR is looking good locally. Not being able to find a way around the build error here though.

arboleya:
@Dhaiwat10 Working fine locally. 

@camsjams Based on the past [fix](https://github.com/FuelLabs/fuels-ts/pull/557) for this issue, do you imagine what might be happening?

> _Although I can't explain, I guess the error is derived from the new `forc` version._

packages/fuel-gauge/src/payable-annotation.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    contract.functions
+      .non_payable()
+      .callParams({
+        forward: {
+          amount: bn(100),
+          assetId: NativeAssetId,
+        },
+      })
+      .call()
+  ).rejects.toThrowError(/Function is not payable/);
```
Torres-ssf, line 45: Looks real good @Dhaiwat10 :clap: 
Just as was pointed out, I also believe we can split this test into 2 parts.

Dhaiwat10:
Let's wait for #800 to get merged, that should fix this failing test.

image

arboleya (approved):

Torres-ssf (approved):

PR #764, opened 2023-02-06
fix: make `walletOrProvider` a required param for the `Contract` constructor

### LINKED ISSUE #755: Make `walletOrProvider` a required param for the `Contract` constructor
Currently, it is possible to pass in `null` to the `walletOrProvider` param when initializing a contract and its default value is also `null`:
https://github.com/FuelLabs/fuels-ts/blob/36f99093def4a06137bad3dfb86baa57b69bb0f4/packages/contract/src/contracts/contract.ts#L23

This shouldn't be the case because a contract cannot work if `walletOrProvider` is `null`. It needs at least one of either a `Wallet` or a `Provider`.

See [this discussion](https://github.com/FuelLabs/fuels-ts/pull/712#discussion_r1086860026) for more context.

Torres-ssf:
- Close #755

### DIFF
--- a/.changeset/lazy-trains-admire.md
+++ b/.changeset/lazy-trains-admire.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/contract": minor
+---
+
+Make parameter walletOrProvider required in Contract constructor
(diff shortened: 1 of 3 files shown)


arboleya (approved):
Luckily, there was only one place using the `null` optionality. 🙂

Dhaiwat10 (approved):

PR #766, opened 2023-02-06
fix: stringify / parse transaction request 

luizstacio:
Currently, when parsing a transaction to JSON and trying to reconstruct it, some of the values generated in the JSON are not compatible with the constructor method. On the previous PR #561 we have fixed by implementing a custom constructor `arraifyFromUint8Array` fixing just the script data, now expanding the scope of what we can do on the Wallet. We also notice the issue happens on `witnesses` data when deploying a contract and can also happen with any value of type Unit8Array, Address, or BN.

In this PR I have removed the `arraifyFromUint8Array` and implemented a `normalizeJSON` utility, implementing a `toJSON` method on `BaseTransaction`. This creates a fully compatible JSON object that can be passed to the `transactionRequestify` and eliminates the need for customs checks on the constructor.

This PR does;
- Implements toJSON on BaseTransaction using normalizeJSON.
- Breaks the util file and util.test.ts into smaller chunks.
- Enforces bytecode to be a Unit8Array during constructor time, avoiding Buffer-like objects.
- Splits the deployContract function into two methods, deployContract, and createTransactionRequest.
- Adds tests for deploy contract transactions, ensuring they can be stringified and parsed back.


packages/providers/src/utils/sleep.ts
```diff
@@ -0,0 +1,7 @@
+export function sleep(time: number = 1000) {
```
arboleya, line 1: I'd discourage using a default value for this parameter since we can't assume anything about it.

> **Note** _We could replace this with [`sleep-promise`](https://www.npmjs.com/package/sleep-promise), but perhaps that'd be trickier to mock in tests, in which case having a simple function may still have its place._

```suggestion
export function sleep(time: number) {
```

packages/contract/src/contracts/contract-factory.ts
```diff
@@ -87,18 +85,29 @@ export default class ContractFactory {
...
+    return {
+      contractId,
+      transactionRequest,
+    };
+  }
 
-    const response = await this.wallet.sendTransaction(request);
+  async deployContract(deployContractOptions?: DeployContractOptions) {
+    if (!this.wallet) {
+      return logger.throwArgumentError('Cannot deploy without wallet', 'wallet', this.wallet);
```
arboleya, line 105:
```suggestion
      return logger.throwArgumentError('Cannot deploy Contract without wallet', this.wallet);
```

arboleya (changes requested):
Good stuff! 🚀 — left minor comments and suggestions.

I loved that you split the utilities into multiple files but somehow missed having isolated unit tests for `utils/{fee,receipts}.test.ts`. Would it be worth adding?

packages/providers/src/utils/json.ts
```diff
@@ -0,0 +1,35 @@
...
+        case 'Object':
+          object[key] = nomarlize(object[key]);
+          break;
+        default:
+          break;
+      }
+    });
+    return object;
+  }
+  return nomarlize({ ...root });
```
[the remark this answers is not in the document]
luizstacio: Root can only be a object by children objects can be a Array always doing the clone `{ ...root }` would create a wrong value. This is why went with a local scoped function, but I can move the function to outside also, would not be a issue.

packages/contract/src/contracts/contract-factory.ts
luizstacio: throwArgumentError requires a message, name and a value. But the message fix makes sense.

packages/providers/src/utils/json.ts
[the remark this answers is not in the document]
luizstacio: @arboleya I improve the code as suggest and enable the root object to be a array by using deepClone instead of shallow clone.

arboleya (approved):

PR #767, opened 2023-02-07
feat: add support for custom fetcher on `Provider`

### LINKED ISSUE #754: Add support to custom fetcher on Provider
### Motivation

Currently, the Provider creates an instance of the [GraphQLClient using URL](https://github.com/FuelLabs/fuels-ts/blob/cef77013e14461b07a525213046859ba1013dd29/packages/providers/src/provider.ts#L215), not leaving room for customizable fetchers or other network configurations like proxy.

Current [graphql-request](https://github.com/jasonkuhrt/graphql-request#using-a-custom-fetch-method) supports optional params that accept some fetch configurations or a custom fetcher.

Enabling the configuration of a custom fetcher enables us on the Wallet project to create a simpler interceptor for redirecting requests from directly fetch to an event base model. But it can also be useful for mocking test environments removing the need for the access network.

### Usage example

```ts
import crossFetch, { Response } from 'cross-fetch';

const customFetch = async (url, options) => {
    const graphqlRequest = JSON.parse(options.body);
    const { operationName, query } = graphqlRequest;
    
    if (operationName=== 'getVersion') {
       const reponseText = JSON.stringify({
           data: { nodeInfo: { nodeVersion: "0.30.0" } },
       });
       const response = new Response(reponseText, options);

       return response;
    }
    
    return crossFetch(url, options);
}

const provider = new Provider('http://localhost:4000/graphql', {
    fetch: customFetch
});
```

### Possible implementations

- Add a second param to the provider called options. This will make the provider configs more flexible for future changes.
- The constructor should accept `fetch` as an option taking the example as a reference.

Dhaiwat10:
## Summary
- This PR adds a new param to the `Provider` constructor called `options`. It only accepts one option called `fetch` for now.
- Closes #754


packages/providers/src/provider.ts
```diff
@@ -190,6 +190,13 @@ export type BuildPredicateOptions = {
   fundTransaction?: boolean;
 } & Pick<TransactionRequestLike, 'gasLimit' | 'gasPrice' | 'maturity'>;
 
+/*
+ * Provider initialization options
+ */
+export type ProviderOptions = {
+  fetch?: unknown;
```
Dhaiwat10, line 197: Marked `fetch`'s type as `unknown` because even the internal type for `fetch` inside `GraphQLClient`'s type declarations is `any`.

arboleya: Could we add some common typings defined by us, or would it restrict the options?

Their [docs](https://github.com/jasonkuhrt/graphql-request#using-a-custom-fetch-method) _(somehow also vague)_ cites [this](https://github.com/valeriangalliat/fetch-cookie/blob/40eb95500ee2c7d9db27e438448742b8c8ce2fbd/src/index.ts#L236-L240C32) as an example, not sure if we can draw inspiration from it.

@luizstacio What do you think?

luizstacio: Typing the fetch method would be tricky as the API goes from simple type to extensive returns types, steam, JSON, etc... As this Option is more advanced and not for common use, I don't see the need to invest too much time on this fetch type.

The basic implementation I would suggest is to type the parameters that are passed to the fetch.

```ts
interface RequestOptions {
      method: 'POST',
      headers: { [key: string]: string },
      body: string
}

fetch: (url: string, options: RequestOptions) => Promise<any>
```

arboleya (changes requested):
There's a failing test here. I thought it could be a CI thing (as you enabled auto-merge), but no. I went ahead and tried running it locally, and the error persists.

I wonder if (and how) it could be working for you.

Dhaiwat10:
@arboleya good catch! This is what I observed: when I _only_ run that failing test in isolation, it doesn't fail. But when we run `pnpm ci:test`, it fails. Need to look into this.

arboleya:
The order of the tests _must never_ interfere with the test's success, as this is a very fragile thing that can easily become flaky if someone adds/removes more cases in the future. Instead, mocks should be properly restored. 

Please, try restoring it on an `afterEach` hook or directly inside the test that's mocking the client — if you opt to do it inside the test case itself, be sure to restore it before the first test assertion. Otherwise, if the assertion fails, the restoration might never occur.

Let's see if the coverage problem persists after this; if it does, we can see what to do.

packages/providers/src/provider.test.ts
```diff
@@ -156,6 +156,20 @@ describe('Provider', () => {
...
+    const providerUrl2 = 'http://127.0.0.1:8080/graphql';
+    const provider = new Provider(providerUrl1);
+    const spyGraphQLClient = jest.spyOn(GraphQL, 'GraphQLClient');
+
+    expect(provider.url).toBe(providerUrl1);
+    provider.connect(providerUrl2);
+    expect(provider.url).toBe(providerUrl2);
+    expect(spyGraphQLClient).toBeCalledWith(providerUrl2, undefined);
+
+    spyGraphQLClient.mockRestore();
```
arboleya, line 170: I tried explaining, but maybe it wasn't clear enough — the mock restoration _must_ have come before the first expect/assertion after the `connect` call because if somehow the assertion fails, the restoration will never occur.

```suggestion

    spyGraphQLClient.mockRestore();

    expect(provider.url).toBe(providerUrl2);
    expect(spyGraphQLClient).toBeCalledWith(providerUrl2, undefined);
```

Dhaiwat10:
@arboleya done! :)

packages/providers/src/provider.test.ts
Dhaiwat10: @arboleya oh I see! I missed that, my bad

Dhaiwat10: Maybe we just need a `try-catch` block here, because if I restore the mock before the assertion it will not work. The `mockRestore` call will 'reset' the function we are trying to inspect and hence the assertion will fail.

Dhaiwat10: The problem with using `afterEach` is that we don't want to mock this method for all the tests. We are only mocking it for one test. If we mock it in a `beforeEach`, we'll have to explicitly restore the mock in all the tests that _don't_ mock this method and need its actual implementation to be intact in order to work properly.

arboleya: You are right.

But assertions inside try-catch blocks can generate false positives unless we use control variables to ensure it doesn't. Another approach could be to have an `afterEach` hook, feel free to do as you prefer.

Control variables:

```ts
let error;

try {
  expect(provider.url).toBe(providerUrl2);
  expect(spyGraphQLClient).toBeCalledWith(providerUrl2);
} catch (err) {
  error = err;
}

spyGraphQLClient.mockRestore();

expect(error).toBeFalsy();
```

Using test hooks:

```ts
describe('Provider', () => {
  afterEach(() => {
    jest.restoreAllMocks();
  });
  // ...
})
```

Dhaiwat10: Figured that would be the way to go! Thanks :) This is a very elegant solution

packages/providers/src/provider.test.ts
```diff
@@ -5,9 +5,15 @@ import { bn } from '@fuel-ts/math';
 import type { Receipt } from '@fuel-ts/transactions';
 import { ReceiptType, TransactionType } from '@fuel-ts/transactions';
 import * as GraphQL from 'graphql-request';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import fetch, { Response } from 'node-fetch';
```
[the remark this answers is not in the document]
arboleya: @camsjams Would this be enough, or do you mean adding it to the `devDependencies` on the `provider` package?
https://github.com/FuelLabs/fuels-ts/blob/9943c5a713f774412136513461836e50548c3e80/package.json#L73

As for the node upgrade, I agree:
 - https://github.com/FuelLabs/fuels-ts/issues/795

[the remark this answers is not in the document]
Torres-ssf: @camsjams I agree we should upgrade Node to the current LTS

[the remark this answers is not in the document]
LuizAsFight: if it's working I would suggest merge this PR as is

then upgrade node + remove `node-fetch` dependency + fix import in another PR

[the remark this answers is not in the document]
arboleya: @LuizAsFight Agreed — I added these notes [here](https://github.com/FuelLabs/fuels-ts/pull/796#pullrequestreview-1306346440) on the related PR.

arboleya (approved):

PR #781, opened 2023-02-11
feat: implement husky pre push hooks

### LINKED ISSUE #780: Implement pre-push git hooks to validate contributions before review
There's been a couple of instances where I have pushed contributions without linting (:upside_down_face:) and then have to wait for someone with actions privileges to kindly start my workflow before I realise the problem and can rectify it. I fully appreciate this is user error and some may opt to run tests/build in watch mode, and potentially also lint on save. However this can be frustrating and cause a lag time in progressing with contributions.

A potential solution to this is to implement a pre-push git hook that runs `build`, `ci:test` and `lint`. This would then validate a contribution before it is pushed. A package like [typicode/husky](https://github.com/typicode/husky) does a good job of implementing this. And it can still be overridden using a `--no-verify` flag

This should therefore increase the success rate of lint/test workflow and reduce action minutes used.

I appreciate this is not everyones preferred workflow, especially to those with workflow priviliges where it may be completely redundant, so I would recommend it is implemented as an optional setup stage.

danielbate:
Closes #780 

Implementation of husky pre-push git hooks to build, test and lint contributions before push.

- Installation of husky as a dev dependency and command to implement hook
- Hook executing `build`, `ci:test` and `lint`
- Updated contribution documentation detailing the setup of the pre-push hook


arboleya (changes requested):
Although I fully understand where this PR comes from, we're afraid that piling up many things on a `pre-push` hook can slow development. Instead of just throwing the attempt away, we can try a middle ground by restricting the `pre-push` to run only the **_`lint`_** script.

On top of this, we could ideally configure proper caching ([eslint](https://eslint.org/docs/latest/use/command-line-interface#caching), [prettier](https://prettier.io/docs/en/cli.html#--cache)) to guarantee subsequent pushes don't cause unwanted delays — the cache location could be:
 - `.husky/.lintcache/eslint`
 - `.husky/.lintcache/prettier`

Will the final result be fast enough?

I'm not sure, but I believe it's worth trying. :)

danielbate:
Thanks for your input @arboleya - I can appreciate that. I have removed both of those checks. I have also been given workflow privileges since reporting the issue in question (#780) which has made this change slightly redundant for myself. 

I have implemented caching for both prettier and eslint to ensure those shouldn't cause unwanted / unnecessary delays. Currently riding off the current `lint:check` and `prettier:check` scripts. I'm wondering wether the cached implementations should be in there own scripts as to not bring in any unwanted functionality to those currently using those scripts. Potentially something like:

- `"lint:check-cache": "eslint . --ext .ts --cache --cache-location=./.husky/.lintcache/eslint"`
- `"prettier:check-cache": "prettier --check --cache --cache-location=./.husky/.lintcache/prettier packages"`
- `"lint:pre-push": "run-s lint:check-cache prettier:check-cache"`

The husky `pre-push` hook would then utilise `lint:pre-push`.

arboleya:
Good stuff, Daniel. Thank you.

Here's a quick benchmark I ran on my machine:

 1. Without cache —— 17.168s <sup>_(1st run)_</sup>
 2. With cache —— 3.757s <sup>_(2nd run)_</sup>

At only ~3s, it shouldn't be a problem to have it as a `pre-push` hook, IMO.

Let's rename the `husky:install` script to `prepare`, enabling npm `auto-install` capabilities and removing the need to `pnpm husky:install` manually. The docs should then be updated accordingly, briefly explaining the hooks in use and what it does while keeping the mention about the `--no-verify`, if that makes sense.

Could you please do this last update?

arboleya:
Also, we don't need separate scripts, like you suggested, as the cache should not interfere with anything.

arboleya (approved):
LGTM - Thank you @danielbate.

Torres-ssf (approved):

Issue #788, opened 2023-02-16
Allow users to pay for transaction fees using predicates


luizstacio:
Users have to use their `Wallet` to pay for a transaction. However, there are many use cases where a user wants to pass a `Predicate`, instead of a `Wallet`, to pay for a transaction. The actual UTXOs used to pay for the transaction must be unlocked in the predicate passed. 

For example: 

```ts
const contract = ExampleContractAbi__factory.connect(wallet);

let response = await contract
    .functions
    .initialize_counter(42) // Build the ABI call
    .call() // Perform the network call
```

Here, we link a `Wallet` to the contract instance. Underneath, it will grab this wallet's funds to pay for the transaction. 

With this new feature, one could pass in a `Predicate` instead of a `Wallet`, and, underneath, the predicate would be evaluated, and, if true, the UTXOs unlocked would be used to pay for the tx fees.

This is based by a feedback we got from an external developer, here's what was said about this feature request: 

> When I want to spend utxos from a predicate, there should be a wallet that offers the same kind of features as a wallet for a simple public key address. Currently, you need to initialize the wallet with a key, and that will also be the address that pays for the fees. If I have a predicate, I generally want to pay for the fees with the coins available in the utxos on the predicate address. A public-key based address should not be required to manage a predicate wallet.

Base issue: https://github.com/FuelLabs/fuels-rs/issues/809

arboleya:
@luizstacio How is this issue different from this other one?
 - https://github.com/FuelLabs/fuels-ts/issues/757

> cc @camsjams

luizstacio:
> 

Issue #757 would include tests using a predicate to execute script call. Once this issue is done and also scripts are done, it would be possible to implement a Script using a Predicate as we do use Wallets.

Issue #457, opened 2022-08-15
docs: move from jekyll pages to a better doc engine


luizstacio:
Items;
- [ ] change typedoc to generate to the new temple engine
- [ ] change the deploy config on gh-pages
- [ ] make a reusable block for `### Full SDK Installation` and reuse them inside packages `README` file

Notes;
- One good template is the [nextra](https://nextra.vercel.app/).

arboleya:
Nextra seems to be an excellent solution, but I have never used it.

I've seen a couple of projects using it, and also on our other repos:
 - https://fuels-wallet.vercel.app/
 - https://fuel-graphql-docs.vercel.app/

> _Congrats on these! @FuelLabs/frontend 🤘 🚀_ 

@FuelLabs/frontend Do you have any pros and cons you'd like to point out about the tool?

@FuelLabs/sdk-ts Anyone else with experience using Nextra?

pedronauck:
Why we don't use the same project we have on our wallet in order to have standard documentation websites across the entire repositories? 🤔

pedronauck:
I guess we could improve the next project we have inside the wallet and try to use it as a base for all projects. Could be a template, a boilerplate, a generator, or even a tool. But it will be essential in orde to have standard docs across the entire organization.

arboleya:
@pedronauck Excellent point.

I like the sound of it, but I wonder how complicated that would be.

arboleya:
Another candidate could be Vite Press:
 - https://vitepress.vuejs.org

It has many goodies for Markdown, `import code snippets` being one of them:
 - https://vitepress.vuejs.org/guide/markdown#import-code-snippets

Dhaiwat10:
I have used Nextra multiple times in the past, but never with anything along the lines of auto-generated docs.

arboleya:
Yes, in this case, we wouldn't be auto-generating docs from JavaDoc style annotations via TypeDoc anymore, but writing everything by hand and using code snippets where applicable to denote methods usage and whatnot.

I would be OK with throwing TypeDoc away and stopping using those styles of API documentation, which, IMO, are not very useful. These generated docs are often vague and mechanical and do not seem to provide advantages beyond what one could get by reading the source code directly.

@Dhaiwat10 What are your thoughts about Nextra? Did you have a good experience using it? Any experience comparing it with other tools such as VitePress or similar?

Torres-ssf:
[...]

PR #789, opened 2023-02-16
feat: move predicate to work as a wallet

luizstacio:
This PR enables Predicate to be used as a Wallet, enabling predicates to be passed to contract instances, do transfers, and pay for transactions.

On this PR;
- Move `BaseWalletLocked` to `Account`.
- Move the implementation of buildPredicate, spend, and another method directly to `Predicate` and add the same API of `Account` by extending it.
- Fix the tests related to Predicate.
- Add new unit tests to Predicate object
- Fix test predicate docs example
- Re enable UTXO validation on fuel-core and fix tests related to it

close: #788


packages/fuel-gauge/src/contract.test.ts
```diff
@@ -638,7 +641,7 @@ describe('Contract', () => {
     const externalWallet = Wallet.generate();
     await seedTestWallet(externalWallet, [
       {
-        amount: bn(1_000_000_000),
+        amount: bn(1_000_000),
```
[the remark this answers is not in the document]
luizstacio: No, just avoid the seeder wallet to drain locally when running multiple times

packages/fuel-gauge/src/doc-examples.test.ts
```diff
@@ -378,51 +377,59 @@ it.skip('can create a predicate and use', async () => {
...
+  const predicateBalance = await predicate.getBalance();
 
   // assert that predicate address now has the expected amount to predicate
   expect(bn(predicateBalance)).toEqual(initialPredicateBalance.add(amountToPredicate));
   // #endregion
 
   // #region typedoc:Predicate-triple-submit
-  await wallet1.submitPredicate(predicate.address, 200);
-  const updatedPredicateBalance = await provider.getBalance(predicate.address, assetId);
+  const depositOnPredicate = await await wallet1.transfer(predicate.address, 200);
```
LuizAsFight, line 429: wait wait

```suggestion
  const depositOnPredicate = await wallet1.transfer(predicate.address, 200);
```

packages/predicate/src/predicate.ts
```diff
@@ -36,4 +46,36 @@ export class Predicate extends AbstractPredicate {
...
+
+    request.inputs?.forEach((input) => {
+      if (input.type === InputType.Coin && hexlify(input.owner) === this.address.toB256()) {
+        // eslint-disable-next-line no-param-reassign
+        input.predicate = this.bytes;
+        // eslint-disable-next-line no-param-reassign
+        input.predicateData = this.predicateData;
+      }
+    });
+
```
LuizAsFight, line 61: can we use `request.inputs?.map` instead to avoid reassigning `forEach` params ?

LuizAsFight (changes requested):

packages/account/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Full SDK Installation
+
+Alternatively, we recommend you install the [complete SDK](https://github.com/FuelLabs/fuels-ts) using the umbrella package:
+
+```sh
+yarn add fuels
+# or
+npm add fuels
+```
```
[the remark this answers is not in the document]
luizstacio: The readme inside the package focus on the package itself. This is why it does not focus on saying about installing fuels. But the doc was wrong I'm fixing it thanks for the note.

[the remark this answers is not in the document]
luizstacio: Other comments about making it reusable would be great to an issue, maybe comment it here: https://github.com/FuelLabs/fuels-ts/issues/457

packages/predicate/src/predicate.ts
luizstacio: The idea here is to reassign the params to inject the new data in the tx, otherwise, it will create new inputs and would need to replace the entire inputs section.

packages/address/src/address.ts
```diff
@@ -142,17 +142,23 @@ export default class Address extends AbstractAddress {
    * thrown if the input string is not nilsy and cannot be resolved to a valid address format
    * @returns a new `Address` instance
    */
-  static fromDynamicInput(addressId: string): Address {
-    if (isPublicKey(addressId)) {
-      return Address.fromPublicKey(addressId);
+  static fromDynamicInput(address: string | AbstractAddress): Address {
```
arboleya, line 145: I'll repeat myself in this regard, but I'd prefer if we could keep it as before, not overloading the `address` parameter type and keeping it as `string` only.

I know there are many places with this kind of overloading, and it might be tricky to adjust them all, but we should at least make an extra effort not to **_add new ones_**.

```suggestion
  static fromDynamicInput(addressId: string): Address {
```

LuizAsFight (approved):

arboleya (changes requested):
Great job! 🚀 

I've just left a single comment about type overloading (_again_). All other things I'd request have been requested from others reviewers and implemented already. 

The only noteworthy exception is that we don't need to add a new package for this code split. We can DRY stuff by keeping the `account.ts` and `account.test.ts` inside the **_`wallet`_** package, removing many boilerplate files that don't need to exist.

packages/address/src/address.ts
luizstacio: This will be removed on the PR to remove the object Address.

LuizAsFight (approved):

packages/address/src/address.ts
arboleya: @luizstacio Adding something only to be removed later isn't optimal.

@camsjams Don't the `Address.fromDynamicInput` method already know things for certain?

arboleya (changes requested):
Can you please move the `account.ts` and `account.test.ts` files inside the `wallet` package instead of creating a new package for holding only these two files?

Let's try to DRY things, reducing the amount of boilerplate code.

packages/address/src/address.ts
luizstacio: If I remove it from here I would need to implement the same logic in the other place. This is why for now it makes sense to receive both.

.changeset/clever-chefs-shop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/wallet": minor
+---
+
+Create Account pacakge. Moves BaseWalletLocked to a sperate package.
```
arboleya, line 5:
```suggestion
Renaming BaseWalletLocked class to Account while abstracting common code.
```

.changeset/honest-ways-learn.md
```diff
@@ -0,0 +1,5 @@
+---
+"fuels": minor
+---
+
+Add packages Account and Predicate to exports
```
arboleya, line 5:
```suggestion
Adding missing export for package Predicate.
```

.changeset/nine-bottles-jog.md
```diff
@@ -0,0 +1,10 @@
+---
+"@fuel-ts/address": minor
+"@fuel-ts/contract": minor
+"@fuel-ts/interfaces": minor
+"@fuel-ts/predicate": minor
+"@fuel-ts/providers": minor
+"@fuel-ts/script": minor
+---
+
+Update use of BaseWalletLocked to Account
```
arboleya, line 10:
```suggestion
Updating usages of BaseWalletLocked after renaming it to Account
```

arboleya (dismissed):
Beautiful. ❤️ 

I left some final and minor suggestions regarding the changelogs. ✨

arboleya (approved):

LuizAsFight (approved):

Torres-ssf (approved):

PR #794, opened 2023-02-17
feat: add status back to messages

pixelcircuits:


packages/providers/src/operations.graphql
```diff
@@ -55,7 +55,7 @@ fragment coinFragment on Coin {
   amount
   assetId
   maturity
-  status
+  coinStatus: status
```
luizstacio, line 58: Why we had to add this alias here?

luizstacio (approved):

packages/providers/src/operations.graphql
pixelcircuits: because both coins and messages get combined into a resource object, but unlike values like `amount`, `status` is different between the two (one is type `CoinStatus` and the other `MessageStatus`.

Issue #799, opened 2023-02-22
Update fuels-ts with latest forc version and fuel-core


luizstacio:
Update with fuel-core with version 0.17.3 and forc 0.35.3

PR #772, opened 2023-02-08
fix: forc bin path on typegen test utility

Torres-ssf:
- Close #771

PR #800, opened 2023-02-22
feat: update forc and fuel-core to latest versions

luizstacio:
This PR;

- Updates fuel-core to version 0.17.3
- Update forc to 0.35.3
- Update contracts not compatible with sway 0.35.3

close: #799

### DIFF
--- a/packages/fuel-gauge/scripts/build.sh
+++ b/packages/fuel-gauge/scripts/build.sh
@@ -6,7 +6,7 @@ for PROJECT_DIR in ./test-projects/* ;
 do
   BIN_DIR="$PROJECT_DIR/out/debug"
 
-  ./../../packages/forc-bin/forc-binaries/forc build -p $PROJECT_DIR
+  pnpm forc build -p $PROJECT_DIR
   if [[ $PROJECT_DIR == *"predicate-"* ]]; then
     pnpm exec ts-node scripts/process-predicate.ts $PROJECT_DIR
   fi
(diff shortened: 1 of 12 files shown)


pixelcircuits (approved):

packages/fuel-gauge/scripts/build.sh
```diff
@@ -6,7 +6,7 @@ for PROJECT_DIR in ./test-projects/* ;
 do
   BIN_DIR="$PROJECT_DIR/out/debug"
 
-  ./../../packages/forc-bin/forc-binaries/forc build -p $PROJECT_DIR
+  pnpm forc build -p $PROJECT_DIR
```
arboleya, line 9: This is the PR that changed this to use an absolute path instead:
 - https://github.com/FuelLabs/fuels-ts/pull/557/files#r1040574105

Since then, we've been having problems again:
 - https://github.com/FuelLabs/fuels-ts/pull/763#issuecomment-1424169679

Also, there have been other situations (under Linux) where `pnpm` wasn't finding the correct binary.

The fix was again to use the full path:
- https://github.com/FuelLabs/fuels-ts/pull/772

TBH, I don't know what's happening, but it looks like a delicate change.

arboleya (approved):
Approving as is, but leaving a comment about `pnpm` and absolute paths regarding the internal `forc` bin, as things seem tricky in this regard.

> cc @Torres-ssf @camsjams @Dhaiwat10

Dhaiwat10 (approved):

packages/fuel-gauge/scripts/build.sh
luizstacio: On my local, the path was not working... but with pnpm it worked. And it looks to be the correct way to do this. If pnpm fails on every Linux, it shall fail on CI also. This looks like to be a machine-specific issue and the solution is related to pnpm not to the project it self.

matt-user (approved):

Issue #748, opened 2023-01-23
Accept address as `String` instead of `Address`


luizstacio:
When using;
- `BaseWalletLocked.<transfer|buildPredicateTransaction
|submitPredicate>`
- `Provider.<getCoins|getResourcesToSpend|getBalance|getBalances|getMessages|buildSpendPredicate|submitSpendPredicate>`
- `BaseTransactionRequest.<updateWitnessByOwner>`

These methods require the Application to instantiate an Address object before executing the methods, as bech32 and hex are both strings and can be differentiated by string evaluation. All the methods should be able also to receive a string. Making address param type from `AbstractAddress` to something like`string | AbstractAddress`. And use `Address.fromDynamicInput` internally for the correctness of the data.

> [!NOTE]
> - Related to https://github.com/FuelLabs/fuels-ts/discussions/1537

arboleya:
I would advocate accepting only `string` instead of `string | AbstractAddress`.

Still on the same wave about _[avoiding parameter type overloading](https://github.com/FuelLabs/fuels-ts/pull/746#pullrequestreview-1264101092)_ and why it [_sounds_](https://github.com/FuelLabs/fuels-ts/pull/746#discussion_r1083677506) like the thing [_to do_](https://github.com/FuelLabs/fuels-ts/pull/746#discussion_r1084296559).

For this issue, this means that:
 1. It is a breaking change
 1. It requires refactoring all places currently using the affected methods
     - _Mostly purging boilerplate-code anyways_
 1. It sucks

Still, I think it is worth it.

Thoughts?

> cc @FuelLabs/sdk-ts

luizstacio:
I think is good to accept both once the code you implement the call can have an Address instance and forcing it to transform into a string to then pass the param, would not be so good as just accepting both.

Dhaiwat10:
@arboleya I would lean towards simplicity, too.

Our methods only accepting `string` inputs and our code _internally_ taking care of all the complexity around parsing the input & processing it further sounds like a better experience for our users than the methods accepting inputs of multiple shapes and our internal code _still_ having to do some processing depending on the shape of the input _anyways_. 

The former approach sounds like it makes things super simple for the users, while the latter sounds like it involves a compromise on both ends.

LuizAsFight:
I agree inputs should be more flexible.. `string | AbstractAddress` sounds good to me


[...]

PR #801, opened 2023-02-23
feat: remove address object and move to plain string

luizstacio:
This is a draft PR open for comments;

Motivation;
- Remove the need to create Address objects, and make the APIs that require addresses more simple, without the complexity of creating objects and dealing with them on comparisons.
- Enforce the bech32 address as input by failing when the address is not in the correct format.
- Looking forward fuel-core will also work with bech32 address removing the need for transformations when sending and receiving addresses.


arboleya:
@luizstacio Clean and smooth, but what led you to rename the `Address` package to `Bech32`? I think `Address` was more descriptive since `bech32` is just a format that could _[semantically]_ live under the `Address` nomenclature/domain.

Thanks for putting this together; it accurately illustrates the idea.

arboleya:
@camsjams Do your [concerns](https://github.com/FuelLabs/fuels-ts/issues/748#issuecomment-1439496827) still persist when considering the approach from this draft?

I'm afraid I can still be missing something from your previous comment, hence why [I was looking](https://github.com/FuelLabs/fuels-ts/issues/748#issuecomment-1439844600) to get some extra code snippets to help clarify any possible misunderstanding.

PR #806, opened 2023-02-26
Torres/fix/improve receipts evaluation

### LINKED ISSUE #802: Improve receipts evaluation
We're hard-coding the assumptions that the receipts are ordered. If that changes in the future, it'll trigger some gnarly errors. 

We should refactor this not to assume order and filter the receipts we're looking for in the `receipts`. 

_Originally posted by @digorithm in https://github.com/FuelLabs/fuels-ts/pull/745#discussion_r1115125928_

Torres-ssf:
- Close #802

### DIFF
--- a/.changeset/slow-carrots-tease.md
+++ b/.changeset/slow-carrots-tease.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/program": minor
+---
+
+improve receipts evaluation at callResultToScriptResult
--- a/packages/program/src/script-request.ts
+++ b/packages/program/src/script-request.ts
@@ -11,14 +11,15 @@ import {
 } from '@fuel-ts/abi-coder';
 import type { BN } from '@fuel-ts/math';
 import type {
+  TransactionResultReturnDataReceipt,
+  TransactionResultRevertReceipt,
   CallResult,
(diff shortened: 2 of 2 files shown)


Dhaiwat10 (approved):
LGTM!

arboleya (approved):
LGTM.

However, I'd like to point out some things — it doesn't need to be done on this PR, each item can become its own issue, or we could have one issue to address them all. Please, feel free to create those as you see fit.

 1. I missed having a unit test case for the `callResultToScriptResult` utility method, which would provide mixed-up unordered receipts to validate the functionality. Everything is currently at the mercy of the e2es, which is not good.
 2. For clarity, we could have a comment describing why we search for those specific results within the receipts array. There's some hidden rationale there.
 3. This method's return _(see below)_ looks ambiguous; nomenclatures are generic. It makes it confusing to understand what is what. It could be cleaned up, improved, and better differentiated.

```ts
const scriptResult: ScriptResult = {
  code: scriptResultReceipt.result,
  gasUsed: scriptResultReceipt.gasUsed,
  receipts,
  scriptResultReceipt,
  returnReceipt,
  callResult,
};
```

PR #814, opened 2023-03-07
chore: export mnemonic + small changes

LuizAsFight:
THIS PR:

- export `fuels-ts/mnemonic` in `fuels` umbrella package
- create/exports `MNEMONIC_SIZES`
- avoid throw error in predicate function `Mnemonic.isMnemonicValid` , it should return always boolean


arboleya (dismissed):

packages/mnemonic/src/mnemonic.ts
```diff
@@ -140,7 +141,11 @@ class Mnemonic {
 
     let i: number = 0;
 
-    assertMnemonic(words);
+    try {
+      assertMnemonic(words);
+    } catch (e) {
+      return false;
+    }
```
luizstacio, line 148:
```suggestion
    try {
      assertMnemonic(words);
    } catch {
      return false;
    }
```

luizstacio (changes requested):

luizstacio (approved):

Issue #784, opened 2023-02-14
Purge GAS_PRICE_FACTOR from codebase


LuizAsFight:
### Motivation

Some places are still using `GAS_PRICE_FACTOR` constant, which is wrong because this factor can be different between chains.

should use `provider.getChain()` instead to get correct gasPriceFactor

PR #820, opened 2023-03-09
refactor: re-organize constants throughout the repo

### LINKED ISSUE #225: Move fix values to constants package
We have some constants around the system that can be moved to the package constants or better organized inside the package.

Ref.: https://github.com/FuelLabs/fuels-ts/pull/224#discussion_r853256590

Dhaiwat10:
## Summary

This PR is an attempt at re-organizing constants throughout our repo, following the approach discussed [here](https://github.com/FuelLabs/fuels-ts/issues/225#issuecomment-1457880613).

- I have moved most of the constants to `env.ts` files _inside_ the respective packages if they are not being used by any other packages.
- Some constants, particularly the ones inside of the `transactions` package are being used by multiple packages throughout our repo. I have moved these constants from the `transactions` package to the `constants` package. I'm treating this `constants` package as a place where 'globally' shared constants should live.
- I need to create a follow-up issue after this PR that would make constants like [`GAS_PRICE_FACTOR`](https://github.com/FuelLabs/fuels-ts/issues/784) _async/dynamic_, instead of them being static/hard-coded.

Closes #225


Torres-ssf (dismissed):
LGTM

packages/constants/src/index.ts
```diff
@@ -1,3 +1,47 @@
...
+  // TODO: set max predicate length value
+  /** Maximum length of predicate, in instructions. */
+  MAX_PREDICATE_LENGTH: 1024 * 1024,
+
+  // TODO: set max predicate data length value
+  /** Maximum length of predicate data, in bytes. */
+  MAX_PREDICATE_DATA_LENGTH: 1024 * 1024,
+
+  FAILED_TRANSFER_TO_ADDRESS_SIGNAL: '0xffffffffffff0001',
+});
```
arboleya, line 47: Shouldn't the existent properties be moved inside the `getEnv` method?
 - `ZeroBytes32`
 - `NativeAssetId`
 - `EmptyRoot`

arboleya (dismissed):
Meticulous. ✨ 

I'm questioning the existence of the `constants` package, though.

Is it essential?

packages/constants/src/index.ts
arboleya: Yes, from the outsider's perspective, it does feel weird.

Haven't looked at it from this angle.

 1. What would you suggest?
 1. Parallel to this, do you think the `constants` package is essential? Is it just a matter of moving things where they belong, or does the package solve a bigger issue? (e.g., cyclical refs)

Dhaiwat10: 1. I suggest omitting the completely static variables from the `getEnv()` function just for the `constants` package to optimize for our SDK's users. These constants can live at the base level of the file as constants, and they can be hence easily imported. Let's use the `getEnv()` function only for values that actually rely on the environment they are being used in.

2. Yes, if there is no `constants` package there were several cases where I was facing cyclical refs/dependencies. That's really the only major reason why I think we need this package, apart from the fact that some of these constants are being used _outside_ of the package where they 'belong' that it somewhat makes sense to have them placed in a global store of some sort.

packages/constants/src/index.ts
```diff
@@ -1,3 +1,50 @@
-export const ZeroBytes32 = '0x0000000000000000000000000000000000000000000000000000000000000000';
-export const NativeAssetId = ZeroBytes32;
-export const EmptyRoot = '0xe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855';
+import { bn } from '@fuel-ts/math';
+
+export const getEnv = () => ({
```
LuizAsFight, line 3: what you think on changing this name to `getConstants`?

`getEnv` reminds me of env variables, which is not the case here

packages/constants/src/index.ts
```diff
@@ -1,3 +1,50 @@
...
+  /** Gas charged per byte of the transaction. */
+  GAS_PER_BYTE: bn(4),
+
+  // TODO: set max script length const
+  /** Maximum length of script, in instructions. */
+  MAX_SCRIPT_LENGTH: 1024 * 1024 * 1024,
+
+  // TODO: set max script length const
+  /** Maximum length of script data, in bytes. */
+  MAX_SCRIPT_DATA_LENGTH: 1024 * 1024 * 1024,
```
LuizAsFight, line 36: this comes from `chain` gql query to node.. we don't need to handle it now (same situation of `GAS_PRICE_FACTOR`), but eventually will be nice to get from the correct place.

packages/constants/src/index.ts
```diff
@@ -1,3 +1,50 @@
...
+
+  /** Maximum gas per transaction. */
+  MAX_GAS_PER_TX: bn(100000000),
+
+  /**
+   * Gas Price factor this is used to calculate
+   * This is used to calculate the gas fee in Native Coins.
+   * Ex.: transactionFee = Math.ceil(<gasUsed> / MAX_GAS_PER_TX);
+   */
+  GAS_PRICE_FACTOR: bn(1000000),
```
LuizAsFight, line 25:
```suggestion
  GAS_PRICE_FACTOR: bn(1000000000),
```

packages/constants/src/index.ts
```diff
@@ -1,3 +1,50 @@
...
+  /** Maximum number of static contracts. */
+  MAX_STATIC_CONTRACTS: 255,
+
+  // TODO: set max predicate length value
+  /** Maximum length of predicate, in instructions. */
+  MAX_PREDICATE_LENGTH: 1024 * 1024,
+
+  // TODO: set max predicate data length value
+  /** Maximum length of predicate data, in bytes. */
+  MAX_PREDICATE_DATA_LENGTH: 1024 * 1024,
```
LuizAsFight, line 47: this comes from `chain` gql query to node.. we don't need to handle it now (same situation of `GAS_PRICE_FACTOR`), but eventually will be nice to get from the correct place.

LuizAsFight (changes requested):
I like the idea of creating `constants` package... one thing to notice is that in code reviews we need to make sure constants are being created in correct place, otherwise will start to get spread around as @arboleya pointed out

packages/constants/src/index.ts
arboleya: I still think we can remove the duality _(of choosing where to put things)_ by **_eliminating_** the `constants` package.

We could solve the cyclic imports with isolated entry points:

```ts
// index.js
// the line below should not exist
// export * from "./configs";

// configs.js
export const ABC = 123

// package.json
"exports": {
  "./configs": {
    "require": "./dist/configs.js",
    "import": "./dist/configs.mjs",
    "types": "./dist/configs.d.ts"
  }
}
```

Other files can use it like:

```ts
import { ABC } from '@fuel-ts/<some-pkg>/configs';

console.log({ ABC })
```

The main `fuels` package can export everything like:

```ts
export * from '@fuel-ts/<some-pkg>/configs'
export * from '@fuel-ts/<other-pkg>/configs'
// ...
```

In short:
 1. Using the `configs` terminology _(instead of `consts` or `env`)_
 2. Exporting individual entry points for config files _(in `package.json`)_
 3. Not exporting configs in the packages' index file _(only on `fuels` index)_
 4. Removing `getEnv` methods and having the variables statically defined

Although we thought about encapsulating all variables inside methods such as `getEnv` in preparation for the future involving async variable fetching, I'm reconsidering it due to the same reason I cited at the beginning:
 - Removing duality

To make things easier, especially because **_most_** variables will **_not_** be fetched remotely, we could pull it from the inside out and do the opposite: have _all variables statically defined_ and use methods **_only_** for ones requiring async fetching.

This would leave a minimum footprint:

```ts
// configs.js
export const ABC = 123
export const XYZ = async () => fetch(...) // async stuff
```

All of this combined should provide the following:
 1. decoupled sync/async support
 1. freedom from cyclic-dependency errors
 1. coherence where variables are defined 
 1. zero _(or minimum?)_ room for interpretation

packages/constants/src/index.ts
Dhaiwat10: Yeah, I have filed an issue for this: #825

packages/constants/src/index.ts
Dhaiwat10: `getConstants()` sounds good to me

packages/constants/src/index.ts
Dhaiwat10: Love the idea of using isolated entry points to solve the cyclic dependencies! TIL. 

I agree with pretty much everything else, too. Just had one minor question regarding this:

```ts
// configs.js
export const ABC = 123
export const XYZ = async () => fetch(...) // async stuff
```

When someone wants to consume `XYZ`, they would have to write `await XYZ()`. Do you think this would be a little confusing since in most cases, people won't expect a variable having a fully uppercase name to be a promise. Maybe we can go with `getXYZ()` for async stuff instead?

arboleya: Perfect. 👌 

```ts
// configs.js
export const ABC = 123;

export const fetchXyz = async () => {
  const XYZ = await fetch(...);
  return { XYZ };
}


// otherfile.ts
import { getXyz } from "./configs"

const doSomething = async () => {
  const { XYZ } = await getXyz();
  console.log({ XYZ });
}
```

.vscode/settings.json
```diff
@@ -10,8 +10,8 @@
     "**/.git": true,
     "**/.DS_Store": true,
     "**/node_modules": true,
-    "**/.turbo": true,
-    "**/dist": true
+    "**/.turbo": true
+    // "**/dist": true
```
arboleya, line 14: Oops

packages/fuels/src/configs.ts
```diff
@@ -0,0 +1,5 @@
+export * from '@fuel-ts/transactions/configs';
+export * from '@fuel-ts/math/configs';
+export * from '@fuel-ts/program/configs';
+export * from '@fuel-ts/wallet/configs';
+export * from '@fuel-ts/address/configs';
```
arboleya, line 5: Maybe I couldn't express this clearly, but `fuels` is the only package that doesn't need a separate entry point for configs. At this point, there is no more risk of cyclic dependencies. All exports can be safely done on `fuels/src/index.ts`.

arboleya: Or you can have a `fuels/src/configs.ts` like the others.

And then `export * from './configs'` in the index.

arboleya (changes requested):
Looking good. 👌 

Almost there; I left some comments.

packages/fuels/src/configs.ts
Dhaiwat10: Gotcha @arboleya, thanks for pointing this out

arboleya (approved):

packages/fuel-gauge/src/doc-examples.test.ts
```diff
@@ -6,19 +6,19 @@ import {
...
   getRandomB256,
-  ZeroBytes32,
   addressify,
   Contract,
   Wallet,
   WalletUnlocked,
   Signer,
   ContractFactory,
+  ZeroBytes32,
+  NativeAssetId,
```
LuizAsFight, line 21: this ordering changes in imports are related to specific config inside your vscode, right @Dhaiwat10?

project should contain a vscode config file, and everyone should use the same... if this is a good config, should go to the project vscode config file.

otherwise next person that change this file can change the order again, and we end up reviewing ordering changes in a lot of prs... same follows for all lints/formatting configs

LuizAsFight: this can be done on other small PR only to update vscode config file... just pointing it out as different configs between the team generates unnecessary reviews

LuizAsFight (approved):

packages/fuel-gauge/src/doc-examples.test.ts
Dhaiwat10: @LuizAsFight good catch. I don't think it's because of my vscode settings but rather because during this PR, I changed where these constants get imported from multiple times and then we eventually ended up resetting to our initial API. 

I don't think we have a setting in our vscode settings file in the repo to automatically arrange imports in a certain way, though. That would've taken care of this.

Want me to create a new issue for this?

PR #824, opened 2023-03-13
refactor: treat the `no-explicit-any` ESLint rule as an error

### LINKED ISSUE #821: Eslint - Treat `no-explicit-any` rule as an error
Currently, our `eslint` _**`warn`s**_ about violations of the [`no-explicit-any`](https://typescript-eslint.io/rules/no-explicit-any/) rule.

We should treat these as **`error`s** instead and _adjust/refactor_ all current usages accordingly.

If using `any` can't be avoided, we can use an `eslint` comment to disable the rule **_explicitly_**:

```ts
// eslint-disable-next-line @typescript-eslint/no-explicit-any
const msg: any = "any is allowed in this line";
```

> _This should ideally be a last resort thing_

Dhaiwat10:
This PR updates the `no-explicit-any` eslint rule in our TypeScript project to throw an error instead of a warning. As a result, any usage of any will now be treated as an error during the build process.

To ensure compliance with this updated rule, I have tried to remove the usage of `any` in all applicable places in our codebase. However, in some cases (particularly files revolving around our `InvocationScope`s and `Script`s), I was unable to remove the usage of `any` without causing issues for our SDK users. Replacing `any` with `unknown` would mean that our users would receive TypeScript errors by default when trying to consume certain functions, unless they explicitly pass in a generic type. In other cases, the return types of certain functions were unclear, making it difficult to determine the appropriate replacement type without causing a huge unnecessary waterfall of TypeScript errors throughout our codebase. Therefore, in these cases, I decided to leave the `any` type in place for the time being.

Closes #821


arboleya:
Ok, it looks like this is the issue:

https://github.com/FuelLabs/fuels-ts/blob/b6b3340963283fe428b8eeee0db379249b9c09c4/packages/providers/codegen.json#L32-L34

In short, `eslint` will raise warnings for files you ask it to lint, even if they are ignored.

Refs:
 - https://stackoverflow.com/a/66434753
 - https://stackoverflow.com/a/57948047
 - https://github.com/eslint/eslint/issues/5623#issuecomment-209789397

Possible solutions:
 1. `npx eslint src --fix` or `npx eslint . --fix` _(passing a folder to it)_
    - I don't know if running `eslint` on a package in isolation like this would consider configs from `<repoRoot>/.eslint{rc.js,ignore}`. I assume it would, but it's worth double-checking.
 1. Remove these hooks from `codegen.json`
    - We don't need to lint anything at a `prebuild` step since the `lint` routine will run right after `build`.

I vote for the 2nd option.

Dhaiwat10:
Thanks for investigating this @arboleya! I vote for the 2nd option, too. Sounds fair.

arboleya (approved):

Torres-ssf (approved):

PR #833, opened 2023-03-16
feat: decode revert reasons and convert errors to specific type

### LINKED ISSUE #829: Parse revert error codes to determine more specific reasons for revert
### Motivation

Every Revert has a specific code that enables it to treat the log in different ways
https://github.com/FuelLabs/fuel-abi-types/blob/a340a110ded4268ca834269cb5916b9ac9b90943/src/error_codes.rs#L3-L16

In the case of a revert of a type required "0xffff_ffff_ffff_0000", we can safely get the previous log data from the log stack.

### Usage example

_No response_

### Possible implementations

1. improve our parsing of the revert to better utilize the different types of revert codes
2. This would allow us to have five types of revert cases now to better help DX, they'd have a better understanding of why something failed:
```
FAILED_REQUIRE_SIGNAL
FAILED_ASSERT_EQ_SIGNAL
FAILED_ASSERT_SIGNAL
FAILED_SEND_MESSAGE_SIGNAL
FAILED_TRANSFER_TO_ADDRESS_SIGNAL
```

Three part update to improve DX for #634:
1. This one
2. [Update enum support](https://github.com/FuelLabs/fuels-ts/issues/828)
3. [Use better enums in errors](https://github.com/FuelLabs/fuels-ts/issues/823)

camsjams:
closes #829


packages/fuel-gauge/src/revert-error.test.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  it('can throw SendMessageRevertError', async () => {
+    await expect(async () => {
+      await contractInstance.functions.failed_message().call();
+    }).rejects.toThrow(SendMessageRevertError);
+  });
+
+  it.skip('can throw TransferToAddressRevertError', async () => {
+    await expect(async () => {
+      await contractInstance.functions.failed_transfer_revert().call();
```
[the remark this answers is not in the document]
arboleya: Adding these notes as a `TODO` might be useful, followed by a link for the Sway docs.

Dhaiwat10 (approved):

packages/fuel-gauge/src/revert-error.test.ts
```diff
@@ -46,30 +49,48 @@ describe('Revert Error Testing', () => {
...
       await contractInstance.functions.validate_inputs(INPUT_TOKEN_ID, INPUT_PRICE).call();
-    }).rejects.toThrow(RequireRevertError);
+    }).rejects.toThrow(AssertFailedRevertError);
+  });
+
+  it('can throw SendMessageRevertError', async () => {
+    await expect(async () => {
+      await contractInstance.functions.failed_message().call();
+    }).rejects.toThrow(SendMessageRevertError);
+  });
```
Torres-ssf, line 83: I think this type of asynchronous code testing may result in a flaky test, just like it was spotted [here](https://github.com/FuelLabs/fuels-ts/issues/845). 

I've also experienced some tests with this syntax occasionally failing in my setup. 

According to [Jest docs](https://jestjs.io/docs/asynchronous#asyncawait), we can simply use:
```ts
await expect(foo()).rejects.toMatch('error');
```

Torres-ssf (changes requested):
Real nice :+1:

packages/program/src/revert/revert-error.ts
```diff
@@ -0,0 +1,97 @@
...
+  constructor(receipt: TransactionResultRevertReceipt, reason: RevertReason) {
+    super(receipt, reason);
+    this.name = 'AssertFailedRevertError';
+  }
+}
+
+export const revertErrorFactory = (
+  receipt: TransactionResultRevertReceipt
+): RevertError | undefined => {
+  const reason = decodeRevertErrorCode(receipt);
```
luizstacio, line 80: I didn't understand by reading the code + tests, if this reason is going to include the log from the contract when using `require` as example, once the log is on the previous receipt and not on the val.

luizstacio (changes requested):

packages/fuel-gauge/src/revert-error.test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  /**
+   * @todo fix this
```
arboleya, line 86: It looks like `@todo` is not a valid TSDoc tag.

I think a usual `TODO: fix this` would do it.

todo

tsconfig.base.json
```diff
@@ -4,13 +4,13 @@
...
     "esModuleInterop": true,
-    "lib": ["ES2021"],
+    "lib": ["es2022"],
     "module": "NodeNext",
     "moduleResolution": "NodeNext",
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
-    "target": "ES2021",
+    "target": "es2022",
```
arboleya, line 13: Two questions:

 1. Why did you change `lib` and `target` configs?
 1. Why `ES2022` and not `ES2023`?

arboleya (changes requested):

tsconfig.base.json
arboleya: I wasn't aware of this Error `cause`; it seems pretty [compatible](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#browser_compatibility) already. TIL.

I was just curious about the  `ES20**` config. I don't have a saying about it as I have no in-depth knowledge about which would be better or why, and neither if it could break something.

arboleya: According to these, it seems ok to target `ES2023`:
 - https://node.green/
 - https://kangax.github.io/compat-table/es2016plus/

Happy new year @camsjams 🎉

packages/program/src/revert/revert-error.ts
luizstacio: When there is a required error, we could get the log and use it as the reason for removing the necessity of the developer to know that he needs to get the logs to see it. The idea is that the `log` for `require` is the reason why it had fail.

Torres-ssf (approved):

arboleya (dismissed):
LGTM.

Why do you prefer e2es over unit tests for PRs like this one?

Dhaiwat10 (approved):

arboleya (approved):

Torres-ssf (approved):

luizstacio (approved):

PR #838, opened 2023-03-18
refact: function selector of function fragments

LuizAsFight:


LuizAsFight:
Closing because this got old... opening a new one from revamp build branch with local linking working

PR #839, opened 2023-03-20
fix: the `B512` ABI coder and typegen now expect its correct Sway type

### LINKED ISSUE #826: Convert a message signed by a wallet into a B512 type
### Motivation

When signing a message with a wallet a long `string` type is created.
To be able to recover the user address that has signed the message we need the type to be converted into a `B512` so that it can be recovered in Sway.

### Usage example

```typescript
const signature = "0x0abc...";
const b512 = B512.new(signature);
console.log(b512);

{
  bytes: ["abcd...", "efgh..."]
}
```

Dhaiwat10:
## Summary
- As pointed out by Luiz [here](https://github.com/FuelLabs/fuels-ts/issues/826#issuecomment-1467191259), the `B512` coder was expecting its Sway type to be of `b512` when in reality it is `struct B512`.
- This fix should fix the typegen and all the B512 inputs should now accept a `string` instead of a `{ bytes: [string, string] }` type.
- I have verified that you can now safely pass signed messages generated by `Wallet.signMessage()` (`string`) to any Sway function expecting a `B512` type. ([Example E2E test](https://github.com/FuelLabs/fuels-ts/commit/7529ba60bd7c1e1cc32071eb19158682e4ac3975#diff-e15776e8fb8591bfffed9b6081ee898226210e3841038a1bf29492b836580598))

Closes #826


Torres-ssf (dismissed):
LGTM

luizstacio (changes requested):
Can we add a test on the fuel-gauge package for this type?

Dhaiwat10:
Oh never mind, I had already begun typing my reply before I refreshed and then saw @arboleya's comment.

arboleya:
@Dhaiwat10 Only now I saw that test — although the test looks good, it's limited to the new type and involves many unrelated classes. If we can send/receive all types correctly, all internals encode/decode usages should follow. If you could please add such a broader test, I think it'd greatly improve the confidence of the `coder` package.

arboleya:
@Dhaiwat10 Awesome! I suspect we had this, but I couldn't remember.

Yes, adding a new test there should be enough.

Dhaiwat10:
@arboleya @luizstacio I have added a test for B512 similar to the other types. Please feel free to review now :)

arboleya (approved):

Torres-ssf (approved):

LuizAsFight (approved):

PR #840, opened 2023-03-20
Add `typegen` subcommand for command examples

tkporter:
Not sure if I should be editing both `_guide` and `guide` -- noticed the example commands were missing the `typegen` subcomment


arboleya:
@tkporter First of all, thank you for your contribution. Unfortunately, it would not be helpful in this manner anymore because we completely changed doc engines in https://github.com/FuelLabs/fuels-ts/pull/811, and I think it will be better to update these files on that PR directly.

@Torres-ssf Makes sense?

PR #842, opened 2023-03-21
fix: add eslint rule require await

### LINKED ISSUE #832: Investigate and adopt eslint rule `require-await`
### Motivation

If we added an eslint rule like:
https://eslint.org/docs/latest/rules/require-await

It probably would have prevented #705 from occuring.

### Usage example

Update eslintrc, determine how bad things are in relation to required changes. We might discover secret bugs

### Possible implementations

I also looked at and use https://github.com/eslint-community/eslint-plugin-promise a bunch, but this linter didn't pick up on this particular mistake.

Relates to issue #705 and fix #831

### LINKED ISSUE #845: Possibly new flaky test
Never seen this one before (_it worked after [re-running](https://github.com/FuelLabs/fuels-ts/actions/runs/4480342044/jobs/7875965168#step:8:1195) the flow_):
 - https://github.com/FuelLabs/fuels-ts/actions/runs/4480342044/jobs/7875505933?pr=844

```
FAIL packages/fuel-gauge/src/auth-testing.test.ts
  ● Auth Testing › can check_msg_sender [with incorrect id]

    expect(received).rejects.toThrow()

    Received promise resolved instead of rejected
    Resolved to value: undefined

      46 |
      47 |   it('can check_msg_sender [with incorrect id]', async () => {
    > 48 |     await expect(async () => {
         |           ^
      49 |       await contractInstance.functions
      50 |         .check_msg_sender({ value: wallet.address.toB256().replace('a', 'b') })
      51 |         .call();

      at expect (node_modules/.pnpm/expect@29.3.1/node_modules/expect/build/index.js:105:15)
      at Object.<anonymous> (packages/fuel-gauge/src/auth-testing.test.ts:48:11)
```

Torres-ssf:
- Close #832
- Close #845


scripts/changeset-version-with-docs.ts
```diff
@@ -2,6 +2,7 @@
 
 import sh from 'shelljs';
 
+// eslint-disable-next-line @typescript-eslint/require-await
```
arboleya, line 5: Isn't this method fully synchronous?

scripts/forc-update.ts
```diff
@@ -2,6 +2,7 @@
 
 import sh from 'shelljs';
 
+// eslint-disable-next-line @typescript-eslint/require-await
```
arboleya, line 5: Same here; it seems to me that these methods using `shelljs` are synchronous, discarding the need for the `async` directive.

packages/keystore/src/aes-ctr-node.ts
```diff
@@ -11,6 +11,7 @@ const ALGORITHM = 'aes-256-ctr';
  *
  * @returns Promise<Keystore> object
  */
+// eslint-disable-next-line @typescript-eslint/require-await
```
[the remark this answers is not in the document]
arboleya: Or else, remove the `async` from the method.

packages/wallet-manager/src/storages/memory-storage.ts
```diff
@@ -5,19 +5,19 @@ class MemoryStorage implements StorageAbstract {
...
   }
 
   async removeItem(key: string) {
-    this.storage.delete(key);
+    await Promise.resolve(this.storage.delete(key));
   }
 
   async clear() {
-    return this.storage.clear();
+    return Promise.resolve(this.storage.clear());
```
arboleya, line 20: All these methods seem to deal with the `storage` variable, an instance of a standard javascript `Map` class. I don't think any of these can become `async` in the future, can they?

packages/wallet-manager/src/wallet-manager.ts
```diff
@@ -190,7 +190,7 @@ export class WalletManager extends EventEmitter {
     // Clean password from state
     this.#passphrase = '';
     // Emit event that wallet is locked
-    this.emit('lock');
+    await Promise.resolve(this.emit('lock'));
```
arboleya, line 193: @luizstacio Do you think we should have `async` stuff here in the future? It looks pretty `sync` to me; I'd guess the `async` notation was a trivial mistake like most things in this PR.

packages/wallet/src/base-unlocked-wallet.ts
```diff
@@ -56,6 +56,7 @@ export class BaseWalletUnlocked extends Account {
    * @param transactionRequestLike - TransactionRequestLike
    * @returns string - Signature a ECDSA 64 bytes
    */
+  // eslint-disable-next-line @typescript-eslint/require-await
```
arboleya, line 59: On this method, I see you preferred to disable the rule instead of using a `Promise.resolve()` approach like on the `signMessage` method (for which I also asked other things).

I'm just trying to understand the differences between all these different approaches to see if we can hopefully standardize all occurrences.

arboleya (changes requested):
Well done. 👌 

The incoherence that can creep in with just one missing `eslint` rule is huge.

I left a lot of questions in an attempt of trying to understand the rationale for it all.

> **_There's also [this](https://github.com/FuelLabs/fuels-ts/issues/845)._**

packages/keystore/src/aes-ctr-node.ts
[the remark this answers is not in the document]
Torres-ssf: Removing was my first thought.  But taking into consideration what was discussed [here](https://github.com/FuelLabs/fuels-ts/issues/832#issuecomment-1475614618) in the issue, I am under the impression that this function may become (real) `async` in the future. I believe @luizstacio can give us more context.

About the disabled rule, it was just a matter of choice. Since adding an `await` where is not needed will make the TS shows the 3 dots warning
Screenshot from 2023-03-21 15-16-13

But yeah, we could also leave it like that.

packages/wallet/src/base-unlocked-wallet.ts
```diff
@@ -47,7 +47,7 @@ export class BaseWalletUnlocked extends Account {
    * @returns Promise<string> - Signature a ECDSA 64 bytes
    */
   async signMessage(message: string): Promise<string> {
-    return this.signer().sign(hashMessage(message));
+    return Promise.resolve(this.signer().sign(hashMessage(message)));
```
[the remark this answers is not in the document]
Torres-ssf: I believe they are the same thing. But TS shows a little warning on the second one.

Screenshot from 2023-03-21 15-08-56

packages/keystore/src/aes-ctr-node.ts
```diff
@@ -32,6 +33,7 @@ export async function encrypt<T>(password: string, data: T): Promise<Keystore> {
  * Given a password and a keystore object, decrypts the text and returns
  * the resulting value
  */
+// eslint-disable-next-line @typescript-eslint/require-await
```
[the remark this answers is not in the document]
Torres-ssf: Same reasons as  https://github.com/FuelLabs/fuels-ts/pull/842#discussion_r1143835548

packages/program/src/functions/invocation-results.ts
```diff
@@ -131,6 +131,7 @@ export class InvocationCallResult<T = any> extends InvocationResult<T> {
     this.callResult = callResult;
   }
 
+  // eslint-disable-next-line @typescript-eslint/require-await
```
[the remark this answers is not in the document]
Torres-ssf: Same reasons as  https://github.com/FuelLabs/fuels-ts/pull/842#discussion_r1143835548

packages/wallet/src/base-unlocked-wallet.ts
Torres-ssf: I see. If we decide to leave all of them as `async`, we can then agree on only one way to bypass this rule for all these cases.

scripts/changeset-version-with-docs.ts
Torres-ssf: Yes, we are right. I am going to just remove the `async` here

scripts/forc-update.ts
Torres-ssf: Yes, we are right. I am going to just remove the `async` here

packages/wallet/src/base-unlocked-wallet.ts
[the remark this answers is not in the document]
arboleya: It's because, in these cases, there is no need to `await` for anything, and the method consumer should do the awaiting. If you keep it like it was before, does it raises warnings? It should be perfectly fine.

```ts
return this.signer().sign(hashMessage(message));
```

packages/wallet-manager/src/storages/memory-storage.ts
Torres-ssf: @camsjams Since we are using the [Eslint rule](https://typescript-eslint.io/rules/require-await/) `require-await`, we need to make all `async` functions either use `await` or return a `Promise.resolve(...)` to satisfy the lint error.

arboleya: Or remove the `async` directive from the method, which in this case, is what might be ideal.

arboleya (approved):

Torres-ssf:
@arboleya @camsjams I mistakenly updated the branch with a push and all reviews were dismissed. Nothing changed here

Dhaiwat10 (approved):

PR #847, opened 2023-03-22
fix: adjusting constants evaluation by removing usage of `process.env`

### LINKED ISSUE #846: Problem with new constants evaluation approach
When trying to use the latest @next tag, I've stumbled on this error:

```
Uncaught ReferenceError: process is not defined
    at account.ts:31:58
```

This is one example that _[accesses the `process` and]_ triggers this, but there are others:

https://github.com/FuelLabs/fuels-ts/blob/1613399e97fc3ce63cdefa00ccff938e10f9fb9a/packages/wallet/src/account.ts#L31

arboleya:
- Close #846

### DIFF
--- a/.changeset/empty-waves-appear.md
+++ b/.changeset/empty-waves-appear.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/wallet": patch
+---
+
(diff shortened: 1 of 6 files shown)


Dhaiwat10 (approved):
Good catch 😅

PR #852, opened 2023-03-25
fix: adding multi-type resolution support [also] for legacy projects

### LINKED ISSUE #851: Can't import `NativeAssetId` constant in `0.36.0`
Users started to get errors like this after upgrading to `0.36.0`.

error

> cc @Dhaiwat10 @LuizAsFight @matt-user

arboleya:
- Close #851

> **Warning** This PR is a hotfix for a **_broken_** release.

It adds backward compatibility for legacy `tsconfigs`.

This is done by:
 1. Rolling back `tsconfigs` to:
    - `"module": "CommonJS"`
    - `"moduleResolution": "node"`
 2. Adding legacy multi-type definition mapping configs


tsconfig.base.json
```diff
@@ -5,8 +5,8 @@
     "declarationMap": true,
     "esModuleInterop": true,
     "lib": ["ES2021"],
-    "module": "NodeNext",
-    "moduleResolution": "NodeNext",
+    "module": "CommonJS",
+    "moduleResolution": "node",
```
Dhaiwat10, line 9: Does this change have any effect on our support for both `ESM` and `CommonJS` environments?

Dhaiwat10: Rephrasing for clarity:

Does this change mean that people will be able to consume our packages _only_ in the CommonJS format now and not ESM?

arboleya: No, we continue exporting bundles for CJS and ESM.

I [changed](https://github.com/FuelLabs/fuels-ts/pull/661/commits/9c2770df05364e54e550ad52bd081c1aa730274c) these properties a [while ago](https://github.com/FuelLabs/fuels-ts/pull/688/commits/a2f06ce1c5b55107fd98c8da5ae478cfdbf1c993) to enable multi-entry points via the `exports` config in `package.json`, but it requires users to change their `tsconfig`, which can cause incompatibilities with existing setups as reported by @matt-user and @LuizAsFight.

This PR achieves the same via the `typesVersions` entry on `package.json`, a more conservative configuration that won't require users to change their `tsconfig`. Hence why I'm reverting these settings to their original form.

Torres-ssf (approved):

tsconfig.base.json
Dhaiwat10: Makes sense. Thanks for the detailed response @arboleya 🙂

Dhaiwat10 (approved):

PR #808, opened 2023-02-27
refact: purge GAS_PRICE_FACTOR + provider refactor

LuizAsFight:

PR #856, opened 2023-03-27
refactor: purge usage of `GAS_PRICE_FACTOR` and other 'dynamic' constants

### LINKED ISSUE #784: Purge GAS_PRICE_FACTOR from codebase

### LINKED ISSUE #825: Dynamically fetch values of constants that rely on the environment
This is a follow-up issue to #820. Now that we have organized our constants throughout the repo, the next step should be to making constants like `GAS_PRICE_FACTOR` dynamic and async.

https://github.com/FuelLabs/fuels-ts/blob/dd7b1cab0e7c4a5234383ce6fc34f041ee6d03a9/packages/transactions/src/consts.ts#L12-L17

Please also see #784 and #808 for more context.

Dhaiwat10:
## Summary

- This PR purges the hardcoded `GAS_PRICE_FACTOR` constant, and instead fetches the gas price factor from the chain via the provider everywhere.
- Please see [this comment](https://github.com/FuelLabs/fuels-ts/pull/856#discussion_r1149415127) for the one major issue with this approach.
- `MAX_GAS_PER_TX` seems to be the only other constant which needs to be fetched _dynamically_ from the chain instead of it using a hardcoded value. I will go about purging its hardcoded usage once we are happy with the approach we have used for `GAS_PRICE_FACTOR` to avoid any unnecessary refactoring work which might need to reverted later.

Closes #825
Closes #784

### DIFF
--- a/packages/program/src/functions/multicall-scope.ts
+++ b/packages/program/src/functions/multicall-scope.ts
@@ -10,11 +10,11 @@ export class MultiCallInvocationScope<TReturn = any> extends BaseInvocationScope
     this.addCalls(funcScopes);
   }
 
-  addCall(funcScope: FunctionInvocationScope) {
+  async addCall(funcScope: FunctionInvocationScope) {
     return super.addCalls([funcScope]);
   }
 
-  addCalls(funcScopes: Array<FunctionInvocationScope>) {
+  async addCalls(funcScopes: Array<FunctionInvocationScope>) {
     return super.addCalls(funcScopes);
   }
 }
--- a/.changeset/fair-walls-mate.md
+++ b/.changeset/fair-walls-mate.md
@@ -1,0 +1,8 @@
+---
+"@fuel-ts/program": patch
+"@fuel-ts/providers": patch
+"@fuel-ts/script": patch
+"@fuel-ts/wallet": patch
+---
+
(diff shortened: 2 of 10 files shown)


packages/program/src/functions/multicall-scope.ts
```diff
@@ -10,11 +10,11 @@ export class MultiCallInvocationScope<TReturn = any> extends BaseInvocationScope
     this.addCalls(funcScopes);
```
Dhaiwat10, line 10: This is a problem. The `constructor` of `MultiCallInvocationScope` calls `addCalls`, which is now an `async` function. This might cause issues. Not sure how to go around this.

I have tried several different approaches including the one shared by @LuizAsFight in #808, but even in that approach eventually I ended up in a place where I had to pass in `gasPriceFactor` to the constructor of `InvocationScope`s. This was also not possible because these constructors were being called from other constructors.

Dhaiwat10:
Closing this one for now since it's been inactive for quite long. Might be worth re-visiting later.

PR #867, opened 2023-03-31
ci: revert changes on CI

Torres-ssf:
Undoing the last changes on the `release` workflow. Just trying to debug and isolate the issue.

### DIFF
--- a/.github/workflows/release-docs.yaml
+++ b/.github/workflows/release-docs.yaml
@@ -1,0 +1,50 @@
+name: "Release"
+
+on:
+  pull_request:
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  release:
+    name: "Release"
+    runs-on: ubuntu-latest
+    permissions:
+      pages: write
+      id-token: write
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+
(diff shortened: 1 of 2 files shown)


Dhaiwat10:
Do we have to debug on `master`? Can you try modifying the workflow file so that, for the sake of debugging, it runs on your PR's branch? Once we have verified that things are running as intended, you can revert that change and we can merge to `master`. What do you think? @Torres-ssf

Torres-ssf:
> Do we have to debug on `master`? Can you try modifying the workflow file so that, for the sake of debugging, it runs on your PR's branch? Once we have verified that things are running as intended, you can revert that change and we can merge to `master`. What do you think? @Torres-ssf

@Dhaiwat10 [I did try that before](https://github.com/FuelLabs/fuels-ts/actions/runs/4417946965/jobs/7744378100).

The strange thing about the [error](https://github.com/FuelLabs/fuels-ts/actions/runs/4574396787/jobs/8076280555#step:7:118) happening now is that it happens before the step in the CI that would deploy the docs.

arboleya:
@Torres-ssf This one you linked to didn't have the changesets steps combined, right? If you could make it so that the release action would run on your PR, as @Dhaiwat10 suggested, that would be a more detailed replication of the real thing to try out.

PR #868, opened 2023-03-31
fix: remove permissions item from release workflow

Torres-ssf:
This tries to fix the CI error from the last merge.

I am removing the `permissions` item from the `release.yaml` workflow to avoid all other permissions being set as `none`

> ref 1: https://github.com/FuelLabs/fuels-ts/actions/runs/4574396787/jobs/8079391507#step:7:167
> ref 2: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#overview

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -11,9 +11,6 @@ jobs:
   release:
     name: "Release"
     runs-on: ubuntu-latest
-    permissions:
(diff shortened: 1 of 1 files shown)


arboleya (approved):
🤞

Dhaiwat10 (approved):

PR #827, opened 2023-03-14
chore: revamping build tooling

arboleya:
# Summary

 - Close #834

This PR thoroughly revises the current build system, double-checks configs, rules, and caching mechanisms, removes `tsup` in favor of good old `tsc`, fulfills all requirements from #834,  and does a few extra optimizations here and there.

I will try to outline the main ones.

# Local install (TL;DR)

More notably, you can now install and run your local versions of `fuels`.

This was the primary motivation that drove everything else.

```console
cd your-app
pnpm add ~/my/local/fuel-ts/packages/fuels
```

# Remember to `build` first

Since our packages export Javascript files now, we need to **_build_** the packages after each change we make. Otherwise, we'd still be looking at outdated compiled Javascript files. To automate this, you can run the following:

```shell
# on repo root
pnpm dev
```

This will `pnpm build` the project once, then watch and re-build it automatically on every change. It uses `nodemon` to watch all packages simultaneously for optimum performance. The building part continues to be handled by Turbo Repo.

# Go to Definition

To make it possible to export Javascript files and have a local working `fuels` package while still being able to `go to definition` inside our Typescript source files, we had to generate `declarationMaps` in our build routine:
 - https://www.typescriptlang.org/tsconfig#declarationMap

DeclarationMaps are SourceMaps for DTS files; they map it back to their source `*.ts` file.

This is the secret hidden part.

# Yes compile, no bundling

Because `tsup` can't generate `declarationMaps`, it was replaced by good ol' `tsc`, which means:
 - We are not **_bundling_** our code anymore

Bundling was a feature of `tsup`, which `tsc` doesn't do.

 All packages are still published in CJS and ESM formats, but now as multiple individual files mirroring the source directory structure instead of a single `.js` file. Frontend frameworks can easily tree-shake it as part of their bundling.

# Forc optimizations

Every time you do a `pnpm install`, it triggers a `preinstall` routine for `packages/forc-bin` that ~will~ would download all `Forc` binaries. Again. All the time. Starting now, it first checks if it needs to download Forc and skips otherwise.

# Build optimizations

In the middle of every `pnpm build` for the `example-contract` package, it would trigger a `pnpm install` again. Since we use pnpm w/ workspaces, that install would occur on the repo root. Triggering `preinstall` hooks again, re-downloading Forc, and all that good stuff. It would delay the build by around ~10s or so, but not anymore.

> **Note** _I suspect this inner `pnpm install` could be causing those `text: busy` errors. (@Dhaiwat10 @camsjams)_

[...]

PR #870, opened 2023-03-31
fix: issue with an imported CommonJS module & Next.js/Vite support

### LINKED ISSUE #841: The requested module 'elliptic' is a CommonJS module

^0.35.0

### Steps to Reproduce

Here's a simple NextJS [Stackblitz example](https://stackblitz.com/edit/nextjs-zwgubp?file=pages/index.js) with only `fuels` installed

1 - Open the link
2 - Wait for the installation to finish
3 - The app gets a runtime error

For the Vitest example, here's the branch with a simple test for getting the current block number https://github.com/0xYami/fuels-react/tree/feat/unit-tests

1 - Clone the repository
2 - Checkout the `feat/unit-tests` branch
3 - Install dependencies - `pnpm i`
4 - Run local testnet - `fuel-core run --db-type in-memory`
5 - Run tests - `pnpm test`
6 - The same error as above occurs

### Expected Behavior

To be able to use in Vitest and NextJS

### Actual Behavior

I'm running into an issue while writing unit tests using [Vitest](https://vitest.dev) for the [React library](https://fuels-react.com) I'm currently working on. I'm not sure exactly but the error seems related to some CommonJS dependencies not correctly bundled (in this case the `elliptic` dependency) to work properly in ESM.

I tried a few things and noticed that a fresh NextJS project with only `fuels` installed (like in the [reproduction](https://stackblitz.com/edit/nextjs-zwgubp?file=pages/index.js)) crashes when using the `Signer` class which imports the `elliptic` library.

And while trying to debug a bit more I also noticed that having the `fuels-ts` repository locally and linking the `fuels` dependency in my React library directly to it make things work.

Here's the error I'm getting in both Vitest and NextJS.

```
SyntaxError: Named export 'ec' not found. The requested module 'elliptic' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'elliptic';
const {concat: n,hexlify: c,arrayify: s}from"@ethersproject/bytes";import{Address: l}from"@fuel-ts/address";import{hash: b}from"@fuel-ts/hasher";import{randomBytes: u}from"@fuel-ts/keystore";import{toBytes: d}from"@fuel-ts/math";import{ec: p}from"elliptic";function o(){return new p("secp256k1")}var a=class{constructor(e){typeof e=="string"&&e.match(/^[0-9a-f]*$/i)&&e.length===64&&(e=`0x${e}`);let r=s(e),t=o().keyFromPrivate(r,"hex");this.compressedPublicKey=c(t.getPublic(!0,"array")),this.publicKey=c(t.getPublic(!1,"array").slice(1)),this.privateKey=c(r),this.address=l.fromPublicKey(this.publicKey)}sign(e){let t=o().keyFromPrivate(s(this.privateKey),"hex").sign(s(e),{canonical:!0}),y=d(t.r,32),i=d(t.s,32);return i[0]|=(t.recoveryParam||0)<<7,c(n([y,i]))}addPoint(e){let r=o().keyFromPublic(s(this.compressedPublicKey)),t=o().keyFromPublic(s(e)),y=r.getPublic().add(t.getPublic());return c(y.encode("array",!0))}static recoverPublicKey(e,r){let t=s(r),y=t.slice(0,32),i=t.slice(32,64),m=(i[0]&128)>>7;return i[0]&=127,o().recoverPubKey(s(e),{r:y,s:i},m).encode("array",!1).slice(1)}static recoverAddress(e,r){return l.fromPublicKey(a.recoverPublicKey(e,r))}static generatePrivateKey(e){return e?b(n([u(32),s(e)])):u(32)}static extendPublicKey(e){let r=o().keyFromPublic(s(e));return c(r.getPublic(!1,"array").slice(1))}},k=a;export{k: Signer,o: getCurve} = pkg;
```

Dhaiwat10:
Closes #841


arboleya (approved):

Dhaiwat10:
Just verified that this works since I am now able to use a locally linked copy of `fuels` thanks to @arboleya's PR (#827) 😄

Torres-ssf (approved):

Issue #871, opened 2023-04-01
Use the sdk can not access to GraphQL.


LinLL:
### What version of fuels-ts are you using?

fuels@0.37.1

### Steps to Reproduce

console.log(await wallet.getBalance())
The SDK will post the query to the URL "https://beta-3.fuel.network/". However, this may result in a 404 Error. The correct URL to be used is "https://beta-3.fuel.network/graphql".

### Expected Behavior

the account of balance

### Actual Behavior

404 error

Dhaiwat10:
@LinLL can you please share the code where you are initializing the `wallet` instance?

LuizAsFight:
> Hi @LinLL
> 
> We purposefully default to localhost for Fuel network to avoid any confusion or misconfiguration, and we recommend that developers be aware of the desired Provider URL at time of implementation.

this issue is not about `localhost`.. it's mentioning url is missing `/graphql` path in the URL when executing `wallet.getBalance()`

thought `wallet.getBalance()` is working fine in dapps we're working, so probably it's something in @LinLL code or version of wallet

LinLL:
> @LinLL can you please also share the code where you are initializing the `wallet` instance?

```
export const FUEL_PROVIDER_URL = 'https://beta-3.fuel.network/graphql';
const provider = new Provider(FUEL_PROVIDER_URL);
const walletFuel = Wallet.fromAddress(wallet['address'], provider);
const balanceAccountPromise  = walletFuel.getBalance();
```

fuels versions 
  "name": "fuels",
  "version": "0.37.1",

PR #872, opened 2023-04-03
category(fix): Add getFuelNetworkUrl function

protocolwhisper:
This solves #871

### DIFF
--- a/packages/wallet/src/configs.ts
+++ b/packages/wallet/src/configs.ts
@@ -1,1 +1,2 @@
 export const FUEL_NETWORK_URL = 'http://127.0.0.1:4000/graphql';
+export const FUEL_NETWORK_BACKUP_URL = "https://beta-3.fuel.network/graphql";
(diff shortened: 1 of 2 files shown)


LuizAsFight:
@protocolwhisper 

you could try something like this in your dapp:
```
  const getProvider = async (url) => {
    const provider = new Provider(url);

    try {
      await provider.getVersion();
      return provider;
    } catch(e) {
      // try with new Url
      return getProvider(MY_BACKUP_NETWORK);
    }
  }
  getProvider(MY_DEFAULT_NETWORK)
```

luizstacio:
You can achieve the same result using a custom fetch on the Provider. https://github.com/FuelLabs/fuels-ts/pull/767

```ts
import crossFetch, { Response } from 'cross-fetch';

const getProviderUrl = () => {
    return 'http://localhost:4000/graphql';
}

const customFetch = async (url, options) => {
    return crossFetch(getProviderUrl(), options);
}

const provider = new Provider('http://localhost:4000/graphql', {
    fetch: customFetch
});
```

PR #873, opened 2023-04-03
fix: docs workspace version

Torres-ssf:
Fixes new Docs workspace version from `1.0.0` to `0.37.1`


Dhaiwat10 (approved):

arboleya (approved):
I'm not sure it will revert all version bumps, but it's worth trying.

arboleya:
It worked just as expected! Awesome.

PR #880, opened 2023-04-05
feat!: use native TypeScript enums where applicable

camsjams:
We should use a native TypeScript enum if an enum from Sway is a basic empty Tuple style:
```rust
pub enum ColorEnum {
    Red: (),
    Green: (),
    Blue: (),
}

```

This allows for an improvement in DX. Instead of:
```typescript
const INPUT = {
  Green: [],
};
```
We can do:
```typescript
// import `ColorEnum` from abi-typegenerated files
const INPUT = ColorEnum.Green;
```


packages/abi-typegen/src/templates/utils/formatEnums.ts
```diff
@@ -1,22 +1,35 @@
-import type { EnumType } from '../../abi/types/EnumType';
+import type { EnumType, RawTypeHash } from '../../abi/types/EnumType';
```
arboleya, line 1:
```suggestion
import type { EnumType } from '../../abi/types/EnumType';
```

packages/abi-typegen/src/templates/utils/formatEnums.ts
```diff
@@ -1,22 +1,35 @@
...
   const { types } = params;
 
+  const typeHash: RawTypeHash = types.reduce(
+    (hash, row) => ({
+      ...hash,
+      [row.rawAbiType.typeId]: row,
+    }),
+    {}
+  );
+
```
arboleya, line 15:
```suggestion

```

packages/abi-typegen/src/templates/utils/formatEnums.ts
```diff
@@ -1,22 +1,35 @@
...
   const enums = types
     .filter((t) => t.name === 'enum')
     .map((t) => {
       const et = t as EnumType; // only enums here
       const structName = et.getStructName();
       const inputValues = et.getStructContents({ types, target: TargetEnum.INPUT });
       const outputValues = et.getStructContents({ types, target: TargetEnum.OUTPUT });
+      const inputNativeValues = et.getNativeEnum(typeHash);
+      const outputNativeValues = et.getNativeEnum(typeHash);
```
arboleya, line 24: Please, preserve a similar signature to these:

```suggestion
      const inputNativeValues = et.getNativeEnum({ types });
      const outputNativeValues = et.getNativeEnum({ types });
```

packages/abi-typegen/src/abi/types/EnumType.ts
```diff
@@ -6,6 +6,8 @@ import { findType } from '../../utils/findType';
 
 import { AType } from './AType';
 
+export type RawTypeHash = { [key: number]: IType['rawAbiType'] };
+
```
arboleya, line 10:
```suggestion

```

packages/abi-typegen/src/abi/types/EnumType.ts
```diff
@@ -40,6 +42,19 @@ export class EnumType extends AType implements IType {
...
+    // `components` array guaranteed to always exist for structs/enums
+    const enumComponents = components as IRawAbiTypeComponent[];
+
+    if (!enumComponents.every(({ type }) => !typeHash[type])) {
+      return undefined;
+    }
+
+    return enumComponents.map(({ name }) => `${name}`).join(', ');
+  }
+
```
arboleya, line 57:
```suggestion
  public getNativeEnum(params: { types: IType[] }) {
    const { types } = params;

    const { components } = this.rawAbiType;

    // `components` array guaranteed to always exist for structs/enums
    const enumComponents = components as IRawAbiTypeComponent[];

    if (!enumComponents.every(({ type: typeId }) => !findType({ types, typeId })) {
      return undefined;
    }

    return enumComponents.map(({ name }) => `${name}`).join(', ');
  }

```

packages/abi-typegen/src/templates/contract/dts.test.ts
```diff
@@ -74,4 +74,19 @@ describe('templates/dts', () => {
...
+
+    const abi = new Abi({
+      filepath: './my-contract-abi.json',
+      outputDir: 'stdout',
+      rawContents,
+      programType: ProgramTypeEnum.CONTRACT,
+    });
+
+    const rendered = renderDtsTemplate({ abi });
+    expect(rendered).toMatch(/export enum MyEnumOutput { Checked, Pending };$/m);
```
arboleya, line 90: Do you think this test is necessary?

If so, the assertion must be improved.

packages/abi-typegen/src/abi/types/EnumType.ts
arboleya: Also, please add two test cases on the `EnumType.ts` file, one covering each `return` scenario.

arboleya (changes requested):
On point. 👌

I left a few notes, most related to using the `findType()` utility inside `getNativeEnum()`.

packages/abi-typegen/src/abi/types/EnumType.ts
```diff
@@ -40,6 +40,29 @@ export class EnumType extends AType implements IType {
...
+    );
+
+    const { components } = this.rawAbiType;
+
+    // `components` array guaranteed to always exist for structs/enums
+    const enumComponents = components as IRawAbiTypeComponent[];
+
+    if (!enumComponents.every(({ type }) => !typeHash[type])) {
+      return undefined;
+    }
```
arboleya, line 61: In my [previous](https://github.com/FuelLabs/fuels-ts/pull/880#discussion_r1160175034) review, I suggested using the `findType` helper instead of adding a duplicate called `typeHash`.

Is anything wrong with that?

```suggestion
    if (!enumComponents.every(({ type: typeId }) => !findType({ types, typeId })) {
      return undefined;
    }
```

arboleya (changes requested):

Torres-ssf (dismissed):

packages/abi-typegen/src/abi/types/EnumType.ts
arboleya: Make sense. I will give this a complete review soon.

Dhaiwat10 (approved):
LGTM 💯

packages/abi-typegen/src/templates/contract/dts.test.ts
arboleya: I misread the assertion in my first review. As it is, the test makes sense.

However, we could reassign the enum to a const, which should have the same effect with fewer chars.

Just nitpicking, nothing to worry about.

```ts
export enum MyEnumInput { Checked = 'Checked', Pending = 'Pending' };
export const MyEnumOutput = MyEnumInput;
```

arboleya (dismissed):
Great job! 🚀 

My only observation is that I think we should avoid `@ts-expect-error` at _all_ costs, even if the code gets a bit verbose.

Torres-ssf (approved):

arboleya (approved):

PR #886, opened 2023-04-07
ci: ensure docs deploy after changeset PR merge 

Torres-ssf:
The Docs deployment should happen consistently when the Changeset PR is merged.


.github/workflows/release.yaml
```diff
@@ -72,17 +72,26 @@ jobs:
...
 
+      # ensure docs are always deployed after merge of changeset PR
+      - name: Set var SHOULD_DEPLOY_DOCS
+        run: |
+          echo "SHOULD_DEPLOY_DOCS=$( \
+            [[ '${{ steps.changesets.outputs.published }}' == 'true' ]] || \
+            git --no-pager log -1 --pretty=%B | grep -q '^ci(changesets):' \
+            && echo 'true' || echo 'false' \
+          )" >> $GITHUB_ENV
+
```
arboleya, line 83: Maintaining an embedded multiline shell script with special line-break formatting such as this can be tricky. I think it'd be simpler to go with two steps for increased clarity:

```suggestion
      - name: Get the last commit message and set env vars
        run: |
          echo LAST_COMMIT_MSG=$(git --no-pager log -1 --pretty=%B) >> $GITHUB_ENV
​          echo SHOULD_DEPLOY_DOCS=false >> $GITHUB_ENV

      - name: Decides if Docs should be deployed
        if: >-
          ${{
            steps.changesets.outputs.published == 'true' ||
            startsWith(env.LAST_COMMIT_MSG, 'ci(changesets):'
          }}
        run: echo SHOULD_DEPLOY_DOCS=true >> $GITHUB_ENV

```

arboleya (changes requested):

arboleya (approved):

Dhaiwat10 (approved):

Issue #881, opened 2023-04-05
Incorrect encoding of Vec when it's not the first parameter in the function


tkporter:
### What version of fuels-ts are you using?

0.38.0

### Steps to Reproduce

Full repro available here: https://github.com/tkporter/fuels-ts-vec-encoding-bug

### Expected Behavior

I expect Vec encoding to work when calling a contract, regardless of the presence of other parameters to the function

### Actual Behavior

When a Vec<_> is the first parameter of a function in a Sway contract, fuels-ts encodes this properly. However, when the Vec isn't the first parameter to the function, the Vec is encoded incorrectly

See the repro repo https://github.com/tkporter/fuels-ts-vec-encoding-bug for a detailed example

PR #890, opened 2023-04-10
fix: bytes offset start in script transaction

### LINKED ISSUE #888: when executing a contract call, constructor undefined error

0.38

### Steps to Reproduce

on this [repo](https://github.com/LuizAsFight/ui-contract-mint), when trying to execute a contract call with:

```
      const contract = TokenContractAbi__factory.connect(
        '0x8b112594e42fcba3f38f8947a2b6d25c0972233fd7b991b5451d6545dd6340fc',
        wallet
      );

      const invocation = await contract.multiCall([
        contract.functions
          .mint_to_address(bn(100), {
            value: wallet.address.toB256(),
          }, bn(100)),
        contract.functions
          .mint_to_address(bn(100), {
            value: wallet.address.toB256(),
          }, bn(100)),
      ]);

      const txRequest = await invocation.getTransactionRequest();

      fuel.sendTransaction(txRequest, { url: 'http://localhost:4000/graphql' }, wallet.address.bech32Address);
```

### Expected Behavior

should do the contract call

### Actual Behavior

image

LuizAsFight:
Closes #888 

- in case value is `undefined` , added `?` to `normalize` switch code. on this way code doesn't break
- set `bytesOffset`(if exists in input), to `ScriptTransactionRequest` created


packages/providers/src/transaction-request/transaction-request.ts
```diff
@@ -398,10 +400,11 @@ export class ScriptTransactionRequest extends BaseTransactionRequest {
   /** determined bytes offset for start of script data */
   bytesOffset: number | undefined;
 
-  constructor({ script, scriptData, ...rest }: ScriptTransactionRequestLike = {}) {
+  constructor({ script, scriptData, bytesOffset, ...rest }: ScriptTransactionRequestLike = {}) {
     super(rest);
     this.script = arrayify(script ?? returnZeroScript.bytes);
     this.scriptData = arrayify(scriptData ?? returnZeroScript.encodeScriptData());
+    this.bytesOffset = bytesOffset;
```
arboleya, line 407: I miss unit tests for this change:
 1. What's the utility of the `bytesOffset` variable?
 1. What changed after you started setting it on the constructor?

packages/providers/src/utils/json.ts
```diff
@@ -9,7 +9,7 @@ import cloneDeep from 'lodash.clonedeep';
 // stringified and parsed back into a Transaction
 function normalize(object: any) {
   Object.keys(object).forEach((key) => {
-    switch (object[key].constructor.name) {
+    switch (object[key]?.constructor.name) {
```
arboleya, line 12: Could you please add unit tests covering this change?

arboleya (changes requested):

packages/providers/src/transaction-request/transaction-request.ts
LuizAsFight:
```
/** determined bytes offset for start of script data */
bytesOffset: number | undefined;
```

`bytesOffset` is the offset to find script data inside transaction raw payload.

before this change, when creating a new transaction (using `transactionRequestify` for instance), it was not getting the bytesOffset, resulting on
1- a bug(fixed using the `?` change in this PR).
2- incorrect encoding when calling contracts (you can see its usage [here](https://github.com/FuelLabs/fuels-ts/pull/891/files#diff-3409481ab69cbec51784c01628082462a0c87c332fd2e0cdf97fb93d9a59b2fdR25)

only change 1 would stop breaking the code, but encoding would continue incorrect

now `bytesOffset` is saved properly in transactionRequest

LuizAsFight:
@arboleya thanks for the review

created tests for covering new code, could you check again ?

packages/providers/src/transaction-request/transaction-request.ts
Dhaiwat10: > 2- incorrect encoding when calling contracts

@LuizAsFight will this fix #881 by any chance? Or is this unrelated?

Dhaiwat10 (approved):
🫡

packages/providers/src/transaction-request/transaction-request.ts
arboleya: @LuizAsFight It seems the only places using `bytesOffset` are the `constructor` and the `setScript` method.

From this, I'd presume this variable currently has no function.

What am I missing?

https://github.com/FuelLabs/fuels-ts/blob/c8736e8c24fca01b497e6d44fbeb087cd6c570d8/packages/providers/src/transaction-request/transaction-request.ts#L403-L408

https://github.com/FuelLabs/fuels-ts/blob/c8736e8c24fca01b497e6d44fbeb087cd6c570d8/packages/providers/src/transaction-request/transaction-request.ts#L442-L449

arboleya: Oh, never mind, VSCode wasn't finding all references properly; it is used [here](https://github.com/FuelLabs/fuels-ts/blob/59aabffe76e7f03f1a4082b09f302a293a322a42/packages/program/src/functions/base-invocation-scope.ts#L28) and [here](https://github.com/FuelLabs/fuels-ts/blob/c7cb8ac2e268b860a41d29927814c24339f8514a/packages/program/src/functions/invocation-scope.ts#L36).

Looking forward to the answer to @Dhaiwat10's [question](https://github.com/FuelLabs/fuels-ts/pull/890#discussion_r1163462497).

arboleya (approved):
Great work. I'm approving with a minor pending question. 👌 

Finally, could you please share how you came across this issue? It seems so deep in the trenches; I wonder how it was working before this PR and also what was the context or what did you do to trigger this problem.

Torres-ssf (approved):

packages/providers/src/transaction-request/transaction-request.ts
LuizAsFight: hey @Dhaiwat10 , no, it's not related

LuizAsFight:
@arboleya to reproduce you can use this [repo](https://github.com/LuizAsFight/ui-contract-mint)

- deploy the contract
- replace contractId in `App.tsx`
- try to click in the "mint" button


ps. make sure you have the wallet extension installed

PR #897, opened 2023-04-13
feat: create tag to deploy pr to npm

LuizAsFight:
### How it is currently
We can only test getting dependency from npm after code is in master using `next` tag.

### Objective
Allow PRs to be tested through npm dependency

#### Benefits: 
- We detect errors earlier in the flow, avoiding hotfixes after merged in `master` and giving more time to address a solution. Some PRs that change our structure/distribution configs of packages/builds can be break only when using from npm
- When developing dapps, we can use new features from SDK in a new PR directly using the PR npm version

### Feature
Added github Action to deploy current PR to npm.

### How to deploy:
1- accress `pr-release.yaml`
2- comment out `if:false` line
3- commit and push to PR

### Output
the action will comment (check the comment in this PR) in current pr informing the **version** published in NPM, then it's just use directly in package.json:
```
dependencies: {
  "fuels": "0.0.0-pr-123-123123132"
}
```

> **Note:** 
this action is not enabled by default, so we don't flood our npm with a lot of versions from all PRs. PR release should be enabled only when we need it

### After merge
When this PR gets merged, it's recommend to use this feature in #827 , which change our builds


arboleya:
Good stuff.

How could we delete these tags once they're not needed anymore?

Torres-ssf (approved):

LuizAsFight:
> Good stuff.
> 
> How could we delete these tags once they're not needed anymore?

that's other thing we could explore in next PRs I guess, but for now it's same of "next" tags, we don't delete, they stay there 4ever

arboleya:
The difference is that the `next` tag is perpetual, and the PR tags should be temporary.

The cleanup routine must consider all packages, starting from `pnpm-workspace.yaml`.

https://github.com/FuelLabs/fuels-ts/blob/0522917f64d05d992b7607740272e4954e991472/pnpm-workspace.yaml#L1-L3

This is necessary because the `changesets` automatically create _GIT_ `tags` for _all_ packages.

```console
$ pnpm info fuels dist-tags
{
  test: '0.0.0-test',
  latest: '0.38.1',
  master: '0.0.0-master-3bed0aad',
  next: '0.0.0-next-20230413191536',
  'pr-897': '0.0.0-pr-897-20230414020502'
}

$ pnpm info @fuel-ts/wallet dist-tags
{
  master: '0.0.0-master-3bed0aad',
  latest: '0.38.1',
  next: '0.0.0-next-20230413191536',
  'pr-897': '0.0.0-pr-897-20230414020502'
}
```

arboleya (approved):

LuizAsFight:
perfect @arboleya , I filled an issue to it > #901

PR #899, opened 2023-04-14
fix: adjusting `tsconfigs` and correcting broken auto-import suggestions

### LINKED ISSUE #898: Auto-import suggesting absolute paths instead of relative
I noticed it's also importing from `src/..` instead of `../`, is that expected? @arboleya

image
image

_Originally posted by @LuizAsFight in https://github.com/FuelLabs/fuels-ts/issues/827#issuecomment-1502595433_

arboleya:
- Close #898


Torres-ssf (dismissed):

Torres-ssf (approved):

.vscode/settings.json
```diff
@@ -6,6 +6,7 @@
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true
   },
+  "typescript.preferences.autoImportFileExcludePatterns": ["index.ts"],
```
Dhaiwat10, line 9: This will only affect _this_ repo, right? Is this a fix for the bug where auto-import would import files from `src/...` instead of `<package_name>`? If yes, are we sure that this issue was limited to our repo only?

Dhaiwat10 (approved):
LGTM, left a non-blocking question.

.vscode/settings.json
arboleya: Yes, this setting affects only this repo — it ignores `index.ts` files when suggesting auto-import paths, reducing circular-dependency risks. These `index.ts` files are the package's entry point for the external world, but internally, all imports should point to their destination file directly.

The fix for the `src/` was done by removing `baseUrl` from `tsconfig.json`.

Dhaiwat10: That makes sense, thanks @arboleya

arboleya: It turns out this didn't work as planned.

The `index.ts` files are ignored even from outside the package.

So imports like this:

```ts
import {
  getSupportedVersions,
  getUserVersions,
  compareUserVersions,
  colorizeUserVersion,
} from '@fuel-ts/versions';
```

Would be suggested like this:

```ts
import { colorizeUserVersion } from '@fuel-ts/versions/src/lib/colorizeUserVersion';
import { compareUserVersions } from '@fuel-ts/versions/src/lib/compareUserVersions';
import { getSupportedVersions } from '@fuel-ts/versions/src/lib/getSupportedVersions';
import { getUserVersions } from '@fuel-ts/versions/src/lib/getUserVersions';
```

I wonder why it wasn't suggesting it like this on my previous tests.

Anyway, I'm already on it.

arboleya: Removed on:
 - https://github.com/FuelLabs/fuels-ts/pull/902

Issue #769, opened 2023-02-08
Remove check for falsy provider attribute in `Contract::getBalance()`


danielbate:
Given the recent change in #764 to make `walletOrProvider` a required param for the `Contract` constructor, there is now a redundant check in the `getBalance` method within that class to check for a falsy `provider` attribute.

https://github.com/FuelLabs/fuels-ts/blob/9eaad60e1aceffc31882175e9e81ad202bfe41f9/packages/contract/src/contracts/contract.ts#L67-L76

This should be removed as we are currently misreporting on the code coverage for that class.

PR #904, opened 2023-04-17
refactor: remove redundant falsy provider check in `Contract`

danielbate:
Closes #769 

Test coverage is not required here as `getBalance()` gets covered in #760, however this PR should be merged first.


Torres-ssf (approved):

arboleya (approved):
Good catch.

> _I'm assuming we didn't have any tests accounting for this [removed] conditional._

PR #497, opened 2022-09-09
feat: add Vec support as an Input

camsjams:
Closes #434 

Here is a basic example of my thought process, also see #434 for greater details
Blank diagram

PR #910, opened 2023-04-19
fix: correctly offset vector input

### LINKED ISSUE #881: Incorrect encoding of Vec when it's not the first parameter in the function

camsjams:
Aside from fixing the issue, to get there I had to write some tests and iron out the exact nuances. The final result looks simpler overall, but there were some fun bugs to kill as I changed the Vec data code.

I've added a lot of unit tests in different places as a result.

## Issue
The `pointer` of the Vector data was wrong, caused by a bug when the index of the first vector param was not zero.

## Solution

By always starting with the base offset, the bug is fixed.

These lines were the culprit, as this would only run if the first vector was also the first input:
```javascript
if (byteIndex === 0 && byteIndex === paramIndex) {
  return baseVectorOffset;
}
```

Fixes #881 

Note: I'd recommend also seeing the original PR here: #497 which includes a still accurate diagram.


Dhaiwat10 (approved):
Finally 😅

PR #895, opened 2023-04-12
docs: add dedicated package for code snippets from contracts section

Torres-ssf:
- Close: #894

PR #924, opened 2023-04-24
refactor: rename `addMissingVariables` to `estimateTxDependencies`, and add docs for tx dep estimation

### LINKED ISSUE #701: Implement transaction dependencies estimation
Similar to the [Rust SDK](https://github.com/FuelLabs/fuels-rs/blob/a424a9eff6fb0e03df56b15926f6dca39f47c175/packages/fuels/tests/contracts.rs#L468), the TS SDK should be able to estimate the transaction dependencies (contract input set and variable outputs). 

This removes the guesswork of setting the variable outputs and contract input set when performing a call.

Dhaiwat10:
# Summary
As discussed [here](https://github.com/FuelLabs/fuels-ts/issues/701#issuecomment-1518139934), this PR renames the `addMissingVariables` helper to `estimateTxDependencies` for better consistency with the Rust SDK, and adds a docs page for transaction dependency estimation.

I have based this PR off of @Torres-ssf's docs snippet branch since it contains the updated guides around contracts. This PR should automatically change base to `master` once his PR is merged.

Closes #701


apps/docs/src/guide/contracts/transaction-dependency-estimation.md
```diff
@@ -0,0 +1,11 @@
+# Transaction Dependency Estimation
+
+Previously, we mentioned that a contract call might require you to manually specify external contracts or variable outputs.
+
+However, the SDK always automatically estimates these dependencies and double-checks if everything is in order whenever you invoke a contract function or attempt to send a transaction.
+
```
Dhaiwat10, line 6: @digorithm I believe that this behaviour here in the TS SDK is slightly different from the Rust SDK. The Rust SDK doesn't seem to automatically call `estimate_tx_dependencies` on each function invocation, while the TS SDK does that every single time automatically.

I feel like this would make the `estimateTxDependencies` function here in the TS SDK more of an _internal_ helper since SDK users will never have to call it manually.

apps/docs/src/guide/contracts/transaction-dependency-estimation.md
```diff
@@ -0,0 +1,11 @@
+# Transaction Dependency Estimation
+
+Previously, we mentioned that a contract call might require you to manually specify external contracts or variable outputs.
```
danielbate, line 3: Would be good to link to or reference what 'Previously' is referring to here incase this is the first page the reader is landing on.

Dhaiwat10: Good catch! @danielbate

packages/providers/src/provider.ts
```diff
@@ -337,13 +341,18 @@ export default class Provider {
...
-   * Will dryRun a transaction and check for missing VariableOutputs
+   * Will dryRun a transaction and check for missing dependencies.
    *
-   * If there are missing VariableOutputs
+   * If there are missing variable outputs,
    * `addVariableOutputs` is called on the transaction.
+   *
+   * If there are missing output contract IDs, those are fetched and added to the transaction.
+   *
+   * TODO: Add support for missing output messages
```
Dhaiwat10, line 351: Will file a follow-up issue for this.

apps/docs/src/guide/contracts/transaction-dependency-estimation.md
digorithm: Yeah. I'd suggest that, in the near future, we should move to _manually_ calling `estimateTxDependencies`; the additional dry-run calls happening without the user's consent could be an undesired behavior. Or it could be an optional value we set at the call time. We can discuss that later.

packages/providers/src/provider.ts
```diff
@@ -341,6 +341,9 @@ export default class Provider {
    *
    * If there are missing VariableOutputs
    * `addVariableOutputs` is called on the transaction.
+   *
+   * If there are missing output contract IDs, those are fetched and added to the transaction.
+   *
```
Torres-ssf, line 350: Is this functionality working? When I was rewriting the `Inter-contract Calls` doc page I was always running into an error before specifically calling the `addContracts` function in the contract function chain call.

Torres-ssf (changes requested):
Nice job @Dhaiwat10 :+1:.
 
But I have one question though

packages/providers/src/provider.ts
Dhaiwat10: @Torres-ssf there seems to be code in place that would take care of that functionality, but we don't have any tests in place for this. 

https://github.com/FuelLabs/fuels-ts/blob/dadc5760549564480b7b580cb5c268030bf7dcd6/packages/providers/src/provider.ts#L374-L376

arboleya:
Just a side note — there are dozens of unrelated commits here. I'd guess this PR wasn't based on the `master` branch.

packages/providers/src/provider.ts
Dhaiwat10: @Torres-ssf thanks for looking into this with so much detail. Unfortunately, I don't have a lot of insight on the dry-run implementation, but this is def something we need to fix. I can file a follow-up issue for this. Does that sound good?

Dhaiwat10:
@arboleya good heavens. I based this PR off of the branch for #895 since I wanted to use some of the work from there. This has blown up 😆

Dhaiwat10:
Closing this in favor of #954. Tried rebasing this PR against `master` but there are way too many merge conflicts. Creating a new PR seems to be the wiser choice.

packages/providers/src/provider.ts
Torres-ssf: @Dhaiwat10 Yes that sounds great

Issue #865, opened 2023-03-30
UTXO caching mechanism


digorithm:
Mirror of the same feature being developed for the Rust SDK: https://github.com/FuelLabs/fuels-rs/issues/903.

PR #930, opened 2023-04-25
feat: optional UTXO caching

camsjams:
closes: #865 

This option is _disabled_ by default (a point of discussion :warning: )

### This can be enabled like so:

Non-zero number
```javascript
// uses cache time in ms
const provider = new Provider('http://127.0.0.1:4000/graphql', { cacheUtxo: 2500 });
```


packages/providers/src/memory-cache.ts
```diff
@@ -0,0 +1,65 @@
...
+  };
+};
+const cache: Cache = {}; // it's a cache hash ~~> cash?
+
+const DEFAULT_TTL_IN_MS = 30 * 1000; // 30seconds
+
+class MemoryCache {
+  ttl: number;
+  constructor(ttlInMs: number | boolean) {
+    if (typeof ttlInMs === 'boolean') {
```
Torres-ssf, line 17: Shouldn't be `typeof ttlInMs === 'boolean' && ttlInMs`?

Do we really need the type overloading here?

Torres-ssf (dismissed):
Hey good job :+1: 

About to go a little bikeshedding here...

packages/providers/src/provider.test.ts
```diff
@@ -244,4 +251,343 @@ describe('Provider', () => {
...
+      provider.operations.getResourcesToSpend = resourcesToSpendMock;
+      await provider.getResourcesToSpend(owner, [], {
+        utxos: [
+          '0xbc90ada45d89ec6648f8304eaf8fa2b03384d3c0efabc192b849658f4689b9c503',
+          '0xbc90ada45d89ec6648f8304eaf8fa2b03384d3c0efabc192b849658f4689b9c507',
+          '0xbc90ada45d89ec6648f8304eaf8fa2b03384d3c0efabc192b849658f4689b9c508',
+        ],
+      });
+
+      expect(resourcesToSpendMock).toHaveBeenCalledWith({
```
arboleya, line 574: > **Warning** This must be reviewed for all tests.

Several tests developed a pattern of having expectations <sup>`expect(...)` calls</sup> inside `catch` blocks. This doesn't work because `expect()`-ations must always run, while `catch` blocks might **_not_** — for instance, if the code inside the `try` block doesn't throw. This can result in false positives.

You can circumvent this by placing them outside the `catch` block.

```ts
let error

try {
  // will always run
  doSomething();
} catch (err) {
  // might not run
  error = err;
}

expect(error).toBeTruthy(); // ensures method threw (or not)
expect(...); // other validations
```

packages/providers/src/memory-cache.ts
arboleya: My experience with [type overloading](https://github.com/FuelLabs/fuels-ts/issues/748#issuecomment-1400789719) is that it makes things harder, not easier.

FWIW, I suggested a minimal alternative down [here](https://github.com/FuelLabs/fuels-ts/pull/930#discussion_r1183519727).

packages/providers/src/provider.test.ts
```diff
@@ -244,4 +251,343 @@ describe('Provider', () => {
...
+    expect(provider.cache).toBeTruthy();
+    expect(provider.cache?.ttl).toEqual(2_500);
+  });
+
+  it('can cacheUtxo [invalid numerical]', () => {
+    const provider = new Provider('http://127.0.0.1:4000/graphql', { cacheUtxo: -500 });
+
+    expect(provider.cache).toBeTruthy();
+    expect(provider.cache?.ttl).toEqual(30_000);
+  });
```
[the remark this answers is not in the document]
arboleya: Agreed.

packages/providers/src/provider.test.ts
arboleya: @camsjams I wrote a simple test utility for this on the Typegen package. I can quickly export it from a shared package if you think it'd be worth using it on this PR — just let me know.

<details>
<summary>Utility</summary>

https://github.com/FuelLabs/fuels-ts/blob/9d70dfa4f2bcef2caeefac9cd358176c43a4b59a/packages/abi-typegen/test/utils/executeAndCatch.ts#L1-L12
</details>


<details>
<summary>Usage 1</summary>

https://github.com/FuelLabs/fuels-ts/blob/9d70dfa4f2bcef2caeefac9cd358176c43a4b59a/packages/abi-typegen/src/runTypegen.test.ts#L196-L209
</details>


<details>
<summary>Usage 2</summary>

https://github.com/FuelLabs/fuels-ts/blob/9d70dfa4f2bcef2caeefac9cd358176c43a4b59a/packages/abi-typegen/src/utils/findType.test.ts#L34-L38
</details>

arboleya: Cool, will do.

arboleya: Here:
 - #978

arboleya (changes requested):
Looking great.

Just this comment:
 - https://github.com/FuelLabs/fuels-ts/pull/930#discussion_r1183521718

IMO, the class should have a method `getAllEntries()` or similar, that returns everything. Consumers could then choose to filter the entries however they like.

packages/providers/src/provider.test.ts
Dhaiwat10, line 1: Looks like all the single quotes in this file have been changed to double quotes. Can you look into this?

Dhaiwat10 (changes requested):
Looking great apart from some formatting changes that crept in to `provider.test.ts` 🫡

arboleya (dismissed):

Dhaiwat10 (approved):

Dhaiwat10:
Noticed something weird. The `test` workflow is failing but apparently there are no logs:

image

Torres-ssf (approved):

arboleya (approved):

danielbate (approved):

Issue #894, opened 2023-04-12
Documentation - Add dedicated packages for code snippets from `Contracts` section


Torres-ssf:
Two things we need to do on this issue:

1. Create a dedicated package for code snippets
2. Move all code snippets related to the documentation Contract section to the new code snippets package

This issue is part of https://github.com/FuelLabs/fuels-ts/issues/730 and I am splitting it into smaller ones

PR #931, opened 2023-04-26
fix: make the second param in `produceBlocks` optional, add docs for block production

### LINKED ISSUE #912: Add block production with custom timestamp to the docs
Recently we introduced a feature to force-produce blocks with custom timestamps. However, we don't mention it in our docs yet; this should be added somewhere in the docs so users can easily find it.

Note: we should also add docs for block production in general, not only with custom timestamps.

cc @Dhaiwat10

Dhaiwat10:
# Summary
- The second param `timeParameters` in the `produceBlocks` helper was _required_ until now. This is now an _optional_ param, in line with the GQL API and the Rust SDK.
- Added a page for block production to our docs.

Closes #912


apps/docs/src/guide/testing/tweaking-the-blockchain.md
```diff
@@ -0,0 +1,23 @@
...
+import { fromUnixToTai64, Provider } from 'fuels';
+
+const latestBlockNumber = await provider.produceBlocks(
+  amountOfBlocksToProduce: 3,
+  {
+    blockTimeInterval: '100',
+    startTime: fromUnixToTai64(1620000000), // Use the `fromUnixToTai64` helper to convert unix timestamps to tai64
+  }
+);
+```
```
Dhaiwat10, line 21: I have gone for an independent code snippet instead of referencing the unit test for this functionality because the unit test is kinda long, and has a lot of code just for assertions:

https://github.com/FuelLabs/fuels-ts/blob/dadc5760549564480b7b580cb5c268030bf7dcd6/packages/providers/src/provider.test.ts#L202-L247

arboleya: The issue with hardcoded snippets is that they can quickly get outdated, and we cannot validate if they are working as expected. The primary motivation behind #894 is to solve these code snippets' clarity x confidence relationship.

From this perspective, could you please write a new test following your rationale, put it inside the apps/docs-snippets package, and then reference it instead?

Dhaiwat10: @arboleya @Torres-ssf thanks for pointing this out, and I agree. Please review the test that I have added in place of the hard-coded snippet, and let me know if it looks good!

arboleya (commented):
Looking great! Very concise code snippet. 👌 

But shouldn't the `testing` directory be called `providers`?

 - `<docs>/src/guide/providers/producing-blocks.md`
 - `<snippets>/src/guide/providers/producing-blocks.md`

If so, the docs about `Producing Blocks` would be a sub-section of it, also on the left menu.

I also think `Producing Blocks` works better than `Tweaking the Blockchain`.

How do you see it all?

Dhaiwat10:
@arboleya My rationale behind putting this new page under _Testing_ and choosing that title was consistency with the Rust SDK: https://rust.fuel.network/v0.41.0/testing/chains.html

That aside, I do feel like this section should fall under _Testing_ because it covers a feature that lets people force-produce blocks which is something someone would only usually do when testing things out. This is not something they can do on a 'real' network.

I do feel like _Producing Blocks_ would be a better title than _Tweaking the Blockchain_, though - at least until we have more features related to tweaking the blockchain.

arboleya:
@Dhaiwat10 Make sense, and thanks for the explanations.

arboleya (approved):
🚀

Issue #753, opened 2023-01-27
Increase test coverage and overall test quality


digorithm:
This is an issue to track a big effort to improve our test suite. Please, open smaller, more specific issues for this and link them to this one.

danielbate:
Closing as this has been superseded by changes to our overall testing approach and infastructure.

Issue #933, opened 2023-04-27
Improve Vector output support with `Vec<>` return


camsjams:
### Motivation

Currently vector outputs from slice do not cover all possible types, but rather only work with primitive numbers.

### Cover these cases, some already work:

- [x] string
- [x] struct
- [x] arrays
- [x] b256
- [x] b512
- [x] bool
- [x] u8
- [x] u16
- [x] u32
- [x] u64
- [x] options
- [x] enums

### Possible implementations

Blocked by:
https://github.com/FuelLabs/fuels-ts/issues/332

See related tasks:
https://github.com/FuelLabs/fuels-rs/pull/848
https://github.com/FuelLabs/fuels-rs/pull/743

See also:
https://github.com/FuelLabs/fuels-ts/pull/936
https://github.com/FuelLabs/fuels-ts/pull/812

PR #936, opened 2023-04-28
test: refactor `VecCoder` tests

danielbate:
Related to #753

This PR will:
- Introduce a seperate test file for the service under test
- Seperate test cases for `encode` and `decode`
- Remove snapshot tests to bring expectations into the same test file
- Remove redundant tests to more clearly report on code coverage


arboleya (approved):
This is pretty good stuff, although we need more test cases for covering more sub-types _(not only `number` or `boolean`)_ — I guess this might overlap with #933. (cc @camsjams)

```
----------------|---------|----------|---------|---------|---------------------------
File            | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
----------------|---------|----------|---------|---------|---------------------------
  vec.ts        |    87.5 |       50 |     100 |   86.95 | 31,40,57
```

danielbate:
@arboleya more than happy to add more test cases. What would we expect this to look like? A single vector containing all possible sub-types? Or a separate test case for each sub type within a vector?

arboleya:
@danielbate Good question; I suspect a single Vector might be cleaner.

However, I'm not sure about which types we currently support:
 - https://github.com/FuelLabs/fuels-ts/issues/933

Additional test cases could be added alongside the issue above.

arboleya:
@camsjams What do you think about the above comments?

PR #938, opened 2023-04-28
test: refactor `EnumCoder` tests

danielbate:
Related to #753

This PR will:
- Introduce a seperate test file for the service under test
- Seperate test cases for `encode` and `decode`
- Remove snapshot tests to bring expectations into the same test file
- Remove redundant tests to more clearly report on code coverage


arboleya (approved):
```
----------------|---------|----------|---------|---------|-----------------------------------------------------------
File            | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
----------------|---------|----------|---------|---------|-------------------------------------------------
  enum.ts                 |   82.97 |    55.55 |     100 |   82.97 | 45-50,55,74,94
```

Torres-ssf (approved):

danielbate:
I'd like to get those other lines covered within this PR, I've reached out to @camsjams to provide further input on what we would expect the inputs/outputs for a native enum to look like, and will add to this PR.

packages/abi-coder/src/coders/enum.test.ts
```diff
@@ -0,0 +1,97 @@
...
+
+import BooleanCoder from './boolean';
+import EnumCoder from './enum';
+import U64Coder from './u64';
+
+describe('EnumCoder', () => {
+  const coder = new EnumCoder('TestEnum', { a: new BooleanCoder(), b: new U64Coder() });
+
+  it('should encode an enum containing a boolean', () => {
+    const expected = new Uint8Array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1]);
```
Dhaiwat10, line 14: Out of pure curiosity, where you are getting these expected values from?

Dhaiwat10 (approved):

packages/abi-coder/src/coders/enum.test.ts
danielbate: @Dhaiwat10 so I used a different approach dependent on the coder. Boolean values were fairly simple, just had to account for array length if it was part of an enum or struct. For number values and representations, I often opted for min and max values to make the calculations easier, hence the introduction of the constants. For the hex coders I used values that were already available. Arrays, Structs and Vecs then became an amalgamation of the passed coders, just accounting for the offset of where each value starts in the encoded `Uint8Array`.

I could also reverse some of the tests already in place for some of the values. However this is potentially riskier as we are using the previous test values as a source of truth.

There's some really good resources for this as well. I haven't touched this kind of encoding for a while so had to do some reading around each data type first. The [ethers documentation](https://docs.ethers.org/v4/api-utils.html) have some good examples available.

PR #960, opened 2023-04-30
docs: add docs for getting contract balance

### LINKED ISSUE #916: Add docs for getting a contract's balance
Currently, we have a method to pull a contract's balance:

```typescript
getBalance(assetId: BytesLike) {
  return this.provider.getContractBalance(this.id, assetId);
}
```

But we don't have that documented anywhere; we should.

Torres-ssf:
- Close #916


arboleya (commented):
Would it be useful to showcase the Sway contract that the snippet example interacts with to give users the full picture of what the call below is doing?

```ts
const scope = contract.functions.return_context_amount().callParams({
  forward: [amountToForward, NativeAssetId],
});
```

I did something similar here:

 - https://fuellabs.github.io/fuels-ts/guide/abi-typegen/using-generated-types.html#using-generated-predicate-types

I'm not sure if, in this case, it'd add noise to the `getBalance` stuff, but in general, I have the feeling that it's good to have a full understanding of the examples.

See if you think.

Dhaiwat10 (dismissed):

danielbate (approved):

apps/docs/src/guide/contracts/contract-balance.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Checking Contract Balance
+
+Consider a simple contract that transfers a specified amount of a given asset to an address:
+
+<<< @/../../docs-snippets/contracts/transfer-to-address/src/main.sw#contract-balance-2{rust:line-numbers}
+
+The `transfer` function has three parameters:
+
+1. `amount_to_transfer`: The amount that is being transfered.
```
arboleya, line 19:
```suggestion
1. `amount_to_transfer`: The amount that is being transferred.
```

apps/docs/src/guide/contracts/contract-balance.md
```diff
@@ -0,0 +1,35 @@
...
+
+2. `asset`: The address of the deployed contract Token.
+
+3. `recipient`: The address from the receive's wallet.
```
arboleya, line 23:
```suggestion
3. `recipient`: The address from the receiver's wallet.
```

apps/docs/src/guide/contracts/contract-balance.md
```diff
@@ -0,0 +1,35 @@
...
+
+The `transfer` function calls the built-in Sway function `transfer_to_address`, which does exactly what the name suggests.
```
arboleya, line 25:
```suggestion
The `transfer` function calls the built-in Sway function `transfer_to_address`, which does precisely what the name suggests.
```

apps/docs/src/guide/contracts/contract-balance.md
```diff
@@ -0,0 +1,35 @@
...
+
+Let's execute this contract and use the `getBalance` method to validate the remaining asset amount that the contract has left to spend.
```
arboleya, line 27:
```suggestion
Let's execute this contract and use the `getBalance` method to validate the remaining asset amount the contract has left to spend.
```

apps/docs/src/guide/contracts/contract-balance.md
```diff
@@ -0,0 +1,35 @@
...
+
+<<< @/../../docs-snippets/src/guide/contracts/contract-balance.test.ts#contract-balance-3{ts:line-numbers}
+
+In this example, we first forward an asset amount greater than the amount required for the transfer and then execute the contract call.
+
+Finally we use the `getBalance` method to confirm that the contract balance is precisely the total forwarded amount minus the transferred amount.
```
arboleya, line 33:
```suggestion
Finally, we use the `getBalance` method to confirm that the contract balance is precisely the total forwarded amount minus the transferred amount.
```

apps/docs/src/guide/contracts/contract-balance.md
```diff
@@ -0,0 +1,35 @@
...
+
+It is important to note that this method returns the total available contract balance, regardless of how many times assets have been sent or spent on costly operations.
```
arboleya, line 35:
```suggestion
It is important to note that this method returns the total available contract balance, regardless of how often assets have been sent or spent on costly operations.
```

arboleya (changes requested):
Very clean and straight to the point. ❤️ 

Left some trivial tyop corrections and suggestions.

apps/docs/src/guide/contracts/contract-balance.md
Torres-ssf: Good catch @arboleya

arboleya (approved):

Dhaiwat10 (approved):

PR #969, opened 2023-05-02
docs: add section comments for devrel

sarahschwartz:
Adds section comments to the documentation so that the devrel team can easily pull content from this book, and adds some small updates to the docs


camiinthisthang (approved):

.changeset/small-jars-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+"docs": patch
+---
+
+add section comments in docs for devrel team
```
Torres-ssf, line 5: @sarahschwartz I believe the changeset should be empty because no changes were made to the source code, and therefore, we do not want to bump versions.

Torres-ssf (changes requested):
Hey @sarahschwartz good stuff :+1:

.changeset/small-jars-hammer.md
danielbate, line 5:
```suggestion
---
---
```

arboleya (commented):
@sarahschwartz Could we see these snippets in use / applied elsewhere?

arboleya (approved):

Torres-ssf (approved):

PR #971, opened 2023-05-03
refactor: remove redundant return from `VecCoder.decode()`

danielbate:
The `decode` function is not currently supported by the `VecCoder`. Currently it throws an error informing the consumer of the non implementation, however a redundant return statement is still in place.

https://github.com/FuelLabs/fuels-ts/blob/e51a26d86b7ec74b7e24e47676522eea9970b538/packages/abi-coder/src/coders/vec.ts#L55-L58

Someone reading through the source may misinterpret the implementation. This is also misreporting on code coverage.

Code coverage of `decode` is provided in #936


arboleya (approved):
Nice catch.

Dhaiwat10 (approved):

PR #974, opened 2023-05-03
docs: replace generic content for the index page of `Provider` docs with proper content

### LINKED ISSUE #952: Replace generic content with proper docs for the index page of the `Provider` section
https://fuellabs.github.io/fuels-ts/guide/providers/ contains some generic content instead of describing `Provider`.

Dhaiwat10:
Closes #952 

We should probably add more content that covers more if not all of the features available through our `Provider`s, but it felt like out of scope for this PR. This is just a quick fix.


arboleya (approved):

.changeset/dry-cherries-float.md
```diff
@@ -0,0 +1,6 @@
+---
+"docs": patch
+"@fuel-ts/providers": patch
+---
+
+Removed generic content from the index page of the docs for `Provider` and replaced it with proper content.
```
arboleya, line 6: Let's use an _**empty**_ changeset here.

> **Note** This makes me wonder — should we ignore the `docs` in the changeset scheme? (cc @Torres-ssf)

arboleya (changes requested):

Torres-ssf (approved):

PR #950, opened 2023-04-28
docs: add dedicated code snippets for docs types section

Torres-ssf:
- Close #947

PR #988, opened 2023-05-08
feat: `StringCoder` to throw for length value mismatch

### LINKED ISSUE #985: `StringCoder` should `throw` when encoding a string with length mismatch
When instantiating a `StringCoder`, a length is passed to the class instance that is used for both encoding and decoding, `StringCoder.length`. When encoding/decoding, if the length of the value to encode does not match the `StringCoder.length`, then the value to be encoded is sliced:

https://github.com/FuelLabs/fuels-ts/blob/64a69a10fb12b927f62f190025145098569de40b/packages/abi-coder/src/coders/string.ts#L18-L30

This does not match the behaviour of either the rust SDK:

https://github.com/FuelLabs/fuels-rs/blob/c20630cc92b4bcffce0a516dc360bfb8fbe9bfbf/packages/fuels-types/src/core.rs#L44-L50

Or of other coders in the TS SDK:

https://github.com/FuelLabs/fuels-ts/blob/64a69a10fb12b927f62f190025145098569de40b/packages/abi-coder/src/coders/array.ts#L27-L29

This could potentially introduce unexpected behaviour. I believe we should be throwing when encoding strings not suited to the instantiated `StringCoder`, bringing the TS SDK inline with the Rust SDK, and with other coders.

danielbate:
Closes #985 

- `StringCoder` will `throw` when encoding a value that does not match the length of the initially instantiated encoder


danielbate:
Will update to `master` and change from draft once #935 has been merged.

arboleya:
@danielbate FYI—You might have to update [these docs](https://github.com/FuelLabs/fuels-ts/pull/950/files#diff-56c7fc414a26163cb26ed28476369f2c791e892237fddb7847c36cb77d2f1085R18-R24) updates <sup>_(Passing Longer Strings)_</sup> in case #950 gets merged first.

danielbate:
> @danielbate FYI—You might have to update [these docs](https://github.com/FuelLabs/fuels-ts/pull/950/files#diff-56c7fc414a26163cb26ed28476369f2c791e892237fddb7847c36cb77d2f1085R18-R24) _(Passing Longer Strings)_ in case #950 gets merged first.
> 
> > cc @Torres-ssf

Good spot 👍🏻

danielbate:
@Torres-ssf as #935 has been merged, I'm going to convert this from draft but I'll make a PR against #950 with the required changes

arboleya (approved):

Torres-ssf (approved):

PR #995, opened 2023-05-10
feat: support `EvmAddress`

### LINKED ISSUE #921: Add support for the `EvmAddress`
Similarly to https://rust.fuel.network/v0.41.0/types/evm_address.html.

danielbate:
Closes #921 

This PR includes the following:

- The `EvmAddress` interface
- Extension of `Address` to include conversion to `EvmAddress`
- Typegen support for `EvmAddress` so that the type itself can be passed to contract functions rather than the struct input/out types generated by the `StructType`
- Documentation


arboleya:
I know this is still a draft, but let me tell you something — you must add your `echo-evm-address` project to the workspace member's array _(see below)_. This will ensure the project is compiled with the rest and fix the `ENOENT: no such file or directory` errors om the tests.

https://github.com/FuelLabs/fuels-ts/blob/28c1cda969c2cf97519d9289798c18843dc3364b/apps/docs-snippets/projects/Forc.toml#L2-L19

packages/abi-typegen/test/fixtures/forc-projects/evm-address/Forc.toml
```diff
@@ -0,0 +1,7 @@
+[project]
+authors = ["Daniel Bate"]
```
Torres-ssf, line 2:
```suggestion
authors = ["FuelLabs"]
```

Torres-ssf (changes requested):
Awesome :rocket:

Torres-ssf (approved):

apps/docs-snippets/projects/echo-evm-address/src/main.sw
```diff
@@ -0,0 +1,22 @@
...
+        EvmAddress::from(0xbebd3baab326f895289ecbd4210cf886ce41952316441ae4cac35f00f0e882a6)
+    }
+
+    fn echo_address_comparison(evm_addr: EvmAddress) -> bool {
+        let evm_addr2 = EvmAddress::from(0xbebd3baab326f895289ecbd4210cf886ce41952316441ae4cac35f00f0e882a6);
+
+        evm_addr == evm_addr2
+    }
+}
+// #endregion evm-address-1
```
arboleya, line 22: Can we DRY the `0xbebd3baa...` address by assigning it to a variable?

packages/abi-typegen/src/abi/types/EvmAddressType.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  public parseComponentsAttributes(_params: { types: IType[] }) {
+    const capitalizedName = 'EvmAddress';
+
+    this.attributes = {
+      inputLabel: capitalizedName,
+      outputLabel: capitalizedName,
+    };
+
+    this.requiredFuelsMembersImports = ['Address', capitalizedName];
```
arboleya, line 24: Is `Address` really needed here?

Unlike `EvmAddress`, it doesn't seem to be required in the [generated DTS](https://github.com/FuelLabs/fuels-ts/pull/995/files#diff-19032ebe72a0da3b4fed02b9a04ed3803e0a189b450a60f1bc9c5330e0635234) file.

packages/address/src/utils.ts
```diff
@@ -103,3 +104,27 @@ export const addressify = (addressLike: AddressLike | ContractIdLike): AbstractA
...
 export const getRandomB256 = () => hexlify(randomBytes(32));
+
+/**
+ * An EvmAddress contains a B256 with the first 12 bytes cleared, this function
+ * takes a b256 and returns itself with the first 12 bytes cleared
+ *
+ * @param b256 - the address to clear
+ * @returns b256 with first 12 bytes cleared
+ */
+export const getEvmB256fromB256 = (b256: B256Address): B256AddressEvm => {
```
arboleya, line 115: Would moving this utility inside the `address` class make sense?

Similar to how `toEvmAddress` was added, this could be moved as `fromEvmAddress`.

arboleya (commented):
Great work!

I left some minor observations.

packages/abi-typegen/src/abi/types/EvmAddressType.ts
danielbate: On second thoughts it definitely isn't required, introduced it as an import so that `Address` _could_ be used for `EvmAddress` conversions. Will remove 👍🏻

packages/address/src/utils.ts
danielbate: So I did this because it seems to follow the currently implemented pattern of the other `Address` class methods that call utility functions. We seem to abstract away most of the actual address conversion to the utility functions, with `Address` handling the shape.

This was probably done to adhere to DRY hence the `AbstractAddress` and the utility functions, however i think that abstraction is a bit redundant. So we probably could refactor all of that logic to `Address`, however then it is starting to become a pretty bloated class.

I _could_ move this logic out of utils if preferred but to me it seems to follow the current implementation.

arboleya: Thank's for the detailed answer.

My concern is less with the placement of the code block and more with the ease of use.

For example, if one wants to instantiate an `Address` from an `EvmAddress` (I suppose this is a valid use case), it can be cumbersome to figure out (or to explain) that using the `getEvmB256fromB256` utility is required.

What if we renamed it to something like `clearFirst12BytesFromB256` (to be explicit), kept it in the utils land, and then called it from a new `fromEvmAddress` static method defined in the `Address` class?

danielbate: Makes sense 👍🏻

arboleya (approved):

Torres-ssf (approved):

Issue #997, opened 2023-05-11
Configurable Constants For Predicates


Torres-ssf:
We need to add support for setting new values for configurable constants on Predicates

```rust
predicate;

configurable {
    FEE: u8 = 10
}

fn main(inputted_fee: u8) -> bool {
    FEE == inputted_fee
}
```

References:

https://github.com/FuelLabs/fuels-rs/pull/844
https://github.com/FuelLabs/fuels-rs/pull/935
https://github.com/FuelLabs/fuels-ts/pull/991

PR #991, opened 2023-05-09
feat: support configurable constants contracts

Torres-ssf:
- Close: #529

PR #998, opened 2023-05-11
feat: support configurable contants for predicates

Torres-ssf:
- Close: #997

### DIFF
--- a/packages/predicate/src/predicate.ts
+++ b/packages/predicate/src/predicate.ts
@@ -78,4 +78,34 @@ export class Predicate<ARGS extends InputValue[]> extends Account {
     this.predicateData = encoded;
     return this;
   }
+
+  setConfigurables(configurables: { [name: string]: unknown }) {
+    try {
+      if (!this.interface) {
+        throw new Error(
+          'Unnable to validate configurables, Predicate was instantiated without an ABI interface'
+        );
+      }
+
+      if (!this.interface.configurables) {
+        throw new Error('Predicate has no configurables to be set');
+      }
+
+      Object.entries(configurables).forEach(([key, value]) => {
+        if (!this.interface?.configurables[key]) {
+          throw new Error(`Predicate has no configurable named: ${key}`);
+        }
+
+        const { fragmentType, offset } = this.interface.configurables[key];
+
+        const encoded = new AbiCoder().getCoder(fragmentType).encode(value);
+
+        this.bytes.set(encoded, offset);
+      });
+    } catch (err) {
+      throw new Error(`Error setting configurable: ${err}`);
+    }
+
+    return this;
+  }
 }
--- a/.changeset/tricky-dots-collect.md
+++ b/.changeset/tricky-dots-collect.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/predicate": minor
+---
+
+support configurable contants for predicates
--- a/packages/fuel-gauge/src/predicate-with-configurable.test.ts
+++ b/packages/fuel-gauge/src/predicate-with-configurable.test.ts
@@ -1,0 +1,92 @@
+import { generateTestWallet } from '@fuel-ts/wallet/test-utils';
+import { readFileSync } from 'fs';
+import type { CoinQuantityLike } from 'fuels';
+import { WalletUnlocked, Predicate, BN, NativeAssetId, Provider } from 'fuels';
+import { join } from 'path';
+
+import predicateAbi from '../test-projects/predicate-with-configurable/out/debug/predicate-with-configurable-abi.json';
+
+const predicateBytecode = readFileSync(
+  join(
+    __dirname,
+    '../test-projects/predicate-with-configurable/out/debug/predicate-with-configurable.bin'
+  )
+);
+
+const defaultValues = {
+  FEE: 10,
+};
+
+describe('Predicate With Configurable', () => {
+  let wallet: WalletUnlocked;
+  beforeAll(async () => {
+    const provider = new Provider('http://127.0.0.1:4000/graphql');
+
+    const quantities: CoinQuantityLike[] = [
+      {
+        amount: 1_000_000,
+        assetId: NativeAssetId,
+      },
+    ];
+
+    wallet = await generateTestWallet(provider, quantities);
+  });
+
+  it('predicate should returns true when input value is same as configurable default value', async () => {
+    const predicate = new Predicate(predicateBytecode, predicateAbi, wallet.provider);
+
+    const amountToTransfer = 200;
+
+    // transfer amount to predicate
+    await wallet.transfer(predicate.address, amountToTransfer);
+
+    // create destination address
+    const destination = WalletUnlocked.generate();
+
+    const initialBalance = await destination.getBalance(NativeAssetId);
+
+    expect(new BN(initialBalance).toNumber()).toEqual(0);
+
+    // set predicate input data to be the same as default configurable value
+    predicate.setData(defaultValues.FEE);
+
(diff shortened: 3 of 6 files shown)


Torres-ssf:
I am trying to solve this using the same approach that worked for contracts [here](https://github.com/FuelLabs/fuels-ts/pull/991). I could validate locally that this same approach also works for Scripts

However, it's not working for Predicates

1. [Here](https://github.com/FuelLabs/fuels-ts/blob/28f0f30bd70af483d9d8f5bcec5b49827cbed463/packages/predicate/src/predicate.ts#L82-L110) I am setting new values for Configurable constants on Predicates

2. [This is the Predicate](https://github.com/FuelLabs/fuels-ts/blob/28f0f30bd70af483d9d8f5bcec5b49827cbed463/packages/fuel-gauge/test-projects/predicate-with-configurable/src/main.sw#L1-L9) that I have created to run the tests

3. And [this](https://github.com/FuelLabs/fuels-ts/blob/st/feat/support-configurable-contants-for-predicates/packages/fuel-gauge/src/predicate-with-configurable.test.ts) is the test suite

I could validate that the Predicate encoded bytes are being updated, however, it seems the value of the configurable constant it's not being updated, since the Predicate fails to validate [here](https://github.com/FuelLabs/fuels-ts/blob/28f0f30bd70af483d9d8f5bcec5b49827cbed463/packages/fuel-gauge/src/predicate-with-configurable.test.ts#L62-L90)

And this is the error payload:

```json
{
   "response":{
      "data":null,
      "errors":[
         {
            "message":"Invalid transaction: The transaction contains a predicate which failed to validate: TransactionId(0xce0ae529824f332b75b8f86ccb3d4e59aefb946a2b6f5f0422b0c6731a6e317f)",
            "locations":[
               {
                  "line":2,
                  "column":3
               }
            ],
            "path":[
               "dryRun"
            ]
         }
      ],
      "status":200,
      "headers":{
         
      }
   },
   "request":{
      "query":"mutation dryRun($encodedTransaction: HexString!, $utxoValidation: Boolean) {\n  dryRun(tx: $encodedTransaction, utxoValidation: $utxoValidation) {\n    ...receiptFragment\n  }\n}\n\nfragment receiptFragment on Receipt {\n  data\n  rawPayload\n}",
      "variables":{
         "encodedTransaction":"0x000000000000000000000000000000000000000005f5e100000000000000000000000000000000040000000000000000000000000000000200000000000000020000000000000001000000000000000000000000000000000000000000000000000000000000000024000000000000000000000000000000ca1f408e07617f9b2ec89b8845ea993c1063a82b4b4f3ba97fcf3c6076b067d8000000000000000044b7f03836228eebf4ccc3befff6b8e5a9802700dd06fd5dd18f4cabb53c945f00000000000000c800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000006400000000000000089000000447000000000000000000005c5dfcc00110fff30071480003614521017344000b614d210d9000001272400002134114005a41000173400011614d211f90000012240000005d4d30005d43f000134104c024400000470000000000000000000032000000000000000000000032000000000000000020ac5cccf59298707dc405629825c9bdc867a006938df2d4baf9b8d88840394e000000000000000144b7f03836228eebf4ccc3befff6b8e5a9802700dd06fd5dd18f4cabb53c945f00000000000000c800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000006400000000000000089000000447000000000000000000005c5dfcc00110fff30071480003614521017344000b614d210d9000001272400002134114005a41000173400011614d211f90000012240000005d4d30005d43f000134104c0244000004700000000000000000000320000000000000000000000320000000000000000a6bea733e8d0ec1adac62f160404f5789138d591f5d08e8c4d9ec2993bf4d21300000000000000c80000000000000000000000000000000000000000000000000000000000000000000000000000000344b7f03836228eebf4ccc3befff6b8e5a9802700dd06fd5dd18f4cabb53c945f000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
         "utxoValidation":false
      }
   }
}
```


Does anyone have any insight on this? @FuelLabs/sdk-ts

apps/docs-snippets/src/guide/predicates/predicates-with-configurable.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    // set predicate data to be the same as the configurable constant
+    predicate.setData(configurable.WHITELISTED);
+
+    // transfering funds to the predicate
+    const tx1 = await wallet.transfer(predicate.address, 500);
+
+    await tx1.waitForResult();
+
+    const destination = WalletUnlocked.generate();
```
arboleya, line 34: I think `wallet` would fit better than `destination`. If you'd like to be very explicit, you could even use `destinationWallet`, but since only one wallet is in context, `wallet` sounds enough.

packages/predicate/src/predicate.ts
```diff
@@ -25,26 +25,25 @@ export class Predicate<ARGS extends InputValue[]> extends Account {
   predicateData: Uint8Array = Uint8Array.from([]);
   interface?: Interface;
 
-  constructor(bytes: BytesLike, types?: JsonAbi, provider?: string | Provider) {
-    const address = Address.fromB256(getContractRoot(bytes));
+  constructor(
+    bytes: BytesLike,
+    types?: JsonAbi,
```
arboleya, line 30: The term `types` could be more descriptive, like `jsonAbi` or similar.

arboleya (changes requested):
Great stuff! 🚀 

The docs say that `configurables` can be a "flexible mechanism for customizing Predicates behavior", but in my understanding, we're deriving (factoring) new Predicates instead. No?

In other words, if the use of `configurables` recomputes the Predicate's `bytes` and `address` and results in a new Predicate entirely, the first sentence might give a wrong idea.

I wonder if my assumptions are all correct. If they are, these details must be better explained in the docs.

Since much of the `configurables` magic is done internally, it can be challenging for users to grasp it unless it is explicitly mentioned.

Torres-ssf:
> Great stuff! rocket
> 
> The docs say that `configurables` can be a "flexible mechanism for customizing Predicates behavior", but in my understanding, we're deriving (factoring) new Predicates instead. No?
> 
> In other words, if the use of `configurables` recomputes the Predicate's `bytes` and `address` and results in a new Predicate entirely, the first sentence might give a wrong idea.
> 
> I wonder if my assumptions are all correct. If they are, these details must be better explained in the docs.
> 
> Since much of the `configurables` magic is done internally, it can be challenging for users to grasp it unless it is explicitly mentioned.

@arboleya Yeah, I agree with you. I've added a more detailed explanation.

packages/predicate/src/predicate.ts
```diff
@@ -74,8 +73,82 @@ export class Predicate<ARGS extends InputValue[]> extends Account {
...
     const abiCoder = new AbiCoder();
-    const encoded = abiCoder.encode(this.types || [], args);
+    const encoded = abiCoder.encode(this.jsonAbi || [], args);
     this.predicateData = encoded;
     return this;
   }
+
+  private static processPredicateData(
+    bytes: BytesLike,
+    JsonAbi?: JsonAbi,
```
arboleya, line 83:
```suggestion
    jsonAbi: JsonAbi,
```

arboleya (commented):
Minor tyop.

SilentCicero:
Great work here. This will be very useful for all our predicate infrastructure.

Dhaiwat10 (approved):

PR #1010, opened 2023-05-22
ci: add check for markdown links

### LINKED ISSUE #731: Add markdown checker
As file structure and external URL changes, we might end up with some broken URLs on our markdown files. We should configure some markdown checker to minimize the need to perform this task manually.

Dhaiwat10:
# Summary
This PR adds a new workflow that checks the markdown files in our codebase for any dead links. 

Closes #731


package.json
```diff
@@ -78,8 +78,6 @@
     "husky": "^8.0.3",
     "jest": "^29.5.0",
     "jest-text-transformer": "^1.0.4",
-    "markdownlint": "^0.23.1",
-    "markdownlint-cli": "^0.27.1",
```
Dhaiwat10, line 82: Removed these two packages since I noticed they weren't being used anywhere.

packages/abi-coder/README.md
```diff
@@ -16,7 +16,9 @@ This module is responsible for encoding and decoding the Application Binary Inte
 
 ## Documentation
 
-See [Fuel-ts Documentation](https://fuellabs.github.io/fuels-ts/packages/fuel-ts-abi-coder/)
```
Dhaiwat10, line 19: These links used to link to the API reference page for the respective package in our old docs. We don't have API references in our new docs anymore, so I have replaced these links with a link to the home page of our docs.

I wouldn't even mind deleting these `Documentation` sections for packages that don't have docs as of yet. Thoughts?

scripts/lint-md-links.ts
```diff
@@ -0,0 +1,43 @@
...
+  function traverseDirectory(currentPath: string): void {
+    const files = fs.readdirSync(currentPath);
+
+    files.forEach((file) => {
+      const filePath = path.join(currentPath, file);
+      const fileStat = fs.statSync(filePath);
+
+      if (fileStat.isDirectory() && file !== 'node_modules') {
+        traverseDirectory(filePath);
+      } else if (fileStat.isFile() && path.extname(file) === '.md' && file !== 'CHANGELOG.md') {
```
Dhaiwat10, line 18: I am ignoring our changelogs.

packages/abi-coder/README.md
arboleya: I agree with deleting them altogether, as we don't have a 1-1 relationship between the packages x doc's sections. Also, pointing them all to the main docs page sounds repetitive and unnecessary, IMO.

scripts/lint-md-links.ts
arboleya: For listing all markdown files, we could use the `glob` package:

https://github.com/FuelLabs/fuels-ts/blob/b71ad9fd05f9fa478d66a18739be1a8b7956a4d4/packages/abi-typegen/src/runTypegen.test.ts#L75

It seems the API changed a bit in the last version. 🤔
 - https://github.com/isaacs/node-glob

arboleya (changes requested):

packages/abi-coder/README.md
Torres-ssf: Yes, we should delete those

arboleya: Thinking about this a second time, I think there might be some utility in keeping these links on all READMEs pointing to the main documentation page — repetitive as it may sound. Why? Because they will serve a purpose when browsing the packages via NPM.

@Dhaiwat10 @Torres-ssf What do you guys think?

 - https://www.npmjs.com/package/@fuel-ts/abi-coder

doc

arboleya:
@Dhaiwat10 About the PR description/mention of the limitation regarding internal links, it shouldn't be a problem. As we spoke in the last sync, Vitepress will check for these already. (cc @Torres-ssf)

Torres-ssf:
> @Dhaiwat10 About the PR description/mention of the limitation regarding internal links, it shouldn't be a problem. As we spoke in the last sync, Vitepress will check for these already. (cc @Torres-ssf)

Exactly @arboleya. Vitepress fails to build if it finds any dead links. If this happens, the project build will also fail. Therefore, we do not need to extend the job here to internal links.

packages/abi-coder/README.md
Torres-ssf: @arboleya, you raise an excellent point that we hadn't fully considered. Maintaining the links in all READMEs could indeed be beneficial for users browsing the packages via NPM.

Dhaiwat10: @arboleya that's a good point, but I'm not really sure if it would still be that useful because most of our users on npm will interact with the umbrella `fuels` package. I think it would be safe to assume that most people reading the `README` of any of these child packages would be browsing the source code on GitHub. This is why I see these readme files more as _internal_ files than external / user-facing.

arboleya: Yes, this one is a divider, but mostly I agree with your points.

I just wonder if having the link could still be a net-positive measure.

Also, although out of scope for this PR, the fact that the NPM packages don't link back to the GH repo is another issue. In the rare cases _(as you said)_ where users would go to the NPM page, it'd be a dead-end that doesn't connect anywhere (not the docs, not the repo). We should address this at some point.

danielbate: I also think there is some merit in keeping the link. Although I agree that most will be coming from `fuels`, the packages are independently indexed by the npm registry which often ranks highly on search engines. So potentially a rare case now but potentially more important in the future with more of our standardly named packages (`wallet`, `contract`, `address` etc.)

Dhaiwat10: Fair points, guys. I'm gonna add the links back since I do think it could be a net-positive.

danielbate (dismissed):

Torres-ssf (dismissed):

scripts/lint-md-links.ts
```diff
@@ -0,0 +1,19 @@
...
+import { globSync } from 'glob';
+import sh from 'shelljs';
+
+(() => {
+  const mdFiles = globSync('**/*.md', {
+    ignore: ['**/node_modules/**', '**/CHANGELOG.md'],
+  });
+  const filesWithLintErrors: string[] = [];
+  mdFiles.forEach((file) => {
+    const { code } = sh.exec(`npx textlint ${file}`);
```
arboleya, line 11:
```suggestion
    const { code } = sh.exec(`pnpm textlint ${file}`);
```

arboleya (changes requested):

danielbate (approved):

PR #1011, opened 2023-05-22
refactor: docs snippets utility functions

danielbate:
- Introduction of `createAndDeployContractFromProject` to reduce code duplication in tests
- Renaming of `helixfied` -> `hexlified`


Torres-ssf (dismissed):

arboleya (approved):

danielbate:
> Could we extract `createAndDeployContractFromProject()` into its own file? Other tests would benefit from having it shared.
> 
> I'd suggest adding it [here](https://github.com/FuelLabs/fuels-ts/tree/master/packages/utils/src/test-utils):
> 
> * `packages/utils/src/test-utils/deployForcProject.ts`

So this couldn't be extracted because of cyclic dependency issues, unless we do want a `fuels` dep in `utils`. I think `utils` should remain as `fuels` agnostic and purely functional helpers. Then `fuel-gauge` and `docs-snippets` can have their own helpers.

Dhaiwat10 (approved):
Great job!

PR #1033, opened 2023-05-31
feat: increase supported predicate length in transactions

danielbate:
In the TS SDK, we are expecting a u16 integer for predicateLength:

https://github.com/FuelLabs/fuels-ts/blob/0b3c342b395bca48b1c274d3d99cb3fc61eef9a3/packages/transactions/src/coders/input.ts#L72-L73

This should be a u32 integer, as dictated by the fuel spec.

The error can be reproduced with the following predicate:

https://github.com/hashcloak/fuel-crypto/blob/wallet_predicate/predicate/src/main.sw


Torres-ssf (approved):

arboleya (approved):
🎯

PR #1037, opened 2023-06-01
refactor: re-shape example contract into Typegen demo app

arboleya:
This PR does two things:
 1. Added plain javascript files middleware for all CLI's bin entry points
    - Includes **packages/{`abi-typegen`,`fuels`,`versions`}**
    - Fixes broken bin symlinks during the first `pnpm install`
    - Ensures that pnpm symlinks will work from the start _(before the first `build`)_
 1. Refact `packages/example-contract` into `apps/demo-typegen`
    - Clean `packages/` directory from packages that are not _actual_ packages
      - Next ones include `fuels-gauge` and `testscases`
    - Updates script `build:types` in `apps/demo-typegen/package.json` to use `fuels` bin name
      - Thanks to the CLI's javascript middlewares
    - In the future, we can have other `demo-*` here (`demo-nextjs`, `demo-cra`, `demo-vite`, etc)


danielbate (approved):
Really like the demo projects idea 💯

Torres-ssf (approved):

PR #1039, opened 2023-06-01
ci: add the ability to build `forc-bin` from a specific git branch

### LINKED ISSUE #1019: Add the ability to checkout a specific git branch as the source for the `forc-bin` binaries
Right now, the `forc-bin` package is only able to fetch binaries from any of the official releases on the `sway` repo. 

We should make the package more flexible so that it lets us checkout a specific git branch on the `sway` repo if needed. Recently when working on #986, I faced the need to be using `forc` from a specific _unreleased_ git branch. I think we'll need to do this again whenever we have to upgrade `fuel-core` in the future.

Dhaiwat10:
## Summary
This PR extends the `forc-bin` package to give us the ability to use `forc` binaries from a specific git branch on the Sway repo.

## Usage example
In order to use the `forc` version from a specific git branch on the Sway repo, modify the contents of the `VERSION` file like this:

```
git:<branch_name>
```

## Breaking changes
Before this PR, we used to use the `config` field in the `forc-bin` package's `package.json` to specify the version we want.

https://github.com/FuelLabs/fuels-ts/blob/af1934f778ed940465b82dd7093ac1ba92dc6668/packages/forc-bin/package.json#L21-L23

This has now been moved to a `VERSION` file at the package's root.

Closes #1019


Dhaiwat10:
@arboleya I had to tweak the format of how we specify git branches in the `VERSION` file because just looking for a `/` won't work for _all_ git branch names since it's not necessary that all of them will contain a `/`.

arboleya:
@Dhaiwat10 To prevent us from accidentally merging/releasing versions pointing to unpublished branches, we might need to add a new workflow that validates if the current `VERSION` is not a branch name and fails otherwise.

arboleya (commented):
@Dhaiwat10 To prevent us from accidentally merging/releasing versions pointing to unpublished branches, we might need to add a new workflow that validates if the current `VERSION` is not a branch name and fails otherwise.

arboleya:
@Dhaiwat10 We could also use a branch name in the `forc-bin/VERSION` file for this PR (initially) to validate that this _"build from source"_ strategy is working on the CI. I wonder if we need to set up some Rust stuff and whatnot.

Dhaiwat10:
> To prevent us from accidentally merging/releasing versions pointing to unpublished branches, we might need to add a new workflow that validates if the current VERSION is not a branch name and fails otherwise.

I think we will be able to avoid this as things are because the `install` step will fail for the `test` workflow.

Dhaiwat10:
> We could also use a branch name in the forc-bin/VERSION file for this PR (initially) to validate that this "build from source" strategy is working on the CI. I wonder if we need to set up some Rust stuff and whatnot.

Good catch. We'll probably need to setup Rust. Let's have a look.

arboleya:
> I think we will be able to avoid this as things are because the `install` step will fail for the `test` workflow.

But isn't the aim to make `install` and `test` work even when using git branches?

Although we couldn't merge the PR while using a branch, we could have the integration validated on the CI.

Dhaiwat10:
@arboleya oops, I misread your comment. I thought you were referring to an 'invalid' branch name. This makes sense. I'll add a safeguard.

arboleya (approved):
Good one! ⚡

danielbate:
@Dhaiwat10 ah got it! The test workflow definitely through me off. LGTM!

danielbate (approved):

arboleya (approved):

PR #1044, opened 2023-06-05
1021 infer types based off of the abijson directly

nedsalk:


danielbate:
This is a great piece of work and will definitely bring a lot of value and extra utility. But a couple of thoughts from me with the intention of isolating some of the utilities and proving it against existing functionalities.

Can we prove it working using any existing test cases from the typegen package? Or if appropriate test cases don't exist, it would be great to have feature tests that fulfil both to prove no loss of utility.

Following on from this, I think this should be approached in iterations. Iteration one would be to satisfy the functionality we already have, but instead using the inference implementation. Then further iterations would introduce additional utilities. It would also be good to know what you view these iterations to be, to give us a bit of a roadmap of the solution and how you are planning on getting there. I am just a bit concerned that because of the amount of potential utility that this brings, it will grow into a huge piece of work. Type hinting functions is a really nice gain from this but I think would be a good example of something that should come in a second iteration. Therefore should this piece of work include any changes to the contract class signature?

nedsalk:
Closing this one in favor of #1114. It was easier for me to copy/paste relevant code from this branch into a new one than adapting this branch to master.

PR #1046, opened 2023-06-06
feat: support Vector input for Scripts, improve overall Vector input support

### LINKED ISSUE #1015: Passing Vec inside Struct to script causes random values inside script

Latest

### Steps to Reproduce

**The script**
`script;

struct SomeStruct {
    some_number: u64,
    some_vec: Vec<u64> 
}

fn main(
    some_contract: ContractId,
    some_struct: SomeStruct,
) {
    log(some_struct.some_number);
    log(some_struct.some_vec.get(0).unwrap());
    require(some_struct.some_vec.get(0).unwrap() == 1, "wtf is going on");
}`

**The script call**
`async function simpleScript(
    someContract: Contract,
    wallet: WalletUnlocked
) {
    const scriptBin = readFileSync(
        join(__dirname, '../../../scripts/simple-script/out/debug/simple-script.bin')
    );

    const scriptInstance = new Script<[ContractIdInput, Number[]], any>(
        scriptBin,
        simpleScriptAbi,
        wallet
    );

    const someStruct = {
        some_number: 1,
        some_vec: [1],
    };

    const request = await scriptInstance.functions
        .main({ value: someContract.id.toHexString() }, someStruct)
        .addContracts([someContract])
        .txParams({ gasPrice: 1 })
        .call();
}`

### Expected Behavior

Log should return 1 or BN: 0x1 for BOTH the some_number and some_vec.

### Actual Behavior

Log returns 1 for some_number and different, random values for some_vec on each script call.

camsjams:
Fixes #1015

This is a total rewrite of how Vectors work, removing the previous version I implemented and packing Vector data alongside its pointer, to be dynamically appended at the last step of the Abi Coder, see `concatWithVectorData`.

- Improves support for Vectors across the board
  - For context, existing support (**currently available**) for Vector is - Vector, Struct in Vector, Enum in Vector, Numbers/Strings in Vector
  - This PR adds more advanced usage, matching what is supported in the Rust SDK, such as Vector in a struct, Vector in a Vector, Vector in an array, Vector in an enum etc
- Adds support for Vectors in Scripts
   - Vector input to Scripts was never tested or explicitly supported by TS-SDK
  - Luckily the change was to set correct offset instead of the default 0
 - Adds partial support for Vectors in Predicates, continuation in #1083
   - Vector input to Predicates was also never tested or explicitly supported by TS-SDK
   - Likewise we just had to set correct offset instead of the default 0
   - Currently this is only tested with Predicate Coins
- Adds new tests to validate Sway types in Vectors across all three Program modes
  - these were copied from Rust SDK, see https://github.com/FuelLabs/fuels-rs/pull/718

Also see:
- https://github.com/FuelLabs/fuels-ts/pull/910
- https://github.com/FuelLabs/fuels-ts/pull/497

### DIFF
--- a/packages/fuel-gauge/src/predicate.test.ts
+++ b/packages/fuel-gauge/src/predicate.test.ts
@@ -448,7 +448,7 @@ describe('Predicate', () => {
     ).rejects.toThrow('Invalid transaction');
   });
 
-  it('can call a Coin predicate which returns true with valid predicate data [main args vector]', async () => {
+  it.skip('can call a Coin predicate which returns true with valid predicate data [main args vector]', async () => {
     const [wallet, receiver] = await setup();
     const amountToPredicate = 100;
     const chainId = await wallet.provider.getChainId();


Dhaiwat10 (approved):
👏🏻 Great work!

packages/abi-coder/src/coders/enum.ts
```diff
@@ -67,7 +69,7 @@ export default class EnumCoder<TCoders extends Record<string, Coder>> extends Co
     const encodedValue = valueCoder.encode(value[caseKey]);
 
     const padding = new Uint8Array(this.#encodedValueSize - valueCoder.encodedLength);
-    return concat([this.#caseIndexCoder.encode(caseIndex), padding, encodedValue]);
+    return concatWithVectorData([this.#caseIndexCoder.encode(caseIndex), padding, encodedValue]);
```
nedsalk, line 72: The mention of vectors (`concatWithVectorData`) inside of array/enum/struct/tuple coders feels a bit weird to me because they shouldn't be concerned with any ABI type except their own that they're decoding. 
The Rust SDK has the concept of [Data](https://github.com/FuelLabs/fuels-rs/blob/b731e9468ce39972a4eef3af86514768aee91f20/packages/fuels-core/src/types/unresolved_bytes.rs), a nice abstraction via which they've managed to keep the "known already from ABI how it'll be encoded" types like u8, b256, etc. separate from vector-related encoding by having the `Data.Inline` and `Data.Dynamic` enum values. The `Inline` values are in essence already known when their coder returns the encoded values, whereas the `Dynamic` has all the encoding figured out except for `RawVec`'s pointer value, which is recursively set via the `.resolve()` method wherever the encoding needs to happen. By doing it like this, all their coders have a single responsibility of encoding their specific type, and their implementations are vector-agnostic, so to speak.

As far as I'm concerned, this PR looks good - everything works and we should merge it, but I'd love to have a further discussion on the design decisions at some later time.

nedsalk (approved):

arboleya:
@camsjams This seems like a reasonably big PR; I'm turning off auto-merge so more than 2x people can review it before it's merged, ok?

packages/fuel-gauge/src/vector-types.test.ts
```diff
@@ -0,0 +1,180 @@
...
+        TUPLE_IN_VEC,
+        VEC_IN_TUPLE,
+        VEC_IN_A_VEC_IN_A_STRUCT_IN_A_VEC
+      )
+      .call();
+
+    expect(value.toString()).toBe('1');
+  });
+
+  it.skip('can use supported vector types [predicate-vector-types]', async () => {
```
nedsalk, line 134: Why is this test being skipped?

packages/predicate/src/predicate.ts
```diff
@@ -73,8 +79,15 @@ export class Predicate<ARGS extends InputValue[]> extends Account {
   }
 
   setData<T extends ARGS>(...args: T) {
+    const paddedCode = new ByteArrayCoder(this.bytes.length).encode(this.bytes);
+    const OFFSET =
+      VM_TX_MEMORY +
+      TRANSACTION_SCRIPT_FIXED_SIZE +
+      TRANSACTION_PREDICATE_COIN_FIXED_SIZE +
+      paddedCode.byteLength -
+      17;
```
Torres-ssf, line 88: I've set up a Sway project to test the PR implementation and noticed a slight difference in the encoding of the main function's arguments for Predicates compared to Scripts.

I'm sharing the Sway project and the TypeScript test I've used in this [Gist](https://gist.github.com/Torres-ssf/338f93901f041fc2a6cb0eba2b598ee2). 

This is working fine and returning the expected output when is a Script. But turning into a Predicate, it will throw the error related to the "The transaction contains a predicate which failed to validate", even though the expected outcome is the predicate returning true, just like happened when it was a Script. 

Looking at the Predicate data argument binary it's possible to notice a slightly difference when compared to the Script data argument binary, even though they are being encoded using the exact same TS array.

```ts
// predicate
0x0000000000002baf000000000000000200000000000000020000000000002c3f000000000000000300000000000000030000000000002c57000000000000000200000000000000020000000000002c87000000000000000200000000000000020000000000002c97000000000000000300000000000000030000000000002caf000000000000000200000000000000020000000000002cdf0000000000000002000000000000000200000000000000180000000000000038000000000000002b000000000000001d000000000000000c00000000000007e4000000000000000c000000000000000800000000000007e30000000000000000000000000000000100000000000000180000000000000038000000000000002b000000000000001d000000000000000c00000000000007e4000000000000000c000000000000000800000000000007e300000000000000000000000000000001

// script
0x0000000000002af0000000000000000200000000000000020000000000002b80000000000000000300000000000000030000000000002b98000000000000000200000000000000020000000000002bc8000000000000000200000000000000020000000000002bd8000000000000000300000000000000030000000000002bf0000000000000000200000000000000020000000000002c200000000000000002000000000000000200000000000000180000000000000038000000000000002b000000000000001d000000000000000c00000000000007e4000000000000000c000000000000000800000000000007e30000000000000000000000000000000100000000000000180000000000000038000000000000002b000000000000001d000000000000000c00000000000007e4000000000000000c000000000000000800000000000007e300000000000000000000000000000001
```

Torres-ssf (changes requested):
@camsjams I'm still in the process of reviewing this PR, but I've identified a potential issue.

Torres-ssf (changes requested):
@camsjams mate, I've updated the script code used in the previous test to handle a more complex case. It appears that with this more complex case, the vector input argument is not being encoded as expected.

I've created a new [Gist](https://gist.github.com/Torres-ssf/babffa0635f2a309d505f5bd2dce62f8) with the updated version. By adding some logs to the script code, I've verified that the code returns false upon entering the second if statement, even though it shouldn't.

I've tested this script code with the Rust SDK, and it seems to be working as expected there.

Please let me know if this PR is not intended to handle cases like this one.

packages/fuel-gauge/src/vector-types.test.ts
arboleya: I had the same doubt.

Dhaiwat10 (dismissed):
🪄

packages/abi-coder/src/utilities.ts
```diff
@@ -14,49 +15,116 @@ export function hasOptionTypes(types: ReadonlyArray<string | ParamType>) {
...
+  results: Uint8Array,
+  dynamicData: DynamicData,
+  baseOffset: number,
+  dataOffset: number
+): Uint8Array {
+  let cumulativeDynamicByteLength = 0;
+  let updatedResults = results;
+  Object.entries(dynamicData).forEach(([pointerIndex, vData]) => {
+    // update value of pointer
+    const pointerOffset = ~~pointerIndex * WORD_SIZE;
```
nedsalk, line 75: See comment above on double tilde.

nedsalk (changes requested):
A nit on the return statements of `AbiCoder` and a request for change/explanation on the use of double tildes.

danielbate (dismissed):
LGTM, really appreciate the more complex test cases being used also

nedsalk (dismissed):

Torres-ssf (approved):

Torres-ssf (approved):

nedsalk (approved):

PR #1055, opened 2023-06-09
chore: removing `publishConfigs`, using `.d.ts` for types

arboleya:
# TL;DR

This PR favors `.d.ts` files (over `.ts` ones) as type definitions, even for local development. This requires rebuilding the workspace to see changes but increases reliability and compatibility with other tools.

 - [x] Upgrade `tsup`
 - [x] Refactor all `tsup` configs
 - [x] Add `postbuilld` script ([code](https://github.com/FuelLabs/fuels-ts/pull/1055/files#diff-21462feba05a79008c54d9a3267569b27b8897540ac86e47110e57f25402d224))
 - [x] Refactor all `package.json` configs
 - [x] Stops using `tsup` for generating types
 - [x] Start using `tsc` for generating types w/ Declaration Maps
 - [x] Create a new app in `internal` directory for validating configs

> **Note** This problem arose when I was working on #1049, which is blocked until this is merged.

# Summary

The attempt to use `.ts` files instead of `.d.ts` for development _(introduced by #984)_, didn't work as expected. VSCode only fully worked for **_main_** entry points _(not secondary ones)_, and other apps/bundlers would malfunction locally—_probably due to the format discrepancies between DTS and TS_.

Because of this, I'm bringing back the use of Declaration Maps files (`d.ts.map`), from my first experiments in #827. Although it adds an extra step, it is more reliable and compatible with other tools. (sorry @camsjams, I tried a lot)

## This approach offers
 1. Single build _(not more `publishConfig` overrides)_
    - _Local packages are 100% identical to published ones!_
 1. Increased reliability (local builds works like published ones)
 1. Increased compatibility with other tools _(NextJS, CRA, Vite)_

## BUT
 1. We need to build the repo to see the changes
    - Command `pnpm dev` automates this
    - TurboRepo cache helps a bit with speed
    - We can further reduce the over-fragmentation of packages in the future

# Final Notes

If you can, please review this other PR before this one:
 - #1056 

I separated that into another PR to make it easier to review this PR without the bulkiness of the demo apps, but I will need to merge that PR into this one before sending it to `master`.

PR #1056, opened 2023-06-09
chore: adding demo apps

arboleya:
This PR adds app demos for:
 - NextJS
 - React - Using CRA
 - React - Using Vite

 > All apps are built on a `pretest` step _(see `package.json` scripts)_, to keep local `build` times faster.

These apps import `fuels` and log stuff to the console, but nothing else.

This validates that our build works locally with such tools via #1055.

I wonder if we want these apps in the repo since they're very bulky. For example, we could do these validations via parallel GitHub workflows. I just needed a quick and reliable way to validate things now, and this did the job, but I am happy to open new issues to do this after this goes through.

Curious to see what @FuelLabs/sdk-ts thinks.


Torres-ssf (dismissed):

danielbate (dismissed):
The bloat is worth it IMO, then we ensure they are kept in sync with the rest of the mono repo.

Dhaiwat10 (dismissed):
I think this is a fair solution for now. I don't mind keeping these examples inside the monorepo if they don't slow any builds down

nedsalk:
Could we ignore these apps in local development builds and just have them built in the pipeline? It seems to me that ignoring them locally would be okay because they aren't the focus of development. We'd then get the best of both worlds: unaffected build time and this PR serving its purpose via the pipeline build.

nedsalk (approved):
Could we ignore these apps in local development builds and just have them built in the pipeline? We'd then get the best of both worlds: unaffected local build time and this PR serving its purpose via the pipeline build validating that the SDK works with the tools.

arboleya:
I left a note about build time on the PR description:

> All apps are built on a `pretest` step _(see `package.json` scripts)_, to keep local `build` times faster.

danielbate (approved):

Torres-ssf (approved):

PR #1059, opened 2023-06-10
fix: Update Multicall script to forc v0.35.5

bitzoic:
This compiles but there may need to be changes to the sdk to handle `raw_slice`.

In large part the same changes as done in https://github.com/FuelLabs/fuels-ts/pull/1058 except it returns a `raw_slice` over the `ScriptReturn` struct. 

In practice, this is the same as the assembly block it is replacing as it was just returning a pointer and length anyway. It's just a matter of whatever is calling this script playing nice and taking the `raw_slice` as a return type instead.


arboleya (approved):
@bitzoic Thanks a lot!

Dhaiwat10:
Thanks @bitzoic :)

PR #1060, opened 2023-06-10
fix: Bump multicall to v0.40.1

bitzoic:
The last PR was bumping to forc v0.35.5. Looks like we need v0.40.1 instead.


Dhaiwat10:
@arboleya force-merging this one since I need it

PR #986, opened 2023-05-07
chore: upgrade `fuel-core` to `0.18.1`, `forc` to `0.40.1`

Dhaiwat10:
## Summary
This PR bumps the `fuel-core` version to `0.18.1`, and `forc` to `0.40.1`. I have used https://github.com/FuelLabs/fuels-rs/pull/950 and https://github.com/FuelLabs/fuel-core/releases/tag/v0.18.0 as references.

## Major changes
- [x] The `resourcesToSpend` API has been reverted back to `coinsToSpend`.
- [x] Signing a transaction now requires `chainId`.
- [x] The `produceBlocks` API has been updated.
- [x] Message outputs have been removed. We don't need to specify them in the outputs anymore.
- [x] The `MessageProof` schema has been updated according to the new proof API.
- [x] Update some contracts according to the [new Storage API](https://github.com/FuelLabs/sway/pull/4464)
- [x] Use the Sway version from this unreleased PR: https://github.com/FuelLabs/sway/pull/4525

Closes #976

PR #1065, opened 2023-06-12
feat: update multicall script

IGI-111:
Refactor multicall script to use the idioms of the latest Sway version.


Voxelot:
FYI it looks like @bitzoic already updated this script on this branch here: https://github.com/FuelLabs/fuels-ts/pull/986

IGI-111:
Good point, better use that one actually, there's some stuff that could be changed there but the base is better, I'll just close this and comment over there.

PR #1066, opened 2023-06-14
fix: add messageId to input message decoder

LuizAsFight:
input message is broken, this PR fixes it by:
- adding `messageId` field to InputMessage
  - add to messageFragment in operations.graphql, allowing to query it from node
  - include in decoder
  - include `messageId` when adding resources to a transaction
  - fix tests to use `messageId` in input

Closes [FRO-195](https://linear.app/fuel-network/issue/FRO-195/after-sending-transaction-its-showing-incorrect-in-transaction-screen)


Torres-ssf (approved):
Thanks for working on this @LuizAsFight

PR #1074, opened 2023-06-23
feat: adding `@fuel-ts/fuel-core` package

arboleya:
This PR adds a new package to the party — `@fuel-ts/fuel-core` is a bin wrapper around the original `fuel-core`.

This will be used to replace Docker and improve our tooling onward.

> **Note** It follows the `packages/forc` implementation.


danielbate (approved):
LGTM 👍🏻 Potentially out of scope of this PR, but following on from discussions yesterday and doesn't seem like it would be too difficult an extension to this feature, would to be able to specify the version to update to.

If a new version brought in breaking changes the only way to rollback would be manually. We could then also run the command in a test suite and pass versions for backwards compatibility checks?

arboleya:
@danielbate Yes. For now, it's only a bin wrapper — I tried to keep this clean initially, but we can think about exposing a couple of API methods for more complex stuff.

Torres-ssf (approved):

Dhaiwat10 (approved):
🪄

arboleya:
@danielbate @Dhaiwat10 @Torres-ssf I had to fix a conflict with `master`, and all reviews were dismissed.

This is the only new [commit](https://github.com/FuelLabs/fuels-ts/pull/1074/commits/cc22f2e5b907ce6a2ea27c7dc8dfb022245f701a).

danielbate (approved):

Dhaiwat10 (approved):

PR #1081, opened 2023-07-03
chore: upgrading typescript to `5.1.6`

arboleya:


package.json
```diff
@@ -74,27 +75,26 @@
...
-    "textlint-rule-no-dead-link": "^5.1.2",
     "textlint": "^13.3.2",
+    "textlint-rule-no-dead-link": "^5.1.2",
     "ts-generator": "^0.1.1",
     "ts-jest": "^29.0.3",
     "tsup": "^6.7.0",
     "tsx": "^3.12.7",
     "turbo": "^1.8.8",
-    "typescript": "^4.8.4"
+    "typescript": "^5.1.6"
```
nedsalk, line 98: Typescript doesn't follow semver; each of their minor versions may contain a breaking change. Because of that, we should write ~5.1.6 instead of ^5.1.6.

nedsalk (changes requested):

Torres-ssf:
> Although all tests are passing, I noticed [this](https://github.com/FuelLabs/fuels-ts/actions/runs/5442918828/jobs/9898770076?pr=1081#step:11:1773):
> 
> > Warning Version 5.1.6 of typescript installed has not been tested with ts-jest. If you're experiencing issues, consider using a supported version (>=4.3.0 <5.0.0-0). Please do not report issues in ts-jest if you are using unsupported versions.
> 
> Could that be an obstacle to upgrading?

@arboleya, that's a valid concern

However, I think it's acceptable to proceed with the upgrade for now, as we can always downgrade Typescript in the future if any problems do arise

package.json
arboleya: Good catch!

danielbate (approved):

PR #1085, opened 2023-07-04
chore: upgrade `fuel-core` to `0.19.0`

danielbate:
This PR bumps the fuel-core version supported by the TS to `0.19.0`,

Main changes include:
- Txs that contain a predicate now call the `estimatePredicates` endpoint prior to dry running the tx
- Contract root is now calculated by partitioning into chunks of 16 KiB, instead of 8 bytes
- Coin and message inputs now contain a `predicateGasUsed` attribute (with the addition of a `max_gas_per_predicate` consensus param to the localnet config)
- Txs requires at least one input coin or message to be spendable
- When calculating contract storage root, the storage key is hashed and then passed to the SMT

Follow ups:
- WIth the change to txs requiring at least one input coin or message to be spendable, the `get` and `call` invocation methods are now very similar and will be refactored however will keep the API the same
- Introduction of a `MerkleTreeKey` class to ensure type safety and reliability when working with the SMT class, now that it expects hashed storage keys
- Refactor the shared `getContractRoot` functionality when obtaining the root for contracts and predicates


packages/contract/src/util.ts
```diff
@@ -15,25 +15,19 @@ export const getContractRoot = (bytecode: BytesLike): string => {
...
+  chunks[chunks.length - 1] = newChunk;
 
   return calcRoot(chunks.map((c) => hexlify(c)));
 };
 
 export const getContractStorageRoot = (storageSlots: StorageSlot[]): string => {
   const tree = new SparseMerkleTree();
 
-  storageSlots.forEach(({ key, value }) => tree.update(key, value));
+  storageSlots.forEach(({ key, value }) => tree.update(sha256(key), value));
```
arboleya, line 30: @danielbate Should we add a `TODO` and a follow-up issue for doing the key-hashing from inside the Merkle classes? In the future, if we start reusing `fuel-merkle` via WASM, this would be automatically solved.

arboleya: Never mind, I've just read the "Follow ups" section from the PR description. 👍

packages/predicate/src/utils/getContractRoot.ts
```diff
@@ -14,19 +14,12 @@ export const getContractRoot = (bytecode: BytesLike, chainId: number): string =>
...
-    paddedChunk.set(lastChunk, 0);
-    chunks[chunks.length - 1] = paddedChunk;
-  }
-
   const chainIdBytes = uint64ToBytesBE(chainId);
+  const lastChunk = chunks[chunks.length - 1];
+  const remainingBytes = bytes.length % chunkSize;
+  const paddedChunkLength = remainingBytes + ((8 - (remainingBytes % 8)) % 8);
+  const newChunk = lastChunk.slice(0, paddedChunkLength);
+  chunks[chunks.length - 1] = newChunk;
```
luizstacio, line 22: It looks like we have this logic repeated here and on the contract can we reuse it?

luizstacio (changes requested):

packages/predicate/src/utils/getContractRoot.ts
danielbate: Agreed, can we include that in follow ups to unblock anyone requiring 0.19.0? I'm planning on working on the follows immediately, there's a couple other ones in the description of this PR.

luizstacio (approved):

PR #1094, opened 2023-07-10
refactor: purge old abi format from codebase

nedsalk:
This PR purges the old `JsonAbi` format from the code base. Due to the relative importance and size of the PR, I've added comments in the PR where I thought some pointers/explanations would be useful.

The idea of the PR itself is simple: we want to use only the latest json-abi (aka `JsonFlatAbi`) format. However, a full revamp of the encoding/decoding had to happen because the original implementation just transformed the `JsonFlatAbi` into the old format and the whole code base continued using the old format.


apps/docs-snippets/src/guide/contracts/logs.test.ts
```diff
@@ -29,7 +29,7 @@ describe(__filename, () => {
     expect(new BN(logs[0]).toNumber()).toBe(value1);
     expect(logs[1]).toBe(value2);
     expect(logs[2]).toBe(value3);
-    expect([logs[3], logs[4], logs[5]]).toEqual(value4);
+    expect(logs[3]).toEqual(value4);
```
nedsalk, line 32: This is a fix and is how it's supposed to work. If you look at the corresponding sway file, you'll see that the fourth log entry is an array, which is what it is here now with `logs[3]`.

packages/abi-coder/src/abi-coder.ts
```diff
@@ -27,23 +24,100 @@ import {
...
-import { filterEmptyParams, getVectorAdjustments, hasOptionTypes } from './utilities';
+import type { JsonAbi, JsonAbiArgument, JsonAbiType } from './json-abi';
 
 const logger = new Logger(versions.FUELS);
 
-export default class AbiCoder {
-  constructor() {
-    logger.checkNew(new.target, AbiCoder);
+export abstract class AbiCoder {
+  private static getImplicitGenericTypeParameters(
```
nedsalk, line 34: Implicitly generic type parameters happen when a generic type parameter gets passed to e.g. an array in sway.
```
struct MyGenericStruct<A> {
 arr: [A; 5]
}
```
In the json abi, this gets encoded as an array of `generic A`, but there are no `typeParameters` in this generated array abi type, so it's _implicitly_ generic with `generic A`, because the presence of `typeParameters` indicates that the abi type is _explicitly_ generic.

packages/abi-coder/src/abi-coder.ts
```diff
@@ -27,23 +24,100 @@ import {
...
+      };
+    });
+  }
+
+  static resolveGenericComponents(abi: JsonAbi, arg: JsonAbiArgument): readonly JsonAbiArgument[] {
+    let abiType = abi.types.find((t) => t.typeId === arg.type)!;
+
+    const implicitGenericTypeParameters = this.getImplicitGenericTypeParameters(abi, abiType);
+    if (implicitGenericTypeParameters.length > 0) {
+      abiType = { ...structuredClone(abiType), typeParameters: implicitGenericTypeParameters };
```
nedsalk, line 98: In order to simplify the recursion of `resolveGenericArgs`, the abi type is modified so that it becomes _explicitly_ generic, as the recursion of `resolveGenericArgs` depends on the generics being explicit.

packages/abi-coder/src/interface.ts
```diff
@@ -4,110 +4,57 @@ import { arrayify } from '@ethersproject/bytes';
...
-  used to decode logs from contracts other than the main contract
+  TODO: Refactor so that there's no need for externalLoggedTypes
+   
+  This is dedicated to external contracts added via `<base-invocation-scope.ts>.addContracts()` method. 
+  This is used to decode logs from contracts other than the main contract
   we're interacting with.
-*/
-  private externalLoggedTypes: { [id: string]: ReadonlyArray<JsonAbiLogFragment> };
+  */
+  private externalLoggedTypes: Record<string, Interface>;
```
nedsalk, line 25: This is now a `Record<string, Interface>` because the full json abi is needed to decode a log. The one with the knowledge how to decode a log is the `Interface`, as it encapsulates the whole json abi. If it remained a `Record<string, JsonAbiLoggedtype>`, then the logged type might be of an abi type that the `Interface` it was given to as an `externalLoggedType` cannot decode.

packages/abi-coder/src/json-abi.ts
nedsalk, line 1: I've renamed `JsonFlatAbi` into `JsonAbi`, as it is the only abi format existing in the codebase now. This file is now only a collection of `JsonAbi`-related interfaces.

packages/abi-coder/src/json-abi.ts
```diff
@@ -3,203 +3,44 @@
...
+  readonly arguments: ReadonlyArray<string>;
 }
 
-export const isFlatJsonAbi = (jsonAbi: JsonAbi): jsonAbi is JsonFlatAbi => !Array.isArray(jsonAbi);
-/**
- * A JSON ABI object
- */
-export type JsonAbi = ReadonlyArray<JsonAbiFragment> | JsonFlatAbi;
-
-export class ABI {
```
nedsalk, line 91: This class served the purpose of bridging `JsonFlatAbi` into the old format, so it's not needed anymore.

packages/abi-coder/src/utilities.test.ts
nedsalk, line 1: The tests in this file are now being implicitly run inside of `interface.test.ts` in the encoding section. The utilities being tested were:
- `ParamType`, which is removed as it's tied to the old abi format,
- `getVectorAdjustments`, whose tests are implicitly run in the aforementioned encoding section's vector tests. Given that `getVectorAdjustments` is an implementation detail, I don't see much value in testing it separately. If we ever decide to remove `getVectorAdjustments` in favor of refactoring to a better solution, we'd have to remove the tests as well as they'd break, which beats the purpose of "refactoring is changing code without breaking functionality". Given that this is now implicitly tested, if we were to try refactoring it into something else and failed, the tests would catch it. However, if we succeeded, we wouldn't have to delete nor introduce any tests.

packages/abi-coder/src/index.ts
```diff
@@ -1,21 +1,17 @@
...
-export { default as StringCoder } from './coders/string';
-export { default as StructCoder } from './coders/struct';
-export { default as TupleCoder } from './coders/tuple';
-export { default as U64Coder } from './coders/u64';
-export { default as VecCoder } from './coders/vec';
-export * from './utilities';
-export * from './fragments/fragment';
-export { default as FunctionFragment } from './fragments/function-fragment';
-export { default as Interface } from './interface';
-export { default as AbiCoder } from './abi-coder';
```
nedsalk, line 19: `AbiCoder` is no longer being exported from the package because it's now an implementation detail used inside of `Interface` to provide encoding/decoding of everything `JsonAbi`-related.
`Interface` is now the single entry point for anything `JsonAbi`-related.

packages/abi-coder/src/index.ts
```diff
@@ -1,21 +1,17 @@
...
+export { BooleanCoder } from './coders/boolean';
+export { ByteCoder } from './coders/byte';
+export { EnumCoder } from './coders/enum';
+export { NumberCoder } from './coders/number';
+export { StringCoder } from './coders/string';
+export { StructCoder } from './coders/struct';
+export { TupleCoder } from './coders/tuple';
+export { U64Coder } from './coders/u64';
+export { VecCoder } from './coders/vec';
+export type { FunctionFragment } from './function-fragment';
```
nedsalk, line 14: Because `Interface` is the entry point through which interaction with `FunctionFragment`s happens, only the type (and not the class) of `FunctionFragment` is exported for the needs of e.g. building a `FunctionInvocationScope`, as it's tied to the `FunctionFragment` specifically.

internal/check-imports/src/references.ts
```diff
@@ -36,7 +36,7 @@ const { log } = console;
 /**
  * abi-coder
  */
-log(AbiCoder);
+log(Interface);
```
nedsalk, line 39: `Interface` is now the single entry point for interacting with a `JsonAbi`, not the `AbiCoder`.

packages/abi-coder/src/abi-coder.test.ts
```diff
@@ -1,491 +0,0 @@
-import { hexlify, concat } from '@ethersproject/bytes';
```
nedsalk, line 1: All of the tests in this file are now part of the parameterized tests in `interface.test.ts`'s encoding testing part.

packages/abi-coder/src/abi-coder.test.ts
nedsalk, line 1: All of the tests in this file are now part of the parameterized tests in interface.test.ts's encoding testing part.

packages/abi-coder/src/json-abi.test.ts
nedsalk, line 1: All of the tests in this file are either implicitly or explicitly tested in `interface.test.ts`.

packages/fuel-gauge/src/call-test-contract.test.ts
```diff
@@ -49,33 +48,11 @@ describe('CallTestContract', () => {
...
   it('function with empty return should resolve undefined', async () => {
-    const contract = await setupContract({
-      abi: [
-        {
-          type: 'function',
-          name: 'return_void',
-        },
-      ],
-    });
+    const contract = await setupContract();
```
nedsalk, line 52: All of the `abi` values passed into this function in these tests were either already existing in the default abi of the contract being set up or I added them into the sway program and rebuilt the project to have them in the abi.

packages/providers/src/transaction-response/getDecodedLogs.ts
```diff
@@ -8,15 +8,13 @@ export function getDecodedLogs<T = unknown>(
   receipts: Array<TransactionResultReceipt>,
   abiInterface: Interface
 ): T[] {
-  return receipts.reduce((logs, r) => {
+  return receipts.reduce((logs: T[], r) => {
     if (r.type === ReceiptType.LogData) {
-      return logs.concat(...abiInterface.decodeLog(r.data, r.val1.toNumber(), r.id));
+      logs.push(abiInterface.decodeLog(r.data, r.val1.toNumber(), r.id)[0]);
```
nedsalk, line 13: It's taking only the `0` element as it's only interested in the value of the log, not the offset.

packages/program/src/multicall/static-out/multicall.bin
Dhaiwat10, line 1: Why did the multicall binaries change if we didn't change its source code? 🤔

packages/program/src/contract-call-script.ts
```diff
@@ -34,17 +31,14 @@ export const contractCallScript = new ScriptRequest<ContractCall[], Uint8Array[]
...
   contractCallScriptBin,
   (contractCalls) => {
-    const scriptDataCoder = new AbiCoder().getCoder(SCRIPT_INPUTS) as StructCoder<any>;
-    const callSlotsLength = (scriptDataCoder.coders.calls as ArrayCoder<any>).length;
-
-    if (contractCalls.length > callSlotsLength) {
-      throw new Error(`At most ${callSlotsLength} calls are supported`);
+    if (contractCalls.length > 5) {
+      throw new Error(`At most ${5} calls are supported`);
     }
```
Dhaiwat10, line 36: Looks like we were reading the number of max calls allowed from the script data coder earlier, but we are not anymore. Why is this the case?

packages/program/src/contract-call-script.ts
```diff
@@ -34,17 +31,14 @@ export const contractCallScript = new ScriptRequest<ContractCall[], Uint8Array[]
...
 
     let refArgData = new Uint8Array();
 
     const scriptCallSlots = [];
-    for (let i = 0; i < callSlotsLength; i += 1) {
+    for (let i = 0; i < 5; i += 1) {
```
Dhaiwat10, line 41: Same concern around hard-coding max call slots.

packages/predicate/src/predicate.ts
```diff
@@ -85,19 +81,18 @@ export class Predicate<ARGS extends InputValue[]> extends Account {
...
+        // predicateTypes = mainFunction.inputs;
       } else {
         logger.throwArgumentError(
           'Cannot use ABI without "main" function',
-          'Function fragments',
-          predicateInterface.fragments
+          'Abi functions',
+          abiInterface.functions
         );
       }
```
Dhaiwat10, line 97: Maybe this re-written if-else block could use some refactoring, since the new code doesn't do anything if `mainFunction` is not `undefined`. Should also remove that commented out line of code.

packages/contract/src/contract-factory.ts
```diff
@@ -44,18 +44,18 @@ export default class ContractFactory {
...
 
-      The same is done at:
-        - ./contract.ts
+         The same is done at:
+         - ./contract.ts
 
-      @see Contract
-    */
+         @see Contract
+         */
```
Dhaiwat10, line 58: Looks like your formatter added some extra spaces here.

packages/abi-coder/src/function-fragment.ts
```diff
@@ -0,0 +1,224 @@
...
+  readonly signature: string;
+  readonly selector: string;
+  readonly name: string;
+  readonly jsonFn: JsonAbiFunction;
+  readonly attributes: readonly JsonAbiFunctionAttribute[];
+
+  private readonly jsonAbi: JsonAbi;
+  constructor(abi: JsonAbi, name: FnName) {
+    this.jsonAbi = abi;
+    this.jsonFn = abi.functions.find((f) => f.name === name)!;
```
Dhaiwat10, line 38: The linter has left these `Forbidden non-null assertion` warnings in several places on this PR. Can you have a look at them?

packages/abi-coder/src/coders/boolean.test.ts
```diff
@@ -47,24 +45,16 @@ describe('BooleanCoder', () => {
...
-      coder.encode(TRUE_DECODED);
-    }).toThrow('Invalid bool');
-  });
-
+  it.each([undefined, null, 0, {}, [], '', 'a', Symbol('asd')])(
+    'should throw an error when encoding an invalid boolean value',
+    (val) => {
+      expect(() => {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
Dhaiwat10, line 53: Let's replace all `@ts-ignore`s with `@ts-expect-error`s

Dhaiwat10 (commented):
Really good job on this PR @nedsalk 👏🏻  I have left some initial comments/questions.

packages/abi-coder/src/function-fragment.ts
nedsalk: This is due to the nature of `.find()` which returns `T | undefined`. I just realized that I can do an `as T` cast, so I'll do that.

packages/abi-coder/test/sway-projects/exhaustive-examples/Forc.toml
```diff
@@ -0,0 +1,7 @@
+[project]
+authors = ["Nedim"]
```
Torres-ssf, line 2:
```suggestion
authors = ["FuelLabs"]
```

nedsalk: hahaha

packages/program/src/contract-call-script.ts
nedsalk: It was an "I'll deal with it later" but never did thing. I'll fix it now.

packages/program/src/multicall/static-out/multicall.bin
nedsalk: The difference to master is because of forc/fuel-core version mismatches. After I merged master into this branch and reran pnpm build, the difference is gone.

Torres-ssf (dismissed):
Great Job :+1:

packages/abi-coder/test/interface.test.ts
danielbate, line 1: We did speak about this over the call, and this does come down to implementation preference. I prefer verbose tests with individual cases and assertions, that are also easily extensible. Based off my experience with extending looped tests in the past, and also a basis for my work in #935 . But that's not to say you should change your work here, as I can also see your case for it, maybe if anyone else has any strong feelings on it.

danielbate (commented):
Great work on this one, well definitely benefit from simplifying this type,  I've left a couple thoughts. Thankyou for implementing the new test dir structure, and great work on the complex type tests 🚀 


One thing also, I think in the future it'd be great even if we could have PRs into a topic branch, then this being the topic into master. I think that would speed up the review process as there is a lot of work here.

packages/abi-coder/src/abi-coder.ts
```diff
@@ -27,24 +24,107 @@ import {
...
-  constructor() {
-    logger.checkNew(new.target, AbiCoder);
+export abstract class AbiCoder {
+  private static getImplicitGenericTypeParameters(
+    abi: JsonAbi,
+    abiType: JsonAbiType,
+    implicitGenericParametersParam: number[] | undefined = undefined
+  ): number[] {
+    const isExplicitGeneric = abiType.typeParameters !== null;
+    if (isExplicitGeneric || abiType.components === null) return [];
```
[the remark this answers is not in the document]
nedsalk: This logic takes into account those types as well, as their `components` property is always `null`.

Also, the whole code starting from `resolveGenericComponents` never touches types that start with `generic` at all, it only replaces them whenever it finds them being used in a specific component or typeArgument. It identifies that a component/typeArgument is generic by checking if its `type` is found inside of the `typeParameters` property of their top parent. If that component/typeArgument is generic based on this logic, its `type` property gets replaced based on the `typeParametersAndArgsMap`.

internal/check-imports/src/references.ts
nedsalk: I was thinking of doing it inside of another PR. Per our private discussion, `AbiInterface` might not hit the mark because we've got `FunctionFragment`, and renaming `FunctionFragment` into `AbiFunction` or something similar feels kind of off, but `ProgramAbi` sounds good, as `ProgramAbiFunction` flows nicely.

packages/abi-coder/src/abi-coder.ts
```diff
@@ -27,24 +24,107 @@ import {
...
   }
 
-  getCoder(param: JsonAbiFragmentType): Coder {
-    switch (param.type) {
+  static getCoder(abi: JsonAbi, argument: JsonAbiArgument): Coder {
+    const abiType = findOrThrow(
+      abi.types,
+      (t) => t.typeId === argument.type,
+      () => logger.throwArgumentError('Invalid type', 'type', argument.type)
+    );
```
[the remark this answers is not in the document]
nedsalk: I have to use `find` anyways to find the corresponding `type`, but if I use `find` then it'll return `T | undefined`, so if I move the switch statement lower in the function, I have to then trick the compiler with an `as T` cast after `find` or I have to litter all of the code with optional chaining `abiType?.type` and then handle all of the `undefined` cases. This way, by using `findOrThrow`, I will definitely get back a `T` and the code is cleaner.

However, now I've changed the messages of both this error and the one at the end of the function to convey more appropriate messages.

internal/check-imports/src/references.ts
danielbate: Sounds good 👍🏻

danielbate (approved):

Torres-ssf (approved):

PR #1097, opened 2023-07-11
chore: remove pnpm lint from pre-push

nedsalk:
Removed `pnpm lint` from `pre-push.sh` after discussions with the team. We agree that the `pnpm lint` in the workflows is enough.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,2 +1,0 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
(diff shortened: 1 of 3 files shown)


danielbate (dismissed):

arboleya (changes requested):
For a full cleanup, you need to:
 1. Remove the `husky` dependency
 1. Remove all cache flags from `lint:check` and `lint:fix` scripts
 1. Remove `prepare` script

arboleya (approved):

Dhaiwat10 (approved):

PR #1079, opened 2023-06-30
test: refactor predicate feature tests

danielbate:
Bringing some of the testing standards from #1043 to the predicate package, this PR does the following:

- Moves all predicate based tests from the `fuel-gauge` package into the predicate package inside a feature `tests` directory
- Refactored predicate based integration tests in `fuel-gage`
- Moved the predicate _unit_ tests also to the feature test directory in the `predicate` package
- Moved the integration and feature tests into more logical file structure and grouped based on functionality
- Refactored fixtures and types into appropriate directories
- Refactored tests to DRY up the code and moved shared code to `utils` directories
- Changed the language used in all test cases to follow the gherkin syntax and standardise

Issue #1043, opened 2023-06-05
Introduce a TS SDK testing standard and approach


danielbate:
As we approach mainnet, it is important we start to implement some stricter testing standards that ensures parity with the protocol/FuelVM and gives us the greatest level of confidence in the changes we are making as uptake in the SDK increases. Additionally, we can consolidate and structure our testing to give greater clarity as to what behaviours we are currently supporting. I would like to propose some MoSCoW testing standards as well as a roadmap of changes to bring the current SDK in line with the standards.

### Where we are now
- Most functions, classes and modules have a related test file that contains a mix of unit and feature tests
- Each package approaches testing slightly differently
- Each test case is using different language and syntax
- Some packages are separating unit and feature tests into separate directories
- Heavy reliance on `fuel-gauge` for feature and integration level testing
- Some integration testing is done within the documentation directories
- Ad-hoc creation of `forc` projects for tests

### Proposed standards
- Functions, classes and modules must have a supporting unit test file
- Unit tests must verify all behaviours and correctness of functions, classes and modules
- Unit tests must mock any applicable functions, classes or modules that are not the service under test
- Unit tests must have 100% coverage of the service under test
- Packages should have a test directory
- Package test directories should contain fixtures and test data to be shared across tests within the package
- Package test directories could contain mocks and stubs of functions, classes and modules
- Package test directories should contain appropriate feature tests
- Feature tests must simulate real word interactions to validate the behaviour of a packages features from the perspective of an SDK consumer
- Integration tests should be used to demonstrate the interoperability of packages to deliver key behaviours to the SDK consumer
- Test Driven Development (TDD) could be used to identify, isolate and resolve erroneous behaviours and to implement bug fixes
- Shared test helpers and utilities should live within the `utils` packages
- Test cases should follow the gherkin syntax (e.g. `Given X, When Y, Then Z` - for unit tests only `when` and `then` are usually required as the `given` is our service under test. We can then utilise the `it()` syntax for unit tests and `test()` syntax for feature and integration tests)
- CI pipeline should enforce a coverage prerequisite to push to `master`

### Tasks (TBD)
- [ ] Document the test standards and approach within the SDK
- [ ] Refactor each package to contain a feature test directory
- [ ] Add test coverage check to CI pipeline

This should bring greater clarity to the behaviours supported by the SDK, give greater confidence in the quality of our tests and increase overall coverage as discussed in #753 . All standards and tasks are up for discussion and I would appreciate any insights 😃

[...]

PR #1102, opened 2023-07-17
fix: calculate transaction fee

Torres-ssf:
Pretty much what the title says. 

The `calculateTransactionFee` method was updated based on the version used by `fuels-wallet`

### DIFF
--- a/.changeset/serious-pans-search.md
+++ b/.changeset/serious-pans-search.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/providers": minor
+---
+
+Fix incorrect gasUsed and fee calculation in calculateTransactionFee function
--- a/packages/providers/src/operations.graphql
+++ b/packages/providers/src/operations.graphql
@@ -168,7 +168,7 @@ query getVersion {
   }
 }
 
-query getInfoAndConsensusParameters {
+query getInfo {
   nodeInfo {
     nodeVersion
     minGasPrice
--- a/packages/providers/src/operations.graphql
+++ b/packages/providers/src/operations.graphql
@@ -168,10 +168,17 @@ query getVersion {
   }
 }
 
-query getInfo {
+query getInfoAndConsensusParameters {
   nodeInfo {
     nodeVersion
     minGasPrice
+  }
+  chain {
+    consensusParameters {
+      gasPerByte
+      maxGasPerTx
+      gasPriceFactor
+    }
   }
 }
 
--- a/packages/providers/src/provider.ts
+++ b/packages/providers/src/provider.ts
@@ -15,7 +15,6 @@ import {
   ReceiptCoder,
   TransactionCoder,
 } from '@fuel-ts/transactions';
-import { MAX_GAS_PER_TX } from '@fuel-ts/transactions/configs';
 import { GraphQLClient } from 'graphql-request';
 import cloneDeep from 'lodash.clonedeep';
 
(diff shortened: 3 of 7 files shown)


arboleya (commented):
Firstly, I believe it came from @camsjams Vector PR, but just in case, I will mention it here — there are a couple of lint annotations:
 - https://github.com/FuelLabs/fuels-ts/actions/runs/5575462945

> Lint errors are like: _`xyz is defined but never used`_

Finally, following our last `fuel-gauge` discussion about where to put tests and whatnot, should we move the `fee.test.ts` over there (to `fuel-gauge` since it's not a unit test?

Torres-ssf:
> Firstly, I believe it came from @camsjams Vector PR, but just in case, I will mention it here — there are a couple of lint annotations:
> 
> * https://github.com/FuelLabs/fuels-ts/actions/runs/5575462945
> 
> > Lint errors are like: _`xyz is defined but never used`_
> 
> Finally, following our last `fuel-gauge` discussion about where to put tests and whatnot, should we move the `fee.test.ts` over there (to `fuel-gauge`) since it's not a unit test?

@arboleya but the test inside `fee.test.ts` are unit tests :thinking: 

@nedsalk just suggested to me making them feature test, so instead of having these unit tests on `fee.test.ts`, we would have the value returned by these functions validated in the test that tests the function that makes the call to `calculateTransactionFee`

arboleya:
@Torres-ssf Are they unit tests? Maybe I thought you were querying the node. Perhaps I got something wrong.

danielbate (approved):
@arboleya LGTM as the tests in that file are the calculations that are passed values that would be from the node. Begs the question as to whether we'd want an E2E that tests for a specific `gasUsed` and `fee` value for a given contract. At a glance most of our tests check that `gasUsed` or `fee` are at least defined. I don't think such a test is necessary for this piece of work so will leave up to @Torres-ssf 👍🏻

arboleya:
@Torres-ssf @danielbate My bad, I mixed things up.

The `fee.test.ts` is indeed a unit.

Unlike the `provider.test.ts` — which uses the node for fetching stuff.

As I said before, I prefer this approach of having the `/test` directory, but that wasn't what we discussed at our last discussion with @camsjams. From what I recall, this test should be on the `fuel-gauge` and import the Provider from `fuels`, like so:

```ts
```

Was it me who understood things wrong? I'm not claiming I'm right.

But we surely need to clearly define these convention rules to avoid misinterpretations.

danielbate:
My understanding is that unit tests can remain and will be alongside `src` files. Feature tests, which will be the entry points for the packages, will go in a `/test` directory. Anything making calls to a node and thus requiring encoding/decoding etc will go in `fuel-gauge` as an integration test. On which should be prioritised and implemented is still under discussion / up to the implementer.

I plan on defining this more clearly in #1043 once #1079 has been completed and merged.

Apologies also I missed the `provider.test.ts`, agreed that under definition this would come under an integration test and should be within `fuel-gauge`.

Torres-ssf:
@arboleya @arboleya I've only touched on one test case from `provider.test.ts` to make it conform with the changes that I made.

No test cases were implemented on it.

Do you think that moving `provider.test.ts` to `fuel-gauge` should be part of this PR scope?

danielbate:
@Torres-ssf I'd be happy to count it as out of scope of this work and instead include it in a provider test refactor piece

arboleya (dismissed):

arboleya:
:shipit:

Torres-ssf:
@arboleya @danielbate I've just revet the renaming of a method and query to avoid conflicts with the fuels wallet

arboleya (approved):

nedsalk (approved):

Issue #1071, opened 2023-06-20
Remove the usage of the old ABI format from the code


nedsalk:
With the merging of #986, the old `JsonAbiFragment` format is not being used anywhere in the code and isn't expected to be passed. However, some parts of the code (e.g. `Contract`) still accept it.
This issue's PR will delete the interface and fix all the tests to use the `JsonFlatAbi` format instead.

PR #1107, opened 2023-07-18
fix: nested implicit generics get encoded properly

### LINKED ISSUE #1106: bug: nested implicit generics fail encoding
Example sway code:
```
abi MyContract {
    fn array_with_generic_struct(arg: ArrWithGenericStruct<b256>);
}

struct ArrWithGenericStruct<U> {
    a: [MyStructWithGeneric<U, u8>; 3]
}

struct MyStructWithGeneric<T, U> {
    bim: T,
    bom: StructA<U, T>,
    bam: StructB<U>
}

struct StructA <T, U>{
  propA1: T,
  propA2: U,
}

struct StructB<T>{
  propB1: T,
}
```
It fails due to the nesting of generics and possibly passing of `generic U` into `T` into `U`.

nedsalk:
### Scope of PR
- Added test which showcases issue
- Fixed the issue
- Refactored generic to specific type resolution from `AbiCoder` into `ResolvedAbiType` class

close #1106


arboleya (commented):
@nedsalk You merged the former PR before I could get to it, so I'm coming from a foggy place and thus unable to give completely relevant feedback other than by feeling — to this extent, I find it _**very**_ confusing and probably ill-advised if you are _changing_ the structure of the ABI before using it, since now we have to do some mental gymnastics to keep two mental models in mind. One follows the spec, and the other follows your interpretation.

> Please, don't forget to self-assign your PRs and optimally add a `close #<NNN>` to the description — this is to both _reference_ and _close_ the related issue when merging the PR.

nedsalk:
I agree, it was a bit out of place, although the underlying logic of what was changed in the private ABI was pretty straightforward. I've changed it now to use `ResolvedAbiType` and now everything stems again from one abi type, as was the case with my initial implementation for #1071, but is a bit cleaner.

arboleya (commented):
I tried reviewing this but couldn't make sense of it.

Perhaps I'd have to retrospectively review the former PR (https://github.com/FuelLabs/fuels-ts/pull/1094) to fill the gaps.

packages/abi-coder/src/abi-coder.ts
```diff
@@ -168,28 +76,30 @@ export abstract class AbiCoder {
...
       return new ArrayCoder(arrayElementCoder, length);
     }
 
     if (abiType.type === VEC_CODER_TYPE) {
-      const typeArgument = components.find((x) => x.name === 'buf')?.typeArguments?.[0];
-      if (!typeArgument) {
+      const arg = findOrThrow(components, (c) => c.name === 'buf').originalTypeArguments?.[0];
+      if (!arg) {
         throw new Error('Expected Vec type to have a type argument');
       }
```
danielbate, line 87: Is this now redundant given `findOrThrow` on the above line?

nedsalk:
@arboleya I've refactored it to a better design now and I think there is no need to reference the old PR anymore as this design is more straightforward. Reading the old design would add to confusion.

packages/abi-coder/src/resolved-abi-type.ts
```diff
@@ -0,0 +1,180 @@
+import { joinSignature } from '@ethersproject/bytes';
```
Dhaiwat10, line 1: The linter is reporting a warning here. Can you fix it?

packages/abi-coder/test/interface.test.ts
```diff
@@ -777,8 +816,8 @@ describe('Abi interface', () => {
 
   describe('abi types', () => {
     it('should return the correct type when it exists', () => {
-      const abiType = exhaustiveExamplesInterface.getTypeById(22);
-      expect(abiType.type).toEqual('enum EnumWithStructs');
+      const abiType = exhaustiveExamplesInterface.getTypeById(0);
+      expect(abiType.type).toEqual('()');
```
Dhaiwat10, line 820: I'm not being able to understand this change from `enum EnumWithStructs` to `()`. Can you help me understand?

nedsalk: This change is because the `typeId` is auto-generated by `forc build` and might change from build to build. This test was failing for me because it **did** change and `typeId` of value `22` was something different than `'enum EnumWithStructs'`. The actual functionality worked, just the assumption of the test that the underlying type `22` was this enum was flaky. That's why I changed to `0` and `'()'`, because the chance that this changes is far lesser than a _high-valued_ `typeId` like `22`. From what I've noticed in abi generation, if `()` exists in the sway file, it'll be the first one to get put into the abi and will have `0`. This is also a flaky assumption because the abi might also not have this particular type in it, but it's a less flaky one, especially in the case of such a large sway project like `exhaustive-examples`.

packages/abi-coder/src/function-fragment.ts
```diff
@@ -215,7 +165,7 @@ export class FunctionFragment<
     if (outputAbiType.type === '()') return [undefined, 0];
 
     const bytes = arrayify(data);
-    const coder = AbiCoder.getCoder(this.jsonAbi, this.jsonFn.output);
+    const coder = AbiCoder.getCoder(new ResolvedAbiType(this.jsonAbi, this.jsonFn.output));
```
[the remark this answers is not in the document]
nedsalk: Hmm, you're right, it was cleaner. However, the `ResolvedAbiType` is supposed to be a self-sufficient representation of a type where all of its generics have been resolved, no matter how deeply nested they are. Instantiating it _one time_ in the `getCoder` method as it currently is wouldn't be possible as this method is used recursively. 

There are two options:
1. Use your proposal, but create a private `getCoderImpl` method which takes a `ResolvedAbiType`, or
2. Overload the signatures of `getCoder` to `getCoder(abi: JsonAbi, argument: JsonAbiArgument)` and `getCoder(resolvedAbiType: ResolvedAbiType)`, and then have the implementation check if the first argument is an `instanceof ResolvedAbiType` and, if it's not, then create it because we're then sure that the `getCoder(abi, argument)` overload was used.

I've initially implemented the second option, but after some thought I realized that it adds more cognitive load than the first, so I've chosen the first option.

Dhaiwat10 (commented):
I agree with @camsjams that this does seem a bit cleaner, but I was wondering if splitting this PR up into two would be a better idea since this PR includes a fix and a decent amount of refactoring. I am struggling to identify the bug fix among all the changes related to the refactor. Thoughts?

I apologise in advance in case I am missing any context from the call that was held for the mob review for the other PR last week, I wasn't there.

nedsalk:
@Dhaiwat10 
>  I am struggling to identify the bug fix among all the changes related to the refactor. 
I have refactored it in order to have the bug fix be more straightforward to implement. It would be a pain to go back to the old, arguably worse solution, implement the fix, merge it, and then do the refactoring.

packages/abi-coder/src/interface.ts
```diff
@@ -20,27 +20,27 @@ export class Interface<TAbi extends JsonAbi = JsonAbi> {
   TODO: Refactor so that there's no need for externalLoggedTypes
    
   This is dedicated to external contracts added via `<base-invocation-scope.ts>.addContracts()` method. 
-  This is used to decode logs from contracts other than the main contract
+  This is used to decode logs from contracts ot her than the main contract
```
Dhaiwat10, line 23: Looks like an unintentional change 😅

packages/abi-coder/src/interface.ts
```diff
@@ -20,27 +20,27 @@ export class Interface<TAbi extends JsonAbi = JsonAbi> {
...
+  readonly jsonAbi: TAbi;
 
   constructor(jsonAbi: TAbi) {
     this.jsonAbi = jsonAbi;
 
     this.externalLoggedTypes = {};
 
     this.functions = Object.fromEntries(
-      jsonAbi.functions.map((x) => [x.name, new FunctionFragment(jsonAbi, x.name)])
+      jsonAbi.functions.map((x) => [x.name, new FunctionFragment(this.jsonAbi, x.name)])
```
Dhaiwat10, line 35: Let's try using either one of `jsonAbi` or `this.jsonAbi` consistently everywhere for more clarity.

packages/abi-coder/src/interface.ts
```diff
@@ -20,27 +20,27 @@ export class Interface<TAbi extends JsonAbi = JsonAbi> {
...
     );
 
     this.configurables = Object.fromEntries(jsonAbi.configurables.map((x) => [x.name, x]));
   }
 
   /**
    * Returns function fragment for a dynamic input.
-   * @param nameOrSignatureOrSelector - name (e.g. 'transfer'), signature (e.g. 'transfer(address,uint256)') or selector (e.g. '0x00000000a9059cbb') of the function fragment
+   * @param nameOrSignatureOrSelector - name (e.g. 'transfer'), signature (e.g. 'transfer(address,uint256)')  or selector (e.g. '0x00000000a9059cbb') of the function fragment
```
Dhaiwat10, line 43: This could be another unintentional change

Dhaiwat10 (dismissed):
Looks good, just some small adjustments that you can make!

danielbate (changes requested):
Looking good, couple small suggestions

Dhaiwat10 (approved):

packages/abi-coder/src/abi-coder.ts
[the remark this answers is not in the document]
nedsalk: This just takes the first element, it's not running any `find` logic. I can use something like checking if the index is 0, but I think this is cleaner.

packages/abi-coder/src/resolved-abi-type.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  private static getResolvedGenericComponents(
+    abi: JsonAbi,
+    arg: JsonAbiArgument,
+    components: readonly JsonAbiArgument[] | null,
+    typeParameters: readonly number[] | null
+  ) {
+    if (components === null) return null;
+    if (typeParameters === null || typeParameters.length === 0)
+      return components.map((c) => new ResolvedAbiType(abi, c));
```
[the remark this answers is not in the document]
nedsalk: I think that should be configured in the linter and then fixed via `pnpm lint:fix`. If our current linting standard allows it, it's okay IMO.

danielbate (approved):

Issue #708, opened 2022-12-24
Create `fuel-ts/tx` package


LuizAsFight:
### Motivation

Currently the SDK returns `Transaction`, `TransactionResponse` and `TransactionResult`, over 3 different requests inside provider module.
- If I need transaction `receipts`, I need to create a `TransactionResponse` and use `waitForResult` method.
- If I need transaction `graphql query` data, I cannot access it because SDK doesn't return. Only way is to build the queries in the frontend app itself
- If I need status, both `TransactionResponse.fetch` and `TransactionResponse.waitForResult` return it, but in different forms.

In order to show correct transaction data in front end, we need to apply a lot of logic, deal with this SDK different ways of getting transaction data, explore transaction specs, deal with `outputs` , `inputs` , `receipts`, process them in different ways depending on transaction type.

### First Solution

To simplify and unify things, In [fuels-wallet](https://github.com/FuelLabs/fuels-wallet) repo a collection of [tx util](https://github.com/FuelLabs/fuels-wallet/blob/master/packages/app/src/systems/Transaction/utils/tx.ts) functions was created.

- It has a `lot` of functional helpers to get whichever transaction data the user wants, and also process them depending on the type.
  - All this helpers are extensively `unit tested`, as I wanted to let it as ready as possible to go in SDK.
- for most simple cases, only `parseTx` method should be enough, returning a `Tx` object:
```
export type Operation = {
  name?: string;
  from?: Address;
  to?: Address;
  assetsSent?: Array<Coin>;
};

export type Tx = {
  operations: Operation[];
  gasUsed: BN;
  fee: BN;
  type: TxType;
  totalAssetsSent: Coin[];
  isTypeMint: boolean;
  isTypeCreate: boolean;
  isTypeScript: boolean;
  status?: TxStatus;
  isStatusPending?: boolean;
  isStatusSuccess?: boolean;
  isStatusFailure?: boolean;
  id?: string;
};
```

- It centralizes `fee` and `gasUsed` calculation.
- it returns the `operations` of the transaction

### Proposal

Create a new package `fuel-ts/tx`. This package will contain
- tx utility methods
- other methods to request transaction data 
  - `getTx` - returns tx data
    - input: `{ id: string; provider: Provider; }`
    - output: `{ gqlTransaction, transaction, tx }`
      - `gqlTransaction` is the data return from `graphql query`
      - `transaction` is the decoded transaction from `gqlTransaction.rawPayload`
[...]

PR #1108, opened 2023-07-18
fix!: improve transaction response

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/708


Dhaiwat10 (dismissed):
👏🏻

nedsalk (approved):

packages/providers/src/transaction-summary/assemble-transaction-summary.ts
```diff
@@ -0,0 +1,133 @@
...
+    transactionWitnesses: transaction?.witnesses || [],
+  });
+
+  const operations = getOperations({
+    transactionType: transaction.type,
+    inputs: transaction.inputs || [],
+    outputs: transaction.outputs || [],
+    receipts,
+    rawPayload: hexlify(transactionBytes),
+    // abiMap, TODO: update this when it's possible to get a contract JsonAbi using graphql
```
LuizAsFight, line 76: we should allow abiMap to be a input from user. if user has the contract ABI in hands we should accept

packages/providers/src/transaction-summary/get-transaction-summary.ts
```diff
@@ -0,0 +1,110 @@
...
+  const transactionInfo = assembleTransactionSummary<TTransactionType>({
+    id: gqlTransaction.id,
+    gasPrice: bn(gqlTransaction.gasPrice),
+    receipts,
+    transaction: decodedTransaction,
+    transactionBytes: arrayify(gqlTransaction.rawPayload),
+    gqlTransactionStatus: gqlTransaction.status,
+    gasPerByte: bn(gasPerByte),
+    gasPriceFactor: bn(gasPriceFactor),
+  });
```
LuizAsFight, line 47: what happens if it's a pending transaction (not processed yet in a block). it won't have receipts, but should work anyway and show the data we can proccess.

for example fees will not be shown because need receipts for that.

packages/providers/src/transaction-response/utils.ts
```diff
@@ -0,0 +1,36 @@
+import { arrayify } from '@ethersproject/bytes';
+import type { BN } from '@fuel-ts/math';
+import { ReceiptType, ReceiptCoder } from '@fuel-ts/transactions';
+
+import type { GqlReceiptFragmentFragment, GqlTransaction } from '../__generated__/operations';
+
+import type { TransactionResultReceipt } from './transaction-response';
+
+export const processGqlReceipt = (
```
LuizAsFight, line 9: this function should be incorporate inside transaction-summary too? I think it makes sense

LuizAsFight (changes requested):
so happy to see this code getting in the sdk 👏🏻

can you remove lint warnings in the code you changed?

packages/providers/src/transaction-summary/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './get-transaction-summary';
+export * from './types';
```
LuizAsFight, line 2: I think we can export also helper files (input, output etc...), 

this adds to dev the ability to use methods separated, on a functional way.

packages/providers/src/transaction-summary/assemble-transaction-summary.ts
```diff
@@ -0,0 +1,133 @@
...
+        isStatusFailure = true;
+        break;
+
+      case 'SubmittedStatus':
+        time = gqlTransactionStatus.time;
+        isStatusPending = true;
+        break;
+      default:
+    }
+  }
```
LuizAsFight, line 111: we can have a status.ts file to handle all status logic

packages/providers/src/transaction-summary/assemble-transaction-summary.ts
```diff
@@ -0,0 +1,133 @@
...
+      return SimplifiedTransactionStatusNameEnum.success;
+    case 'SubmittedStatus':
+      return SimplifiedTransactionStatusNameEnum.submitted;
+    case 'SqueezedOutStatus':
+      return SimplifiedTransactionStatusNameEnum.squeezedout;
+    default:
+      throw new Error('Unknown transaction status');
+  }
+};
+
```
LuizAsFight, line 38: we can have a status.ts file to handle all status logic

LuizAsFight (changes requested):

packages/providers/src/transaction-summary/get-transaction-summary.ts
Torres-ssf: Makes sense @LuizAsFight, nice catch :+1:

Torres-ssf: @LuizAsFight Actually this seems to be working as expected already. 

What is the `graphql` answer when fetching a transaction that was not processed yet? 

If the transaction was not processed yet, is the `transaction` property inside of the `graphql` response going to be `null` / `undefined`? Or its status property will be of the type `SubmittedStatus`?

1. `null` or `undefined`: Since there is no data to process we can't do much else here besides throwing an error. We could brainstorm a more appropriate error message than `transaction not found`.
2. `SubmittedStatus`: The helper `assembleTransactionSummary` will process the transaction with the data that has in hand. The `fee` and `gasUsed` are going to default to `bn(0)`, since our helpers related to this calculation assign this value when there are no `Result` receipts to calculate the `gasUsed`, and of course, the `fee`.

Please let me know if I am missing something.

LuizAsFight: @Torres-ssf it will be `SubmittedStatus`

I'm guessing it will work too, but would be nice to check this, maybe running a node that doesn't instantly process blocks

Torres-ssf: @LuizAsFight I've just made some tests on a chain that does not instantly process blocks and could validate that this is working as expected

arboleya (dismissed):
Great stuff! 🤘

Dhaiwat10 (dismissed):
Awesome PR!

packages/providers/src/transaction-response/utils.ts
```diff
@@ -0,0 +1,36 @@
+import { arrayify } from '@ethersproject/bytes';
```
LuizAsFight, line 1: should this file be removed now that `processGqlReceipt` is inside of `transaction-summary?`

packages/fuel-gauge/src/transaction-summary.test.ts
```diff
@@ -0,0 +1,151 @@
...
+describe('TransactionSummary', () => {
+  let provider: Provider;
+  let wallet: WalletUnlocked;
+
+  beforeEach(async () => {
+    provider = new Provider(FUEL_NETWORK_URL);
+    wallet = await generateTestWallet(provider, [[1_000, BaseAssetId]]);
+  });
+
+  it('should ensure getTransactionSummary executes just fine', async () => {
```
LuizAsFight, line 25: can you differentiate the 3 equal titles for tests in this file?

LuizAsFight (changes requested):

packages/fuel-gauge/src/transaction-summary.test.ts
```diff
@@ -0,0 +1,151 @@
...
+    expect(transactionSummary.time).toBeUndefined();
+    expect(transactionSummary.status).toBeUndefined();
+    expect(transactionSummary.receipts).toStrictEqual(expect.any(Array<TransactionResultReceipt>));
+    expect(transactionSummary.isTypeMint).toBe(false);
+    expect(transactionSummary.isTypeCreate).toBe(false);
+    expect(transactionSummary.isTypeScript).toBe(true);
+    expect(transactionSummary.isStatusFailure).toBe(false);
+    expect(transactionSummary.isStatusSuccess).toBe(false);
+    expect(transactionSummary.isStatusPending).toBe(false);
+    expect(transactionSummary.transaction).toStrictEqual(transactionRequest.toTransaction());
```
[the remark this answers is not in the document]
Torres-ssf: @camsjams, my apologies for the confusion. I had previously marked this change request as resolved without making the necessary change. This was an oversight on my part. 

The requested changes have now been made at: https://github.com/FuelLabs/fuels-ts/pull/1108/commits/0d4ef1425c22dd4fbce316210272f7060cfd7e03.

arboleya (approved):

LuizAsFight (approved):

Issue #1067, opened 2023-06-16
Update the `Receipt` with new variant `Mint` and `Burn`


xgreenx:
This issue is a reminder to update the serialization and deserialization rules for `Receipt` after finishing of https://github.com/FuelLabs/fuel-vm/issues/496.

It may also requires update of some other places related to it.

Issue #956, opened 2023-04-28
Investigate support for missing contract IDs in `estimateTxDependencies`


Dhaiwat10:
The `estimateTxDependencies` helper on the `Provider` class (formerly named `addMissingVariables) has some code in place that tries to add any missing contract IDs that it finds to the TX, but this functionality doesn't seem to be working as expected.

More context here: https://github.com/FuelLabs/fuels-ts/pull/924#discussion_r1180374928

PR #1112, opened 2023-07-20
chore: upgrade `fuel-core` to `0.20.3`

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/1067
- Close: https://github.com/FuelLabs/fuels-ts/issues/956

The main changes include:

- Two new receipt types, `Mint` and `Burn`, have been added.
- The `rawPayload` property has been removed from the `Receipt` struct. From now on, when the SDK fetches the receipts (such as through a `dryRun` call), it should create a receipt object by filtering them based on their type.
- Fixed a critical bug that incorrectly calculated gas charging per unit, which resulted in transactions costing significantly more.
- Undo the temporary fix made on #1090 since the bug was already fixed on https://github.com/FuelLabs/fuel-vm/pull/505


packages/providers/src/utils/receipts.ts
```diff
@@ -35,3 +59,212 @@ export const getReceiptsWithMissingData = (receipts: Array<TransactionResultRece
...
+
+    case GqlReceiptType.Panic: {
+      const panicReceipt: ReceiptPanic = {
+        type: ReceiptType.Panic,
+        id: hexOrZero(receipt.contract?.id),
+        reason: bn(receipt.reason),
+        pc: bn(receipt.pc),
+        is: bn(receipt.is),
+        contractId: hexOrZero(receipt.contractId),
+      };
```
Torres-ssf, line 118: The issue https://github.com/FuelLabs/fuels-ts/issues/956 was happening because of the absence of `contractId` in the Panic Receipt `rawpayload` property, which was causing `estimateTxDependencies` failing to identify the `contractId` and add it to the transaction request.

However, with the removal of the `rawpayload` property in `fuel-core 0.20`, the way we assemble receipts has changed. We no longer decode them using the `rawpayload`, but instead assemble them using the Receipt's existing properties from the GraphQL response. 

As a result of the `contractId` property being present on the GraphQL response, `estimateTxDependencies` now successfully identifies and adds missing contract ids to the transaction request.

This change resolves the issue, so I am closing https://github.com/FuelLabs/fuels-ts/issues/956 with this PR.

Dhaiwat10 (approved):
Extremely well-done and meticulous!

Dhaiwat10 (approved):

Torres-ssf:
I believe @luizstacio's review here is crucial before merging this one

luizstacio (approved):

PR #1113, opened 2023-07-21
fix: cross-fetch version

LuizAsFight:
cross-fetch 3.1.8 is breaking when using chrome extension executing requests inside service worker

overriding it to 4.0.0 fixed the problem

ref: [cross-fetch](https://github.com/lquixada/cross-fetch/issues/78)


arboleya:
CI was not passing; hitting "update branch" appears to have fixed it.

arboleya (commented):
Would it be possible to move the override to some `packages/*/package.json` instead of having it globally set on the root `package.json`?

I wonder what package uses such an outdated `cross-fetch` version from before Oct 5, 2020.

LuizAsFight:
> Would it be possible to move the override to some `packages/*/package.json` instead of having it globally set on the root `package.json`?
> 
> I wonder what package uses such an outdated `cross-fetch` version from before Oct 5, 2020.

[graphql-request](https://github.com/jasonkuhrt/graphql-request/blob/main/package.json) latest uses it.

arboleya:
Good find! Indeed, they're using this [outdated version](https://github.com/jasonkuhrt/graphql-request/blob/e5c8e7f43066f238f96016f8c3b5c6c3cd0141ce/package.json#L66), there's this [issue](https://github.com/jasonkuhrt/graphql-request/issues/326) about it.

For now, we could still move the overriding inside the `provider` package.

Congrats on this hotfix, @LuizAsFight!

LuizAsFight:
@arboleya 

https://pnpm.io/package_json#pnpmoverrides
> Note that the overrides field can only be set at the root of the project.

arboleya (approved):

danielbate (approved):

PR #1114, opened 2023-07-21
feat: ABI type inference - base type inference logic and inferred function inputs

### LINKED ISSUE #1021: Infer types based off of the abi.json directly
## Motivation

The current `abi-typegen` implementation does per-abi on-disk file generation. It generates a factory and a declarations file per abi, as well as a shared common.d.ts file. 

The proposal is to infer types based off of the abi files directly. Given that the abi files contain everything necessary to define a sway program and they are `.json` files, we can, in the ideal case, import them directly into our code as `.json` files and not have the need for `abi-typegen` altogether (pending the implementation of [this typescript feature request](https://github.com/microsoft/TypeScript/issues/32063)), or we can reduce `abi-typegen`'s functionality to just copying over the abis, changing their extension to `.ts` and assigning them to an object `as const`, and then creating a `fuelFactory.ts` file where they all would be imported and passed to an instance of a generic `FuelFactory` class that would then be able to call all of the programs and would be fully typesafe based solely off of type inference.

### What is possible to implement now

Implementation details aside, the end result, after running `fuels typegen` could possibly look like this:

```typescript
import { counterContractAbi } from './abis/counterContractAbi'
import { FuelFactory } from 'fuels'

const factory = new FuelFactory([{programName: "counterContract", ...counterContractAbi} as const]);

// typesafe contract names
const counterContract = factory.programs("counterContract").connect(CONTRACT_ID, wallet);

// typesafe function names and whatever else you wish for
const count = counterContract.count().txParams({gasPrice: 1}).call();
```

The `const` assertions in the `FuelFactory` constructor are necessary for the narrowing of inference from i.e. `{name: string}` to `{name: "count"}` and for the whole thing to work.

### What would be possible if TS issue 32063 was implemented
The process would be so straightforward that there wouldn't be any need for `abi-typegen` at all.

```typescript
import counterContractAbi  from './abis/counterContractAbi.json' as const;

const factory = new FuelFactory([{programName: "counter", ...counterContractAbi} as const]);

// ... as in first example
```

The difference between the first and second example is that in the first example we'd be importing from a `counterContractAbi.ts` file which would look like this:

```typescript
export const counterContractAbi = {...abi-stuff...} as const;
```

Whereas the second example is importing the `counterContractAbi.json` file directly and has the `as const` assertion on the import.

Although the first example could also be done without `abi-typegen`, it seems to me that it might be burdensome for some developers to think about changing extensions, assigning the abis to variables and adding `as const` at the end of them - so much so that it justifies having a tool for it.

However, after the mentioned TypeScript feature gets implement, I think the process would be so straightforward and intuitive that a simple example in our documentation would be enough and the `abi-typegen` package wouldn't be necessary anymore.

## Implementation details

In essence, what would be done is something this:

```typescript
type NamedJsonFlatAbi = { programName: string } & JsonFlatAbi;
type TupleToUnion<T> = T extends Array<infer ITEMS> ? ITEMS : never;
type Filter<T, Condition> = T extends Condition ? T : never;

class FuelFactory<TAbisArray extends NamedJsonFlatAbi[] = NamedJsonFlatAbi[], TAbis = TupleToUnion<TAbisArray>> {

    #abis:TAbisArray;

    constructor(...abis: TAbisArray) {
        this.#abis = abis;
    }

    programs<ProgramName extends TAbis["programName"], 
    SelectedProgramAbi = Filter<TAbis, {programName: ProgramName}>
    >(programName: ProgramName, id, wallet): Contract<SelectedProgramAbi> {
        const abi = #abis.find(a => a.contractName === contractName);

        return new Contract<SelectedProgramAbi>(id, abi, wallet);
    }
}
```

Now, based off of the inferred `SelectedProgramAbi` type passed into `Contract<SelectedProgramAbi>`, all other properties of the contract (functions, multicall, etc.) can be inferred to the narrowest type possible.

Note: currently all three types of sway programs get typegen-ed into a `Contract`, and that is why the method `programs` is returning `Contract`.

So, from what it looks like, the only thing that would change on existing infrastructure are the types going in/being returned from our classes. The implementation would ideally stay the same in the first step, and the new top-level `FuelFactory` abstraction would later on enable better cohesion and coupling between code that is currently possibly hard to connect in an appropriate way.

## The future

Besides Awesome Fuel TypeSafety:tm:, `FuelFactory` would enable setting whatever logic we choose that should apply to all or a part of calls just by adding a second `options` parameter into the constructor. Some things that come to mind:

``` typescript
const factory = new FuelFactory([...abis], {
    defaultTxParameters: {
        gasPrice: 1 // no need to do the fn.txParams() part
    },

    shouldExecuteCall: async (params, ctx) => {
        const { minGasPrice } = await ctx.provider.getNodeInfo();
        return minGasPrice <= params.txParams.gasPrice;
    }, // We don't want to pay too much
    
    services: {
        emailService // let's add services and call them if the transaction was a success! See onAfterCallExecuted.
    },
       
    onAfterCallExecuted: async (response, callCtx, factoryCtx) => {
        if(callCtx.program !== "veryImportantContract") return;
        if(callCtx.program.function !== "veryImportantFunction" ) return;
        if(!response.success) return;

        await emailService.EmailCEO(`Job is done, you have earned: 
        ${response.tokenAmount} of ${response.tokenName}.`);
    }
});
```

nedsalk:
## NOTE: This PR is merging into branch `ns/feat/abi-type-inference` 

Due to the importance, size and impact of this PR  and the subsequent PRs on the developer experience of users, I've created a separate feature branch [feat/abi-type-inference](https://github.com/FuelLabs/fuels-ts/tree/feat/abi-type-inference) into which all PRs will be merged. When we're satisfied with the results, we'll merge that branch into master.
Issue #1021 will serve as a reference where all the tasks related to this whole feature will be listed.

This PR represents a big improvement to the developer experience of our users and is a stepping stone to trimming down and eventually deleting the `abi-typegen` package altogether.

The idea of this endeavor is to load the whole abi into the type system and parse it, which results in correct type safety of everything functions, their inputs and outputs, configurables, ...) on our `Interface` and program types: `Contract`, `Script`, `Predicate`.

This PR's scope is to make the functions and their inputs type safe on the `Interface` level. Subsequent PRs will deal with programs, configurables, and everything else relevant.

In order to ensure that the type inference logic covers all use cases, I've used the ABIs found in `abi-typegen` and also added some examples from the `exhaustiveExamples` sway project where it was necessary. For the time being, I've just copied these abis and referenced them in the tests. In the future this should be automated and based off of their sway programs directly so that the abis never get stale. The ABIs and their corresponding compilation tests can be found in the `abi-coder/test/type-inference-tests` folder. These tests get run during `pretest` and the relevant changes to get them running are found in `package.json` and `tsconfig.json`.

The type inference logic is located in the `type-inferrer` folder.


packages/abi-coder/src/abi-coder.ts
```diff
@@ -204,10 +114,10 @@ export abstract class AbiCoder {
       return new TupleCoder(coders);
     }
 
-    return logger.throwArgumentError('Coder not found', 'type', { abiType, abi });
+    return logger.throwArgumentError('Coder not found', 'abiType', { abiType, abi });
   }
 
-  private static getCoders(components: readonly JsonAbiArgument[], abi: JsonAbi) {
+  private static getCoders(abi: JsonAbi, components: readonly JsonAbiArgument[]) {
```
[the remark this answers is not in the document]
nedsalk: I played around with the signature of this method a lot because I was trying out different designs that affected it. In the end I went back to this one, but changed the order... I thought it was already like this, no specific intent. Anyways, it makes sense to have it like this as you can't have coders for components without the abi, thus making the abi the "parent", "more important".

packages/abi-coder/src/function-fragment.ts
```diff
@@ -18,27 +18,29 @@ import type {
...
+import { ResolvedAbiType } from './resolved-abi-type';
 import type { Uint8ArrayWithDynamicData } from './utilities';
 import { isPointerType, unpackDynamicData, findOrThrow } from './utilities';
 
 const logger = new Logger(versions.FUELS);
 
 export class FunctionFragment<
-  TAbi extends JsonAbi = JsonAbi,
-  FnName extends TAbi['functions'][number]['name'] = string
+  Input extends object | never = never,
```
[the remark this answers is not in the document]
nedsalk: Yes, now all inputs are objects by necessity, because TS doesn't allow giving generic names to parameterized tuples: `[a: number, b: string]` -> `<myGenericParam>:<myGenericParamType>[]` (pseudo code, not allowed). Due to this, I had to change the input into an object with keys that match parameter names:
```rs
// sway  
struct MyStruct {
prop1: bool;
prop2: b512;
}
fn my_fn(a: u32, b: b256, c: MyStruct);
```
```ts
// input for my_fn
type InferredInputForMyFn = {
 a: number;
 b: string;
 c: {
  prop1: boolean;
  prop2: string;
 }
}
```

packages/abi-coder/src/abi-coder.ts
```diff
@@ -158,7 +67,8 @@ export abstract class AbiCoder {
     }
 
     // ABI types underneath MUST have components by definition
-    const components = this.resolveGenericComponents(abi, argument);
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const components = abiType.components!;
```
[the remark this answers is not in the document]
nedsalk: Note: this discussion is for #1107. I've explained above in the description of the PR why I've got that PR in here as well. However, now I see that it's rather cumbersome to discern between them so I'll try updating this PR to remove the other PR's work.

As for the question:
The components get resolved inside of `ResolvedAbiType`, if that's what you're asking.
`ResolvedAbiType` is given an abi and an argument of type `JsonAbiArgument` into its constructor, and it resolves all the possible generics in the underlying argument's type.

package.json
```diff
@@ -42,51 +42,51 @@
...
-    "textlint": "^13.3.2",
+    "textlint": "^13.3.3",
     "textlint-rule-no-dead-link": "^5.1.2",
     "ts-generator": "^0.1.1",
-    "ts-jest": "^29.0.3",
+    "ts-jest": "^29.1.1",
     "tsup": "^6.7.0",
     "tsx": "^3.12.7",
-    "turbo": "^1.8.8",
+    "turbo": "^1.10.9",
```
arboleya, line 89: Can we do library upgrades in a separate PR against `master`?

packages/abi-coder/package.json
```diff
@@ -17,6 +17,7 @@
     "dist"
   ],
   "scripts": {
+    "pretest": "tsc --noEmit",
```
arboleya, line 20: Why is this necessary?

.changeset/tasty-taxis-sneeze.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/abi-coder": patch
+---
+
+encoding bugfix
```
arboleya, line 5: This PR must exclude work from other PRs.

If you need things from https://github.com/FuelLabs/fuels-ts/pull/1107, you must make `1106-bug-nested-implicit-generics-fail-encoding` the base branch for https://github.com/FuelLabs/fuels-ts/pull/1116 and update this PR accordingly to dedupe things.

packages/abi-coder/test/type-inference-tests/array-of-enums-abi.ts
```diff
@@ -0,0 +1,79 @@
+export const arrayOfEnumsAbi = {
```
arboleya, line 1: Instead of manually creating and versioning a Typescript version of all and each `*-abi.json`, we should have a script that does this on the fly on a `pretest` step or something similar. This will reduce the noise and also ensure that nothing is outdated concerning each respective forc project.

arboleya (changes requested):

.changeset/tasty-taxis-sneeze.md
nedsalk: This is a leftover from that PR that I didn't remove.

packages/abi-coder/test/type-inference-tests/array-of-enums-abi.ts
nedsalk: I agree, I was planning on adding that and removing these files when I implement the json->ts typegen aspect of the whole feature.

packages/abi-coder/package.json
nedsalk: Because this is how the testing of `*.compilation-test.ts` files is happening. If those files compile with tsc, the type inference works as expected.

packages/abi-coder/src/type-inferrer/abi-type-inferrer.ts
digorithm, line 1: Coming from a lack of deep knowledge in these darker corners of Typescript, this type-system-abuse _feels_ obscene to me, haha. These are my main concerns from a readability and maintainability point of view.

With that said, the (true) endgame, eliminating the need to invoke a CLI command to generate the TS types and having an experience similar to the Rust SDK, might 100% justify incurring this complexity, as the DevEx will be better. 

But even at this point (i.e., before the endgame), the developer experience seems good, or at the very least, unchanged from what the current state of things is, which is good; making these big internal changes without affecting the public APIs that much is a goal we should keep in mind.

If the TS devs on the team (e.g., @FuelLabs/sdk-ts) are okay with this type-inference as-is, and they can read and maintain this afterward, then awesome. Their opinion and judgment should be heavier than mine in this case.

danielbate: Myself, @arboleya and @Torres-ssf had a call with @nedsalk yesterday and did express some concerns around the maintainability. But are of course all in agreement that the endgame utility would be great for the SDK. However a concern with that is we are still some way from that, further to this PR, as detailed by the tasks #1021. Is it worth adding some detail to those tasks to get a better understanding of the remaining work required to get to that point?

I think this PR acts as a great proof of concept for what is feasibly possible with type inference and I'm really excited to see it's implementation in the SDK. However I would like to see some more modularity and documentation around the custom types to aid with readability and maintainability. @arboleya shared a really good example of a package utilising a very modular type inference system with great use of doc blocks - [https://github.com/wagmi-dev/abitype](https://github.com/wagmi-dev/abitype). As further tasks are completed and inference is added for scripts, predicates and composables, this will become quite a heavy file. Once we bring typedoc as well it'd add a lot of noise to the file. IMO a good start would be to split the inferrer and map types up, and the type maps from each other, each type with it's own doc block.

Props to the test implementations though, great depth and range of scenarios 👍🏻

arboleya: Advancements in this PR have been tremendous! It dramatically illustrates the possibilities and opens the doors for highly advanced type-inference black magic. That's for sure.

However, despite fully appreciating all the hard work on this front, I don't think it's mature enough to go through and much work would still be needed to get us to replace Typegen fully. I think we should put it on pause as a **_`successful`_** experiment for now that we can refer back to.

Overall, the achievements here already represent a big win, IMO.

nedsalk: @digorithm Hehe, it might seem obscure, but it's just regular old (...unpleasant perhaps...) type-level programming in Typescript. But I believe that after I implement @danielbate's suggestions that it should be a lot clearer. Also, 15 minutes of [type challenges](https://github.com/type-challenges/type-challenges) a day, and after a month (or even less) everyone can read it.

@danielbate Thanks for the ideas! I can of course split the file into sub-files and document them in more depth. The idea of this was to not export any mappers as they are implementation details, but I see the value in readability that it would bring, especially if each one was accompanied by a micro-example of sorts.

@arboleya Thanks! What are the aspects of this that are not mature? Also, as far as additional work goes, relative to this there is not much additional complex work to be done. The subsequent PRs would be reaping the benefits of this one, they would be easy to review as this is the big one, and we'd be able to infer contracts, scripts, predicates, configurables, logged types, multicalls and anything else that works with the abi. Of course, as already mentioned in the meetings, a typegen cli tool would still be a thing _(for the time being)_, but it would be simplified to "copy-paste the json file, change to `.ts` extension and hexlify binary".

digorithm: > However I would like to see some more modularity and documentation around the custom types to aid with readability and maintainability.

Yup. What @danielbate said here. Readability and maintainability take precedence. 

> Also, 15 minutes of [type challenges](https://github.com/type-challenges/type-challenges) a day, and after a month (or even less) everyone can read it.

This is a nice and fun thing, my computer scientist side gets all tingly when I see problems being solved purely at a type level, but we shouldn't expect that from every developer being onboarded on the project. Ramp-up/onboarding speed is _extremely_ important for a project of this scale; I've seen projects die _because_ of it. That is why I say readability and maintainability take precedence over using cool obscure CS tools and techniques. Make it readable and maintainable (either through modularization or documentation, ideally both) and this will be amazing.

nedsalk: Also, no matter how this PR goes - be it merged or paused - I think it's wise to complete it based off of your suggestions, as whoever continues doing work on it will be grateful that I did it.

nedsalk: @digorithm 
>Make it readable and maintainable (either through modularization or documentation, ideally both) and this will be amazing.

Okay, so I'll document the heck of this as best as I can and then I'll ask for the review of the team again.

packages/abi-coder/src/function-fragment.ts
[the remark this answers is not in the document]
nedsalk: I can change the type inference to be arrays as it was previously, but we would lose info on names of sway function arguments, because TS can't generate tuples with named elements (e.g. `[arg1: number, arg2: string]` - this is valid TS syntax, but it can't be generated dynamically). 

However, your idea on allowing both approaches is great! We can do a hybrid approach with [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads). This wouldn't break **any** existing application, but we can make it so that the first approach suggested by the IDE is now the "object for all inputs" instead of "spread array inputs".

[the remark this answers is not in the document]
nedsalk: > However, your idea on allowing both approaches is great! We can do a hybrid approach with [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads). This wouldn't break **any** existing application, but we can make it so that the first approach suggested by the IDE is now the "object for all inputs" instead of "spread array inputs".

I've updated the code to now have the overload and it works! The `Input` type that @camsjams  commented on is still an object (or, more specifically `Record<string, unknown>`), but the actual `encodeArguments` method can now accept both a tuple as it was accepting pre-type-inference, as well as a typesafe object. The typesafe tuple doesn't have named elements, but that's okay because I've ordered the overloads in a way that the IDE first suggests the object as an input, where the parameter names are keys of the object. The overload is specifically there to not introduce breaking changes - and it does its job as far as that is concerned.

nedsalk:
Closed until there is completely no need for a CLI tool (after implementation of [TS issue 32063](https://github.com/microsoft/TypeScript/issues/32063)). Will be revisited then.

Issue #835, opened 2023-03-17
Re-consider having API docs for certain packages


arboleya:
At first, we spoke about eliminating Typedocs completely because the majority of files are _not_ documented, and displaying purely mechanical and hollow documentation only adds noise, burying the rest underneath it.

However, a few places might have relevant docs to show.

These docs could be:
 1. Incorporated in the Guide in `.md` format _(typedocs stay out)_
 1. Re-enabled and put in a different menu section _(typedocs returns)_

Either way, we must first understand what packages fall in this category.

Then we can see what to do.

arboleya:
@luizstacio Want to chime in and list the packages you had in mind?

> cc @Torres-ssf @FuelLabs/sdk-ts

danielbate:
I would like to see API documentation for certain packages within the TS SDK. Contextualised documentation and cookbooks (as we have currently) goes hand in hand with API documentation, and something I usually expect some level of with complex libraries. Done properly as well this is something that can just work alongside us as well rather than adding unnecessary overhead.

At a high level, I would like to see the entry point classes and methods documented for our core packages, as well as the parameters and config that they require, this could include:
- Contract interaction, contract factory, methods to call contracts and the configuration / parameters required
- Predicate instantiation within the SDK, transactional methods with predicates and setting data / configurables
- Providers and it's methods, then coins, messages and resources (do we want to expose the GQL schema and autogenerated types here?)
- Transactions including `ScriptTransactionRequest` and building out custom transactions (i.e. the methods available to the transaction request classes and params they expect, and any potential config)
- Wallets, different types of wallet and the functions available to those types
- Address, different types of addresses and the helpers to convert between them

I agree with the above points that it should be under a separate menu item to our current documentation menu items. I also think it is easier to maintain as it is directly to our business logic when using something like typedoc. And then easily incorporated to vitepress by outputting markdown.

I also agree with trying to reduce noise in code. We should do this in iterations, at first I would like to see the above just referenced and documented, potentially without any additional explanation to what is already available to us when highlighting functions in IDEs. Any additional noise could be added using type doc references?

arboleya:
@danielbate Thanks. Each of these bullet points can become an individual task. The first task should also add back `typedocs` and integrate it with `vitepress` in the _current_ build routine.

@Torres-ssf I can't remember — did we make any progress experimenting with **_`typedocs` x `vitepress`_** in the past?

danielbate:
[...]

PR #1115, opened 2023-07-21
feat: implement typedoc in vitepress

danielbate:
Following discussions in #835, this reimplements typedoc to provide API documentation to be housed by our current vitepress implementation by:
- Installing typedoc along with the modular and markdown plugins and adding to vitepress config
- Implementing typedoc config in the `Address` and `Interface` packages
- Add typedoc syntax to classes we want to hide/show
- Improve docs on classes/methods
- Generate API docs

Screenshot 2023-07-21 at 14 04 07

> This is a proof of concept hence having only implemented it in 2 packages. Should we be happy with the implementation, we can extend this by autogenerating the vitepress links (the plugin that does is out of date with the latest version of typedoc but is not difficult to implement). There is also room for improvement of our usage of typedoc syntax inside these classes.


arboleya (commented):
Looking beautiful.

I didn't review yet, just quickly glanced over it, here are some preliminary comments:
 1. Should we commit the generated MD files inside `src/api/classes`? Perhaps these could be generated only in CI before publishing. What do you think?
 2. There seem to be a couple of annotations like [this](https://github.com/FuelLabs/fuels-ts/pull/1115/files#diff-792e21671d9c81763592694d7fa6cd14a5ab30941084746ab340aebe8f8052ee):

annotation

arboleya (commented):
> 1. Should we commit the generated MD files inside `src/api/classes`?

Another indication this may be optimal is that these files contain dynamic links such as:

 - https://github.com/FuelLabs/fuels-ts/blob/bef034a9/packages/address/src/address.ts#L27

Due to the blog part (`bef034a9`), files will appear modified even when unchanged.

Other than that, everything is looking good! Very clean. 👌

arboleya:
@danielbate Because you're already integrated `typedoc` in the existent docs build routine:

https://github.com/FuelLabs/fuels-ts/blob/bef034a9204170032043ddf375f8ec326a53152b/apps/docs/package.json#L8

You just need to ignore these two dirs and voilà:
 - src/api/classes
 - src/api/modules

Everything else is already taken care of.

> cc @Torres-ssf

danielbate:
@camsjams @arboleya will get the built docs files removed. Only risk is including classes/methods in the docs that were intended to be hidden/internal only, as they won't be included in the review process. But on balance seems worth it to keep the source clean.

Torres-ssf:
> @danielbate Because you're already integrated `typedoc` in the existent docs build routine:
> 
> https://github.com/FuelLabs/fuels-ts/blob/bef034a9204170032043ddf375f8ec326a53152b/apps/docs/package.json#L8
> 
> You just need to ignore these two dirs and voilà:
> 
> * src/api/classes
> * src/api/modules
> 
> Everything else is already taken care of.
> 
> > @Torres-ssf can correct me if I'm missing something.

@arboleya Yes if we are about to not commit the generated API files, we need to also make Typedoc to generate them on the `dev` script.

arboleya:
@Torres-ssf Good point!

@danielbate Maybe something like this would do the trick:

```json
  "dev": "nodemon --config nodemon.config.json -x 'typedoc && vitepress dev'",
```

danielbate:
@Torres-ssf @arboleya Any reason they were included prior, as I can see they were removed when we moved to vitepress in #811?

arboleya:
IIRC, @Torres-ssf changed the way we publish to GH pages, which made it possible.

danielbate:
I'll pick up the new eslint warnings generated by the implementation of tsdoc in #1125

arboleya (approved):

Torres-ssf (approved):

PR #1123, opened 2023-07-25
feat: add helper method to Abi Interface

LuizAsFight:
### Problem:
As now the [old ABI format is removed](https://github.com/FuelLabs/fuels-ts/pull/1094/files), we don't have easy access to type of ABI fragments anymore

### Solution:
implements a new helper method `getTypeById`, the objective is to provide an easy way to get abi types: 

### Usage:
```
const abiInterface = new Interface(abi);

const abiType = abiInterface.getTypeById(10);
```


arboleya:
@nedsalk More breaking changes?

arboleya (approved):

PR #1126, opened 2023-07-25
docs: add docs around custom storage slots

### LINKED ISSUE #1118: Add documentation around storage slots
Follow-up from #1093:

Currently, there is no mention of `DeployContractOptions` in the docs or how to use Storage Slots. This is an important thing that we are missing.

Dhaiwat10:
Closes #1118


packages/fuel-gauge/src/storage-test-contract.test.ts
```diff
@@ -36,4 +40,46 @@ describe('StorageTestContract', () => {
...
+  it('can increment counter - using custom inline storage slots', async () => {
+    const provider = new Provider('http://127.0.0.1:4000/graphql');
+    const wallet = await generateTestWallet(provider, [[1_000, NativeAssetId]]);
+    const bytecode = readFileSync(
+      join(__dirname, '../test-projects/storage-test-contract/out/debug/storage-test.bin')
+    );
+    const factory = new ContractFactory(bytecode, abi, wallet);
+    // #region contract-deployment-storage-slots-inline
+    const contract = await factory.deployContract({
+      storageSlots: [
```
Dhaiwat10, line 53: I had to opt against avoiding duplication of code here for the setup because if I did that the snippets would not be as clear.

danielbate (approved):

arboleya (approved):

PR #1136, opened 2023-07-26
feat: deprecate tx funding call option

### LINKED ISSUE #1135: Deprecate tx funding call option
In `fuel-core` 0.19.0, txs [now require at least one spendable input to be considered valid](url).

We addressed this in #1085 by not switching the `fundTransaction` config flag to `true`, so that a spendable input is added. 

We should go a step further and remove this configurable option, as all txs now require a spendable input. We should also alter references to funding txs in our documentation to reflect this.

danielbate:
Closes #1135

### DIFF
--- a/apps/docs-snippets/src/guide/contracts/read-only-calls.test.ts
+++ b/apps/docs-snippets/src/guide/contracts/read-only-calls.test.ts
@@ -13,8 +13,7 @@ describe(__filename, () => {
   it('should successfully execute a read only call', async () => {
     // #region read-only-call-1
     const { value } = await contract.functions.echo_u8(15).get();
-
-    expect(value).toEqual(15);
     // #endregion read-only-call-1
+    expect(value).toEqual(15);
   });
 });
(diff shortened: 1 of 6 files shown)


apps/docs-snippets/src/guide/contracts/read-only-calls.test.ts
```diff
@@ -13,8 +13,7 @@ describe(__filename, () => {
   it('should successfully execute a read only call', async () => {
     // #region read-only-call-1
     const { value } = await contract.functions.echo_u8(15).get();
-
-    expect(value).toEqual(15);
     // #endregion read-only-call-1
+    expect(value).toEqual(15);
```
arboleya, line 17: Was this on purpose?

Why should it be removed from the docs code snippet?

arboleya (approved):

apps/docs-snippets/src/guide/contracts/read-only-calls.test.ts
danielbate: _Personally_ I like to keep assertions out of the snippet to focus on the functionality that the docs are referencing rather than an assertion. Potentially a weak arugment, but if someone is not used to jest syntax then this additional line is confusing.

nedsalk (approved):

PR #1137, opened 2023-07-26
feat: refactor predicate and contract root calculations

danielbate:
- Moves the shared root calculation logic for predicates and contracts to the `utils` package
- Renames `getContractRoot` to `getPredicateRoot` for predicates


packages/predicate/src/utils/getPredicateRoot.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const getPredicateRoot = (bytecode: BytesLike, chainId: number): string => {
+  const chunkSize = 16 * 1024;
+  const bytes = arrayify(bytecode);
+  const chunks = chunkAndPadBytes(bytes, chunkSize);
+  const chainIdBytes = uint64ToBytesBE(chainId);
+  const codeRoot = calcRoot(chunks.map((c) => hexlify(c)));
+
+  const contractRoot = hash(concat(['0x4655454C', chainIdBytes, codeRoot]));
+  return contractRoot;
```
Dhaiwat10, line 15: Let's also rename this variable just like the function:

```suggestion
  const predicateRoot = hash(concat(['0x4655454C', chainIdBytes, codeRoot]));
  return predicateRoot;
```

Dhaiwat10 (changes requested):

packages/predicate/src/utils/getPredicateRoot.ts
danielbate: Good spot

Dhaiwat10 (approved):

nedsalk (approved):

PR #1139, opened 2023-07-26
feat: add docs script to remove unwanted files

danielbate:
On the whole typedoc only generates files for classes/modules/types within the SDK. However it is quite opinionated on the file structure and therefore also generates index/module files per package. There are unnecessary for us as we define our own structure as we house typedoc in vitepress and also those consuming our files like the docs hub project (cc @sarahschwartz).

We have a couple options:
- Write a vitepress or typedoc plugin to manipulate the outputted structure
- Write a script that purges the files after build

I've opted for the latter for simplicity and future compatability.


Torres-ssf (approved):

apps/docs/package.json
```diff
@@ -5,7 +5,7 @@
   "description": "",
   "scripts": {
     "dev": "nodemon --config nodemon.config.json -x 'typedoc && vitepress dev'",
-    "build": "typedoc && vitepress build",
+    "build": "typedoc && vitepress build && tsx ./scripts/purge-docs.ts",
```
arboleya, line 8: The purge needs to go before the vite build to avoid creating HTML files for deleted items.
```suggestion
    "build": "typedoc && tsx ./scripts/typedoc-postbuild.ts && vitepress build",
```

apps/docs/scripts/purge-docs.ts
arboleya, line 1: Do you think we could rename `purge-docs` to something more specific?

 - `typedoc-cleanup.ts`
 - `typedoc-postbuild.ts`

apps/docs/scripts/purge-docs.ts
```diff
@@ -0,0 +1,20 @@
+import { unlink } from 'fs';
+import { join } from 'path';
+
+const docsDir = join(__dirname, '../src/');
+
+/**
+ * Array of files to remove from docs. This could be files that would be do not want to be exposed in the docs or indexed,
+ * or files that are generated by other scripts/packages.
+ */
```
arboleya, line 9:
```suggestion
/**
 * Post build script to trim off undesired leftovers from Typedoc.
 */
```

arboleya (changes requested):
Tidy. Left some minor suggestions, sorry :)

apps/docs/scripts/purge-docs.ts
danielbate: On reflection purge-docs is a little overzealous

arboleya (approved):
:shipit:

Torres-ssf (approved):

PR #1140, opened 2023-07-26
refactor: rename `keystore` package to `crypto`

Torres-ssf:
This PR is responsible for renaming the `keystore` package to `crypto`. 

This change is being made to better reflect the broader scope of the package since the keystore name can be considered more specific, it implies a focus on key management and storage.

However, the package incorporates broader cryptographic operations.


arboleya (approved):
Makes sense.

It is also easier to combine all browser-specific crypto stuff in a single place. 👍

PR #1144, opened 2023-07-27
refactor: type check crypto functions

Torres-ssf:
This PR aims to improve the consistency and reliability of our crypto functions. 

Our `crypto` (former `keystore`) package exports multiple versions of each function for different environments (CMJ/ESM, browser). This can lead to an increased risk of inconsistencies across versions.

To address this, I have made the following changes:

- Added the `CryptoApi` interface, standardizing the expected signatures of our crypto functions.
- Created an object, `api`, and define its type with the `CryptoApi` interface. This object ensures that all functions exported align with the correct interface.
- Continued to allow direct function exports from `api`. This ensures crypto functions can be consumed the same way they are now.


packages/crypto/src/browser/cryptoApiWrapper.ts
```diff
@@ -0,0 +1,15 @@
...
+const cryptoApiWrapper: CryptoApi = {
+  bufferFromString,
+  stringFromBuffer,
+  decrypt,
+  encrypt,
+  keyFromPassword,
+  randomBytes,
+};
+
+export default cryptoApiWrapper;
```
arboleya, line 15: Would this be valid? If yes, we could rename this file to `index.ts`.

```suggestion
export default { ...cryptoApiWrapper };
```

arboleya (approved):

Dhaiwat10 (approved):

PR #1151, opened 2023-07-27
docs: add docs around more complex args for predicates

### LINKED ISSUE #1147: Add docs around passing more complex & multiple args to a predicate
Right now our docs only cover passing a single & really simple arg to a predicate: https://fuellabs.github.io/fuels-ts/guide/predicates/send-and-spend-funds-from-predicates.html 

We should add some docs around passing in more complex arguments like a vector or a struct, and also passing in multiple args.

Dhaiwat10:
Closes #1147 

PS: I haven't added a section for Vectors since I found that [this test](https://github.com/FuelLabs/fuels-ts/blob/2fb4536c2252670d9aa6027e2da57d01eb7eb14b/packages/fuel-gauge/src/predicate/predicate-arguments.test.ts#L330) covering us passing a vector as an argument to a predicate is currently skipped. Does anyone know why?


arboleya (approved):
It looks like the test you mentioned was [never run](https://github.com/FuelLabs/fuels-ts/pull/1046/files#diff-f3faefcb1340d572b80a9b1634b0a2fac41d8eef915ab09d6a6354b474fba323).

Perhaps @camsjams is still working on it.

nedsalk (approved):

PR #1152, opened 2023-07-28
fix: standalone installation of `forc` and `fuel-core` bin wrappers

arboleya:
This PR converts `forc` and `fuel-core` bin wrappers to plain Javascript packages.

These two ~have~-_**had**_ custom `pnpm install` routines like this below, but having to _**`build` before `install`ing**_ causes a whole set of undesired side-effects. Using Javascript simplifies this chicken-egg compilation loop-hole.

```json
{
  "scripts": {
    "preinstall": "tsup",
    "install": "dist/install.js"
  }
}
```


nedsalk (dismissed):
I was experimenting with [bunjs](https://bun.sh/) on our repository just yesterday and this is where it failed when running `bun install`. Nice.

nedsalk (approved):

PR #1153, opened 2023-07-29
feat: add `FuelError` package and related utilities

### LINKED ISSUE #1110: Implement new Error handling package
Currently there is no standardized way to throw an error in the SDK. Rather, wherever an error has to be thrown, it gets thrown with `throw new Error(message)`, e.g.:
```ts
if (!gqlTransaction) {
  throw new Error('transaction not found');
}
```
This issue serves as a place to specify and implement _the_ way to throw exceptions.
My initial idea was to create custom errors like this: `new TransactionNotFoundError()`, `new EncodingFailedError()`, etc. However, after discussing privately with @arboleya and @LuizAsFight, they proposed a better way, which is to use one error class that has a `code` field that describes which error code it is:
```ts
  throw new SdkError("TransactionNotFound"); // typesafe string literals
```
or
```ts
throw new SdkError(SDK_ERRORS.TransactionNotFound); // enum of error codes
```

where `SdkError` is:
```ts
export class SdkError extends Error {
code: ERROR_CODES;
 constructor(code: ERROR_CODES, message?: string) {
  super(message);
  this.code = code;
  this.name = "SdkError";
 }
}  

type ERROR_CODES = 
| "TransactionNotFound" 
| "EncodingFailed"
| "NodeUnreachable"   
```

I'm a fan of the first approach as the user just has to import the `SdkError` and write the error code string, whereas in the second approach one has to import the `SDK_ERRORS` enum alongside `SdkError` every time an error has to be thrown. We can achieve the same level of type safety by resorting to string literals instead of enum values.

One concern that I have is error code ownership. Ideally each package should itself define what its failure modes are (e.g. `EncodingFailed` is part of the `abi-coder` package and it should be defined there, not somewhere else). If we do it like in the examples above, then we'll have to define the error codes in the same package where the `SdkError` class is defined, which is less than ideal. Maybe making it generic somehow (`class SdkError<T extends TemplateErrorMessage>`) would solve this issue, but I can't come up with an idea on how to properly do it without resorting to a lot of boilerplate.
One way to do it that may be clean enough (although it breaks error code ownership) would be to define error codes for each package separately and unite them in the `ERROR_CODES` type:
```ts
type AbiCoder_Error_Code = "EncodingFailed";
type Provider_Error_Code = "NodeUnreachable";

type ERROR_CODES = AbiCoder_Error_Code | Provider_Error_Code;
```
This looks simple and good enough.

Another **important** thing to keep in mind is that we should strive towards synchronizing the error naming with the rust sdk.

After we define this, we'll have to update our documentation to contain all the errors and their explanations. However, as @danielbate pointed out, we should save the documentation updating part for when typedoc is ready.

Please add your thoughts.

nedsalk:
This PR adds a new `errors` package based off of #1110.

Thanks @arboleya for your contributions!

An additional question remains: what should we do with the usage of `@ethersproject/logger` for throwing errors? Currently, throwing errors is done in two ways inside of the SDK, either via `throw new Error...` or `logger.throw...`. I've tried writing examples in the `abi-coder` package, but this ambiguity stopped me. @arboleya added examples in `address` package where he substituted `logger` with `FuelError` as a showcase, but we should still have a discussion because we're using the logger alongside our `versions` and currently `FuelError` doesn't have any version info in it.

This PR is mergable as-is, regardless of how we deal with the logger issue, as anything logger-related can be built on top of it. After we resolve the logger issue, errors throughout the SDK can be standardized to use only one approach. We can still reference this package in any subsequent PRs whenever we see someone of us throwing a new error directly and just say "Hey, can you think of a properly-named error code and use `FuelError` instead of `Error`?"

Note that updating the errors throughout the `fuels-ts` packages is outside of the scope of this PR, as that isn't a small endeavor because careful thought needs to be put into error code naming. Also, sometimes the actual code implementation where an error is thrown might not 100% reflect the _error code_ being used, which might necessitate some refactoring to have it make more sense.

Closes #1110.


jest.config.ts
```diff
@@ -16,6 +16,7 @@ const config: Config.InitialOptions = {
   transform: {
     '\\.hbs$': 'jest-text-transformer',
   },
+  setupFilesAfterEnv: [],
```
arboleya:
```suggestion
```

packages/address/src/address.ts
```diff
@@ -32,11 +29,14 @@ export default class Address extends AbstractAddress {
    */
   constructor(address: Bech32Address) {
     super();
-    logger.checkNew(new.target, Address);
+    // logger.checkNew(new.target, Address);
```
arboleya, line 32:
```suggestion
```

packages/errors/src/index.ts
```diff
@@ -0,0 +1,40 @@
...
+    const error = e as FuelError;
+
+    if (error.code === undefined) {
+      throw new FuelError(ErrorCode.PARSE_FAILED, "missing 'code' property");
+    }
+
+    const enumValues = Object.values(ErrorCode)
+    const codeIsUnknown = enumValues.includes(error.code)
+
+    if (!codeIsUnknown) {
```
arboleya: I think this is on me. 🤦‍♂️ 

```suggestion
    const codeIsUnknown = !enumValues.includes(error.code)

    if (codeIsUnknown) {
```

packages/errors/src/test-utils/expect-to-throw-fuel-error.ts
```diff
@@ -0,0 +1,20 @@
...
+  if (error === undefined) {
+    // this is a safeguard to not
+    expect(error).toBeDefined();
+    return;
+  }
+  expect(error).toBeInstanceOf(FuelError);
+  expect(error.code).toEqual(expectedError.code);
+  (Object.keys(expectedError) as Array<keyof typeof expectedError>).forEach((key) => {
+    expect(expectedError[key]).toStrictEqual(error[key]);
+  });
```
arboleya, line 19: No need to validate for `undefiend`; the other validations will fail for it.

``` suggestion
  const matcher = { name: 'FuelError', ...expectedError };
  expect(error).toBeInstanceOf(FuelError);
  expect(error?.toObject()).toMatchObject(matcher);
```

packages/errors/src/test-utils/expect-to-throw-fuel-error.test.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import { expectToThrowFuelError } from './expect-to-throw-fuel-error';
+
+
+describe('FuelError', () => {
+  it('should compare whole or partial error objects', () => {
+    const error = new FuelError(FuelError.CODES.PARSE_FAILED, 'It happens')
+    expectToThrowFuelError(() => {throw error}, error)
+    expectToThrowFuelError(() => {throw error}, { code: FuelError.CODES.PARSE_FAILED })
+  });
```
arboleya, line 32:
```suggestion
  });
  
  it('should fail with non-fuel errors', () => {
    const expected = { code: FuelError.CODES.PARSE_FAILED };
    const fnThrower = () => {
      throw new Error('x'); // not a FuelError
    };
    expect(expectToThrowFuelError(fnThrower, expected)).rejects.toThrow(
      /Expected constructor:.+FuelError/
    );
  });

  it('should fail on the absence of errors', () => {
    const code = FuelError.CODES.PARSE_FAILED;
    const fnThrower = () => {
      /* no throws here */
    };
    expect(expectToThrowFuelError(fnThrower, { code })).rejects.toThrow(
      /Received value:.+undefined/
    );
  });
```

arboleya (changes requested):
I left a few more notes.

There's also a (unused import) lint issue in:
 - `packages/address/src/utils.ts`

packages/errors/src/test-utils/expect-to-throw-fuel-error.test.ts
nedsalk: I've taken the 'should fail with non-fuel errors' test but not the second one as it was already handled in 'fails when passed fn doesnt throw'

packages/utils/src/test-utils/safeExec.ts
arboleya, line 1: Nice! 🤘

nedsalk:
> @nedsalk We could have a read-only `version` property on `FuelError`. It can be read from `@fuel-ts/versions`.

I've added it to `FuelError`.

packages/errors/src/test-utils/expect-to-throw-fuel-error.ts
```diff
@@ -0,0 +1,47 @@
...
+
+function assertExpectations(
+  thrownError: FuelError,
+  expectedError: FuelError | (Partial<FuelError> & Required<Pick<FuelError, 'code'>>)
+) {
+  expect(thrownError.name).toEqual('FuelError');
+
+  expect(enumValues).toContain(expectedError.code);
+
+  (Object.keys(expectedError) as Array<keyof typeof expectedError>).forEach((key) => {
```
nedsalk, line 16: `expect().toMatchObject()` was giving me some weird errors. This does the same thing and it's straightforward; possibly even more so as the reader doesn't have to understand `toMatchObject`.

packages/errors/src/test-utils/expect-to-throw-fuel-error.ts
```diff
@@ -0,0 +1,47 @@
...
+    expect(thrownError?.[key]).toEqual(expectedError[key]);
+  });
+}
+
+export function expectToThrowFuelError<
```
nedsalk, line 21: With this implementation the `expectToThrowFuelError` function can be used in both a synchronous and asynchronous context in the same way. If you pass an async `lambda`, you'll just have to do `expectToThrowFuelError`. This is the same behavior as doing `await expect(async () => {...}).rejects.toThrow()`, etc.

packages/address/src/address.test.ts
```diff
@@ -145,7 +147,12 @@ describe('Address utils', () => {
   });
 
   test('toB256 (b256 to b256)', () => {
-    expect(() => utils.toB256(ADDRESS_B256 as Bech32Address)).toThrow();
+    const address = ADDRESS_B256 as Bech32Address;
+    const expectedError = new FuelError(
+      FuelError.CODES.INVALID_BECH32_ADDRESS,
+      `Invalid Bech32 Address: ${address}`
```
danielbate, line 153: Would be nice if the message was derived from the code 🤔

packages/errors/src/error-codes.ts
danielbate, line 1: If these could be individually doc-blocked that will be really beneficial to typedoc and those hitting these errors to explain why they are seeing the error without digging through source

danielbate (dismissed):
This will definitely provide a lot of value, I think standardising the approach and error codes is great but I’d also like to see us start to standardise error messages. Potentially out of scope in this PR but we could also have a switch statement in the constructor that sets the message derived from the code, also allowing a message to be passed in it’s entirety until we have complete standardisation. We could then allow for some string interpolation:

```ts
switch(code) {
    case INVALID_ADDRESS: return `Invalid {$type} address`;
    case INVALID_SIZE: return `Invalid size for {$type}`;
...
}
````

I'd like to see us move away from `logger.throw…` in place of our own `throw new FuelError` and therefore only utilising our own package for error handling, and agree the version could come from the versions package. 

🚀

packages/address/src/address.test.ts
nedsalk: It seems reasonable to me that we currently focus primarily on the `ErrorCode`s and leave the error messages as they are. After we go through all packages with `FuelError`, `ErrorCode` patterns will start to emerge and we'll standardize them a bit more, as one `ErrorCode` can be used for multiple errors with different messages. That's why I'd leave it as it currently is and return to it afterwards.

packages/errors/src/error-codes.ts
nedsalk: I think we'll have to do that after we replace every error throwing with `FuelError`. We'll then have a proper set of error codes that's unlikely to change, and then we can document them properly. I've added a checkbox to #1156 which covers this discussion.

packages/errors/src/test-utils/expect-to-throw-fuel-error.ts
arboleya: This added a _lot_ of unnecessary complexity to `expectToThrowFuelError` method, and the goal usage deviates from the default way of `expect`-ing async throws in Jest.

arboleya (dismissed):
Although I understand your motivations, I think we should keep the `expect` signatures standard to avoid confusion. This, in turn, brings back the simplicity of the `expectToThrowFuelError` utility.

Here's another take:
 - https://github.com/FuelLabs/fuels-ts/pull/1166

packages/address/src/address.test.ts
arboleya: @danielbate What do you mean when you say the message should be **_derived_** from the code?

packages/errors/src/error-codes.ts
arboleya: @danielbate That's a good idea.

@nedsalk There's no reason we can't add these doc blocks right now to set the example<sup>(*)</sup>. It will set the standard for the subsequent PRs, and instead of leaving everything to be documented in the end, each new PR can document the codes they're adding.

> <sup>(*)</sup> This is the same approach I took when replacing the old errors on the `address` package—to leave an implementation example and set the standard for what should be done.

packages/address/src/address.test.ts
arboleya: Yes, messages will probably only sometimes derive from the code.

We can review this as we go.

packages/errors/src/error-codes.ts
nedsalk: @arboleya You're right, it makes sense. Didn't think about that. I'll add them.

packages/errors/src/test-utils/expect-to-throw-fuel-error.ts
nedsalk: Please review my [comment](https://github.com/FuelLabs/fuels-ts/pull/1166#discussion_r1285856794) on your PR.

Dhaiwat10 (dismissed):
Clean 👏🏻

Dhaiwat10:
Please make sure to file a follow-up issue for the logger issue you mentioned in the PR description, and maybe an epic issue for the conversion of `new Error` to `new FuelError` throughout the repo across different packages.

danielbate (approved):

nedsalk:
@Dhaiwat10 

> Please make sure to file a follow-up issue for the logger issue you mentioned in the PR description, and maybe an epic issue for the conversion of `new Error` to `new FuelError` throughout the repo across different packages.

I've opened up an issue for the logger: #1207.
@arboleya already created an epic for the whole repo: #1156.

Dhaiwat10 (approved):

PR #1154, opened 2023-07-31
chore: adding error test utility

arboleya:
This PR is a practical experiment on favoring test utilities instead of Jest plugins.

It also applies custom errors in `packages/address` to showcase the `packages/errors` usage.


apps/docs/package.json
```diff
@@ -13,6 +13,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "fuels": "workspace:*",
```
arboleya, line 16: This ensures we build the docs after all other packages.

> Had some weird problems locally, and this fixed it.

PR #1157, opened 2023-08-01
feat: supporting hex encoding on bufferFromString and stringFromBuffer

Torres-ssf:
This PR adds support for `hex` encoding in the browser versions of `bufferFromString` and `stringFromBuffer` utility functions. Additionally, these utility functions have been isolated into their own separate file for cleaner code organization.

Changes:
- Added `hex` encoding support to `bufferFromString` and `stringFromBuffer` functions for browser environments. These functions can now seamlessly handle `hex` encoded inputs along with previously supported `utf-8` and `base64` encodings.
- Isolated `bufferFromString` and `stringFromBuffer` functions into a separate file.
- Added unit tests to ensure that `bufferFromString` and `stringFromBuffer` functions work as expected


arboleya (commented):
These two files are identical:
 - `crypto/src/browser/bufferFromString.test.ts`
 - `crypto/src/node/bufferFromString.test.ts`

This makes me think we could have all tests in a `test` folder.

The same tests should work for node/browser since they follow a common interface.

PR #1160, opened 2023-08-01
feat: Add a `deployContract` method to contract factories generated by typegen

### LINKED ISSUE #1150: Add a deployInstance method to contract factories & export contract bytecode through a separate file via typegen
Follow-up from #1093:

To reduce & simplify the boilerplate required to deploy a contract after using typegen, adding a `deployInstance` method to the contract factories would be beneficial. This would take the boilerplate required to deploy a contract from this:

```ts
import bytecode from '../.../out/debug/bytecode.bin';

const factory = new ContractFactory(bytecode, FuelCounterAbi__factory.abi, wallet);
const instance = await factory.deployContract();
```

To this:

```ts
const instance = await FuelCounterAbi__factory.deployInstance(wallet);
```

We will need to include the contract's bytecode and storage slots in the generated factory to enable this.

Any thoughts? @FuelLabs/sdk-ts

Dhaiwat10:
Closes #1150 

This PR adds a new method called `deployInstance` to contract factories generated by typegen. The primary motivation behind this small feature is to make it slightly easier to deploy contracts for users & to make typegen more of a 'complete' tool for working with contracts:

```ts
// Without 'deployInstance'
import bytecode from '../.../out/debug/bytecode.bin';

const factory = new ContractFactory(bytecode, FuelCounterAbi__factory.abi, wallet);
const instance = await factory.deployContract();

// Now
import bytecode from '../types/FuelCounter.hex.ts';

const contract  = await FuelCounterAbi__factory.deployInstance(bytecode, wallet);
```

Note that we are exporting the contract's bytecode through a separate `*.hex.ts` file (generated by typegen). We have opted against including the bytecode in the contract factory's source file to avoid unnecessary bloat when the bytecode is not needed. More context [here](https://github.com/FuelLabs/fuels-ts/issues/1150#issuecomment-1659181698)


apps/demo-typegen/src/demo.test.ts
```diff
@@ -29,5 +25,19 @@ describe('ExampleContract', () => {
...
+    // Deploy
+    const contract = await DemoContractAbi__factory.deployInstance(bytecode, wallet);
+
+    // Call
+    const { value } = await contract.functions.return_input(1337).call();
+
+    // Assert
+    expect(value.toHex()).toEqual(toHex(1337));
+  });
 });
```
Dhaiwat10, line 42: I have added a test for `deployInstance` here in the `demo-typegen` app.

packages/abi-typegen/src/utils/collectBinFilePaths.test.ts
```diff
@@ -54,17 +54,20 @@ describe('collectBinFilePaths.ts', () => {
     expect(validateBinFile).toHaveBeenCalledTimes(1);
   });
 
-  test('should not collect bin files for Contracts', () => {
+  test('should collect bin files for Contracts', () => {
     const { validateBinFile } = mockDeps();
```
Dhaiwat10, line 58: We need to collect bin files for contracts now to export them via `*.hex.ts` files.

Dhaiwat10: Maybe we can make this opt-in via a CLI flag.

danielbate (dismissed):
LGTM 👍🏻

packages/abi-typegen/src/templates/contract/factory.hbs
```diff
@@ -17,4 +17,12 @@ export class {{capitalizedName}}__factory {
   ): {{capitalizedName}} {
     return new Contract(id, _abi, accountOrProvider) as unknown as {{capitalizedName}}
   }
+  static async deployInstance(
```
arboleya, line 20: Why `deployInstance` instead of `deployContract` just like the underlying method?

arboleya (changes requested):
Very clean. Good stuff! 🎯 

Could you mention please add a brief mention of it in the [docs](https://fuellabs.github.io/fuels-ts/guide/abi-typegen/using-generated-types.html)?

packages/abi-typegen/src/templates/contract/factory.hbs
Dhaiwat10: Great point, I prefer `deployContract` too. `deployInstance` can actually be confusing as it can mean we are deploying an instance of the factory itself and not a contract _from_ it. 

I think I just went with `deployInstance` because that's the name mentioned in the feedback issue that Nick opened 🙂

packages/abi-typegen/src/utils/collectBinFilePaths.test.ts
arboleya: Since the hex/bytecode will not be included automatically, we don't need a special flag.

apps/demo-typegen/src/demo.test.ts
arboleya: Perfect

arboleya (dismissed):

arboleya (approved):

nedsalk (approved):
Looks nice!

PR #1166, opened 2023-08-07
chore: simplifying `FuelError` implementation

arboleya:
In Jest, async thrown errors must be validated like:

```ts
expect(asyncMethod).rejects.toThrow(/error/);
```

Also, having empty `.toThrow()` expectations is unreliable. We should match it against the desired error.

This PR standardizes this usage and:

 1. Simplify `expectToThrowFuelError` implementation _(it got way too noisy)_
 2. Review and adjust all errors accordingly
 3. Put tests inside a `describe` block


packages/errors/src/test-utils/expect-to-throw-fuel-error.ts
```diff
@@ -1,47 +1,35 @@
...
-import type { FuelError } from '../index';
-import { ErrorCode } from '../index';
+import { ErrorCode, type FuelError } from '../index';
 
-const enumValues = Object.values(ErrorCode);
+const errorCodes = Object.values(ErrorCode);
 
-function assertExpectations(
-  thrownError: FuelError,
+export const expectToThrowFuelError = async (
```
nedsalk, line 7: The problem with making this function async is that all tests using it **must** be async, even when they're testing synchronous code. Jest _does_ fail when the expectations of a synchronous test that's using this function aren't met, but it doesn't fail gracefully; rather, it exits out of the whole test runner and stops running all the other tests. That's a no-go. In order to get a feel for what I mean, please do the following:
1. Go to `address.test.ts` and change a test to fail (e.g. `'clearFirst12BytesFromB256 (invalid B256)'`, change the expected error message).
2. Run the following command: `pnpm jest --no-cache --runInBand --coverage -- ./packages/address/src/address.test.ts`
3. Modify the test function to be async and await the call of `expectToThrowFuelError`,
4. Rerun the command in step 2

Notice that in the second case the test runner ran all the other tests, whereas in the first case it abruptly shut down and showed a serialized error that _happens_ to be a `JestAssertionError`; if some other error was thrown (e.g. the condition `if (!error) {throw new Error("Passed-in lambda didn't throw.")}` was true), that one would've been shown. Which means that this is just a "catch-all" mechanism in `jest` in the case that something goes awry. I believe that this is because `jest` was expecting a synchronous test, but `expectToThrowFuelError` fails asynchronously and this breaks `jest`.

If we continue using this function as you've proposed, then all _synchronous_ tests that are failing when using it are going to be breaking the test runner completely. Another side effect of this that I've noticed is that these tests "pass" inside of the vscode jest extension's UI even when they're failing internally (i.e. breaking the test runner). So you get a green checkmark but the underlying runner in its terminal says that the test failed. This is bad DX, given that we're providing this function as an utility.

My implementation works without any unexpected outcomes both in synchronous and asynchronous use cases. It's also type safe, in the sense that when the user passes in an asynchronous lambda, the inferred type is a promise, but when they pass in a synchronous lambda, returns void. This is how it should work. Also I disagree that it's complex; it's rather quite straightforward: 
- throw if lambda doesn't throw, or 
- assert the expectations:
--  in the catch block (synchronous),
--  in the `onrejected` callback (asynchronous)
- And infer based on the passed lambda if the function should return a `Promise<void>` or just `void` with a simple type conditional, so that the user knows if they should await or not.

packages/errors/src/test-utils/expect-to-throw-fuel-error.test.ts
```diff
@@ -2,67 +2,72 @@ import { FuelError } from '..';
...
+    const expectedError = { message: error.message }; // no code here
+    const fn = () => expectToThrowFuelError(throwFn, expectedError as FuelError);
+    expect(fn).rejects.toThrow('Expected error must contain a code.');
+  });
 
-  expect(() => expectToThrowFuelError(thrower, expected)).toThrow();
-});
+  it('fails when partial error object doesnt contain code (async)', () => {
+    const expectedError = { message: error.message }; // no `code` here
+    const fn = () => expectToThrowFuelError(asyncThrowFn, expectedError as FuelError);
```
nedsalk, line 42: The benefit of using 
```ts
expect(
    expectToThrowFuelError(
      asyncThrowFn,
      // @ts-expect-error code property is required per type definition
      { message: error.message }
    )
```
instead of 
```ts
const fn = () => expectToThrowFuelError(asyncThrowFn, expectedError as FuelError);
```
with regards to `expectedError`  that in the first case the `// ts-expect-error` is also testing the type system's requirements that a `code` property must be provided, whereas casting it directly to `FuelError` doesn't test that.g

packages/errors/src/test-utils/expect-to-throw-fuel-error.ts
arboleya: Could you please push a _failing test_ to this PR that summarizes your point?

packages/errors/src/test-utils/expect-to-throw-fuel-error.test.ts
arboleya: TBH, I find `@ts-*` tags an ugly thing. But you have a point.

nedsalk:
@camsjams It's currently a back-and-forth between @arboleya and me to come up with a solution that we're both happy with, so we can't merge it yet.

packages/errors/src/test-utils/expect-to-throw-fuel-error.ts
arboleya: The tests would have been enough; thank you for the detailed answer.

Then it is a non-negotiable thing; my little hack must die prematurely.

I will return the base branch to `master` to re-contextualize myself and get back to you soon.

arboleya: Instead of trying to unify the sync/async usages, perhaps we should split them.

I start exploring this direction in this other PR:
 - https://github.com/FuelLabs/fuels-ts/pull/1193

I'm closing this PR/suggestion in favor of that one.

We can continue the conversation there.

PR #1164, opened 2023-08-06
feat!: refactor contract calls and multicall to JS + fuel-asm

camsjams:
This PR converts the Sway file to asm, allowing for:
- a contract call script that doesn't require a Sway compile step
- easier upgrades
- smaller payload size for contract calls
- unlimited mulitcalls (currently 5 maximum)
- IMO the most important changes to look at are all within `packages/program/src/contract-call-script.ts`

### Notes:
- Uses fuel-asm-wasm from @xgreenx and @arboleya (thanks :tada: )
- Added basic wrapper for fuel-asm for DX ergonomics
- removes all usage and sources of multicall Sway contract
- added tests for multicalls greater than 5 now :rocket: 
- **10 file changes are just deletions**, many are just adjusting unit tests :+1: 
- reducing the size of the contract call script changed the gas required

PR #1169, opened 2023-08-07
build: build and cache all forc projects

### LINKED ISSUE #1095: Update pnpm run dev to not build sway projects

nedsalk:
Added a bash script which finds all forc projects in the repo, hashes their `*.sw` and `Forc.toml` files, stores that on disk, creates a workspace `Forc.toml` file with all those projects and builds them. Later on when a project's sway or `Forc.toml` file changes, the hash will change for that project and a new workspace file will be created which will only contain those projects whose files have changed and only those projects will be rebuilt. If there aren't any changes to any files that get hashed, the hashes will be the same and no rebuild will happen. The script is also tracking the `forc` version being used in the repo so it will rebuild the local caches on dev machines when we update the forc version.

- The `.forc-build-cache` folder is in `.gitignore` so it won't be committed to version control.
- I've removed all other `Forc.toml` workspace files as the auto-generated `Forc.toml` is the one to rule them all.
- I've also removed/updated all `package.json` `prebuild` and `pretest` scripts where forc builds were happening.
- I've updated some shell scripts where forc builds were happening to not build anymore; rather, they assume that forc builds already happened and then they do any additional logic based on it.
- I've renamed two projects in the `abi-typegen` package because there were naming conflicts with projects in `fuel-gauge`. This is due to `forc` not allowing projects with the same name in one workspace.

This solution doesn't interfere with building single projects. You can go to a specific project and build it; it will just be rebuilt by this caching solution again when you run `pnpm build` in the root directory.
It also doesn't interfere with CI. The cache folder does get created in CI, but it's not used afterwards for anything.

As a side note, this caching solution has trimmed down ~15s off of build times on my computer.

Closes #1095.


package.json
```diff
@@ -10,8 +10,10 @@
   },
   "packageManager": "pnpm@8.1.1",
   "scripts": {
-    "dev": "nodemon --config nodemon.config.json -x 'turbo run build --filter=!fuel-gauge'",
-    "build": "turbo run build",
+    "prpa": "eslint . --ext .ts",
```
danielbate, line 13: What does `prpa` mean?

package.json
```diff
@@ -10,8 +10,10 @@
...
+    "dev": "nodemon --config nodemon.config.json -x 'pnpm build --filter=!fuel-gauge'",
+    "build": "pnpm build-forc-projects && turbo run build",
+    "build-forc-projects": "./scripts/build-script/build-all-with-cache.sh",
```
danielbate, line 16:
```suggestion
    "build:forc": "./scripts/build-script/build-all-with-cache.sh",
```
We tend to use the `build:some-thing` notation more consistently in the SDK.

.gitignore
```diff
@@ -1,3 +1,6 @@
+#CACHE
```
danielbate, line 1:
```suggestion
# Cache
```
Nit, sorry

danielbate (commented):
Couple comments, the script and approach seems good to me if something like this is not already built in.

arboleya (changes requested):
I can't see why we'd want to re-implement a hand-made cache solution in Shell Script, considering TurboRepo has it built-in and is already caching our dev pipelines.

We want to avoid building the Forc projects during the `dev` command. Looking at the most straightforward implementation with the least amount of friction added, we should preserve things as they were, segregated per packages, with scripts explicitly set in their `package.json` in an obvious manner.

Some things that come to mind:
 1. Try standardizing all script names — for example, `build:forc` as @danielbate suggested
 2. Call it from a `pretest` step for relevant projects
     - `"pretest": "pnpm build:forc"`
 3. Review and adjust any TurboRepo configs/pipelines to match the goal
 4. Ensure there's no relationship between `forc:build` _(and `pretest`)_, `build`, and `dev`.
 5. Normalize exception for the `fuel-gauge` project _(currently skipped in `build` script)_, making it behave like the others.

package.json
```diff
@@ -10,8 +10,10 @@
   },
   "packageManager": "pnpm@8.1.1",
   "scripts": {
-    "dev": "nodemon --config nodemon.config.json -x 'turbo run build --filter=!fuel-gauge'",
-    "build": "turbo run build",
+    "prpa": "eslint . --ext .ts",
+    "dev": "nodemon --config nodemon.config.json -x 'pnpm build --filter=!fuel-gauge'",
```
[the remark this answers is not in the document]
nedsalk: It doesn't affect it and I wanted `pnpm build` to encapsulate `turbo run build` as well as building the forc projects.

arboleya:
@nedsalk Are you still going to address #1095 issue in this PR, or will you create a new one?

nedsalk:
@arboleya This is the PR for it, but it's currently on hold.

arboleya:
Precisely, I'm asking because the last commit is 2x weeks old.

Can we help you with something to get it finished?

nedsalk:
> Can we help you with something to get it finished?

Thanks @arboleya! Sorry for not answering for so long, but I first wanted to gather my thoughts on it before I can comment any further.

After going through what [you've mentioned](https://github.com/FuelLabs/fuels-ts/pull/1169#pullrequestreview-1572278032), I remembered that I already tested the caching capabilities of turbo and this was/is what I noticed:
1. Turbo indeed caches forc builds,
2. Turbo rebuilds forc projects even when a non-sway file is changed

The reason why I went down this route was to avoid building sway projects when a non-sway file changes. This was primarily driven by not wanting to rebuild the `call-test-script` in the `script` package every time I change anything that's in the dependency chain below `script`, as that sway project rebuilding was being run via a `build` script. I figured that I could easily expand the solution to include all forc projects.

However, after #1164 got merged, there is indeed no need at all to build forc projects in `build` scripts. I tried out your 
> 4. Ensure there's no relationship between forc:build (and pretest), build, and dev.

proposal in a separate branch and it works - it gets cached, with the caveat that turbo rebuilds forc projects even when a non-sway file is changed. But, given that it's a `pretest` step, it's not really that impactful on the dev workflow given that the `pretest` scripts are primarily run during `pnpm ci:test`, which runs the full test suite anyways.

I'll close this PR in favor of #1235 which is in the spirit of your comment.

PR #1171, opened 2023-08-07
feat: accept `DeployContractOptions` on `deployContract` of typegen contract factories

### LINKED ISSUE #1170: Extend the functionality of `deployContract` in typegen contract factory - `DeployContractOptions`
The `deployContract` helper exposed via typegen contract factories is fairly limited. It does not let users specify any `DeployContractOptions` like storage slots or configurable constants.

Dhaiwat10:
Closes #1170

This PR extends `deployContract` so that it lets users specify `DeployContractOptions` like storage slots and configurable constants like so:

```ts
const contract = await DemoContractAbi__factory.deployContract(bytecode, wallet, {
      storageSlots,
    });
 ```

### CODE REVIEW

**Review by nedsalk — APPROVED** [review]

File: apps/docs/src/guide/abi-typegen/using-generated-types.md [hunk 1]
```diff
@@ -40,6 +40,16 @@ const contract = await MyContract__factory.deployContract(bytecode, wallet);
...
+You can also pass in [`DeployContractOptions`](https://github.com/FuelLabs/fuels-ts/blob/a64b67b9fb2d7f764ab9151a21d2266bf2df3643/packages/contract/src/contract-factory.ts#L19-L24) like storage slots and configurable constants to the `deployContract` method:
+
+```ts
+import storageSlots from "../contract/out/debug/storage-slots.json";
+
+const contract = await MyContract__factory.deployContract(bytecode, wallet, {
+  storageSlots,
+});
+```
+
```
danielbate, line 52: Can we move this to a docs-snippets test so that it does not get outdated?

danielbate (changes requested):
Looking good, just one small comment

apps/docs/src/guide/abi-typegen/using-generated-types.md
Dhaiwat10: @danielbate agreed, but I also found out that the rest of the snippets in this file are hard-coded as well and have a todo attached to them about moving them to doc-snippets. Do you mind if I tackle all of them at once in a follow-up issue after this PR?

danielbate: Makes sense 👍🏻

Issue #1050, opened 2023-06-06
Read transaction parameters from chain and not hardcoded values


camsjams:
### Motivation

The max Inputs setting for Transaction Parameters (aka Consensus Parameters) can be defined in fuel-core chain config:
https://github.com/FuelLabs/fuels-ts/blob/cm/issue-1015/.docker/fuel-core/chainConfig.json#L275

But the TS-SDK hardcodes a value of 255 which just happens to match that of the chain config:
https://github.com/FuelLabs/fuels-ts/blob/cm/issue-1015/packages/abi-coder/src/constants.ts#L21

### Possible implementations

We should update the functionality to use the actual value defined in the local chain if possible.

Dhaiwat10:
@camsjams this value would have to be fetched asynchronously from the node just like the rest of the consensus parameters, right? Reading directly from a `chainConfig` file would only work for a local node.

digorithm:
@Dhaiwat10 @camsjams should this be closed?

digorithm:
That's what I thought. Dhai merged some PR related to this. Not sure about @Torres-ssf's work, though. I'll wait for him to update us before closing this issue.

Torres-ssf:
@digorithm Some constants were purged here:

- https://github.com/FuelLabs/fuels-ts/pull/1272
- https://github.com/FuelLabs/fuels-ts/pull/1271

However, it seems that the hardcoded value for `MAX_INPUTS` is still being used to calculate `VM_TX_MEMORY`.

Dhaiwat10:
Like @Torres-ssf mentioned, we still need to purge this one last constant. Filed an issue for it: #1313 

I guess we can close this issue though, since the supporting work for this has been done already

PR #1181, opened 2023-08-09
feat!: fetch and cache chain info on `Provider` when initialized

### LINKED ISSUE #1176: Fetch & cache consensus parameters when `Provider` is initialized
This API change/feature attempts to solve #1050

Similar to the Rust SDK's `Provider::connect`:
```rs
/// Connects to an existing node at the given address.
    pub async fn connect(url: impl AsRef<str>) -> Result<Provider> {
        let client = FuelClient::new(url).map_err(|err| error!(InfrastructureError, "{err}"))?;
        let consensus_parameters = client.chain_info().await?.consensus_parameters.into();
        Ok(Provider::new(client, consensus_parameters))
    }
```
In the TS SDK, we could have:

```typescript 
// new API
const provider = await Provider.connect('https://node.url');

// old API
const provider = new Provider('https://node.url')
```
### Motivation behind this API change
Making the initialization of the provider asynchronous will let us internally fetch consensus parameters from the chain and cache them on our provider object for later use. 

This means that all of our internal logic related to contract calls can stay synchronous as it is with minimal modifications, and just consume these cached consensus parameters from the provider object whenever needed.

Please correct me if I am wrong - but I think the consensus parameters of the chain should not change often. Still, if the user wants to invalidate the cache and wants the provider to use the latest consensus parameters, we can expose a method called `invalidateConsensusParamsCache` from the `Provider` class that lets them do that explicitly.

### LINKED ISSUE #1260: Add a `getGasConfig` helper on the `Provider` class
Follow-up from #1181 and [this comment](https://github.com/FuelLabs/fuels-ts/pull/1181/files#r1301934987):

### Possible implementation:

```ts
getGasConfig() {
  // get cached info
  const chain = this.getChain();
  const nodeInfo = this.getNodeInfo();

  const { minGasPrice } = nodeInfo;
  const { maxGasPerTx, maxGasPerPredicate, gasPriceFactor, gasPerByte } = chain;

  // return relevant gas info
  return {
    minGasPrice, maxGasPerTx, maxGasPerPredicate, gasPriceFactor, gasPerByte
  }
}
```

### Possible usage:

```ts
const provider = await Provider.create(url);
const gasConfig = provider.getGasConfig();

// now use correct config to create things with correct default gasConfig
const scriptRequest = new ScriptTransactionRequest({ gasPrice: gasConfig.minGasPrice, gasLimit: maxGasPerTx });
```

Dhaiwat10:
## Summary
This PR caches the `chainInfo` for a given Fuel node on the `Provider` instance whenever a `Provider` is initialized.

> Note: I know this PR has a *lot* of changes (~ 80 files), but 90% of these changes are changes made to our test files so that they are adapted to the breaking changes. I recommend focusing on reviewing these files: `provider.ts`, `script-invocation-scope.ts`, `base-invocation-scope.ts`.

## Motivation
Please see https://github.com/FuelLabs/fuels-ts/issues/1176 for a detailed breakdown of the problem at hand, but tldr: we are currently using a lot of hardcoded values for some constants that we use when creating transactions. These hardcoded values served us well until now, but we now need to fetch them from the chain dynamically. This has been requested as a top priority by the frontend team.

## Breaking Changes
Because of the nature of the changes introduced in this PR, and the fact that `Provider` is used so widely in our SDK, there are multiple breaking changes that we need to be aware of and need to communicate to our users:

#### 1. Initializing the `Provider`
```ts
// Used to be:
const provider = new Provider(url)

// New API:
const provider = await Provider.create(url)
```

#### 2. `Wallet methods`
All of these methods now require a `Provider` to be passed in:
```ts
const provider = await Provider.create(url);

// Some of these methods used to accept a URL instead of a `Provider` object.
// Note that the `provider` param *has* to be a `Provider` object now.
WalletUnlocked.fromSeed(seed, provider, path);
WalletUnlocked.fromMnemonic(mnemonic, provider, path, passphrase);
WalletUnlocked.fromExtendedKey(extendedKey, provider);
await WalletUnlocked.fromEncryptedJson(jsonWallet, password, provider);

Wallet.fromAddress(address, provider);
Wallet.fromPrivateKey(pk, provider);

Wallet.generate({ provider });
```

#### 3. `Account` class
```ts
new Account(address, provider);
```

#### 4. `PrivateKeyVault`
```ts
// these are the options that are accepted by the `PrivateKeyVault` constructor. `provider` is a required input now.
interface PkVaultOptions {
  secret?: string;
  accounts?: Array<string>;
  provider: Provider;
}
```

#### 5. `MnemonicVault`
Same as `PrivateKeyVault`. The constructor requires a `provider` to be passed in.

#### 6. `WalletManager`
The `VaultConfig` now requires a `provider` to be passed in.
```ts
export type VaultConfig = {
  type: string;
  title?: string;
  secret?: string;
  provider: Provider;
};
```

#### 7. `Predicate`
```ts
// `provider` is no longer optional. there is a change in order of the parameters here, + we don't require a `chainId` to be passed anymore. 
new Predicate(bytes, provider, jsonAbi);
```

### Other notes
To demonstrate the fact that we can now use fresh on-chain values for some of the previosuly hard-coded constants, I have refactored the usage of the `VM_TX_MEMORY` constant. This constant is used to calculate the contract call script's data offset.

Closes #1176
Closes #1260


apps/demo-typegen/src/demo.test.ts
```diff
@@ -9,7 +9,7 @@ import bytecode from './generated-types/DemoContractAbi.hex';
 
 describe('ExampleContract', () => {
   it('should return the input', async () => {
-    const provider = new Provider('http://127.0.0.1:4000/graphql');
+    const provider = await Provider.connect('http://127.0.0.1:4000/graphql');
```
Dhaiwat10, line 12: For end-users, this is the _big_ breaking change that this PR will introduce.

packages/providers/src/provider.ts
```diff
@@ -209,15 +211,34 @@ export default class Provider {
...
     this.operations = this.createOperations(url, options);
     this.cache = options.cacheUtxo ? new MemoryCache(options.cacheUtxo) : undefined;
   }
 
+  static async connect(url: string, options: ProviderOptions = {}) {
+    const provider = new Provider(url, options);
+    const { consensusParameters } = await provider.getChain();
+    provider.consensusParamsCache = consensusParameters;
+    return provider;
+  }
```
Dhaiwat10, line 231: Instead of telling users to create providers using `new Provider()`, we will now encourage users to run `await Provider.connect('url')` instead.

This new `connect` method calls the constructor and does some additional work afterwards. It fetches the consensus params from the chain, and simply stores them on the `Provider` object for later use.

packages/providers/src/provider.ts
```diff
@@ -227,13 +248,6 @@ export default class Provider {
     return getOperationsSdk(gqlClient);
   }
 
-  /**
-   * Connect provider to a different Fuel node url
-   */
-  connect(url: string) {
-    this.operations = this.createOperations(url);
-  }
-
```
Dhaiwat10, line 236: The new `connect` method is not to be confused with this _old_ method with the same name. I have preserved this functionality and put it under a different name - `updateUrl`.

packages/providers/src/provider.ts
```diff
@@ -209,15 +211,34 @@ export default class Provider {
...
+    const provider = new Provider(url, options);
+    const { consensusParameters } = await provider.getChain();
+    provider.consensusParamsCache = consensusParameters;
+    return provider;
+  }
+
+  async invalidateConsensusParamsCache() {
+    const { consensusParameters } = await this.getChain();
+    this.consensusParamsCache = consensusParameters;
+  }
```
Dhaiwat10, line 236: Since the consensus parameters will never change unless the URL that the `Provider` object is pointing to changes, I don't think we need to ever re-fetch them over an interval. Still, if a user wants to update them explicitly, I have exposed this method that lets them do so.

packages/providers/src/provider.ts
```diff
@@ -209,15 +211,34 @@ export default class Provider {
...
+
+  updateUrl(url: string) {
+    this.operations = this.createOperations(url);
+  }
```
Dhaiwat10, line 240: We should probably invalidate the consensus params cache when we update the URL. I'm noting down this as a TODO.

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -16,9 +16,18 @@ import { InvocationCallResult, FunctionInvocationResult } from './invocation-res
...
+  if (!consensusParams) {
+    throw new Error(
+      'Provider consensus params cache empty! Pls make sure you ran `await Provider.connect()` and not just `new Provider()`'
+    );
+  }
+
   const data = func.encodeArguments(
     args as Array<InputValue>,
-    contractCallScript.getScriptDataOffset() + bytesOffset
+    contractCallScript.getScriptDataOffset(consensusParams.maxInputs.toNumber()) + bytesOffset
```
Dhaiwat10, line 30: We can now easily access and pass down consensus params everywhere we have access to the provider object, like so.

packages/abi-coder/src/constants.ts
```diff
@@ -13,6 +13,12 @@ export const MAX_INPUTS = 255;
 export const ASSET_ID_LEN = BYTES_32;
 export const CONTRACT_ID_LEN = BYTES_32;
 
+export const calculateVmTxMemory = ({ maxInputs }: { maxInputs: number }) =>
+  BYTES_32 + // Tx ID
+  WORD_SIZE + // Tx size
+  // Asset ID/Balance coin input pairs
+  maxInputs * (ASSET_ID_LEN + WORD_SIZE);
+
```
Dhaiwat10, line 21: As requested in https://github.com/FuelLabs/fuels-ts/issues/1050, we don't need to use  hard-coded value for `MAX_INPUTS` anymore. We can safely use a value fetched from the chain, passed down from a provider object.

packages/program/src/functions/base-invocation-scope.ts
Dhaiwat10: This is also where we are passing down the `maxInputs` fetched from the chain, down to the `getScriptDataOffset` function and then the `calculateVmTxMemory` function so that we don't use a hard-coded value for `MAX_INPUTS` anymore.

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -16,9 +16,18 @@ import { InvocationCallResult, FunctionInvocationResult } from './invocation-res
...
   const { program, args, forward, func, callParameters, bytesOffset } = funcScope.getCallConfig();
 
+  const provider = program.provider as Provider;
+  const consensusParams = provider.consensusParamsCache;
+
+  if (!consensusParams) {
+    throw new Error(
+      'Provider consensus params cache empty! Pls make sure you ran `await Provider.connect()` and not just `new Provider()`'
+    );
+  }
```
danielbate, line 26: This feels a bit unintuitive to me, we should probably ensure that `Provider::connect()` is called within the `constructor()` and thus has the consensus params available otherwise I can see us needing this check in multiple places. And then we can also ensure type safety that `consensusParams` is always set on the `Provider` instance.

danielbate, line 26: This feels a bit unintuitive to me, we should probably ensure that `Provider::connect()` is called within the `constructor()` and thus has the consensus params available otherwise I can see us needing this check in multiple places. And then we can also ensure type safety that consensusParams is always set on the Provider instance.

danielbate (commented):
Looking good, I am in favour of bringing us closer to the Rust SDK and improving the UX. However couple comments on the approach:

- If `connect()` is now paramount to populate `consensusParameters`, we should make the constructor private and therefore `Provider` becomes instantiatable only via `connect()`. Therefore we won't need to encourage one way or the other and it is enforced. We could introduce a `ProviderFactory` to handle the provision of a packaged provider but this is potentially an unnecessary level of abstraction if we are in preference of dev ex and bringing us closer to the Rust SDK.
- Or we maintain the API, and call `connect()` from within the constructor (not my preference but lots of breaking changes / tutorials to think of).
- Another thought is that now `consensusParameters` and the provider itself are now tightly bound. Should we be making `url` read only, and restricting the ability to update a url as that itself should refetch the params. Or we make the url update in turn refetch. But I feel as if we are restricting those, we should just be reinstantiating a `Provider` class?

packages/program/src/functions/base-invocation-scope.ts
Torres-ssf: I completely agree with @danielbate. This field `consensusParams` should always be present on an instance of a Provider

packages/providers/src/provider.ts
Torres-ssf: Could be possible that by updating the provider url we would have a provider with a different chain config? If so, maybe it makes sense to make `updateUrl` also async and to work similarly with `connect`, to fetch the `consesusParameters` once again, since they could be different from the last provider instance.

packages/providers/src/provider.ts
Torres-ssf: Perhaps I'm overlooking something, but `consensusParameters` appears to me to be more of a class attribute rather than a cache. If that's the case, maybe we should consider removing the cache suffix.

I am under the impression that these `consensusParameters` should be fetched every time a Provider is instantiated with (`await Provider.connect('url')) and also whenever we update the Provider URL since a different provider would imply a different chain configuration. And only in these 2 cases.

Please let me know if I'm missing any details or if I'm oversimplifying things.

packages/providers/src/provider.ts
Torres-ssf: It would make sense to make the constructor private and force users to only instantiate the Provider class by `await Provider.connect('url')`. This way we make sure that the `consensusParams` is always going to be available within a Provider instance

packages/providers/src/provider.ts
```diff
@@ -64,30 +64,32 @@ export type ContractResult = {
...
+  maxOutputs: BN;
+  maxWitnesses: BN;
+  maxGasPerTx: BN;
+  maxScriptLength: BN;
+  maxScriptDataLength: BN;
+  maxStorageSlots: BN;
+  maxPredicateLength: BN;
+  maxPredicateDataLength: BN;
+  maxGasPerPredicate: BN;
+  gasPriceFactor: BN;
```
Torres-ssf, line 79: I reckon that with the solution proposed here, we can also close https://github.com/FuelLabs/fuels-ts/issues/784

packages/providers/src/provider.ts
arboleya: I think it's cool to cache consensus parameters using the provider URL as the key. It can improve startup times if we are constructing new `Provider`s for the same URLs in different places of the application.

packages/program/src/functions/base-invocation-scope.ts
arboleya: It looks like we have some chicken-egg problem here.

Maybe we could fetch the needed params differently and then pass them to the `Provider` constructor.

```ts
Provider.connect = (url: stirng) => {
  const xyz = await Provider.getChainInfo(url)
  const p = new Provider(url, ...xyz)
  return p;
}
```

Dhaiwat10:
A bunch of spaghetti code still, don't review yet pls :P

apps/demo-typegen/src/demo.test.ts
Dhaiwat10: @camsjams I actually tried what you are proposing in #1173, and while I did make it work for contracts - it is a very hacky solution and I could not make it work for script calls.

After having tried that, I found it easier to just fetch these consensus parameters and pass them down instead of doing it later on and trying to make a lot of things that are sync right now async. This is also what the Rust SDK does.

I agree that us having to rewrite docs, and users having to rewrite their code is a bit of a hassle but I believe it might be worth doing it now rather than later on.

packages/providers/src/provider.ts
Torres-ssf: By moving to this approach and making sure that a `provider` always has the `consensusParameter`, we get rid of the constants like `GAS_PRICE_FACTOR`. 

Since these values are going to be always present on the provider, we can easily access them on the `BaseInvocationScope`, `Account`, and `TransactionResponse` classes

packages/wallet/src/wallets.ts
```diff
@@ -100,15 +100,19 @@ export class WalletUnlocked extends BaseWalletUnlocked {
...
 
     return new WalletUnlocked(<string>hdWallet.privateKey, provider);
   }
 
-  static async fromEncryptedJson(jsonWallet: string, password: string): Promise<WalletUnlocked> {
+  static async fromEncryptedJson(
+    jsonWallet: string,
+    password: string,
+    provider: Provider
+  ): Promise<WalletUnlocked> {
```
Dhaiwat10, line 113: ❗️

packages/wallet/src/wallets.ts
```diff
@@ -100,15 +100,19 @@ export class WalletUnlocked extends BaseWalletUnlocked {
    * @param provider - The provider URL or a Provider instance (optional).
    * @returns An instance of WalletUnlocked.
    */
-  static fromExtendedKey(extendedKey: string, provider?: Provider): WalletUnlocked {
+  static fromExtendedKey(extendedKey: string, provider: Provider): WalletUnlocked {
```
Dhaiwat10, line 103: ❗️

packages/wallet/src/wallets.ts
```diff
@@ -75,16 +75,16 @@ export class WalletUnlocked extends BaseWalletUnlocked {
...
-   * @param provider - The provider URL or a Provider instance (optional).
    * @returns An instance of WalletUnlocked.
    */
   static fromMnemonic(
     mnemonic: string,
+    provider: Provider,
     path?: string,
-    passphrase?: BytesLike,
-    provider?: Provider
+    passphrase?: BytesLike
```
Dhaiwat10, line 87: ❗️

packages/wallet/src/wallets.ts
```diff
@@ -60,11 +60,11 @@ export class WalletUnlocked extends BaseWalletUnlocked {
    * Create a Wallet Unlocked from a seed.
    *
    * @param seed - The seed phrase.
+   * @param provider - The provider URL or a Provider instance.
    * @param path - The derivation path (optional).
-   * @param provider - The provider URL or a Provider instance (optional).
    * @returns An instance of WalletUnlocked.
    */
-  static fromSeed(seed: string, path?: string, provider?: Provider): WalletUnlocked {
+  static fromSeed(seed: string, provider: Provider, path?: string): WalletUnlocked {
```
Dhaiwat10, line 67: ❗️

packages/wallet/src/wallets.ts
```diff
@@ -50,7 +50,7 @@ export class WalletUnlocked extends BaseWalletUnlocked {
    * @param generateOptions - Options to customize the generation process (optional).
    * @returns An instance of WalletUnlocked.
    */
-  static generate(generateOptions?: GenerateOptions): WalletUnlocked {
+  static generate(generateOptions: GenerateOptions): WalletUnlocked {
```
Dhaiwat10, line 53: ❗️

packages/wallet/src/wallet.ts
```diff
@@ -30,7 +26,7 @@ export class Wallet {
    * @param provider - The provider URL or a Provider instance.
    * @returns An unlocked wallet instance.
    */
-  static fromPrivateKey(privateKey: BytesLike, provider: string | Provider = FUEL_NETWORK_URL) {
+  static fromPrivateKey(privateKey: BytesLike, provider: Provider) {
```
Dhaiwat10, line 29: ❗️

packages/wallet/src/wallet.ts
```diff
@@ -16,10 +15,7 @@ export class Wallet {
    * @param provider - The provider URL or a Provider instance.
    * @returns A locked wallet instance.
    */
-  static fromAddress(
-    address: string | AbstractAddress,
-    provider: string | Provider = FUEL_NETWORK_URL
-  ): WalletLocked {
+  static fromAddress(address: string | AbstractAddress, provider: Provider): WalletLocked {
     return new WalletLocked(address, provider);
```
Dhaiwat10, line 19: ❗️

packages/wallet/src/base-unlocked-wallet.ts
```diff
@@ -38,7 +37,7 @@ export class BaseWalletUnlocked extends Account {
    * @param privateKey - The private key of the wallet.
    * @param provider - The provider URL or a Provider instance.
    */
-  constructor(privateKey: BytesLike, provider: string | Provider = FUEL_NETWORK_URL) {
+  constructor(privateKey: BytesLike, provider: Provider) {
```
Dhaiwat10, line 40: ❗️

packages/wallet/src/account.ts
```diff
@@ -54,7 +53,7 @@ export class Account extends AbstractAccount {
    * @param address - The address of the account.
    * @param provider - The provider URL or a Provider instance.
    */
-  constructor(address: string | AbstractAddress, provider: string | Provider = FUEL_NETWORK_URL) {
+  constructor(address: string | AbstractAddress, provider: Provider) {
```
Dhaiwat10, line 56: ❗️

packages/wallet-manager/src/vaults/privatekey-vault.ts
```diff
@@ -19,33 +22,41 @@ export class PrivateKeyVault implements Vault<PkVaultOptions> {
...
+    this.provider = options.provider;
     if (options.secret) {
       this.#privateKeys = [options.secret];
     } else {
-      this.#privateKeys = options.accounts || [Wallet.generate().privateKey];
+      this.#privateKeys = options.accounts || [
+        Wallet.generate({
+          provider: options.provider,
+        }).privateKey,
+      ];
```
Dhaiwat10, line 33: ❗️

packages/wallet-manager/src/vaults/mnemonic-vault.ts
```diff
@@ -18,12 +20,14 @@ export class MnemonicVault implements Vault<MnemonicVaultOptions> {
   pathKey = '{}';
   rootPath: string = `m/44'/1179993420'/${this.pathKey}'/0/0`;
   numberOfAccounts: number = 0;
+  provider: Provider;
```
Dhaiwat10, line 23: ❗️

packages/providers/src/provider.ts
```diff
@@ -205,28 +207,75 @@ export type ProviderCallParams = {
...
+  /**
+   * Creates a new instance of the Provider class. This is the recommended way to initialize a Provider.
+   * @param url - GraphQL endpoint of the Fuel node
+   * @param options - Additional options for the provider
+   */
+  static async connect(url: string, options: ProviderOptions = {}) {
+    const consensusParameters = await this.getConsensusParamsWithoutInstance(url);
+    const provider = new Provider(url, consensusParameters, options);
+    return provider;
+  }
```
Dhaiwat10, line 251: ❗️❗️

packages/providers/src/provider.ts
```diff
@@ -64,30 +64,32 @@ export type ContractResult = {
   bytecode: string;
 };
 
+type ConsensusParameters = {
```
LuizAsFight, line 67: is it possible/good to use the generated type from graphql?

packages/providers/src/provider.ts
LuizAsFight, line 251: we could apply this strategy to don't have breaking changes:

```suggestion
  static async create({ url, options }: { url: string, options?: ProviderOptions }) {
    const provider = new Provider(url, options);
    await provider.connect();
    return provider;
  }
```

on this way:
- use a more generic name `create` that can be used in other classes as well to maintain the pattern
- no breaking changes as constructor is not changed (everyone using `new Provider(url)`)
- use input object to pass params
- `updateUrl` function is not needed as `connect` should be centralizing the connection and caching the provider info when needed 
- no breaking changes in `connect` method as won't be changed to `static async`. 
  - `connect` now will receive a optional `url` param, this allow us to connect to current URL and also change to other URL
  - when connecting will get the url from:
```
  connect({ url }: { url: string }) {
    const urlToConnect = input?.url || this.url; 
    // logic to connect to URL, get consensus params if it's not cached, etc..
  }
```

ps. we can also adopt same pattern to use a `static async create` as a way to handle the async part that we don't wanna do in the constructor. other places in the code might need a similar refactor to avoid hardcoded consts

it's being done also in TransactionResponse -> https://github.com/FuelLabs/fuels-ts/pull/1206

packages/providers/src/provider.ts
```diff
@@ -205,28 +207,75 @@ export type ProviderCallParams = {
...
+    return provider;
+  }
+
+  /**
+   * Re-fetches the consensus parameters from the chain for the current URL.
+   */
+  async invalidateConsensusParamsCache() {
+    const { consensusParameters } = await this.getChain();
+    this.consensusParamsCache[this.url] = consensusParameters;
+  }
```
LuizAsFight, line 259: would be nice to reuse the `connect` method

```suggestion
  async refetchConsensusParams() {
    // clear the cache for current URL
    delete this.consensusParamsCache[this.url];
    
    // connect will handle getting the consensusParams if it's not cached
    await this.connect();
  }
```

packages/providers/src/provider.ts
```diff
@@ -205,28 +207,75 @@ export type ProviderCallParams = {
...
+
+  /**
+   * Returns the cached consensus parameters for the current URL.
+   */
+  getConsensusParams() {
+    return this.consensusParamsCache[this.url];
+  }
```
LuizAsFight, line 266: this method is not needed if we use `getChain`

what will change is:
we need a new method to fetch from server
```
  async fetchChain(): Promise<ChainInfo> {
    const { chain } = await this.operations.getChain();
    return processGqlChain(chain);
  }

  // then getChain can be sync and handle cache / error
  getChain() {
    const chain = this.chainCache[this.url];
    if (!chain) throw new Error('Provider is not connected, please connect() it first.');
    return chain;
  }
```

packages/providers/src/provider.ts
```diff
@@ -205,28 +207,75 @@ export type ProviderCallParams = {
...
+ */
+type ConsensusParamsCache = Record<string, ConsensusParameters>;
+
 /**
  * A provider for connecting to a node
  */
 export default class Provider {
   operations: ReturnType<typeof getOperationsSdk>;
   cache?: MemoryCache;
+  consensusParamsCache: ConsensusParamsCache = {};
```
LuizAsFight, line 221: instead of caching `consensusParams` , we can anticipate the needings and cache whole `chainInfo` , as we have it in hands.

we can also create same approach and for `nodeInfo`, from `provider.getNodeInfo()`

packages/providers/src/provider.ts
```diff
@@ -240,6 +289,17 @@ export default class Provider {
...
+   * A method to get the consensus parameters for a given node URL when we don't have access to an instance of the Provider class.
+   */
+  private static async getConsensusParamsWithoutInstance(url: string) {
+    const gqlClient = new GraphQLClient(url);
+    const operations = getOperationsSdk(gqlClient);
+    const { chain } = await operations.getChain();
+    const { consensusParameters } = processGqlChain(chain);
+    return consensusParameters;
+  }
+
```
LuizAsFight, line 302: maybe this method can be removed as:

- to the purpose of getting params the user can: 
```
const provider = await Provider.create(url);
const chain = provider.getChain();
```

packages/providers/src/provider.ts
```diff
@@ -205,28 +207,75 @@ export type ProviderCallParams = {
...
 
+/**
+ * URL - Consensus Params mapping.
+ */
+type ConsensusParamsCache = Record<string, ConsensusParameters>;
+
 /**
  * A provider for connecting to a node
  */
 export default class Provider {
```
LuizAsFight, line 218: I see the provider has a `processNodeInfoAndConsensusParameters` 

it should be split in 2 functions that do its part completely and independently:

`processNodeInfo`
`processChain`

then will be used to cache the correct data

LuizAsFight: would also be nice to add a `getGasConfig` helper method in provider, that would be used like:
```
const provider = await Provider.create(url);
const gasConfig = provider.getGasConfig();

// now use correct config to create things with correct default gasConfig
const scriptRequest = new ScriptTransactionRequest({ gasPrice: gasConfig.minGasPrice, gasLimit: maxGasPerTx });
```

implementation like: 
```
getGasConfig() {
  // get cached info
  const chain = this.getChain();
  const nodeInfo = this.getNodeInfo();

  const { minGasPrice } = nodeInfo;
  const { maxGasPerTx, maxGasPerPredicate, gasPriceFactor, gasPerByte } = chain;

  // return relevant gas info
  return {
    minGasPrice, maxGasPerTx, maxGasPerPredicate, gasPriceFactor, gasPerByte
  }
}
```

LuizAsFight (changes requested):

packages/providers/src/provider.ts
```diff
@@ -240,6 +289,17 @@ export default class Provider {
     return getOperationsSdk(gqlClient);
```
nedsalk, line 289: Could you have done the caching on the level of the request by adding `requestMiddleware` to `GraphQLClient`? `requestMiddleware` gets called on every request the client makes and it seems like the ideal place to insert caching.

```ts
  private createOperations(url: string, options: ProviderOptions = {}) {
    this.url = url;

    const gqlClient = new GraphQLClient(url, {
      requestMiddleware: async (request) => {
        if (!this.consensusParamsCache[url]) {
          const consensusParameters = await Provider.getConsensusParamsWithoutInstance(url);
          this.consensusParamsCache[url] = consensusParameters;
        }
        return request;
      },
      fetch: options.fetch,
    });
    return getOperationsSdk(gqlClient);
  }
```

This way there wouldn't be any breaking changes to the public API of the `Provider` class as far as I can see, nor would there be changes in how to instantiate it.
The caching would happen on first **real** request instead of during instantiation of the `Provider`, which kind of makes sense because why would the provider fetch and cache consensus parameters for a node if it won't be contacting the node to do some actual querying/mutations/subscriptions?
Also, `createOperations` is called whenever `url` is changed, so we'd be sure that the first call to the new `url` would update the cache as well.


If somebody is actually interested in the consensus parameters, it can be put behind a method:

```ts
class Provider {
...
#consensusParametersCache: Record<string, ConsensusParameters | undefined>;

public async getConsensusParameters(): Promise<ConsensusParameters> {
    const cached = this.consensusParametersCache[this.url];
    if(cached) return cached;
    const consensusParameters = await Provider.getConsensusParamsWithoutInstance(this.url);
    this.consensusParametersCache[this.url] = consensusParameters;
    return consensusParameters;
  }
}
```

packages/providers/src/provider.ts
Dhaiwat10: @LuizAsFight the type coming directly from graphql has all the fields with the type `string`. This is why I am using the types returned from `procesGqlChain`'s results - `ChainInfo`: https://github.com/FuelLabs/fuels-ts/blob/ff8c4dfeafd23744f2a6f747910975540cd381cd/packages/providers/src/provider.ts#L70

packages/providers/src/provider.ts
Dhaiwat10: @LuizAsFight actually this method is for internal use only and cannot be accessed by users. It is in place because inside of the new `connect` method, we need to fetch the consensus parameters _before_ calling the `Provider` constructor.

packages/providers/src/provider.ts
Dhaiwat10: Great point!

packages/providers/src/provider.ts
Dhaiwat10: This is very interesting @nedsalk. I'm intrigued. I'm gonna have a go at implementing this. Tagging the rest of the team so that they can have a look and share their opinions too cc @FuelLabs/sdk-ts

packages/providers/src/provider.ts
LuizAsFight: yeah I understand I just don't see the use of it after applying the other comments I mentioned

if we can keep this logic only in `Provider.getChain` it's better. the code I provided would fit the needs of this method
```
const provider = await Provider.create(url);
const chain = provider.getChain();
```

LuizAsFight: > It is in place because inside of the new `connect` method, we need to fetch the consensus parameters _before_ calling the `Provider` constructor.

you don't need if your static **create** function looks like this:
```
  static async create({ url, options }: { url: string, options?: ProviderOptions }) {
    const provider = new Provider(url, options);
    await provider.connect();
    return provider;
  }
```

packages/program/src/functions/base-invocation-scope.ts
LuizAsFight: @arboleya should be something like

```
  static async create({ url, options }: { url: string, options?: ProviderOptions }) {
    const provider = new Provider(url, options);
    await provider.connect();
    return provider;
  }
  ```

File: packages/providers/src/provider.ts (hunk 21 above)
Comment (LuizAsFight) [reply packages/providers/src/provider.ts:251]: > I like this idea @LuizAsFight, but wouldn't this _still_ be a breaking change for most users? For example, anyone who has implemented any contract calls will need to update their code and add `await provider.create()` everywhere.

 yeah about this I'm ok with changing to remove `new Provider` and always use `Provider.create` if you prefer on that way

Comment (LuizAsFight) [reply packages/providers/src/provider.ts:251]: keeping `connect` function as I mentioned in other comments instead of changing to `updateUrl` will also help with breaking changes

File: packages/providers/src/provider.ts (hunk 28 above)
Comment (Dhaiwat10) [reply packages/providers/src/provider.ts:289]: @nedsalk I have an update on this. Like I mentioned in the weekly sync on Monday, this approach is really smart but will not work for us as of now. This is due to the fact that we need the consensus parameters when we _prepare_ a transaction and not only when we _send_ it. We only call the GQL API to _send_ the transaction. Hence, with the approach that you suggested, we won't be able to fetch the consensus parameters in time for things to work. Maybe we can make it work with a huge refactor and a re-design of the whole architecture (this would be possible in the Rust SDK, for example), but that would be outside the scope of this PR. If you want, we can still track this idea in this issue - #1103  to revisit later on.

**Comment by Dhaiwat10** [comment]
Marking this as a draft briefly while I make some adjustments and put the PR in a better place to be reviewed.

File: packages/providers/src/provider.ts (hunk 28 above)
Comment (nedsalk) [reply packages/providers/src/provider.ts:289]: Thanks for the investigation and the explanation!

Given that this PR changes the way we create a `Provider` seemingly just so that we can satisfy the design that we find problematic, I think that we should put it on hold and focus on fixing that transaction design. If this PR was just introducing an internal change, I wouldn't raise this issue; however, this is introducing a breaking change but its purpose is lost because the consensus parameters we're looking to cache in the end aren't even used due to us fixating the values on transaction creation. 
The consensus parameters passed to a transaction on creation (be they cached or hardcoded) can become different to the latest ones the `Provider` has because X amount of time can pass between transaction creation and sending - and there is no way to change them as they're set _on creation_. When a transaction can be given consensus parameters the moment before it's sent instead of on its creation (e.g. by modifying `Provider.sendTransaction` to pass the consensus parameters to it and calculate the offsets at that moment), then it would make sense to revisit this issue and PR. Also, this PR is concerned with caching - meaning that everything works even without it but is just slower - so I don't see a big issue in postponing it.

@FuelLabs/sdk-ts please chime in.

File: packages/providers/src/provider.ts (hunk 27 above)
Comment (Dhaiwat10) [reply packages/providers/src/provider.ts:218]: @LuizAsFight this is a great suggestion and I will get this delivered to you ASAP. But I do think we can create a separate PR for this? Does that sound OK?

File: packages/providers/src/provider.ts (hunk 21 above)
Comment (Dhaiwat10) [reply packages/providers/src/provider.ts]: @LuizAsFight Tysm for your detailed review of the PR and so many helpful suggestions mate! After careful consideration of all of them and some discussions with the team, I have incorporated some of them but have decided to keep some things as they were. The team has agreed that it is better to have this *loud* breaking change instead of keeping it _silent_ and leaving users with some code that might seem ok but actually needs some additions in order to work now. That itself would be a breaking change, and potentially even dangerous.

File: packages/providers/src/provider.ts (hunk 28 above)
Comment (Dhaiwat10) [reply packages/providers/src/provider.ts]: @nedsalk Although I completely see your point and I used to share this opinion a few weeks ago, at the moment I disagree mate. I think it will slow us down *a lot*. A massive architectural/design change like that will have to be extremely well thought-out. We need to ensure nothing breaks. We need this fix to go out for beta-4 and we don't have enough time to carefully craft a complete re-design.

Also, the architecture that this PR follows actually brings us closer to the Rust SDK. Which is why I believe this PR still takes us in a overall positive direction.

File: packages/providers/test/provider.test.ts [hunk 29]
```diff
@@ -651,4 +667,26 @@ describe('Provider', () => {
     );
     expect(messageProof).toMatchSnapshot();
   });
+
+  it('connect method', async () => {
```
danielbate, line 671:
```suggestion
  it('can connect', async () => {
```
Nit

packages/providers/test/provider.test.ts
```diff
@@ -651,4 +667,26 @@ describe('Provider', () => {
...
+    const provider = await Provider.connect('http://127.0.0.1:4000/graphql');
+
+    // check if the provider was initialized properly
+    expect(provider).toBeInstanceOf(Provider);
+    expect(provider.url).toEqual('http://127.0.0.1:4000/graphql');
+    expect(provider.chainInfoCache['http://127.0.0.1:4000/graphql']).toBeDefined();
+  });
+
+  it('invalidChainInfoCache', async () => {
```
danielbate, line 680:
```suggestion
  it('can invalidate the chain info cache', async () => {
```
Nit

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -73,23 +73,29 @@ export class BaseInvocationScope<TReturn = any> {
...
    */
   protected get calls() {
+    const script = getContractCallScript(this.functionInvocationScopes);
+    const provider = this.program.provider as Provider;
+    const consensusParams = provider.getCachedChainInfo().consensusParameters;
+    if (!consensusParams) {
+      throw new Error(
+        'Provider chain info cache is empty. Please make sure to initialize the `Provider` properly by running `await Provider.connect()``'
+      );
+    }
```
danielbate, line 83: Should we be using the `@fuel-ts/errors` package now for top level throws?

packages/script/src/script-invocation-scope.ts
```diff
@@ -25,13 +26,21 @@ export class ScriptInvocationScope<
 
   private buildScriptRequest() {
     const programBytes = (this.program as AbstractScript).bytes;
+    const chainInfoCache = (this.program.provider as Provider).getCachedChainInfo();
+
+    if (!chainInfoCache) {
+      throw new Error('Provider must have chain info cache.');
```
danielbate, line 32: As above regarding `@fuel-ts/errors`

packages/wallet-manager/package.json
```diff
@@ -27,7 +27,8 @@
     "@fuel-ts/crypto": "workspace:*",
     "@fuel-ts/mnemonic": "workspace:*",
     "@fuel-ts/wallet": "workspace:*",
-    "events": "^3.3.0"
+    "events": "^3.3.0",
+    "@fuel-ts/providers": "workspace:*"
```
danielbate, line 31:
```suggestion
    "@fuel-ts/providers": "workspace:*",
    "events": "^3.3.0"
```
Nit sry

danielbate (commented):
Great work Dhai, the additional context in the PR description was also appreciated.

Added some comments and a fix for the skipped test.

packages/wallet-manager/src/vaults/privatekey-vault.ts
```diff
@@ -19,33 +22,41 @@ export class PrivateKeyVault implements Vault<PkVaultOptions> {
...
     }
   }
 
   serialize(): PkVaultOptions {
     return {
       accounts: this.#privateKeys,
+      provider: this.provider,
```
Torres-ssf, line 40: Why do we need to add the provider when serializing a wallet?

packages/wallet/src/account.ts
```diff
@@ -66,16 +65,8 @@ export class Account extends AbstractAccount {
...
-  connect(provider: string | Provider): Provider {
-    if (typeof provider === 'string') {
-      if (this.provider) {
-        this.provider.connect(provider);
-      } else {
-        this.provider = new Provider(provider);
-      }
-    } else {
-      this.provider = provider;
-    }
```
Torres-ssf, line 78: :rocket:

Torres-ssf (approved):
LGTM :+1:

packages/providers/src/provider.ts
nedsalk: Sounds good then!

apps/docs-snippets/src/guide/contracts/calls-with-different-wallets.test.ts
```diff
@@ -23,8 +26,13 @@ describe(__filename, () => {
...
 
-  it('should successfully update contract instance provider', () => {
-    const newProvider = new Provider('http://provider:9999');
+  /**
+   * This test is currently skipped because Provider.connect will throw an error for the dummy url.
+   * Locally, we only have one provider, so we can't test this.
+   * TODO: Figure out how to test this.
+   */
+  it.skip('should successfully update contract instance provider', async () => {
+    const newProvider = await Provider.connect('http://provider:9999');
```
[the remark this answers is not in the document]
Dhaiwat10: @danielbate this worked, thanks mate! I think we can keep it this way. Shouldn't be too much of an issue 🤔

packages/wallet-manager/src/vaults/privatekey-vault.ts
Dhaiwat10: @Torres-ssf do you mean why are we returning it?

Torres-ssf (approved):

apps/docs-snippets/src/guide/types/bits512.test.ts
```diff
@@ -18,10 +18,14 @@ describe(__filename, () => {
     // #context }
     // #endregion bits512-1
 
+    const provider = await contract.provider;
```
arboleya, line 21: The same question about this `await`.

apps/docs/src/getting-started.md
```diff
@@ -35,13 +35,14 @@ import { useEffect, useState } from "react";
...
 
-  const provider = new Provider("https://beta-3.fuel.network/graphql");
-  const myWallet = Wallet.fromAddress("0x...", provider);
-
   useEffect(() => {
-    myWallet.getBalances().then((data) => {
-      setBalance(new BN(data[0].amount).toNumber());
-    });
+    async () => {
+      const provider = await Provider.connect("https://beta-3.fuel.network/graphql");
```
arboleya, line 40: Should this point to the beta-4 url?

apps/docs/src/index.md
```diff
@@ -61,10 +61,11 @@ console.log(Wallet.fromPrivateKey(PRIVATE_KEY));
 <!-- TODO: stop using hardcoded snippets -->
 
 ```ts:line-numbers
-import { Wallet, Contract, BigNumberish, BN } from "fuels";
+import { Provider, Wallet, Contract, BigNumberish, BN } from "fuels";
 import abi from "./abi.json";
 
-const wallet = Wallet.fromPrivateKey(PRIVATE_KEY); // private key with coins
+const provider = await Provider.connect('https://beta-3.fuel.network/graphql');
```
arboleya, line 67: Another possible URL to point to beta-4.

There might be others.

packages/fuel-gauge/src/contract-factory.test.ts
```diff
@@ -7,7 +7,7 @@ import storageSlots from '../fixtures/forc-projects/storage-test-contract/out/de
 
 describe('Contract Factory', () => {
   const createContractFactory = async () => {
-    const provider = new Provider('http://127.0.0.1:4000/graphql');
+    const provider = await Provider.connect('http://127.0.0.1:4000/graphql');
```
arboleya, line 10: Should the `FUEL_NETWORK_URL` constant replace this URL?

Several other places are using hardcoded strings.

> This can be another PR though.

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -75,8 +74,18 @@ export class BaseInvocationScope<TReturn = any> {
    * @returns An array of contract calls.
    */
   protected get calls() {
+    const script = getContractCallScript(this.functionInvocationScopes);
+    const provider = this.program.provider as Provider;
```
arboleya, line 78: Why is this `as Provider` cast necessary?

packages/wallet-manager/src/vaults/privatekey-vault.ts
arboleya: Yeah, I'm with @Torres-ssf on this one.

I am trying to understand why we must return the provider in this method.

arboleya (changes requested):

arboleya:
Also, don't forget to add `- close #NNN` tags for issues #1050 and #1176 on the PR description.

packages/fuel-gauge/src/contract-factory.test.ts
Dhaiwat10: @arboleya Yeah, this is indeed a problem but one that is prevalent throughout our codebase. I will file a follow-up for this and get this fixed in a separate PR.

apps/docs-snippets/src/guide/contracts/simulate-transactions.test.ts
```diff
@@ -13,7 +13,11 @@ describe(__filename, () => {
     const amountToForward = 40;
     const amountToTransfer = 10;
 
-    const someAddress = Wallet.generate().address.toB256();
+    const provider = await contract.provider;
```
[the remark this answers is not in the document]
Dhaiwat10: Good catch, it's a side-effect of find-and-replace 😅

packages/program/src/functions/base-invocation-scope.ts
Dhaiwat10: @arboleya this cast is necessary because the `AbstractProgram` class has its own custom definition of the `provider` field, which only includes the `sendTransaction` method and nothing else:

https://github.com/FuelLabs/fuels-ts/blob/5310e8b5226a74058b85a6f1ca3671b6b5b5c5b7/packages/interfaces/src/index.ts#L63-L66

I believe this might be to avoid cyclic dependencies with the `providers` package.

arboleya: I see; maybe an `AbstractProvider` could be added to the `interfaces` package to solve this?

> I'm not a fan of using abstract classes. TBH, I'm against _[the vast majority]_ of abstractions, but it is what it is. I reckon the `interfaces` package only exists for circumventing cyclic dependencies — something to consider.

packages/providers/test/provider.test.ts
```diff
@@ -281,29 +296,30 @@ describe('Provider', () => {
...
   });
 
-  it('can cacheUtxo [invalid numerical]', () => {
-    expect(() => new Provider('http://127.0.0.1:4000/graphql', { cacheUtxo: -500 })).toThrow(
-      'Invalid TTL: -500. Use a value greater than zero.'
+  it('can cacheUtxo [invalid numerical]', async () => {
+    const { error } = await safeExec(() =>
+      Provider.connect('http://127.0.0.1:4000/graphql', { cacheUtxo: -500 })
     );
+    expect(error?.message).toMatch(/Invalid TTL: -500\. Use a value greater than zero/);
```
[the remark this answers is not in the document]
Dhaiwat10: @arboleya makes sense, but I don't think this was added in this PR. It just shows up in the changed files because I had to change that test to accomodate the new Provider API.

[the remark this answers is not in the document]
arboleya: That's fine. We can do another check as we go.

packages/program/src/functions/base-invocation-scope.ts
Dhaiwat10: @arboleya Fair, I'll file a follow-up for this too and fix it in a separate PR

packages/wallet-manager/src/vaults/privatekey-vault.ts
Dhaiwat10: @arboleya @Torres-ssf this is because we need a `provider` to create a `PrivateKeyVault` now. It used to be optional. I actually tried omitting `provider` from the returned values by `serialize`, but it quickly resulted in usages like this one here breaking:

image


What do you guys suggest?

arboleya: Ok, then it's necessary. 👍

Dhaiwat10:
@arboleya I think we can close #1176 but not #1050 quite yet. I believe #1050 should be closed once we eliminate the usage of _all_ the hardcoded transaction parameter values. This PR only replaces `VM_TX_MEMORY` as a proof of concept, but doesn't replace all of them. Maybe we should maintain a list of all the hardcoded values in #1050, and then tick them off one-by-one.

packages/program/src/functions/base-invocation-scope.ts
arboleya: - https://github.com/FuelLabs/fuels-ts/issues/1257

packages/fuel-gauge/src/contract-factory.test.ts
arboleya: - https://github.com/FuelLabs/fuels-ts/issues/1258

arboleya (dismissed):
Thanks for all the adjustments. :shipit:

packages/providers/src/provider.ts
Dhaiwat10: Follow-ups: #1259, #1260

Torres-ssf (dismissed):

LuizAsFight (approved):

LuizAsFight (dismissed):
some of my comments were not implemented. I see a follow up issues but there's also other things not covered on the issues

here's a suggestion of how `provider.ts` would look, you can copy/paste over yours, then check the git diff changes to understand better
https://gist.github.com/LuizAsFight/9c9d35c59e85ed08cef58dc6f4a99fee

feel free to dismiss my change requested if the team disagrees with it OR if you guys will do it in next PRs

cc @Torres-ssf @arboleya

Dhaiwat10:
@LuizAsFight thanks so much for reviewing again and sharing that gist mate. 🙏🏻  

Two things:

1. We opted against going with the approach you suggested with the `create` method and avoiding a breaking change on the constructor itself. We opted against this because this won't break existing code for users at compile-time since there are no changes to the constructor's function signature. This will _still_ break their code at run-time, though. For such a significant breaking change, we think this is less ideal than just changing the function signature and signalling that something has changed. If the approach you suggested was avoiding a breaking change altogether, we would go with it but in this case it looks like a breaking change is unavoidable either way - it will just be disguised under something else. We think it's better to just loudly throw and break things now rather than later in this case.

2. I also saw that you added a helper for gas config. I have already filed a follow-up issue for this and we'll get this added ASAP.

Dhaiwat10:
TLDR since you guys last reviewed:

- Made the `Provider` constructor private
- Removed `chainInfo` from the constructor
- Renamed `connect` to `create` (h/t @LuizAsFight 🙏🏻)
- Deleted `refreshChainInfoCache`, `getChainInfoWithoutInstance`, `getCachedChainInfo`
- Added a `nodeInfoCache` on the `Provider` class
- Added `fetchChain`, `fetchNode`, `getNode`
- Refactored `getChain`
- Added `getGasConfig` (h/t @LuizAsFight @Torres-ssf 🙏🏻)
- Removed `getNodeInfo` and the `getInfo` GQL query as it's not needed anymore. I replaced this with a clearer query for `nodeInfo`

arboleya (commented):
1. I think it's worth doing a final search/replace for removing the residual hardcoded `http://127.0.0.1:4000/graphql` addresses in favor of the `FUEL_NETWORK_URL` constant.

2. We should not postpone the resolution of the broken tests since it can hide critical problems with the taken approach. Addressing and uncommenting those should also fix the failing checks of the next point below.

3. There are a couple of failing checks here (the last ones), probably due to the commented-out tests mentioned above:
   - https://github.com/FuelLabs/fuels-ts/actions/runs/6232185591<br/><br/>

   <details>
     <summary>screenshot</summary>

   checks

   </details>

Torres-ssf (dismissed):

arboleya (approved):

arboleya (approved):

Torres-ssf (approved):

PR #1182, opened 2023-08-10
feat: implement playwright for browser compatibility testing

### LINKED ISSUE #284: Create test to `browser compatibility`

danielbate:
Proof of concept implementation of browser compatibility testing for #284


packages/merkle/src/binary/binaryMerkleTree.node.browser.node.browser.test.ts
Torres-ssf, line 1: it seems `node.browser` has being used 2 times on this test file name

packages/merkle/src/sparse/sparseMerkleTree.node.browser.node.browser.test.ts
Torres-ssf, line 1: it seems `node.browser` has being used 2 times on this test file name

packages/merkle/src/sum/sumMerkleTree.node.browser.node.browser.test.ts
Torres-ssf, line 1: it seems `node.browser` has being used 2 times on this test file name

scripts/merge-coverage.ts
```diff
@@ -0,0 +1,23 @@
+import { readdirSync, renameSync, rmSync } from 'fs';
+import { join } from 'path';
+
+const coverageDir = join(__dirname, '../coverage/');
+const environmentsDir = join(coverageDir, '/environments/');
+const validEnvironments = ['browser', 'node'];
+
+const main = () => {
+  const environments = readdirSync(environmentsDir);
+  environments.forEach(async (environment) => {
```
arboleya, line 10: If we stop `await`ing inside the inner method.

```suggestion
  environments.forEach((environment) => {
```

jest-playwright.config.ts
```diff
@@ -0,0 +1,10 @@
+const config = {
+  browsers: ['chromium'],
+  collectCoverage: false,
```
arboleya, line 3: Should `collectCoverage` be true?

package.json
```diff
@@ -14,8 +14,16 @@
     "build": "turbo run build",
     "ci:test": "./scripts/ci-test.sh",
     "pretest": "turbo run pretest",
-    "test": "jest --no-cache --runInBand --coverage",
-    "test:watch": "jest --no-cache --watchAll",
+    "test": "jest --no-cache --runInBand --config './jest.all.config.ts'",
```
Torres-ssf, line 17: When I run `pnpm ci:test` the `.browser` test files are ignored. Also, all `.node.browser` test files are running only on the `node` environment. Is this the expected behavior? Are we going to run all our tests on both environments? Or do we aim to only run some specific tests on the browser?

Running `pnpm test packages/{packageName}` runs the `.browser` tests but the `.node.browser` tests seem to be running only on the  `node` environment.

packages/merkle/src/binary/binaryMerkleTree.node.browser.node.browser.test.ts
Torres-ssf, line 1: it seems `node.browser` has been used 2 times on this test file name

packages/merkle/src/sparse/sparseMerkleTree.node.browser.node.browser.test.ts
Torres-ssf, line 1: it seems `node.browser` has been used 2 times on this test file name

packages/merkle/src/sum/sumMerkleTree.node.browser.node.browser.test.ts
Torres-ssf, line 1: it seems `node.browser` has been used 2 times on this test file name

Torres-ssf (changes requested):
I left some notes but I am still going through the review process

packages/merkle/src/binary/binaryMerkleTree.node.browser.node.browser.test.ts
danielbate: Good spot

package.json
danielbate: These are good points and probably require some additional description _somewhere_. I'll definitely add to the PR description but maybe even in the contributing section of docs. But potentially the commands need a rethink.

1. `pnpm ci:test` default behaviour will only run the `node` tests. It can also run the `browser` tests with `pnpm ci:test --browser` as seen in the altered GitHub action:
https://github.com/FuelLabs/fuels-ts/blob/4a4bda6698cd9279a8479576234b305d6620dfe9/.github/workflows/test.yaml#L45
2. So the file path suffix allows it to be picked up by a config. This way a test can run on multiple environments. I believe there are many tests that we want to run across environments, therefore have suffixed them with `...node.browser.test.ts`. Anything that we only require on browser, should just be `...browser.test.ts`, similarly with `...node.test.ts`.
3. `pnpm test` is currently using `jest.all.config.ts' configuration which should be running all tests regardless of environment suffix, and run them using `node`. Which is the previous behaviour of `pnpm test`. However if it is not picking up certain tests, I will investigate that.

jest-playwright.config.ts
danielbate: I've used jest's own coverage collection configuration rather than `NYC`. We are just using `NYC` for merging and reporting.

package.json
arboleya: I think `test:ci` should default to all tests, like the `pnpm test` command.

Then, we could have the option to run only a group in isolation.

```sh
pnpm ci:test # run all tests
pnpm ci:test --browser # only browser tests
pnpm ci:test --node # only node tests
```

.github/workflows/test.yaml
```diff
@@ -34,37 +34,49 @@ jobs:
...
+        run: pnpm test:validate
+
       - name: Find PR Number
         uses: jwalton/gh-find-current-pr@v1
         id: findPr
 
       - name: Run tests and collect coverage
-        if: ${{ !steps.findPr.outputs.number }}
-        run: pnpm ci:test -- --json --coverage --testLocationInResults --outputFile=report.master.json
+        run: pnpm ci:test && pnpm ci:test --browser && pnpm test:merge-coverage && pnpm test:report-coverage
```
arboleya, line 45: To save CI time, we should run `--browser` and `--node` tests in parallel I CI.

Then, we'd have a third job awaiting the other two to finish and merge their coverage afterward.

I remember @Dhaiwat10 started this implementation on his first draft for this.

jest-playwright.config.ts
arboleya: Oh, can `jest` collect **_browser_** coverage now?

In the past, when using `mocha`, I needed Istanbul/NYC to instrument the compiled [js] files _before_ sending them to the browser to collect coverage back. It wasn't possible otherwise.

I'm wondering why `jest-playwright` would expose an option to [collect coverage](https://github.com/playwright-community/jest-playwright#tracking-the-coverage) this way, via Istanbul instrumentation, if that was possible already with `jest` alone?

danielbate: Apologies at first look I believed this referred to `NYC`. So jest does have [Istanbul integrated](https://istanbul.js.org/docs/tutorials/jest/) for coverage collection / reporting hence not needing the additional configuration that mocha does, and there are references in our deps to `jest-babel` that would be required for the transforms, however I need to have a bit of a further dig as to why that is a requirement. As currently I am seeing no difference in our own reporting if that flag is flipped...

.github/workflows/test.yaml
arboleya: Also, in @Dhaiwat10 first draft, we had 4x groups that could be run simultaneously while ensuring that all e2es (browser x node) wouldn't interfere with one another:

 - `browser:unit`
 - `browser:e2e` — requires `fuel-core` running
 - `node:unit`
 - `node:e2e` — requires `fuel-core` running

The full spec can be seen in the first draft PR:
 - https://github.com/FuelLabs/fuels-ts/pull/728

package.json
danielbate: @arboleya @Torres-ssf Would you expect `pnpm ci:test` to run all tests as node, or all the tests in both environments, even if a test requires both environments?

danielbate:
Thanks for the thoughts so far @FuelLabs/sdk-ts - going to be converting to draft to amend a few points raised on call
- Change approach to grouping so we don't use the file name and required tests are loaded in
- Create a test to confirm the node / browser functionality
- Change `pnpm:test` to run node and browser tests sequentially and remove the all as node configuration
- Resolve the coverage workflow

danielbate:
I've done a more complete writeup of progress so far in https://github.com/FuelLabs/fuels-ts/issues/284#issuecomment-1699072679

danielbate:
Closing as playwright was found to not be a suitable alternative as has been superseded by the investigation in #1310.  However the approach to grouping, coverage merging and CI can be taken across.

PR #1183, opened 2023-08-14
feat!: improve vector output support for contracts

### LINKED ISSUE #933: Improve Vector output support with `Vec<>` return

camsjams:
This adds additional level of support, see details in #933 

#### Previous Vector support
Currently (on master) we support:
- u8
- u16
- u32
- u64

See also follow up for bytes: https://github.com/FuelLabs/fuels-ts/issues/1188
See also follow up for Multicalls with Vector output: https://github.com/FuelLabs/fuels-ts/issues/1189 / https://github.com/FuelLabs/fuels-ts/issues/1192

Closes #933

### DIFF
--- a/packages/abi-coder/src/function-fragment.ts
+++ b/packages/abi-coder/src/function-fragment.ts
@@ -85,12 +85,16 @@ export class FunctionFragment<
   }
 
   #getOutputEncodedLength(): number {
-    const heapCoder = AbiCoder.getCoder(this.jsonAbi, this.jsonFn.output);
-    if (heapCoder instanceof VecCoder) {
-      return heapCoder.coder.encodedLength;
-    }
+    try {
+      const heapCoder = AbiCoder.getCoder(this.jsonAbi, this.jsonFn.output);
+      if (heapCoder instanceof VecCoder) {
+        return heapCoder.coder.encodedLength;
+      }
 
-    return heapCoder.encodedLength;
+      return heapCoder.encodedLength;
+    } catch (e) {
+      return 0;
+    }
   }
 
   encodeArguments(values: InputValue[], offset = 0): Uint8Array {
--- a/packages/fuel-gauge/src/contract.test.ts
+++ b/packages/fuel-gauge/src/contract.test.ts
@@ -172,7 +172,7 @@ describe('Contract', () => {
     }
 
     expect(spy).toHaveBeenCalled();
-    expect(interfaceSpy).not.toHaveBeenCalled();
+    expect(interfaceSpy).toHaveBeenCalled();
   });
 
   it('generates function methods on a complex contract', async () => {
@@ -193,7 +193,7 @@ describe('Contract', () => {
     }
 
     expect(spy).toHaveBeenCalled();
-    expect(interfaceSpy).not.toHaveBeenCalled();
+    expect(interfaceSpy).toHaveBeenCalled();
   });
 
   it('assigns a provider if passed', () => {


nedsalk (changes requested):
I haven't analyzed the whole PR, but I'll have to request changes because the PR didn't touch `interface.test.ts`. I understand that the logic is tested via `fuel-gauge` tests, but we need to verify on the `Interface` level that decoding outputs works.

However, the way `interface.test.ts`'s `encoding/decoding` testing is currently set up is not ideal, as it is using `FunctionFragment.decodeArguments()` instead of `FunctionFragment.decodeOutput()`. Because of that, I have created [this PR](https://github.com/FuelLabs/fuels-ts/pull/1191) where I've updated the `exhaustive-examples` sway project to always return the inputs of a function (either directly, or if it's multiple arguments then it returns a tuple). Afterwards, I've updated the test function accordingly to use `decodeOutput`, and I've removed the skipping of vector decoding as this PR is for that purpose. Please take a look at the PR where I've gone into more depth about my findings, as some vector decoding tests are passing, but some are failing.

_note: I wrote a detailed comment on that PR but exited the tab. I'll link the PRs and incrementally edit/save the PR. :(_

nedsalk:
> They don't prove the code works for these changes, if I deleted a change in one of the other packages, the interface tests would still pass

As is the case with all unit/feature tests in the codebase that are testing how an utility / core package export works. 

> The fuel-gauge tests are exhaustive for this feature addition

I've updated the PR to add more tests to the `fuel-gauge` package. These test cases are the same as the ones failing in the `interface.test.ts` file and they are failing in the end-to-end tests as well.

> Interface tests are isolated to the AbiCoder package which is not where the bulk of the changes occurred

One **core** change occurred in the abi-coder package: vectors are now decodeable. On that note, the `interface.test.ts` is the _all-encompassing_ test file that's testing the guarantees of the abi-coder package from a "public API" point of view, and it was previously skipping decoding vectors because support was lacking. I've written my PR to draw attention to that fact, as now you've implemented `VecCode.decode`, which means that vector decoding now works and all "full usage" test scenarios on the level of `Interface`/`FunctionFragment` should be working. The test that you've written in `vec.test.ts` is not enough to guarantee decoding of any arbitrary vector output from any ABI function that is given to `Interface` via `new Interface(abi)`.

> This is existing tech debt and shouldn't be piled onto my PR

Look at it this way: if I had created a separate sway project with the failing use cases and a separate test file that was specifically testing with `FunctionFragment.decodeOutput` and it was using this tailor-made sway project's ABI, would you call it tech debt? The only additional thing I did was refactor the existing tests to use `decodeOutput` instead of `decodeArguments`, but they're both using `AbiCoder.decode` underneath, which is calling `VecCoder.decode`, which is one of the core changes in this PR.

> For readability of the reviewers, we should add tests in a single PR that could encompass Vectors and Bytes (which are upcoming) so that the focus here is on new functionality

Sorry, I don't quite get this part fully. However, _new_ functionality was introduced in `VecCoder`, and it should be appropriately tested in the package where it was introduced. The appropriate way to do it is via `interface.test.ts` tests, where the tests were already existing but were being skipped.

> IMO PR etiquette requires that changes should be for issues created in a new PR, not solving existing issues ... within reason :)

I believe my "solving existing issues" was within reason. I just refactored a test to be closer to real life usage (by using `decodeOutputs` instead of `decodeArguments`) and pointed out that the previously-skipped decoding tests are failing. If you want, I can revert the changes to that specific parameterized test and create new tests that would essentially be doing the same thing, as described above. I can then later on deal in a separate PR with the cleanup and consolidation into the code refactoring I've introduced in my PR.

-------------------

One main thing I've noticed in your `fuel-gauge` tests is that they're all expecting the output of a function to be a vector of some sort, whereas the additional failing tests that I've added have more complex outputs like vectors in tuples, enums, structs, nested vectors. Is this by design? If it is, then I don't have a problem with the failing tests. I can update the PR to skip these complex use cases and they can be tackled in a subsequent PR.
Also, out of the 10 vector tests in `interface.test.ts` that were skipping decoding, 3 have passed after including decoding. These are when a function output is a `Vec<u8>`, `Vec<bool>`, and `[Vec<u32>; 1]`, which corresponds to my observation that simple vector output cases work as expected. However, I suspect that the last one is succeeding because it's an array with only _one_ vector. If there were more elements, I think it would fail.

packages/abi-coder/src/coders/vec.ts
```diff
@@ -44,7 +46,20 @@ export class VecCoder<TCoder extends Coder> extends Coder<
     return concatWithDynamicData(parts);
   }
 
-  decode(_data: Uint8Array, _offset: number): [DecodedValueOf<TCoder>, number] {
-    this.throwError('unexpected Vec decode', 'not implemented');
+  decode(data: Uint8Array, offset: number): [DecodedValueOf<TCoder>, number] {
+    // const ptr = data.slice(0, 8);
+    // const cap = data.slice(8, 16);
```
nedsalk, line 51: nit: remove commented-out code

packages/abi-coder/src/function-fragment.ts
```diff
@@ -66,6 +67,21 @@ export class FunctionFragment<
     return this.jsonFn.inputs.length > 1 || isPointerType(inputTypes[0]?.type || '');
   }
 
+  isOutputDataHeap(): boolean {
+    const outputType = this.jsonAbi.types.find((t) => t.typeId === this.jsonFn.output.type);
```
nedsalk, line 71: we have the `findOrThrow` utility to get rid of the `undefined` in `T | undefined`. It should throw anyways if it doesn't find the type.

packages/abi-coder/src/function-fragment.ts
```diff
@@ -66,6 +67,21 @@ export class FunctionFragment<
     return this.jsonFn.inputs.length > 1 || isPointerType(inputTypes[0]?.type || '');
   }
 
+  isOutputDataHeap(): boolean {
```
nedsalk, line 70: Just an idea:
```ts
class FunctionFragment {
...
  outputMetadata: {
    isHeapType: boolean,
    encodedLength: number
  };

  constructor(...) {
    ...
    this.outputMetadata = getOutputMetadata(abiFn.output);
  }
}
```

packages/program/src/contract-call-script.ts
```diff
@@ -91,55 +120,112 @@ const getMainCallReceipt = (
...
+            ] as TransactionResultReturnDataReceipt;
+            return concat([encodedScriptReturn, arrayify(nextReturnData.data)]);
+          }
+
+          return [encodedScriptReturn];
+        }
+
+        return [new Uint8Array()];
+      })
+      .filter((v) => v !== undefined);
```
nedsalk, line 159: All of the `return` statements in `flatMap` return an array. How is it possible for this to ever be `undefined` so that filtering has to be done for it?

packages/program/src/contract-call-script.ts
```diff
@@ -91,55 +120,112 @@ const getMainCallReceipt = (
...
+
+    totalHeapCalls += 1;
+    return (
+      sum +
+      getSingleCallInstructions(DEFAULT_OPCODE_PARAMS, {
+        isHeap: true,
+        encodedLength: call.outputEncodedLength,
+      }).byteLength()
+    );
+  }, 0);
```
nedsalk, line 190:
```suggestion
  let totalHeapCalls = contractCalls.filter(call => call.isOutputDataHeap).length;
  const heapCallsInstructionsLength = contractCalls.reduce((sum, call) => {
    if (!call.isOutputDataHeap) {
      return sum;
    }

    return (
      sum +
      getSingleCallInstructions(DEFAULT_OPCODE_PARAMS, {
        isHeap: true,
        encodedLength: call.outputEncodedLength,
      }).byteLength()
    );
  }, 0);
```

This is better because it's not modifying an external variable (`totalHeapCalls`) from `.reduce()`, thus keeping it pure.

packages/program/src/contract-call-script.ts
```diff
@@ -91,55 +120,112 @@ const getMainCallReceipt = (
...
+
+        return [new Uint8Array()];
+      })
+      .filter((v) => v !== undefined);
+  };
 
 export const decodeContractCallScriptResult = (
   callResult: CallResult,
   contractId: AbstractAddress,
+  isOutputDataHeap = false,
```
nedsalk, line 165: It feels weird to me to assign a default value to `isOutputDataHeap`. Why would `false` be the default and not `true`? If we take it an abstraction layer above and say that the signature of the function instead is:

```ts
export const decodeContractCallScriptResult = (
    callResult: CallResult,
    contractId: AbstractAddress,
    functionFragment: FunctionFragment,
): Uint8Array[] => {...}
```

we wouldn't be able to assign a default value to `FunctionFragment` the same way it was done to `isOutputDataHeap`... but the `isOutputDataHeap` value is actually received from `FunctionFragment` in the `InvocationResult.getDecodedValue` method.

It seems to me that the signature with `FunctionFragment` is truer to the use case, as `FunctionFragment` itself encapsulates the actual abi function based on which the decoding is happening.

nedsalk (approved):

nedsalk (commented):
The things that I understand look good to me. I've just added some questions, suggested changes and ideas. I can't comment much on the usage of `fuel-asm` as it's not fully familiar to me, but it obviously works given that the tests are passing :).

Dhaiwat10 (dismissed):

nedsalk (approved):

packages/program/src/contract-call-script.ts
nedsalk: I meant more conceptually, IMO this information isn't something that `decodeContractCallScriptResult` should assert to have a default value, as it's working with `FunctionFragment`s (directly if my suggestion is taken or indirectly as it currently is in `InvocationResult.getDecodedValue`) who, in the end, are owners of that information. It should just say, "give me a `FunctionFragment` (or a boolean indicating that the output data type is a heap type), and I'll decode the result." If we left it as it is, one could provide a `callResult` where the data is actually a heap type and use the default `isOutputDataHeap = false` value, and the decoding would fail.  Default values are for situations where a function will work as expected both when the user provided their own value and when they used the default one (e.g. an `options` parameter).

nedsalk (approved):

Dhaiwat10 (approved):

PR #1193, opened 2023-08-15
chore: splitting utilities in sync/async

arboleya:
This PR is an experiment splitting some utilities into sync/async versions.

We have two versions to look at:

 - **Hybrid**: Having a method that can behave sync/async depending on the received arguments
   - _This is the former implementation from #1153._
 - **Split**: Having two methods _(1x sync, 1x async)_, to be used independently
   - _This is the new angle suggested in this PR_

In this case, the **_split_** is proposed to replace the **_hybrid_**.

Here are some of the [pros and cons](https://github.com/FuelLabs/fuels-ts/pull/1193#issuecomment-1679983095).

Diff highlights [here](https://github.com/FuelLabs/fuels-ts/pull/1193/files#diff-d9704bbfee72536a1a911ce2bfc67c1648c22b7c35be201165f9cceaf344dab3) and source snippets bellow.

# Hybrid

 - Full source file [here](https://github.com/FuelLabs/fuels-ts/blob/aa/chore/splitting-sync-async-utilities/packages/errors/src/test-utils/expect-to-throw-fuel-error.ts)

https://github.com/FuelLabs/fuels-ts/blob/ff8c4dfeafd23744f2a6f747910975540cd381cd/packages/errors/src/test-utils/expect-to-throw-fuel-error.ts#L30-L56

# Split

 - Full source file [here](https://github.com/FuelLabs/fuels-ts/blob/aa/chore/splitting-sync-async-utilities/packages/errors/src/test-utils/expect-to-throw-fuel-error.ts)

https://github.com/FuelLabs/fuels-ts/blob/84cc66b8a58baa40a7232bb39e5fef4398c27f29/packages/errors/src/test-utils/expect-to-throw-fuel-error.ts#L38-L55


nedsalk:
I understand the sentiment in this approach and I'm okay with it. 

My question is: should we really separate our testing utilities into separate sync and async implementations or write one utility that handles both its sync and async cases and use the type system's inference capabilities to let the user know which case they're dealing with? (See [the implementation of `expectToThrowFuelError` in the base branch](https://github.com/FuelLabs/fuels-ts/pull/1153/files/3b9188ca9e5b916a727bbdfe3681ec23cdda1b24#diff-d9704bbfee72536a1a911ce2bfc67c1648c22b7c35be201165f9cceaf344dab3)).

I lean more on the "one implementation covering both cases" side, but it's not something I have a hard opinion on. One might say that it's harder to maintain because you have to think about the type inference side of things as well - and they'd be right, but I like the benefit for the end user in that they have only one utility function to think about. In most cases, the type inference isn't really that problematic to wrap your head around and it'd be relatively easy to maintain. However, given the ubiquitous nature of the `*async` suffix in programming, it would be intuitive for the user to have two _same_ utilities as well (`utility`, `utilityAsync`).

@FuelLabs/sdk-ts please chime in.

arboleya:
@nedsalk You front-ran my description for the PR. 😅 

But yes, that's all there is to it, and I highlighted both implementations in the description:

 1. Having a method that can behave sync/async depending on the received arguments 
 2. Having two methods: `sync` and `async`, to be used independently

Regardless of how this goes, I think the `safeExec` utility should follow the pattern. If this becomes too cumbersome, I can file this PR later on against `master` as well, just wanted to get it out there since I'd be primarily unavailable this week.

danielbate:
As a general rule I wouldn't like to see us trying utilising a hybrid approach for sync/async code for most cases. It can add a lot of noise to the function signature and abstracts the function away from it's actual utility. Maintaining simplicity and a simpler API I think is more beneficial in most cases.

However for test utilities like the former implementation of the above, I appreciate the flexibility it offers to test cases and am in favour of it.

arboleya:
@danielbate How are test utilities different from anything else?

danielbate:
I would regard them as different as the test utility contents shouldn't dictate the clarity of the test. For example, so long as it is semantically named, a utility like `getForcProject` or `getTestWallet`; these utilities bootstrap the tests allowing for the real value to be within the test case itself. How they go about doing that I think is an implementation detail that is irrelevant to the test case, and as long as the test assertions are setup correctly and show a valid change then it is a valid and comprehendible test.

Whereas inside of our service, the entire implementation is relevant to scrutiny as it has a direct implication on the validity and value of the service.

arboleya:
The methods `getForcProject` or `getTestWallet` aren't a good reference since they have only one behavior.

The usability criteria of something we expose shouldn't change just because it is intended to be used in tests. It is all the same, and unlike internal test utilities, it's good to reinforce that this one is also being exported for external consumption.

When we think about it, everything in the Javascript ecosystem is habitually split in sync/async, maybe because of how ambiguous things can get otherwise. Besides going against this habit, the hybrid implementation also looks _hacky_, unless we disregard how multiple `@ts-expect-error` and `eslintignore` directives are required to make it play nice.

PR #1202, opened 2023-08-21
feat: push API documentation to separate branch for docs hub

danielbate:


Torres-ssf (approved):

.github/workflows/release.yaml
```diff
@@ -94,3 +94,22 @@ jobs:
...
+      - name: Checkout API Docs
+        if: env.SHOULD_DEPLOY_DOCS == 'true'
+        uses: actions/checkout@v3
+        with:
+          ref: docs
+        
+      - name: Generate and Push API Docs
+        if: env.SHOULD_DEPLOY_DOCS == 'true'
+        run: |
+          pnpm install
```
arboleya, line 107: Shouldn't we ensure the `docs` branch is up-to-date with `master` before `install`?

.github/workflows/release.yaml
```diff
@@ -94,3 +94,22 @@ jobs:
...
+          pnpm build
+          rm -f apps/docs/.gitignore
+          git add apps/docs/src/api/
+          git add apps/docs/.typedoc/api-links.json
```
arboleya, line 111: Could this lead to conflicts down the road?

IIUC, the version from `master` will always contain `apps/docs/.gitignore`.

Meanwhile, we will always try to delete it, so the next sync will probably fail.

danielbate: The remove is not committed though. Just allows for the markdown and links files to be committed to `docs`. I can also `restore` the `apps/docs/.gitignore` for completeness. We could have conflicts in `docs` if we ever did start committing the contents of `apps/docs/src/api/`.

.github/workflows/release.yaml
arboleya: I couldn't understand how this was resolved, so I am re-opening this to ask:
 - How are you getting the new commits from `master` into `docs` before building it?

arboleya (approved):

.github/workflows/release.yaml
arboleya: Shouldn't it be a `merge` instead? Or, more specifically:

```
git merge origin master --no-edit
```

The `--no-edit` flag is to auto-accept the default commit message in a non-interactive manner.

danielbate: Agreed with `--no-edit` and will implement. I was doing a `pull` to ensure all `master` changes had been fetched.

arboleya: Yes, but the `pull` would `pull` the _remote_ `master` changes into the _local_ `master` branch, right? 😅 

The `merge` is to _remote_ `master` -> _local_ `docs`.

danielbate: To be fair, I am also happy using `merge` if that is the preferred implementation, I just use `pull` for this reason where base branch is the branch I am pulling, and do not need to investigate the fetched changes.

Torres-ssf (approved):

.github/workflows/release.yaml
```diff
@@ -94,3 +94,24 @@ jobs:
...
+        run: |
+          git pull origin master --no-edit
+          pnpm install
+          pnpm build
+          rm -f apps/docs/.gitignore
+          git add apps/docs/src/api/
+          git add apps/docs/.typedoc/api-links.json
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config user.name "github-actions[bot]"
+          git commit -m "docs: API docs"
```
arboleya, line 115: Could we have the version number on the commit message here?

- "docs: API docs - `v0.50.1`"

Let's say we want — _in the future, at some point_ — to make all versions available with a dropdown for users to select which version they're using.

danielbate: Good idea

danielbate: I'll add to a follow-up issue

arboleya:
@danielbate It looks like there's something off:
 - https://github.com/FuelLabs/fuels-ts/actions/runs/6030751395/job/16363160462

Not sure if simply adding a `--no-rebase` will be enough. Maybe it will. :)

PR #1204, opened 2023-08-21
feat: add `launchNode` and `launchNodeAndGetWallets`

### LINKED ISSUE #1172: Provide testing utility that let users run a node and faucet wallets locally
Follow-up from #1093:

> There is currently no easy function to run a fuel-core node, create a bunch of wallets and faucet them, this is an extremely common practice for contract testing.

We should provide a method called `launchNodeAndGetWallets` that launches a `fuel-core` node, faucets a bunch of wallets and returns those wallets back.

### Possbile implementation (taken from #1093):

```ts
import { Provider, NativeAssetId, WalletUnlocked, Wallet } from 'fuels';
const { spawn } = require('child_process');
const fs = require('fs').promises;

// Run fuel-core and stop it when the process has ended.
export function fuel_core(
    chainConfig: String,
    params:String[] = ['--db-type', 'in-memory']
    ): Promise<any> {
    return new Promise(resolve => {
        // The string to search for GraphQL service start.
        const graph_ql_start = 'graphql_api::service';

        // Spawn a fuel-core instance with a specific chainConfig.
        const child = spawn('fuel-core', [
            'run', 
            '--chain', chainConfig,
            ...params
        ]);

        // Set the std encoding to UTF8.
        child.stderr.setEncoding('utf8');

        // Cleanup function where fuel-core is stopped.
        const cleanup = function() {
            child.kill('SIGINT');
        };

        // Look for a specific graphql start point in the output.
        child.stderr.on('data', function(chunk:any) {
            // Look for the graphql service start.
            if (chunk.indexOf(graph_ql_start) !== -1) {
                // Resolve with the cleanup method.
                resolve(cleanup);
            }
        });

        // Process exit.
        process.on('exit', cleanup);

        // Catches ctrl+c event.
        process.on('SIGINT', cleanup);

        // Catches "kill pid" (for example: nodemon restart).
        process.on('SIGUSR1', cleanup);
        process.on('SIGUSR2', cleanup);

        // Catches uncaught exceptions.
        process.on('uncaughtException', cleanup);
    });
}

// Build a test environment.
export async function launch_node_get_wallets(inputConfig = {}): Promise<{
    cleanup: Function,
    wallet: WalletUnlocked,
    wallets: WalletUnlocked[],
    provider: Provider }> {

    // Handle config object. 
    const config = Object.assign({
        numWallets: 10,
        chainConfig: null,
        chainConfigPath: '.chainConfig.json',
        fuelCoreParams: undefined,
        providerURL: 'http://127.0.0.1:4000/graphql',
    }, inputConfig);

    // Create a number of wallets.
    const signers = (new Array(config.numWallets))
        .fill(0)
        .map(() => Wallet.generate())
        .map(v => v.signer());

    // Default chian config.
    const defaultChainConfig = {
        "chain_name": "local_testnet",
        "block_gas_limit": 5000000000,
        "initial_state": {
            "coins": signers.map(signer => {
                return {
                    "owner": signer.address.toHexString(),
                    "amount": "0xFFFFFFFFFFFFFFFF",
                    "asset_id": NativeAssetId,
                };
            }),
        },
        "transaction_parameters": {
            "contract_max_size": 16777216,
            "max_inputs": 255,
            "max_outputs": 255,
            "max_witnesses": 255,
            "max_gas_per_tx": 500000000,
            "max_script_length": 1048576,
            "max_script_data_length": 1048576,
            "max_static_contracts": 255,
            "max_storage_slots": 255,
            "max_predicate_length": 1048576,
            "max_predicate_data_length": 1048576,
            "gas_price_factor": 1000000000,
            "gas_per_byte": 4,
            "max_message_data_length": 1048576
        },
        "consensus": {
            "PoA": {
                "signing_key": "0x94ffcc53b892684acefaebc8a3d4a595e528a8cf664eeb3ef36f1020b0809d0d"
            }
        }
    };

    // Write a temporary chain configuration file.
    await fs.writeFile(
        config.chainConfigPath,
        JSON.stringify(config.chainConfig ? config.chainConfig : defaultChainConfig), 
        'utf8'
    );

    // Run the fuel-core node.
    const close = await fuel_core(config.chainConfigPath, config.fuelCoreParams);

    // Create a provider object.
    const provider = new Provider(config.providerURL);

    // Connect all the wallets.
    const wallets = signers.map(signer => Wallet.fromPrivateKey(signer.privateKey, provider));

    // Define cleanup method.
    async function cleanup() {
        await fs.unlink(config.chainConfigPath);
        close();
    }

    // Return the environment.
    return {
        cleanup,
        wallet: wallets[0],
        wallets,
        provider,
    };
}
```

Dhaiwat10:
## Summary

This PR adds two new testing utility functions to `@fuel-ts/wallet/test-utils`.

### 1. `launchNodeAndGetWallets`
This function would be the quickest way for someone to start a `fuel-core` node directly from the TS SDK, and get some fauceted wallets back.

```ts
const { stop, provider, wallets } = await launchNodeAndGetWallets();

// do whatever you want with the wallets and the provider

stop(); // stops the node
```

You can also configure the node to your liking:

```ts
await launchNodeAndGetWallets({
    launchNodeOptions: {
      consensusKey: 'abcd',
      chainConfigPath: '.chainConfig.json',
      dbPath: '.fuel-core-db',
      args: ['--some-arg', '--some-other-arg']
    }
  });
```

### 2. `launchNode`
This is a subset of `launchNodeAndGetWallets`, and just launches a node without the bootstrapping element.

Closes #1172


packages/wallet/src/test-utils/defaultChainConfig.ts
Dhaiwat10, line 1: This default chaing config is a direct copy of the chain config file we use for our local dev environment.

packages/wallet/src/test-utils/defaultChainConfig.ts
```diff
@@ -0,0 +1,410 @@
+export const defaultChainConfig = {
+  chain_name: 'local_testnet',
+  block_gas_limit: 5000000000,
+  initial_state: {
+    coins: [
+      {
+        owner: '0x94ffcc53b892684acefaebc8a3d4a595e528a8cf664eeb3ef36f1020b0809d0d',
```
Dhaiwat10, line 7: This default chain config is a direct copy of the chain config file we use for our local dev environment. It is there just to provide a good default to users who don't want customization, and want something up and running quickly.

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -0,0 +1,123 @@
...
+
+const generateWallets = async (count: number, provider: Provider) => {
+  const wallets: WalletUnlocked[] = [];
+  for (let i = 0; i < count; i += 1) {
+    const wallet = await generateTestWallet(provider, [[1_000, BaseAssetId]]);
+    wallets.push(wallet);
+  }
+  return wallets;
+};
+
```
nedsalk, line 90:
```suggestion
const generateWallets = async (count: number, provider: Provider) => {
  const calls: Promise<WalletUnlocked>[] = [];
  for (let i = 0; i < count; i += 1) {
    calls.push(generateTestWallet(provider, [[1_000, BaseAssetId]]));
  }
  return Promise.all(calls);
};
```

This parallellizes wallet generation so that there is no need to wait for one wallet to be generated to start with the generation of the next one.

packages/wallet/src/test-utils/launchNode.test.ts
```diff
@@ -0,0 +1,19 @@
...
+import { WalletUnlocked } from '../wallets';
+
+import { launchNodeAndGetWallets } from './launchNode';
+
+/**
+ * This test is skipped because it requires fuel-core to be installed.
+ * TODO: Figure out a way to run this test using pnpm fuels-core in CI.
+ */
+test.skip('launchNodeAndGetWallets', async () => {
+  const { stop, provider, wallets } = await launchNodeAndGetWallets();
```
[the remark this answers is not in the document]
arboleya: We could add a new property to `LaunchNodeOptions`:

```ts
useSystemFuelCore: boolean; // default = false
```

Then use each binary accordingly:

```ts
    const command = (useSystemForc ? 'fuel-core' : './node_modules/.bin/fuels-core')
    const child = spawn(command, [
      'run',
      ...
```

We cannot use `pnpm` as the user may be using another package manager, so we refer to the binary directly from inside `node_modules/.bin` — I believe the `wallet` package must depend on `@fuel-ts/fuel-core` for this.

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -0,0 +1,123 @@
...
+  const provider = new Provider('http://127.0.0.1:4000/graphql');
+  const wallets = await generateWallets(10, provider);
+
+  const cleanup = () => {
+    closeNode();
+    spawn('rm', ['.chainConfig.json']);
+    spawn('rm', ['-rf', '.fuel-core-db']);
+  };
+
```
nedsalk, line 121:
```suggestion
export const launchNodeAndGetWallets = async ({
  launchNodeOptions,
  walletCount = 10
}: {
  launchNodeOptions?: LaunchNodeOptions;
  walletCount?: number
} = {}) => {
  // Write a temporary chain configuration file.
  await fs.writeFile('.chainConfig.json', JSON.stringify(defaultChainConfig), 'utf8');

  // Create a temp db directory.
  await fs.mkdir('.fuel-core-db', { recursive: true });

  const defaultNodeOptions: LaunchNodeOptions = {
    chainConfigPath: '.chainConfig.json',
    consensusKey: '0xa449b1ffee0e2205fa924c6740cc48b3b473aa28587df6dab12abc245d1f5298',
    dbPath: '.fuel-core-db',
  };

  const closeNode = await launchNode(launchNodeOptions ?? defaultNodeOptions);

  const provider = new Provider('http://127.0.0.1:4000/graphql');
  const wallets = await generateWallets(walletCount, provider);

  const cleanup = () => {
    closeNode();
    spawn('rm', ['.chainConfig.json']);
    spawn('rm', ['-rf', '.fuel-core-db']);
  };

  return { wallets, stop: cleanup, provider };
};
```

This is just to enable the user to generate as many wallets as they wish by providing `walletCount`.

nedsalk (commented):
This is awesome stuff! Can't wait to see its fruits in our tests. I've added some suggestions. I've refrained from approving until you resolve the conversation around the skipped test, as it's an important one.

packages/wallet/src/test-utils/launchNode.ts
Dhaiwat10: @nedsalk I just tried this out and the client is reporting some UTXO-related error: `Transaction is not inserted. UTXO is not existing: 0x000000000000000000000000000000000000000000000000000000000000000002: {"response":{"data":null,"errors":[{"message":"Transaction is not inserted. UTXO is not existing:`

Looks like it's not a good idea to try and generate wallets in parallel 🤔

packages/wallet/src/test-utils/launchNode.ts
Dhaiwat10: ✅

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -0,0 +1,123 @@
...
+  return wallets;
+};
+
+/**
+ * Launches a fuel-core node and returns a provider, 10 wallets, and a cleanup function to stop the node.
+ * */
+export const launchNodeAndGetWallets = async ({
+  launchNodeOptions,
+}: {
+  launchNodeOptions?: LaunchNodeOptions;
```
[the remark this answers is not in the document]
Dhaiwat10: ✅

packages/wallet/src/test-utils/launchNode.ts
nedsalk: Hmm, in theory it should work and it would be handy to have it, especially during testing if we want to run one node and have each test use an unique wallet to avoid coupling tests, in which case we'd need _a lot_ of wallets. It also seems like an interesting problem to investigate. Could you please just create an issue after you merge this PR?

packages/wallet/src/test-utils/launchNode.test.ts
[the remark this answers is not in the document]
arboleya: FWIW, after syncing some requirements with @luizstacio, I'm also following this approach in Tooling PR.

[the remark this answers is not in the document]
Dhaiwat10: No I agree this is a great solution! We should both go ahead with this implementation. I am almost there. I'm facing some trouble because the cleanup logic that I have in place is not being able to cleanup the process started with the `pnpm` binary properly.

packages/wallet/src/test-utils/launchNode.ts
Dhaiwat10: @nedsalk sure, I'll file a follow-up

packages/wallet/src/test-utils/launchNode.test.ts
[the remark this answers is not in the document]
arboleya: I'm using something like this in the Tooling PR:

```ts
import kill from 'tree-kill';

const myProcess = spawn(...);

kill(Number(myProcess.pid));
```

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -23,18 +24,22 @@ type LaunchNodeOptions = {
...
  * @param dbPath - the path to the folder to use for the fuel-core db.
  * @param args - additional arguments to pass to fuel-core
+ * @param useSystemFuelCore - whether to use the system fuel-core binary or the one provided by the \@fuel-ts/fuel-core package.
  * */
 export const launchNode = async ({
   chainConfigPath,
   consensusKey,
   dbPath,
   args = defaultFuelCoreArgs,
+  useSystemFuelCore = true,
```
arboleya, line 34: I think the default should be `false` for increased compatibility <sup>(*)</sup>.

Only in atypical situations will users want otherwise.

> <sup>(*)</sup> _Same approach being applied in the Tooling_.

Dhaiwat10: Makes sense, mb!

packages/wallet/src/test-utils/launchNode.test.ts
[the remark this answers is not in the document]
Dhaiwat10: Thanks @arboleya, this worked!

packages/wallet/package.json
```diff
@@ -55,7 +56,7 @@
     "@fuel-ts/transactions": "workspace:*",
     "@fuel-ts/utils": "workspace:*",
     "@fuels/vm-asm": "0.34.1",
-    "@fuel-ts/crypto": "workspace:*",
+    "tree-kill": "^1.2.2",
```
arboleya, line 59: You need to add `@fuel-ts/fuel-core` as a dependency here to guarantee that it will work even outside the mono repo or in case users install the `wallet` package in isolation.

arboleya (changes requested):

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -0,0 +1,133 @@
...
+ * */
+export const launchNodeAndGetWallets = async ({
+  launchNodeOptions,
+  walletCount = 10,
+}: {
+  launchNodeOptions?: Partial<LaunchNodeOptions>;
+  walletCount?: number;
+} = {}) => {
+  // Write a temporary chain configuration file.
+  await fs.writeFile('.chainConfig.json', JSON.stringify(defaultChainConfig), 'utf8');
```
luizstacio, line 110: You can use the OS temp folder here, also  as is a temp file it could include on the name a uuid to avoid conflict.

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  // Create a temp db directory.
+  await fs.mkdir('.fuel-core-db', { recursive: true });
```
luizstacio, line 113: As this is meant to be temporary you can change the fuel-core args to use --db-type in-memory in this way no need to have a database folder

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  const defaultNodeOptions: LaunchNodeOptions = {
+    chainConfigPath: '.chainConfig.json',
+    consensusKey: '0xa449b1ffee0e2205fa924c6740cc48b3b473aa28587df6dab12abc245d1f5298',
+    dbPath: '.fuel-core-db',
+  };
+
+  const closeNode = await launchNode({ ...defaultNodeOptions, ...launchNodeOptions });
+
+  const provider = new Provider('http://127.0.0.1:4000/graphql');
```
luizstacio, line 123: As you launch a node, the return could include the `ip` and `port` instead of having fixed one.

luizstacio: And for the default params you can use the `0.0.0.0` as IP and for the port you can get the next available port on from the OS.

packages/wallet/src/test-utils/launchNode.ts
Torres-ssf: @Dhaiwat10 @nedsalk The cause of this is the flow process used to generate a new test wallet. Basically, the steps are generating a new wallet and using UTXOs from the Genesis wallet to transfer funds to it. 

The process to fund this new wallet is basically fetching UTXOs (Unspent Tokens) from the Genesis wallet and then creating a transfer transaction to this new wallet. 

The change from this transaction (i.e., the new UTXOs) becomes available to the Genesis wallet only after the transaction has been fully processed, and the used UTXOs on the transaction, are now spent.

So when you try to execute all loops concurrently you are creating a race condition. Many of these processes are going to use the same UTXOs to create the transfer transaction, and the transaction that is processed first will succeed but all other ones are going to fail since they are now using a token that was spent. 

This explains the error message: `UTXO is not existing.`

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -106,27 +136,41 @@ export const launchNodeAndGetWallets = async ({
...
+    fsSync.mkdirSync(osTempDir);
+  }
 
-  // Create a temp db directory.
-  await fs.mkdir('.fuel-core-db', { recursive: true });
+  const subDirName = '.fuels-ts'; // Change to your desired subfolder name
+  const subDirPath = path.join(osTempDir, subDirName);
+
+  if (!fsSync.existsSync(subDirPath)) {
+    fsSync.mkdirSync(subDirPath);
```
luizstacio, line 149:
```suggestion
    fsSync.mkdirSync(subDirPath, { recursive: true });
```

You can improve this by just informing node.js to do it recursively. In this way, node.js is going to create all the folders even if the path does not exist. If exists is going to just ignore it.

Also in this way no need to the first checks line 141..143

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -0,0 +1,177 @@
...
+      '--consensus-key',
+      consensusKey,
+      '--chain',
+      chainConfigPath,
+      ...args,
+    ]);
+
+    // Cleanup function where fuel-core is stopped.
+    const cleanup = () => {
+      kill(Number(child.pid));
```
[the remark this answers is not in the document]
arboleya: We could only if we wanted to re-implement the process tree traversal by hand.

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -23,27 +28,48 @@ type LaunchNodeOptions = {
...
 
     const command = useSystemFuelCore ? 'fuel-core' : './node_modules/.bin/fuels-core';
 
+    const ipToUse = ip || '0.0.0.0';
+
+    const portToUse =
+      port ||
+      (
+        await getPortPromise({
+          port: 4000, // tries 4000 first, then 4001, then 4002, etc.
```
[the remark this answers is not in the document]
arboleya: The suggested approach provides a frictionless experience.

AFAIK, all modern tools will search for free ports on your machine. (`vite dev`, `next dev`, etc).

What are your concerns in this regard?

packages/wallet/src/test-utils/defaultChainConfig.ts
arboleya: I agree, but could you please create a new issue for this?

I'm also _[erroneously]_ duplicating this file in another PR I'm working on.

This will give us time to ponder the best way to resolve all occurrences simultaneously.

packages/wallet/src/test-utils/launchNode.ts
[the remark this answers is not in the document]
Dhaiwat10: I see your point @camsjams but I chose to use this package instead of vanilla Node.js because this package goes also kills any child process spawned by the process we are trying to kill, like @arboleya mentioned. I believe Anderson is also using the same package in the contracts tooling PR

packages/wallet/src/test-utils/launchNode.ts
Dhaiwat10: Great catch @luizstacio

packages/wallet/src/test-utils/launchNode.ts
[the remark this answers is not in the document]
arboleya: FWIW, the source and tests look concise, and the package is massively [used](https://www.npmjs.com/package/tree-kill).

Considering it is still in vogue, the fact it is this old might tell me it is really good.

The downloads/issues ratio also transmits the feeling of an outstanding solid package.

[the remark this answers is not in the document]
arboleya: Not in the tooling PR, because there it's not aimed at tests.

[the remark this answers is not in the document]
Dhaiwat10: Can we make it dev dependency in this PR too? Won't we need this logic at run-time?

packages/wallet/src/test-utils/launchNode.ts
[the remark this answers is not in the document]
luizstacio: Yeah searching for port is a very good experience. Imagine that if you don't have this here, but want to use this functions to run all tests we have right now in parallel, you would need to specify a port for each test, and make sure manually they don't conflict. In this way you can just run.

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  const subDirName = '.fuels-ts'; // Change to your desired subfolder name
+  const subDirPath = path.join(osTempDir, subDirName);
+
+  fsSync.mkdirSync(subDirPath, { recursive: true });
+
+  const chainConfigFilePath = path.join(subDirPath, '.chainConfig.json');
+
+  // Write a temporary chain configuration file.
+  await fs.writeFile(chainConfigFilePath, JSON.stringify(defaultChainConfig), 'utf8');
```
luizstacio, line 150: It would be good to allow chainConfig to be provided by the user. And get the consensusKey from a `Wallet.generate()` or a user-provided one.

These things will help tests to have specific chain states before the tests, like particular messages.

Dhaiwat10: @luizstacio you can already do both of these things by passing in the respective `launchNodeOptions` like so:

const { stop, provider, wallets } = await launchNodeAndGetWallets({
    launchNodeOptions: {
      chainConfigPath: './.fuel-core/configs/chainConfig.json',
      consensusKey: 'key'
    },
  });

Dhaiwat10: We only use the default chain config and the default consensus key in case the user hasn't provided one. I did some cleanup though. Please have a look! :)

packages/wallet/package.json
```diff
@@ -54,8 +56,10 @@
     "@fuel-ts/signer": "workspace:*",
     "@fuel-ts/transactions": "workspace:*",
     "@fuel-ts/utils": "workspace:*",
+    "@fuels/vm-asm": "0.34.1",
```
arboleya, line 59: These packages don't seem to be necessary:
 - `@fuel-ts/crypto`
 - `@fuels/vm-asm`

Please remove them.

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  const subDirName = '.fuels-ts'; // Change to your desired subfolder name
+  const subDirPath = path.join(osTempDir, subDirName);
+
+  fsSync.mkdirSync(subDirPath, { recursive: true });
+
+  const defaultChainConfigFilePath = path.join(subDirPath, '.chainConfig.json');
+
+  // Don't create a temporary default chain configuration file if one is provided.
+  if (launchNodeOptions?.chainConfigPath) {
```
arboleya, line 150: Shouldn't this be the opposite?

```ts
if (!launchNodeOptions?.chainConfigPath) {
```

packages/wallet/src/test-utils/launchNode.test.ts
```diff
@@ -0,0 +1,21 @@
+import { Provider } from '@fuel-ts/providers';
+
+import { WalletUnlocked } from '../wallets';
+
+import { launchNodeAndGetWallets } from './launchNode';
+
+test('launchNodeAndGetWallets', async () => {
```
arboleya, line 7: Please add more tests covering the duality between `default` configs vs.`user` configs.

 - User **_not_** passing a `chainConfig`
 - User passing `useSystemFuelCore = true`
 - User not passing `walletCount`
 - User passing a different `walletCount` (other than 10)
 - Method `launchNodeAndGetWallets` being called without params

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -0,0 +1,177 @@
...
+
+import type { WalletUnlocked } from '../wallets';
+
+import { defaultChainConfig } from './defaultChainConfig';
+import { generateTestWallet } from './generateTestWallet';
+
+const defaultFuelCoreArgs = ['--vm-backtrace', '--utxo-validation', '--manual_blocks_enabled'];
+
+type LaunchNodeOptions = {
+  chainConfigPath: string;
```
arboleya, line 19: Shouldn't this be optional?

```suggestion
  chainConfigPath?: string;
```

arboleya (changes requested):

packages/wallet/package.json
arboleya: Never mind, both are used.

I mixed things up, my bad.

packages/wallet/src/test-utils/launchNode.ts
Dhaiwat10: @arboleya it is already optional for the `launchNodeAndGetWallets` function because `launchNodeOptions` is a `Partial`: 

```ts
launchNodeOptions?: Partial<LaunchNodeOptions>;
```

Dhaiwat10: Do you think the logic for going with a default `chainConfig` when one is not provide should live in the `launchNode` function?

Dhaiwat10: Same for `consensusKey`

arboleya: Yes, for both. 👍

packages/wallet/src/test-utils/launchNode.test.ts
Dhaiwat10: @arboleya great points.

- I actually had a test case for `useSystemFuelCore = true` in place but it won't work in our CI :( 
- Shouldn't the last test case you mentioned cover the user not passing a chainconfig and not passing a walletcount?

packages/wallet/src/test-utils/launchNode.ts
Dhaiwat10: Oopsie woopise horrible mistake

packages/wallet/src/test-utils/launchNode.test.ts
arboleya: Hmm I don't think so, perhaps. Let's see.

I should've split the observations for each method, here it goes:

`launchNode()`:
 - User not passing a chainConfig
 - User passing useSystemFuelCore = true

`launchNodeAndGetWallets()`:
 - User not passing `walletCount`
 - User passing a different `walletCount` _(other than 10, the default)_
 - Method being called without any params

Makes sense?

arboleya: With relation to the `useSystemFuelCore=true` case, one alternative would be to mock the `spawn` method and assert the 1st argument to ensure it's `fuel-core`. It may be tricky to mock this one, though.

Pls, let me know if it becomes a PITA.

Dhaiwat10: @arboleya the test cases look good collectively, but I feel like we only really need to test the `launchNodeAndGetWallets` function properly since it internally always calls `launchNode` and just passes down the arguments. If there is a problem anywhere in this, it will be caught in our assertions. I was thinking we can have these three test cases + one for `useSystemFuelCore: true`

image


Thoughts?

arboleya: Sure, these might be enough, and I also don't want to hold this back.

If the `useSystemFuelCore` test becomes difficult, let's go without it.

Just make **_sure_** it's working with your local `fuel-core`, and voilà.

You can also check for possible missed spots by checking the coverage report.

```shell
# tab1
pnpm node:run

# tab 2
pnpm test packages/wallet/src/test-utils/launchNode.test.ts
open ./coverage/lcov-report/index.html
```

arboleya: Never mind, you already did it.

I missed the last commits when sending the previous comment.

arboleya (approved):

luizstacio (approved):

packages/wallet/src/test-utils/launchNode.ts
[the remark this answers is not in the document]
arboleya: I share your concerns about inadvertently adding 3rd party libraries.

Like I said in my previous [analisys](https://github.com/FuelLabs/fuels-ts/pull/1204#discussion_r1311556677) about `tree-kill`, this library also looks remarkably solid.

Look at the numbers:
 - ~8 million [downloads](https://www.npmjs.com/package/portfinder) per week
 - 3x open [issues](https://github.com/http-party/node-portfinder/issues)
 - 2k+ [dependent](https://www.npmjs.com/package/portfinder?activeTab=dependents) packages

Implementing everything ourselves is not optimal, so we must find a middle ground.

What are your criteria for using 3rd party libraries vs. implementing things ourselves?

PR #1210, opened 2023-08-24
refactor: transaction response wait strategy

Torres-ssf:
This PR refactors the `fetch` method in the `TransactionResponse` class. 

The main goal is to improve its behavior to continuously poll until a non-null response is obtained from the fetch operation. This ensures that any fetch operation returns the desired `Transaction` data.

Previously, the `waitForResult` method was responsible for waiting for both a non-null response and a transaction status other than `submitted`. With this refactor, its responsibility has been narrowed to only ensuring the transaction status isn't `submitted`.


packages/providers/src/transaction-response/transaction-response.ts
```diff
@@ -199,11 +192,40 @@ export class TransactionResponse {
...
+    }
+  }
+
+  async waitUntilResponseReceived(): Promise<void> {
+    while (!this.gqlTransaction?.status?.type) {
+      await this.sleepBasedOnAttempts();
+      await this.fetch();
+      this.attempts += 1;
+    }
+  }
```
LuizAsFight, line 218: on this way attemps will advance twice as fast.

one solution is to have 2 different attempts like: `fetchAttempts` then pass it like `sleepBaseOnAttemps(this.fetchAttemps)`

LuizAsFight (changes requested):

packages/errors/src/error-codes.ts
```diff
@@ -18,6 +18,10 @@ export enum ErrorCode {
 
   PARSE_FAILED = 'parse-failed',
 
+  // transaction
+
+  TRANSACTION_FAILED = 'transaction-failed',
```
arboleya, line 23:
```suggestion
  // transaction
  TRANSACTION_FAILED = 'transaction-failed',
```

packages/providers/src/transaction-response/transaction-response.ts
```diff
@@ -109,12 +113,20 @@ export class TransactionResponse {
...
-    const transaction = await this.provider.operations.getTransactionWithReceipts({
+  async fetch(): Promise<GqlTransaction> {
+    const response = await this.provider.operations.getTransactionWithReceipts({
       transactionId: this.id,
     });
 
-    return transaction;
+    if (!response.transaction) {
+      this.fetchAttempts += 1;
+      await this.sleepBasedOnAttempts(this.fetchAttempts);
```
arboleya, line 123: Nit.

```suggestion
      await this.sleepBasedOnAttempts(++this.fetchAttempts);
```

packages/providers/src/transaction-response/transaction-response.ts
```diff
@@ -173,15 +165,37 @@ export class TransactionResponse {
...
+   * @returns The completed transaction result
+   */
+  async waitForResult<TTransactionType = void>(
+    contractsAbiMap?: AbiMap
+  ): Promise<TransactionResult<TTransactionType>> {
+    await this.fetch();
+
+    if (this.gqlTransaction?.status?.type === 'SubmittedStatus') {
+      this.resultAttempts += 1;
+      await this.sleepBasedOnAttempts(this.resultAttempts);
```
arboleya, line 190: Nit.
Nit.

```suggestion
      await this.sleepBasedOnAttempts(++this.resultAttempts);
```

packages/providers/src/transaction-response/transaction-response.ts
```diff
@@ -199,11 +213,24 @@ export class TransactionResponse {
...
     return result;
   }
+
+  private async sleepBasedOnAttempts(attempts: number): Promise<void> {
+    // This code implements a similar approach from the fuel-core await_transaction_commit
+    // https://github.com/FuelLabs/fuel-core/blob/cb37f9ce9a81e033bde0dc43f91494bc3974fb1b/fuel-client/src/client.rs#L356
+    // double the interval duration on each attempt until max is reached
+    //
+    // This can wait forever, it would be great to implement a max timeout here, but it would require
+    // improve request handler as response Error not mean that the tx fail.
```
arboleya, line 231: The code here is simple; the previous comments felt redundant to me.

Except for the `maxTimeout` idea that I turned into a TODO comment.


```suggestion
  /*
    TODO: Consider adding `maxTimeout` or `maxAttempts` parameter.

    The aim is to avoid perpetual execution; when the limit
    is reached, we can throw accordingly.
  */
  private async sleepBasedOnAttempts(attempts: number): Promise<void> {
```

arboleya (commented):
🪄

arboleya (dismissed):

packages/providers/src/transaction-response/transaction-response.ts
```diff
@@ -91,7 +91,11 @@ export class TransactionResponse {
   /** Gas used on the transaction */
   gasUsed: BN = bn(0);
   /** Number off attempts to get the committed tx */
-  attempts: number = 0;
+  fetchAttempts: number = 0;
+
+  resultAttempts: number = 0;
+
+  gqlTransaction?: GqlTransaction;
```
danielbate, line 98: Please can we get some commented descriptions on these as typedoc is exposing this class

packages/providers/src/transaction-response/transaction-response.ts
```diff
@@ -199,11 +211,24 @@ export class TransactionResponse {
...
         `Transaction failed: ${(<FailureStatus>result.gqlTransaction.status).reason}`
       );
     }
 
     return result;
   }
+
+  /*
+    TODO: Consider adding `maxTimeout` or `maxAttempts` parameter.
+
```
danielbate, line 225:
```suggestion
    @hidden

    TODO: Consider adding `maxTimeout` or `maxAttempts` parameter.

```

danielbate (changes requested):

packages/providers/src/transaction-response/transaction-response.ts
arboleya: @danielbate Is the `@hidden` necessary here?

This block is a multiline comment `/*`, not a DocTag `/**`.

arboleya: @Torres-ssf I am unr-resolving this one, ok? I'm hoping @danielbate can clarify this matter with multiline comments, as I believe it could be problematic if these are treated as DocBlocks.

Torres-ssf: @arboleya ok makes sense

LuizAsFight (approved):
beautiful

packages/providers/src/transaction-response/transaction-response.ts
danielbate: Apologies both, you'e correct @arboleya

Torres-ssf: Awesome @danielbate, thanks :pray:

arboleya (approved):

digorithm (approved):
Nice -- lgtm!

Dhaiwat10 (approved):

Issue #779, opened 2023-02-10
Script `forc:check` fails silently


arboleya:
The `pnpm forc:check` script _(and the related CI action)_ always passes, regardless of errors.

Example:
 - https://github.com/FuelLabs/fuels-ts/actions/runs/4145700900/jobs/7170416899#step:4:14

<details>
  <summary>Screenshot</summary>

![ci](https://user-images.githubusercontent.com/26660/218167629-4c4ee18d-dd38-4d88-a0be-7bd1aabca76e.png)

</details>

digorithm:
Should this be closed?

arboleya:
No, it hasn't been addressed yet.

arboleya:
The problem is with this script, which seems to have _never_ worked since it'd always `exit 0` regardless.

I pushed a quick fix here:
 - https://github.com/FuelLabs/fuels-ts/pull/1300

Now, it [fails](https://github.com/FuelLabs/fuels-ts/actions/runs/6329057387/job/17188571857?pr=1300#step:4:678) as expected, and it's a matter of fixing all poorly formatted files.

Any help is appreciated. 🙏

PR #1211, opened 2023-08-24
chore: formatted all sway projects

nedsalk:
Just ran `./scripts/forc-format.sh` in the root directory.

We should put [`forc fmt --check`](https://fuellabs.github.io/sway/v0.44.1/book/forc/plugins/forc_fmt.html) somewhere in our CI to ensure proper formatting of all our sway projects, but that's for another PR. The rust-sdk has all of their sway projects in one workspace so they just run `forc fmt --check` on that workspace.


Torres-ssf (dismissed):

arboleya (commented):
This is the issue:
 - https://github.com/FuelLabs/fuels-ts/issues/779

https://github.com/FuelLabs/fuels-ts/blob/7dc8cca11b8320dcff3ed3402c01b9dfe44dbc25/.github/workflows/test.yaml#L28-L29

danielbate (approved):

danielbate (approved):

Torres-ssf (approved):

PR #1213, opened 2023-08-24
fix: maxGasPerTx value

LuizAsFight:


arboleya:
@Dhaiwat10 FYI - Although I believe you're already on top of this.

Torres-ssf (approved):

PR #1214, opened 2023-08-27
refactor: slim down `interfaces` package's `AbstractProgram`

nedsalk:
In the spirit of eventually removing the `interfaces` package altogether, I removed two methods from `AbstractProgram.interface`:
- `encodeFunctionData` - it was unused throughout the solution,
- `decodeFunctionResult` - the places it was used in could use `FunctionFragment.decodeOutput` directly


arboleya (commented):
When did these methods become obsolete?

The changes in this PR are minimal; I wonder how/why we got here.

I will defer this one to @camsjams, who is more familiar with these corners of the SDK and is probably on top of the historical reasoning behind this design.

> cc @luizstacio - in case you have something to add.

Dhaiwat10 (approved):
Your reasoning here sounds fair but I would wait for an approval from @camsjams and @arboleya

PR #1216, opened 2023-08-29
feat: improve predicate use on custom transactions

### LINKED ISSUE #1020: Configure ScriptCallHandler's inputs and outputs as in rust sdk
### Motivation

Hi there
In our CLOB on predicates in a few cases like fulfillment or cancelation of order, we should have the possibility to configure `ScriptCallHandler` with `inputs` and `outputs`

Here is the `cancel_order` func on Rust language using Rust SDK

```rs
 pub async fn cancel_order(
        predicate: &Predicate,
        wallet: &WalletUnlocked,
        asset0: AssetId,
        amount0: u64,
    ) -> Result<FuelCallResponse<()>, fuels::prelude::Error> {
        let provider = wallet.provider().unwrap();

        let inputs = predicate
            .clone()
            .set_provider(provider.clone())
            .get_asset_inputs_for_amount(asset0, amount0, None)
            .await
            .unwrap();

        let outputs = wallet.get_asset_outputs_for_amount(wallet.address(), asset0, 0);
        println!("inputs = {:?}", inputs);
        println!("outputs = {:?}", outputs);
        let script_call = ScriptCallHandler::new(
            vec![],
            UnresolvedBytes::default(),
            wallet.clone(),
            provider.clone(),
            Default::default(),
        )
        .with_inputs(inputs)
        .with_outputs(outputs)
        .tx_params(TxParameters::default().set_gas_price(1));

        script_call.call().await
    }
```

logs will be like that

```
alice_address = 0xccc4f9249536cf89ab0f31bb52fa06c00ea0387dcac5830924ecf56f2dded3a5
Predicate root = Bech32Address { hrp: "fuel", hash: c04e498ac26cbff0b93d9c218cf8bdc18e1190953ef5f3e5ace8146deded519c }
USDC AssetId (asset0) = 0x56fb8789a590ea9c12af6fe6dc2b43f347700b049d4f823fd4476c6f366af201
UNI AssetId (asset1) = 0x5381bbd1cff41519062c8531ec30e8ea1a2d752e59e4ac884068d3821e9f0093
    

    inputs = [
        ResourcePredicate { 
            resource: Coin(Coin {
                 amount: 1000000000, 
                 block_created: 766024, 
                 asset_id: 56fb8789a590ea9c12af6fe6dc2b43f347700b049d4f823fd4476c6f366af201, 
                 utxo_id: UtxoId { tx_id: 2744628915b0d89ca8e79133a4598f72b32be9e884a723dad67a56c882da48ff, output_index: 2 }, 
                 maturity: 0, 
                 owner: Bech32Address { hrp: "fuel", hash: c04e498ac26cbff0b93d9c218cf8bdc18e1190953ef5f3e5ace8146deded519c }, 
                 status: Unspent 
                }), 
            code: [...], 
            data: UnresolvedBytes { data: [] } 
        }
    ]

    outputs = [
        Coin { 
            to: ccc4f9249536cf89ab0f31bb52fa06c00ea0387dcac5830924ecf56f2dded3a5, 
            amount: 0, 
            asset_id: 56fb8789a590ea9c12af6fe6dc2b43f347700b049d4f823fd4476c6f366af201 
        }, 
        Change { 
            to: ccc4f9249536cf89ab0f31bb52fa06c00ea0387dcac5830924ecf56f2dded3a5, 
            amount: 0, 
            asset_id: 56fb8789a590ea9c12af6fe6dc2b43f347700b049d4f823fd4476c6f366af201 
        }
    ]

```

In the predicate, this code is in charge of cancelation 
```rs
    let mut i = 0u8;
    let inputs: u8 = input_count();
    while i < inputs  {
        if input_owner(i).unwrap() == Address::from(MAKER) {
            return true;
        }
        i += 1u8;
    }
```

So, ts-sdk has func `` that can help us to make cancelation:
```
const cancelTx = await wallet.transfer(predicate.address, 0, token0.assetId, {
    gasPrice: 1,
  });
await cancelTx.waitForResult();
```

We did this script to test that

```ts
import BN from "../src/utils/BN";
import { TOKENS_BY_SYMBOL } from "../src/constants";
import { LimitOrderPredicateAbi__factory } from "../src/predicates";
import { Predicate, Provider, Wallet } from "fuels";
import { nodeUrl, privateKey } from "../src/config";

(async () => {
  const token0 = TOKENS_BY_SYMBOL.USDC;
  const amount0 = BN.parseUnits(20, token0.decimals);
  const token1 = TOKENS_BY_SYMBOL.BTC;
  const amount1 = BN.parseUnits(0.001, token1.decimals);
  const exp = BN.parseUnits(1, 9 + token0.decimals - token1.decimals);
  let price = amount1.times(exp).div(amount0);

  const wallet = Wallet.fromPrivateKey(privateKey, nodeUrl);

  // console.log(price.toString());
  const configurableConstants = {
    ASSET0: token0.assetId,
    ASSET1: token1.assetId,
    MAKER: wallet.address.toB256(),
    PRICE: price.toFixed(0),
    ASSET0_DECINALS: token0.decimals,
    ASSET1_DECINALS: token1.decimals,
  };
  console.log(configurableConstants);
  const predicate = new Predicate(
    LimitOrderPredicateAbi__factory.bin,
    LimitOrderPredicateAbi__factory.abi,
    new Provider(nodeUrl),
    configurableConstants
  );

  console.log(predicate.address.toB256());

  const initialPredicateBalance = await predicate.getBalance(token0.assetId);
  console.log("initialPredicateBalance", initialPredicateBalance.toString());

  const depositTx = await wallet
    .transfer(predicate.address, amount0.toFixed(0), token0.assetId, { gasPrice: 1 })
    .catch((e) => console.error(`depositTx ${e}`));
  await depositTx?.waitForResult();
  //
  const feeTx = await wallet
    .transfer(predicate.address, 1, TOKENS_BY_SYMBOL.ETH.assetId, { gasPrice: 1 })
    .catch((e) => console.error(`feeTx ${e}`));
  await feeTx?.waitForResult();

  const predicateBalances = await predicate.getBalances();
  console.log(
    "predicateBalances",
    predicateBalances.map((v) => ({
      amount: v.amount.toString(),
      assetId: v.assetId.toString(),
    }))
  );
  //---------------------
  const cancelTx = await wallet.transfer(predicate.address, 0, token0.assetId, {
    gasPrice: 1,
  });
  await cancelTx.waitForResult();

  const finalPredicateBalance = await predicate.getBalances();
  console.log(
    "finalPredicateBalance",
    finalPredicateBalance.map((v) => v.amount.toString())
  );
})();

```

But this test fails with this error
```
/Users/alexey/projects/spark-ts-sdk/node_modules/graphql-request/src/index.ts:441
    throw new ClientError(
          ^
ClientError: TransactionOutputCoinAssetIdNotFound(56fb8789a590ea9c12af6fe6dc2b43f347700b049d4f823fd4476c6f366af201): {"response":{"data":null,"errors":[{"message":"TransactionOutputCoinAssetIdNotFound(56fb8789a590ea9c12af6fe6dc2b43f347700b049d4f823fd4476c6f366af201)","locations":[{"line":2,"column":3}],"path":["dryRun"]}],"status":200,"headers":{}},"request":{"query":"mutation dryRun($encodedTransaction: HexString!, $utxoValidation: Boolean) {\n  dryRun(tx: $encodedTransaction, utxoValidation: $utxoValidation) {\n    ...receiptFragment\n  }\n}\n\nfragment receiptFragment on Receipt {\n  data\n  rawPayload\n}","variables":{"encodedTransaction":"0x000000000000000000000000000000010000000005f5e1000000000000000000000000000000000400000000000000000000000000000001000000000000000200000000000000010000000000000000000000000000000000000000000000000000000000000000240000000000000000000000000000009835cda5c81ed1492fe4dc3696a7b39b6e94ac2c3ba4f739bafc69562061b31700000000000000042ce05bde9ada2d2c58b5eb9f08be34df19375d618626c324f75dfdbd226c8d8800000000000004300000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000057fe0ee890797038951426b07a7c0f16d135635e0bc27cae1461aae80ca6b7e6000000000000000056fb8789a590ea9c12af6fe6dc2b43f347700b049d4f823fd4476c6f366af20100000000000000032ce05bde9ada2d2c58b5eb9f08be34df19375d618626c324f75dfdbd226c8d88000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","utxoValidation":false}}}
    at makeRequest (/Users/alexey/projects/spark-ts-sdk/node_modules/graphql-request/src/index.ts:441:11)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async H.estimateTxDependencies (/Users/alexey/projects/spark-ts-sdk/node_modules/@fuel-ts/providers/src/provider.ts:387:39)
    at async n.sendTransaction (/Users/alexey/projects/spark-ts-sdk/node_modules/@fuel-ts/wallet/src/base-unlocked-wallet.ts:85:5) {
  response: {
    data: null,
    errors: [ [Object] ],
    status: 200,
    headers: Headers { [Symbol(map)]: [Object: null prototype] }
  },
  request: {
    query: 'mutation dryRun($encodedTransaction: HexString!, $utxoValidation: Boolean) {\n' +
      '  dryRun(tx: $encodedTransaction, utxoValidation: $utxoValidation) {\n' +
      '    ...receiptFragment\n' +
      '  }\n' +
      '}\n' +
      '\n' +
      'fragment receiptFragment on Receipt {\n' +
      '  data\n' +
      '  rawPayload\n' +
      '}',
    variables: {
      encodedTransaction: '0x000000000000000000000000000000010000000005f5e1000000000000000000000000000000000400000000000000000000000000000001000000000000000200000000000000010000000000000000000000000000000000000000000000000000000000000000240000000000000000000000000000009835cda5c81ed1492fe4dc3696a7b39b6e94ac2c3ba4f739bafc69562061b31700000000000000042ce05bde9ada2d2c58b5eb9f08be34df19375d618626c324f75dfdbd226c8d8800000000000004300000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000057fe0ee890797038951426b07a7c0f16d135635e0bc27cae1461aae80ca6b7e6000000000000000056fb8789a590ea9c12af6fe6dc2b43f347700b049d4f823fd4476c6f366af20100000000000000032ce05bde9ada2d2c58b5eb9f08be34df19375d618626c324f75dfdbd226c8d88000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000',
      utxoValidation: false
    }
  }
}

```

Forum topics:
https://forum.fuel.network/t/how-to-configure-predicate-calls-inputs-and-outputs-in-ts-sdk/2688
https://forum.fuel.network/t/how-to-get-asset-outputs-for-amount-and-get-asset-inputs-for-amount-with-ts-dsk/2638

### Usage example

```
async function cancelOrder(
    predicate: Predicate<[]>, //not sure about type
    wallet: WalletUnlocked,
    asset0: string,
    amount0: BN,
) {
    let inputs = await predicate.getAssetInputsForAmount(asset0, amount0)
    let outputs = wallet.getAssetOutputsForAmount(wallet.address, asset0, 0);
    let scriptCall = new ScriptCallHandler(
        [], //script binary
        UnresolvedBytes.default(), //encoded args
        wallet, //account
        wallet.provider,//provider
    )
    scriptCall
        .with_inputs(inputs)
        .with_outputs(outputs)
        .tx_params(TxParameters.default().set_gas_price(1));

    await scriptCall.call()
}
```

Torres-ssf:
- Close #1020

This PR intends to simplify the use of predicates when executing custom transactions.

Currently, the most straightforward way to use a predicate in the SDK is by instantiating the `Predicate` class and executing the transfer method. This approach internally creates a transaction, fetches resources from the predicate, and adds an output `Coin` from the intended asset Id destined to the intended address.

However, this approach falls short in more complex scenarios. For example, consider a situation where you want another person to pay for the transaction fee of the previously described transaction. Using only the Predicate instance with the transfer method makes this impossible to achieve, as detailed in issue #1020.

While it is possible to achieve this by creating and setting up an instance of `ScriptTransactionRequest`, it involves manually setting up the predicate transaction inputs, as shown in the code snippet below:

```ts
const predicateUtxos = (await predicate.getResourcesToSpend([
  [amount, BaseAssetId],
])) as Coin[];

const predicateInputs: TransactionRequestInput[] = predicateUtxos.map((utxo) => ({
  id: utxo.id,
  type: InputType.Coin,
  amount: utxo.amount,
  assetId: utxo.assetId,
  owner: utxo.owner.toB256(),
  txPointer: '0x00000000000000000000000000000000',
  witnessIndex: 0,
  maturity: 0,
  predicate: predicate.bytes, // predicate's UTXOs should have the predicate field,
  predicateData: predicate.predicateData,
}));
```
The proposed solution simplifies this process by introducing the `addPredicateResources` method, as illustrated in the following code snippet:

```ts
    const predicateUtxos = await predicate.getResourcesToSpend([
      [amount, BaseAssetId],
    ]);

    request
      .addPredicateResources(predicateUtxos, predicate.bytes, predicate.predicateData)
```

The `addPredicateResources` method works similar to the `addResources` method (formerly `addPredicateResourcesInputsAndOutputs`). 

However, a separate method for adding predicate resources was implemented because, when creating predicate inputs, the predicate's bytes and data bytes must be present. 

Therefore, the `addPredicateResources` method explicitly requires at least the predicate bytes, since the data bytes are required only when the used predicate main function receives arguments.


packages/providers/src/transaction-request/transaction-request.ts
```diff
@@ -226,130 +228,188 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
...
+  }
+
+  /**
+   * Adds multiple resources to the transaction by adding coin/message inputs and change
+   * outputs from the related assetIds.
+   *
+   * @param resources - The resources to add.
+   * @returns This transaction.
+   */
+  addPredicateResource(resource: Resource, predicate: BytesLike, predicateData?: BytesLike) {
```
nedsalk, line 374: If we have a predicate that requires data, the user still wouldn't have to provide the data because it's optional, and the transaction would fail.

Why not do:
```ts
addPredicateResources(resources, predicate: Predicate)
```

and then you'll have both `predicate.bytes` and `predicate.predicateData` available and this would no longer be an issue?

apps/docs-snippets/src/utils.ts
```diff
@@ -41,7 +41,7 @@ export const getTestWallet = async (seedQuantities?: CoinQuantityLike[]) => {
   });
 
   // add the UTXO inputs to the transaction request
-  request.addResourceInputsAndOutputs(resources);
+  request.addResources(resources);
```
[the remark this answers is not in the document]
Torres-ssf: I made this change because I noticed that the original method name `addResourceInputsAndOutputs` does not completely describe the function's behavior. Specifically, it doesn't clarify the type of output being added, which could lead to confusion.

We have two possible types of inputs that are resources: `Coin` and `Message`. However, it is not clear what kind of output is being added. So since it was not completely descriptive, I've gone for a shorter name.

In order to improve readability I've implemented 3 methods: `addCoinInput`, `addMessageInput`, and `addChangeOutput`.  Each one handles specifically what their name suggests. Therefore, the `addResources` method is now just a helper method that calls these more specific methods.

Also, since I decided to add a specific add resources method for predicates, to specifically require the predicate byte and avoid users forgetting to pass it when calling `addResources`, I got under the impression that `addResourceInputsAndOutputsForPredicate` would end up being too long/descriptive.

Maybe I am just nitpicking here. I do not have a strong opinion on this, Let me hear what you guys think, if you think `addResourceInputsAndOutputs` does a better job here, I roll back the change.

packages/providers/src/transaction-request/transaction-request.ts
Torres-ssf: @nedsalk I had the same idea. 

However, the package `predicate` already makes use of the package `provider`, where the `BaseTransactionRequest` lives. So I ended up receiving the following:

Screenshot from 2023-08-30 10-26-18

nedsalk: It's weird that the `BaseTransactionRequest` says, "you can give me a predicate" via `addPredicateResource` but it itself doesn't know what a `Predicate` is (because it's got no dependency to the `predicate` package). It seems to me that the root issue is the `interfaces` package which is obfuscating all of these dependencies and we don't know what is **really** dependent on what so that we can more easily split packages, move code from one package to another, etc.

Now that I think of it, if we had the `Predicate` available, we could have used only the `addResources` method, made it generic and created a special input type for this method which checks if the source of the resources is a predicate and, if it is, mandate the passing of that predicate as well.

nedsalk: It's weird that the `BaseTransactionRequest` says, "you can give me a predicate" via `addPredicateResource` but it itself doesn't know what a predicate is (because it's got no dependency to the `predicate` package). It seems to me that the root issue is the `interfaces` package which is obfuscating all of these dependencies and we don't know what is really dependent on what so that we can more easily split packages, move code from one package to another, etc.

Dhaiwat10 (dismissed):

nedsalk:
I've created a potential improvement upon this PR, see it here: #1219.

Torres-ssf:
> I've created a potential improvement upon this PR, see it here: #1219.

I've made the `Predicate` class implement the abstract class `AbstractPredicate`. 

I believe it was created with this purpose but was currently not used. Since `AbstractPredicate` comes from the `interfaces` package, there is no circular dependency in using it on the `providers` package.

Therefore, I've refactored `addPredicateResource` to receive an `AbstractPredicate` instead of the predicate bytes and predicate data. This change will make the code more maintainable and easier to understand because it leverages the existing `AbstractPredicate` interface and avoids passing around raw bytes and data.

@nedsalk thanks for the proposal on https://github.com/FuelLabs/fuels-ts/pull/1219. The idea is really good, however, this approach is way too hacky

Dhaiwat10 (approved):

arboleya (approved):

PR #1219, opened 2023-08-30
feat!: Remain with only one method to add all types of resources

nedsalk:
This PR was initially merging into #1216 as a suggestion for a better way to add resources so that there is no need for the additional `addPredicateResource`/`addPredicateResources` method pair. However, that PR got merged so the base of this PR is now `master`.

The problem in the current implementation can be shown in the examples below:
```ts
// example 1
const predicate1 = new Predicate(...);
const predicate2 = new Predicate(...);

const request = new ScriptTransactionRequest(...);
const predicate1Resources = await predicate1.getResourcesToSpend(...);

// problem: passing first predicate's resources with the second predicate
request.addPredicateResources(predicate1Resources, predicate2);

// example 2
const wallet = new WalletUnlocked(...);
const walletResources = await wallet.getResourcesToSpend(...);

// problem: passing in a wallet's resources with a predicate
request.addPredicateResource(walletResources, predicate1);

// example 3
const predicate1Coins = predicate1Resources.filter(isCoin);

// problem: adding coins of first predicate alongside second predicate
// The same problem exists in addMessageInput as well 
// and could be applied to the wallet resource + predicate problem also
predicate1Coins.forEach(coin => request.addCoinInput(coin, predicate2));
```


packages/wallet/src/account.ts
```diff
@@ -86,11 +86,17 @@ export class Account extends AbstractAccount {
    * @param excludedIds - IDs of resources to be excluded from the query.
    * @returns A promise that resolves to an array of Resources.
    */
-  async getResourcesToSpend(
+  async getResourcesToSpend<T extends this>(
```
nedsalk, line 89: This is a neat trick that lets the compiler infer which subclass is used to call the method. If it's `Predicate` calling it, the return would be `AccountResource<Predicate>[]`, and if it's `WalletUnlocked`, it would be `AccountResource<WalletUnlocked>[]`.

packages/wallet/src/account.ts
```diff
@@ -33,7 +33,7 @@ import {
 } from './utils';
 
 type TxParamsType = Pick<TransactionRequestLike, 'gasLimit' | 'gasPrice' | 'maturity'>;
-
+export type AccountResource<T extends Account> = Resource & { account: ThisType<T> };
```
nedsalk, line 36: I was thinking of defining this where `Resource` is defined (in the `providers` package), but that's not the place for it as `AccountResource` is related to accounts specifically.

packages/providers/src/transaction-request/transaction-request.ts
```diff
@@ -336,12 +341,19 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
...
+   * change output for the related assetId or predicate, if one of it was not added yet.
    *
    * @param resource - The resource to add.
    * @returns This transaction.
    */
-  addResource(resource: Resource) {
+  addResource<T extends AbstractAccount>(resource: AccountResource<T>) {
+    const account = resource.account;
+
+    if ((account as AbstractPredicate).bytes) {
```
nedsalk, line 352: It's not possible to use `account instanceof AbstractPredicate` because `Predicate` itself just implements `AbstractPredicate`, it doesn't extend it; it extends `Account`.

packages/providers/src/transaction-request/transaction-request.ts
```diff
@@ -255,12 +260,12 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
...
    * @param predicateData - Predicate data bytes.
    */
-  addCoinInput(coin: Coin, predicate?: AbstractPredicate) {
+  addCoinInput<A extends AbstractAccount>(coin: AccountResource<Coin, A>) {
     const { assetId, owner, amount } = coin;
 
     let witnessIndex;
 
-    if (predicate) {
+    if ((coin.account as AbstractPredicate).bytes) {
```
nedsalk, line 268: It's not possible to use `account instanceof AbstractPredicate` because `Predicate` itself just implements `AbstractPredicate`, it doesn't extend it; it extends `Account`.

packages/providers/src/resource.ts
```diff
@@ -25,6 +26,9 @@ export type RawMessage = {
 
 export type RawResource = RawCoin | RawMessage;
 export type Resource = Coin | MessageCoin;
+export type AccountResource<R extends Resource, T extends AbstractAccount> = R & {
```
nedsalk, line 29: Ideally this would be defined wherever `Account` is defined, but then we'd have a problem with cyclic dependencies as `@fuels-ts/wallet`, where `Account` is defined, already depends on `@fuels-ts/providers`.
This type can easily be sorted out whenever the `interfaces` package's removal of the `AbstractAccount` happens, because that removal would entail a lot of reordering and this type would be caught in it.

Torres-ssf (changes requested):
Great job @nedsalk and thanks for suggesting this :pray:. I appreciate the thought you've put into this.

It does definately makes more sense having just `addResources` instead of having `addPredicateResources` as well.

However, I believe there's a more straightforward approach. By adding the props `predicate` and `predicateData` to both `Coin` and `MessageCoin` types, we can eliminate the need to inject the entire `Account` object within `getResourcesToSpend`.

We won't need to change the `Resources` type

We won't be needing to change the function signature for `addResource` and  `addResources`.

Moreover, this proposed approach could help us to reduce and isolate the type gymnastics to just a single step, inside `getResourcesToSpend`

packages/providers/src/resource.ts
```diff
@@ -1,4 +1,5 @@
 import type { BytesLike } from '@ethersproject/bytes';
+import type { AbstractAccount } from '@fuel-ts/interfaces';
 
 import type { Coin } from './coin';
```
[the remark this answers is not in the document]
nedsalk: I've created the `CoinPredicate` type.

nedsalk:
@Torres-ssf Adding the `predicate` and `predicateData` onto a `Coin` directly inside of `predicate.getResourcesToSpend` was problematic; I added two tests which showcase the problem. Instead, I put that data behind a `getPredicateContent` callback.

Also, instead of adding that info onto a `Coin` directly, inspired by the rust-sdk I created a `CoinPredicate` and `MessageCoinPredicate` and updated the `Resource` type accordingly.

Additionally, I made `addCoinInput` and `addMessageInput` private because there's no need to expose them as users can use `addResources`. I also made `addResource` private because IMO there's no need to have both `addResource` and `addResources` - it's double the work to maintain (docs, types, etc) with not much benefit. Rather, if the user wants to add a _single_ resource, they can just do `addResources([singleResource])`.

packages/providers/src/resource.ts
```diff
@@ -1,4 +1,5 @@
...
 import type { MessageCoin } from './message';
```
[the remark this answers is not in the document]
nedsalk: I created the `MessageCoinPredicate` type.

packages/providers/src/transaction-request/transaction-request.ts
```diff
@@ -340,7 +342,7 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
    * @param resource - The resource to add.
    * @returns This transaction.
    */
-  addResource(resource: Resource) {
+  private addResource(resource: Resource) {
```
Torres-ssf, line 345: Why we are making `addResource` private?

packages/providers/src/resource.ts
```diff
@@ -3,28 +3,39 @@ import type { BytesLike } from '@ethersproject/bytes';
...
+  getPredicateContent: () => { predicate: Uint8Array; predicateData: Uint8Array };
+}
+
+export interface MessageCoinPredicate extends MessageCoin {
+  getPredicateContent: () => { predicate: Uint8Array; predicateData: Uint8Array };
+}
 
 export type RawResource = RawCoin | RawMessage;
-export type Resource = Coin | MessageCoin;
+export type PredicateResource = CoinPredicate | MessageCoinPredicate;
```
Torres-ssf, line 35: Why do we need this `PredicateResource`? 

Wouldn't be simpler to just set `CoinPredicate` and `MessageCoinPredicate` to be a `Resource` as well?

packages/providers/src/resource.ts
```diff
@@ -3,28 +3,39 @@ import type { BytesLike } from '@ethersproject/bytes';
...
-};
+}
+
+export interface CoinPredicate extends Coin {
+  getPredicateContent: () => { predicate: Uint8Array; predicateData: Uint8Array };
+}
+
+export interface MessageCoinPredicate extends MessageCoin {
+  getPredicateContent: () => { predicate: Uint8Array; predicateData: Uint8Array };
+}
```
Torres-ssf, line 32: Why do we need this method? 

We can simply make `Coin` and `MessagePredicate` extend `Coin` and `MessageCoin` and define both `predicate` and `predicateData` props.

packages/providers/src/transaction-request/transaction-request.ts
```diff
@@ -297,34 +297,36 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
...
-      type: InputType.Message,
-      sender: sender.toB256(),
-      recipient: recipient.toB256(),
-      amount,
-      witnessIndex,
-      predicate: predicate?.bytes,
-      predicateData: predicate?.predicateData,
-    };
+      input.witnessIndex = witnessIndex;
+    }
```
Torres-ssf, line 329: If we move to the approach of making both `CoinPredicate` and `MessageCoinPredicate` be types of `Resource`, removing the `getPredicateContent` method, and given that the param `message` is being spread, I reckon there won't be a need to specifically set `predicate` and `predicateData` within the input object.

packages/providers/src/transaction-request/transaction-request.ts
```diff
@@ -254,34 +252,36 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
...
       owner: owner.toB256(),
       amount,
       assetId,
       txPointer: '0x00000000000000000000000000000000',
-      witnessIndex,
-      predicate: predicate?.bytes,
-      predicateData: predicate?.predicateData,
+      witnessIndex: 0,
+      predicate: undefined,
+      predicateData: undefined,
```
Torres-ssf, line 267: Maybe I am missing something here, but I reckon that by not defining these 2 props, their value would be `undefined` already.

Torres-ssf (commented):
@nedsalk I've left some questions here to better understand your motivations behind it.

packages/providers/src/resource.ts
nedsalk: If we assign `predicate` and `predicateData` to a predicate's resources the moment we fetch it in `getResourcesToSpend`, updating that predicate's data via `predicate.setData` before adding those resources to a transaction request  won't affect those resources' `predicateData`. Returning those properties via a callback would allow that.

```ts
const predicate = new Predicate(...);
const resources = await predicate.getResourcesToSpend(...); 

// someData would not be reflected in the resources above 
// if we set predicateData right away upon fetching,
// but with the getPredicateContent callback, 
// you can change the data as much as you like before adding it to transactionRequest,
// because it's behind a callback which holds a reference to the predicate's properties.
predicate.setData(someData);

transactionRequest.addResources(resources);
```

This idea is illustrated in the two tests added in `packages/fuel-gauge/src/predicate-resources-with-custom-transaction.test.ts`.

packages/providers/src/resource.ts
nedsalk: If we assign `predicate` and `predicateData` to a predicate's resources the moment we fetch it in `getResourcesToSpend`, updating that predicate's data via `predicate.setData` before adding those resources to a transaction request  won't affect those resources' `predicateData`. Returning those properties via a callback would allow that.

```ts
const predicate = new Predicate(...);
const resources = await predicate.getResourcesToSpend(...); 

// someData would not be reflected in the resources above 
// if we set predicateData right away upon fetching,
// but with the getPredicateContent callback, 
// you can change the data as much as you like before adding it to transactionRequest,
// because it's behind a callback which holds a reference to the predicate's properties.
predicate.setData(someData);

transactionRequest.addResources(resources);
```

This idea is illustrated in the two tests added in `packages/fuel-gauge/src/predicate-resources-with-custom-transaction.test.ts`.

packages/providers/src/transaction-request/transaction-request.ts
nedsalk: Indeed. I just wanted it to be explicit - a preference.

packages/providers/src/transaction-request/transaction-request.ts
nedsalk: It all depends on how we deal with [this](https://github.com/FuelLabs/fuels-ts/pull/1219#discussion_r1360678241).

packages/providers/src/transaction-request/transaction-request.ts
```diff
@@ -297,34 +297,36 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
...
-    let witnessIndex;
+    const input: MessageTransactionRequestInput = {
+      ...message,
+      type: InputType.Message,
+      sender: sender.toB256(),
+      recipient: recipient.toB256(),
+      amount,
+      witnessIndex: 0,
+      predicate: undefined,
+      predicateData: undefined,
```
Torres-ssf, line 313: I reckon we can remove this explicit undefined set

packages/providers/src/transaction-request/transaction-request.ts
Torres-ssf, line 267: I reckon we can remove this explicit undefined set

Torres-ssf (changes requested):

packages/providers/src/transaction-request/transaction-request.ts
nedsalk: Done

packages/providers/src/transaction-request/transaction-request.ts
nedsalk: Done

Torres-ssf (dismissed):

packages/predicate/src/predicate.ts
```diff
@@ -61,46 +60,26 @@ export class Predicate<ARGS extends InputValue[]> extends Account implements Abs
...
-   * @returns A promise that resolves to the call result.
-   */
-  simulateTransaction(transactionRequestLike: TransactionRequestLike): Promise<CallResult> {
-    const transactionRequest = this.populateTransactionPredicateData(transactionRequestLike);
-    return super.simulateTransaction(transactionRequest);
+    return resources.map((r) => ({
+      ...r,
+      getPredicateContent: () => ({ predicate: this.bytes, predicateData: this.predicateData }),
+    }));
   }
```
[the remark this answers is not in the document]
nedsalk: This is the only place where this callback should be implemented so it's necessary.

nedsalk:
@Torres-ssf I've got these [failing tests](https://github.com/FuelLabs/fuels-ts/actions/runs/6781522988/job/18431976145?pr=1219#step:11:3708) that I tried to fix but couldn't find the root cause. I'll get back to it asap.

nedsalk:
#1428 fixes the remaining failing tests. Before this PR, all transactions with _only_ a predicate coin/message would have an _(unnecessary)_ `0x` witness in their `witnesses` field who's `offset` field was then used to get the chargeable bytes (and thus the fee) of the transaction. Because that witness isn't being added anymore due to the refactoring done in `TransactionRequest.addCoinInput/addMessageInput`, the fee calculation tests for predicates failed.

nedsalk:
@camsjams Thanks for the review! I'll postpone the merge of this branch until after beta-5, however, as I don't want any unknown unknowns from this branch to mess it up.

packages/predicate/src/predicate.ts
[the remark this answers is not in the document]
nedsalk: Whoops, GitHub showed me this question in such a way that I thought it was aimed at `getPredicateContent` on the right side, not these removed methods.

Anyways, I don't see a reason why we wouldn't remove them as `Predicate.populateTransactionPredicateData` was removed due to not being necessary, and these methods are just wrappers around the base method with the additional call to `Predicate.populateTransactionPredicateData`.

nedsalk:
This PR's logic on adding predicate data is in conflict with #1455, because `populateTransactionPredicateData` got additional significance. I'll put this on wait until #1437 is merged. In a nutshell, `populateTransactionPredicateData` shouldn't exist because there is a presupposition that a transaction containing the predicate will be sent via `predicate.sendTransaction`, which isn't necessarily the case as it can also be sent via `provider.sendTransaction`.

Torres-ssf:
@nedsalk Do you think it will be better to close this PR and open a new one, from a fresh branch? 

I reckon this will make it easy for others to go through the review process.

nedsalk:
I was thinking of it as well, but decided to not close it due to the amount of comments. However, now I realize that this PR depends on some other "in the air" stuff like #1506, so I'll close it to not pollute the PR list until those other things are resolved.

PR #1220, opened 2023-08-30
fix: stateRoot calculation on sparse merkle tree 

luizstacio:
This PR fixes the calculation of the contract's state root. 

Before this PR, there was a bug where it was not possible to deploy any contracts where a storage key is was initialized with the value `0`. We managed to track this problem down to the implementation of the `update` and `delete` functions in our `SparseMerkleTree` class. The logic was incorrect, and it has now been fixed to replicate the logic used by our Rust SDK.


cold-briu:
Some context: 

Reported in [sway/4634](https://github.com/FuelLabs/sway/issues/4634)

Storage initialization values won't be automatically set by the sdk since the initialization values are output into a separate json file from the abi. 

The sdk should automatically include storage slots. Otherwise, an additional config step is required to deploy the contract.

Torres-ssf (approved):

packages/utils/src/test-utils/getForcProject.ts
```diff
@@ -32,6 +35,16 @@ export const getProjectAbi = (params: IGetForcProjectParams) => {
...
 };
 
+export const getProjectStorageSlots = (params: IGetForcProjectParams) => {
+  const projectPath = getProjectStorageSlotsPath(params);
+  try {
+    const storageSlots = JSON.parse(readFileSync(projectPath, 'utf-8'));
+    return storageSlots;
+  } catch (error) {
+    return [];
+  }
```
arboleya, line 45: Could we check for `existSync(projectPath)` instead of the try/catch?

arboleya (changes requested):

apps/docs-snippets/src/guide/contracts/multicalls.test.ts
```diff
@@ -33,7 +33,9 @@ describe(__filename, () => {
     );
 
     echoContract = await factory1.deployContract();
-    counterContract = await factory2.deployContract();
+    counterContract = await factory2.deployContract({
+      storageSlots: counterArtifacts.storageSlots,
+    });
```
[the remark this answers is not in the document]
Dhaiwat10: @arboleya it seems like we already have a test covering this in `fuel-gauge` (https://github.com/FuelLabs/fuels-ts/blob/4175de71dbe1cd5df9967d17afe8e7779c246225/packages/fuel-gauge/src/storage-test-contract.test.ts#L25) but it didn't have any counters that were initialized with 0. I have added that change now

arboleya (dismissed):

apps/docs-snippets/src/guide/contracts/multicalls.test.ts
```diff
@@ -33,7 +33,9 @@ describe(__filename, () => {
     );
 
     echoContract = await factory1.deployContract();
-    counterContract = await factory2.deployContract();
+    counterContract = await factory2.deployContract({
```
luizstacio, line 36: I think we should change the counter contract and remove the try_read and the unwrap_or to make sure the initialization worked.

Dhaiwat10: @luizstacio I removed this from the contract and now some new unrelated test that uses this contract started failing haha

Dhaiwat10: It passes if I pass in the storage slots imported from the storage slots json file

Dhaiwat10: Is this expected behaviour? Should it fail without proper initialization when deploying the contract?

arboleya: You managed to fix it, right? I see the tests are passing now.

Can we merge and release it via [#1232](https://github.com/FuelLabs/fuels-ts/pull/1232)?

luizstacio: @Dhaiwat10 yeah is expected that the other tests fails

arboleya (approved):

Dhaiwat10 (approved):

PR #1223, opened 2023-08-31
fix: docs release workflow

danielbate:
The current docs release workflow [is failing](https://github.com/FuelLabs/fuels-ts/actions/runs/6030751395/job/16363160462). This appears to be due to not having the correctly fetched version of `master`, as `docs` is based off master and should merge without conflict. This PR resolves that workflow.


.github/workflows/release.yaml
```diff
@@ -100,18 +100,19 @@ jobs:
         uses: actions/checkout@v3
         with:
           ref: docs
-        
+          fetch-depth: 0
```
danielbate, line 103: I appreciate this is not an optimal approach could become very slow, however I have tried a few different fetch configurations without success (in the CI). I would like to use this configuration for now to unblock the workflow and investigate in isolation.

arboleya (approved):

Torres-ssf (approved):

Issue #1156, opened 2023-08-01
Standardize error handling across all packages


arboleya:
This is a follow-up issue after:
 - #1110

# Documentation

Each package implementation should also document the added `ErrorCode`.

# Packages

 - [x] abi-coder
 - [x] abi-typegen
 - [x] address #1153
 - [x] contract
 - [x] crypto
 - [x] errors #1153
 - [x] forc
 - [x] fuel-core
 - ~fuel-gauge~
 - [x] fuels
 - [x] hasher
 - [x] hdwallet
 - [x] interfaces
 - [x] math
 - [x] merkle
 - [x] mnemonic
 - [x] predicate
 - [x] program
 - [x] providers
 - [x] script
 - [x] signer
 - ~testcases~
 - [x] transactions
 - [x] utils
 - [x] versions
 - [x] wallet
 - [x] wallet-manager
 - [x] wordlist

PR #1228, opened 2023-08-31
chore: simplifying fuel error implementation

arboleya:
Re-writing of #1193 following internal conversations.

 1. Simplify `expectToThrowFuelError` test utility
 1. Re-write all tests accordingly
 1. Improve package README

Bonus:
 1. Adding new `eslint` rule to prevent floating promises ([here](https://github.com/FuelLabs/fuels-ts/blob/83e954f3ddf7acb6e5e96d68ed107d377584c302/.eslintrc.js#L54))
 1. Fixing all floating promises

PR #1230, opened 2023-09-01
refactor: standardize error handling across all packages

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/1156
- Close: #1234


arboleya:
Thanks a lot for taking on this! Seriously.

Yesterday, while I was working on https://github.com/FuelLabs/fuels-ts/pull/1228, I remembered what @danielbate said [here](https://github.com/FuelLabs/fuels-ts/pull/1153#pullrequestreview-1563902845).

I initially [suggested](https://github.com/FuelLabs/fuels-ts/pull/1153#discussion_r1285724857) leaving it for later, but a whole month went by and _later_ became the **present**. Since you rolled up your sleeves and took the matter upon yourself, I'd like to know your thoughts.

More specifically, I sense a bit of friction being created according to the first pseudo-code example I wrote [here](https://github.com/FuelLabs/fuels-ts/issues/1110#issuecomment-1646311520). You can see that the rationale for i18n would be more coherent if we had only a single message per error code, just like @danielbate's suggested.

We don't need to go in this direction necessarily.

If we don't, I can adjust these docs [here](https://github.com/FuelLabs/fuels-ts/blob/aa/chore/simplifying-fuel-error-implementation/packages/errors/README.md#external-usage).

Let's chat.

Torres-ssf:
> Let's chat.

@arboleya Absolutely, I'm glad that you brought this up. It seems to me that having a single message per error code, as @danielbate suggested, would indeed make more sense to the internationalization (i18n) process. The more fragmented and diverse the error messages, the more complex it will be to manage translations and ensure consistency.

I believe that there are clear benefits in moving in this direction. It will help us maintain a clean and organized codebase and will probably save us time and effort down the line.

If we are all on the same page, I can refactor the `FuelError` and start the standardization of error codes with their respective desired messages. This move will set the stage for a more coherent and efficient i18n process.

arboleya:
On the other hand, some cases may be problematic.

A flex approach may be best; my only concern is not over-engineering it.

Let's think this through and see what comes up.

nedsalk:
Grouping multiple failure modes under one error code sometimes simply makes sense and is cleaner and more understandable (e.g. [here](https://github.com/FuelLabs/fuels-ts/pull/1217#discussion_r1309787961)). I believe that having the ability to group error messages under one error code would result in a healthier code base because the developer's mindset would be "Is this essentially the same failure mode (error code) but just with a different reason (error message)?", which would probably lead to better code structure, as they're thinking about failure mode encapsulation behind as few error codes as it makes sense. It would also make it easier to understand the failure modes of the domain as they're grouped under those error codes instead of being one-offs for each specific situation (e.g. the linked example above: `INVALID_MULTICALL` vs `ONLY_ONE_HEAP_TYPE_CALL_ALLOWED` & `HEAP_TYPE_RETURN_MUST_BE_LAST`).

The problem regarding internationalization can be solved by grouping error message _keys_ under error codes and using these keys when instantiating a `FuelError` instead of a plain string message ( `new FuelError(code, messageKey)` instead of `new FuelError(code, message)` ). This would allow us to both simply group multiple error messages under one error code and also provide easy internationalization capabilities. I've created a functioning POC in [this branch](https://github.com/FuelLabs/fuels-ts/tree/ns/feat/add-error-message-keys). I didn't modify the `fuel-error.test.ts` files much as it's depending on `expectToThrowFuelError` which is undergoing a change [here](https://github.com/FuelLabs/fuels-ts/pull/1228). Some changes of the type of input on `expectToThrowFuelError` would probably need to happen after that PR merges because of the change in `FuelError` to `FuelError<ErrorCode, MessageKey>`, but nothing special.

Based off of @arboleya's [examples](https://github.com/FuelLabs/fuels-ts/issues/1110#issuecomment-1646311520), this is what it would look like:

Internal usage:
```ts
import { FuelError, ErrorCode } from '@fuel-ts/error';

export function singleImport() {
  throw new FuelError(FuelError.CODES.INVALID_ADDRESS, 'unknown-format');
}

export function createAddress(address: string) {
  throw new FuelError(ErrorCode.INVALID_ADDRESS, 'invalid-bech32', { address });
}

```

External usage:
```ts
import { FuelError, ErrorCode, Address, errorMessages } from 'fuels';

type Locale = 'PT_BR' | 'BS_BA' | 'EN_US';

const currentLocale: Locale = 'PT_BR';

type I18N = {
  [L in Locale]: {
    [EC in ErrorCode]: {
      [K in keyof (typeof errorMessages)[EC]]: string;
    };
  };
};

const i18nDict: I18N = {
  EN_US: errorMessages,
  PT_BR: {
    [ErrorCode.INVALID_ADDRESS]: {
      'invalid-bech32': 'Endereço Bech32 inválido.',
      'unknown-format':
        'Formato de endereço desconhecido: apenas Bech32, B256 ou Public Key (512) são suportados.',
    },
  },
  BS_BA: {
    [ErrorCode.INVALID_ADDRESS]: {
      'invalid-bech32': 'Nevažeća Bech32 adresa.',
      'unknown-format':
        'Nepoznati adresni format: podržani su samo Bech32, B256 i Public Key (512).',
    },
  },
};


function translateError(e: unknown) {
  const { code, messageKey } = FuelError.parse(e);
  return i18nDict[currentLocale][code][messageKey];
}

(function main() {
  try {
    const address = new Address("fuel-invalid-bech32");
    console.log(address);
  } catch (e) {
    const prettyError = translateError(e);
    console.log({ prettyError });
  }
})();
```

nedsalk:
Also, based off of the reasoning mentioned [in the comment above](https://github.com/FuelLabs/fuels-ts/pull/1230#issuecomment-1704374986) regarding error codes and what I've mentioned in the original `FuelError` PR's [description](https://github.com/FuelLabs/fuels-ts/pull/1153):

> Note that updating the errors throughout the fuels-ts packages is outside of the scope of this PR, as that isn't a small endeavor because careful thought needs to be put into error code naming. Also, sometimes the actual code implementation where an error is thrown might not 100% reflect the error code being used, which might necessitate some refactoring to have it make more sense.

I think that this PR should be split into per-package PRs where we can discuss the choice of error code names, messages, the refactoring, etc. in a more focused manner. We don't have to rush to finish every package right away because 
> sometimes the actual code implementation where an error is thrown might not 100% reflect the error code being used, which might necessitate some refactoring to have it make more sense.

Even if refactoring would sometimes be out of scope due to its complexity, at least new issues could be created where the one who did a specific package's error codes could explain why something needs refactoring, that they think that there's the ability to do more of it but it requires substantial effort which would potentially look something like this or that, etc.

Torres-ssf:
@arboleya, I'll address the modifications to the `FuelError` in a separate PR. 

For the current PR, I've standardized the usage of `FuelError` throughout the SDK and enhanced several error messages.

packages/errors/src/index.ts
```diff
@@ -1,2 +1,4 @@
 export { ErrorCode } from './error-codes';
 export { FuelError } from './fuel-error';
+export * from './test-utils/safeExec';
+export * from './test-utils/expect-to-throw-fuel-error';
```
arboleya, line 4: These should be exported only in the `test-utils.ts` file.

packages/errors/src/test-utils.ts
```diff
@@ -1 +0,0 @@
-export * from './test-utils/expect-to-throw-fuel-error';
```
arboleya, line 1: This export should remain in this file, following [this](https://github.com/FuelLabs/fuels-ts/pull/1230/files/2a16499381e169379c3625cf409085911b1ec298..b29792c3b4d5106f7335110d0ebfaabc1cdee878#r1322872516).

packages/wallet/src/wallet.test.ts
```diff
@@ -2,7 +2,7 @@ import { BaseAssetId } from '@fuel-ts/address/configs';
 import { bn } from '@fuel-ts/math';
 import type { TransactionRequestLike, TransactionResponse } from '@fuel-ts/providers';
 import { transactionRequestify, Provider } from '@fuel-ts/providers';
-import { safeExec } from '@fuel-ts/utils/test-utils';
+import { safeExec } from '@fuel-ts/errors';
```
arboleya: There's a check failure in line 5. Sorting the imports should fix it.

arboleya (changes requested):
Good stuff.

I left some notes; they should all be simple to fix.

arboleya (dismissed):

arboleya (dismissed):

arboleya (approved):

arboleya (approved):

danielbate (approved):
👏🏻

arboleya (approved):

danielbate (approved):

PR #1246, opened 2023-09-12
test: add throw scenarios to `simulate` and `dryRun`

danielbate:
[The following forum post ](https://forum.fuel.network/t/cant-make-dryrun-with-wallet-that-doesnt-have-eth-on-balance/3068) asks whether a wallet with no funding should be able to perform a read only call. Current specs detail that this is not possible, as per our own TS documentation.

This PR will add some test cases to the demo typegen application to directly replicate the above forum post, as well as additional expectations to fuel-gauge for both `simulate` and `dryRun` throw expectations. This is already explained in the documentation.


arboleya (approved):
@danielbate You could have used the `safeExec` utility in these try/catch validations, but as you are not yet dealing with `FuelErrors`, these expectations will probably need to be aligned with:
 - https://github.com/FuelLabs/fuels-ts/pull/1230

> cc @Torres-ssf

digorithm (approved):
:shipit:

danielbate:
@arboleya good idea 👍🏻

arboleya (approved):

Dhaiwat10 (approved):

Torres-ssf (approved):

PR #1247, opened 2023-09-12
feat: improve predicate vector input support

### LINKED ISSUE #1083: Improve Predicate Vector input support
### Motivation

Vector code works partially with Predicates. The vectors can be encoded and sent to the the Sway programs well part of the time, but there is an offset issue that occurs occasionally when trying to send the payloads.

See #1046

### Usage example

The offset here isn't calculating properly, see the `17`: https://github.com/FuelLabs/fuels-ts/blob/master/packages/predicate/src/predicate.ts#L117-L122

### Possible implementations
Start with running these skipped tests:
- https://github.com/FuelLabs/fuels-ts/blob/master/packages/fuel-gauge/src/vector-types.test.ts#L134 (see its sway file: https://github.com/FuelLabs/fuels-ts/blob/master/packages/fuel-gauge/fixtures/forc-projects/predicate-vector-types/src/main.sw)
- And https://github.com/FuelLabs/fuels-ts/blob/master/packages/fuel-gauge/src/predicate/predicate-arguments.test.ts#L335

danielbate:
Closes #1083


packages/predicate/src/predicate.ts
```diff
@@ -116,11 +117,7 @@ export class Predicate<ARGS extends InputValue[]> extends Account implements Abs
     const paddedCode = new ByteArrayCoder(this.bytes.length).encode(this.bytes);
 
     const OFFSET =
-      VM_TX_MEMORY +
-      TRANSACTION_SCRIPT_FIXED_SIZE +
-      TRANSACTION_PREDICATE_COIN_FIXED_SIZE +
-      paddedCode.byteLength -
-      17;
+      VM_TX_MEMORY + SCRIPT_FIXED_SIZE + INPUT_COIN_FIXED_SIZE + paddedCode.byteLength + WORD_SIZE;
```
danielbate, line 120: `-17` was a red herring here. In the `packages/fuel-gauge/src/predicate/predicate-arguments.test.ts', this test was passing once the `predicateGasUsed` fixed size was added to `INPUT_COIN_FIXED_SIZE`. However, only because `-17` was pointing to a previous byte that also referenced the expected number of `43`, which was pretty lucky. Rather than the required `+2` bytes that was needed to correctly offset to `predicateData` - deduced this by using the `chunkByBytes` helper and affirmed by the fact that the `packages/fuel-gauge/src/vector-types.test.ts` is also now passing (with some changes to the sway predicate to ensure that the tx could actually validate).

However, now that both these tests are passing, and the fixed sizes seem to match that of the Rust SDK (VM TX memory = 10240, script fixed size = 104, Input coin fixed size = 176 + predicate length + predicate data length), I'm having trouble deducing what the additional byte relates to? 🤔

packages/predicate/src/predicate.ts
```diff
@@ -116,11 +117,7 @@ export class Predicate<ARGS extends InputValue[]> extends Account implements Abs
     const paddedCode = new ByteArrayCoder(this.bytes.length).encode(this.bytes);
 
     const OFFSET =
-      VM_TX_MEMORY +
-      TRANSACTION_SCRIPT_FIXED_SIZE +
-      TRANSACTION_PREDICATE_COIN_FIXED_SIZE +
-      paddedCode.byteLength -
-      17;
+      VM_TX_MEMORY + SCRIPT_FIXED_SIZE + INPUT_COIN_FIXED_SIZE + WORD_SIZE + paddedCode.byteLength;
```
danielbate, line 120: `-17` was a red herring here. In the `packages/fuel-gauge/src/predicate/predicate-arguments.test.ts`, this test was passing once the `predicateGasUsed` fixed size was added to `INPUT_COIN_FIXED_SIZE`. However, only because `-17` was pointing to a previous byte that also referenced the expected number of `43`, which was pretty lucky. Rather than the required `+1` byte that was needed to correctly offset to `predicateData` - deduced this by using the `chunkByBytes()` helper and affirmed by the fact that the `packages/fuel-gauge/src/vector-types.test.ts` is also now passing (with some changes to the sway predicate to ensure that the tx could actually validate).

However, now that both these tests are passing, and the fixed sizes seem to match that of the Rust SDK (VM TX memory = 10240, script fixed size = 104, Input coin fixed size = 176 + predicate length + predicate data length), I'm having trouble deducing what the additional byte relates to? 🤔 If anyone has time to double check my numbers against the specs when reviewing that would be really helpful.

danielbate: Thanks for checking @camsjams, and yes that was the one I was alluding to. I'm also happy with that outcome so will resolve the conversation.

arboleya (approved):

Issue #1245, opened 2023-09-12
Dry run fails `Error: Cannot read properties of undefined (reading 'status')`


sdankel:
### What version of fuels-ts are you using?

0.56.1

### Steps to Reproduce

When it's a dryRun we don't receive the gqlTransaction but only receipts

![image](https://github.com/FuelLabs/fuels-ts/assets/47993817/80dbed49-e667-46a5-9314-01aabd5db261)

1. On this branch: https://github.com/FuelLabs/sway-playground/pull/34
2. Run the app, set toolchain to `nightly` in the dropdown, and add this sample contract:

```
// ANCHOR: body
contract;

use std::hash::*;

////////////////////////////////////////
// Event declarations
////////////////////////////////////////
//
// Events allow clients to react to changes in the contract.
// Unlike Solidity, events are simply structs.
//
/// Emitted when a token is sent.
struct Sent {
    from: Address,
    to: Address,
    amount: u64,  
}

////////////////////////////////////////
// ABI method declarations
////////////////////////////////////////
/// ABI for a subcurrency.
abi Token {
    // Mint new tokens and send to an address.
    // Can only be called by the contract creator.
    #[storage(read, write)]
    fn mint(receiver: Address, amount: u64);

    // Sends an amount of an existing token.
    // Can be called from any address.
    #[storage(read, write)]
    fn send(receiver: Address, amount: u64);
}

////////////////////////////////////////
// Constants
////////////////////////////////////////
/// Address of contract creator.
const MINTER = Address::from(0x9299da6c73e6dc03eeabcce242bb347de3f5f56cd1c70926d76526d7ed199b8b);

////////////////////////////////////////
// Contract storage
////////////////////////////////////////
// Contract storage persists across transactions.
storage {
    balances: StorageMap<Address, u64> = StorageMap::<Address, u64> {},
}

////////////////////////////////////////
// ABI definitions
////////////////////////////////////////
/// Contract implements the `Token` ABI.
impl Token for Contract {
    #[storage(read, write)]
    fn mint(receiver: Address, amount: u64) {
        let sender = msg_sender().unwrap();
        let sender: Address = match sender {
            Identity::Address(addr) => {
                assert(addr == MINTER);
                addr
            },
            _ => revert(0),
        };

        // Increase the balance of receiver
        storage.balances.insert(receiver, storage.balances.get(receiver).try_read().unwrap_or(0) + amount);
    }

    #[storage(read, write)]
[...]

PR #1250, opened 2023-09-13
fix: `ScriptResultDecoderError` on dry run calls

Torres-ssf:
- Close https://github.com/FuelLabs/fuels-ts/issues/1245

### DIFF
--- a/packages/fuel-gauge/src/contract.test.ts
+++ b/packages/fuel-gauge/src/contract.test.ts
@@ -965,5 +965,13 @@ describe('Contract', () => {
         })
         .dryRun<BN>()
     ).rejects.toThrowError('Expected returnReceipt');
+
+    await expect(
+      invocationScope
+        .txParams({
+          gasLimit,
+        })
+        .simulate<BN>()
+    ).rejects.toThrowError('Expected returnReceipt');
   });
 });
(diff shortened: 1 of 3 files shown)


packages/fuel-gauge/src/contract.test.ts
```diff
@@ -946,4 +946,24 @@ describe('Contract', () => {
 
     expect(finalBalance).toBe(initialBalance + amountToContract);
   });
+
+  it('should ensure ScriptResultDecoderError works for dryRun and simulate calls', async () => {
```
danielbate, line 950: Says `simulate` but we only seem to `dryRun` in the test case. I appreciate these are coupled at the moment but they could be decoupled at some point, is it worth having a separate test case for `simulate`?

danielbate, line 950: Says `simulate` but we only seem to `dryRun` in the test case. I appreciate these are coupled at the moment but they could be decoupled at some point, is it worth having a separate test case for `simulate`?

Torres-ssf: I've just improved the test case. But at the end of the day both `dryRun` and `simulate` will run the `operation.dryRun`.

nedsalk (approved):

arboleya (approved):
Awesome!

PR #1252, opened 2023-09-13
feat: add `date` to `TransactionSummary`

Torres-ssf:
This PR adds the property `date` to the `TransactionSummary`.
> request from @LuizAsFight

### DIFF
--- a/packages/providers/src/transaction-summary/date.test.ts
+++ b/packages/providers/src/transaction-summary/date.test.ts
@@ -2,7 +2,7 @@ import * as tai64Mod from 'tai64';
 
 import * as dateMod from './date';
 
-const { tai64toDate, dateToTai64 } = dateMod;
+const { fromTai64ToDate: tai64toDate, fromDateToTai64: dateToTai64 } = dateMod;
 
 describe('transaction-summary/date', () => {
   afterEach(jest.restoreAllMocks);
--- a/packages/providers/src/transaction-summary/date.test.ts
+++ b/packages/providers/src/transaction-summary/date.test.ts
@@ -2,7 +2,7 @@ import * as tai64Mod from 'tai64';
 
 import * as dateMod from './date';
 
-const { tai64toDate, formatDate, dateToTai64 } = dateMod;
+const { tai64toDate, dateToTai64 } = dateMod;
 
 describe('transaction-summary/date', () => {
   afterEach(jest.restoreAllMocks);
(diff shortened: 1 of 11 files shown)


LuizAsFight (changes requested):
small suggestion:

just adding a `date` field to `transactionSummary` would do it.

date can be formatted in so many different ways, we can just return the `date` and then the dapp can format in the specific way it wants

arboleya (approved):

packages/providers/src/transaction-summary/date.ts
```diff
@@ -0,0 +1,15 @@
...
+import { TAI64 } from 'tai64';
+
+dayjs.extend(relativeTime);
+
+export const tai64toDate = (tai64Timestamp: string) => {
+  const timestamp = TAI64.fromString(tai64Timestamp, 10).toUnix();
+  return dayjs(timestamp * 1000);
+};
+
+export const formatDate = (tai64Timestamp: string) => tai64toDate(tai64Timestamp).fromNow();
```
nedsalk, line 12: Instead of using `dayjs`, it's enough to just convert to ISO 8601 format and have this function's implementation be:
```ts
export function formatDate(tai64Timestamp: string) {
const unixTimestamp = TAI64.fromString(tai64Timestamp, 10).toUnix();
const date = new Date(unixTimestamp * 1000);
return date.toISOString();
}
```

We can then rename `timeFormatted` into `timeISO` and we're good to go. There's no need depend on `dayjs`, or any other library for that matter, as it's sufficient to convert TAI64 into ISO 8601. Whoever wants to use the `timeISO` value can do so and they can use their favorite date library to do with it whatever they wish.

nedsalk: Didn't see [Luiz's comment](https://github.com/FuelLabs/fuels-ts/pull/1252#pullrequestreview-1625093044). His suggestion is better.

arboleya (approved):

Dhaiwat10 (approved):

packages/providers/src/transaction-summary/date.test.ts
```diff
@@ -0,0 +1,38 @@
+import * as tai64Mod from 'tai64';
```
nedsalk, line 1: The tests are testing that libraries are being called, how many times and with which values, but not the actual conversion of values. Suggestion:

```suggestion
import { fromDateToTai64, fromTai64ToDate } from './date';

describe('transaction-summary/date', () => {
  const times = {
    tai64: '4611686020121838636',
    unix: 1694450695000,
  };
  it('should convert TAI64 to Date correctly', () => {
    const date = fromTai64ToDate(times.tai64);

    const expectedDate = new Date(times.unix);
    expect(date).toEqual(expectedDate);
  });

  it('should convert Date to TAI64 correctly', () => {
    const date = new Date(times.unix);

    const tai64 = fromDateToTai64(date);

    expect(tai64).toEqual(times.tai64);
  });
});
```

Torres-ssf: @nedsalk It seems your review came only moments after I merged the PR. 

But yeah, since we are using an external lib I am not validating the actual values, only if the appropriate method was called.

PR #1255, opened 2023-09-14
feat: upgrade ethers v6

danielbate:


packages/abi-coder/src/coders/b256.ts
```diff
@@ -12,7 +12,7 @@ export class B256Coder extends Coder<string, string> {
   encode(value: string): Uint8Array {
     let encodedValue;
     try {
-      encodedValue = arrayify(value);
+      encodedValue = getBytesCopy(value);
```
danielbate, line 15: `getBytesCopy` has been used instead of `getBytes` as it's functionality is closer to that of the previous `arrayify`. `arrayify` always returns a new typed array representing the initially passed value, `getBytes` will mutate the passed value. This posed problems where we passed in a `Buffer` to instantiate program types, as a `Buffer` was then returned by `getBytes` which would throw type errors further along the flow. Hence `getBytesCopy` should be used to ensure a byte array is returned.

packages/abi-coder/src/coders/byte.ts
```diff
@@ -43,14 +43,14 @@ export class ByteCoder extends Coder<number[], Uint8Array> {
...
-    const data: Uint8Array[] = [arrayify(value)];
+    const data: Uint8Array[] = [Uint8Array.from(value)];
 
     const paddingLength = (WORD_SIZE - (value.length % WORD_SIZE)) % WORD_SIZE;
     if (paddingLength) {
       data.push(new Uint8Array(paddingLength));
     }
 
-    return concat(data);
+    return concatBytes(data);
```
danielbate, line 53: Usage of new utility function `concatBytes`, where the ethers v6 `concat` is no longer a viable solution as it will only return a hex string.

packages/contract/src/contract-factory.ts
```diff
@@ -93,8 +92,8 @@ export default class ContractFactory {
   createTransactionRequest(deployContractOptions?: DeployContractOptions) {
     const storageSlots = deployContractOptions?.storageSlots
       ?.map(({ key, value }) => ({
-        key: includeHexPrefix(key),
-        value: includeHexPrefix(value),
+        key: hexlifyWithPrefix(key),
```
danielbate, line 95: Preferential nomenclature.

packages/contract/src/util.ts
```diff
@@ -64,11 +63,9 @@ export const getContractId = (
...
  */
-export const includeHexPrefix = (value: string, options?: DataOptions) =>
-  hexlify(value, {
-    ...options,
-    allowMissingPrefix: true,
-  });
+export const hexlifyWithPrefix = (value: string) => {
+  if (value.startsWith('0x')) return hexlify(value);
+  return hexlify(`0x${value}`);
+};
```
danielbate, line 71: `hexlify` v6 no longer takes any configuration options but the v5 hex prefix option is mirrored here for use in storage slots.

packages/fuel-gauge/src/advanced-logging.test.ts
```diff
@@ -100,7 +100,7 @@ describe('Advanced Logging', () => {
   it('can get log data from a downstream Contract', async () => {
     const INPUT = 3;
     const { value, logs } = await contractInstance.functions
-      .test_log_from_other_contract(INPUT, otherContractInstance.id)
+      .test_log_from_other_contract(INPUT, otherContractInstance.id.toB256())
```
danielbate, line 103: This is an interesting one, potentially up for discussion.

So the second parameter to this contract function is expecting a `b256` address. Therefore, a `B256Coder` will be instantiated to encode this value. Previously, we would pass in the `contact.id` which is an `Address` type. This was fine previously because `arrayify` which is used inside of the `B256Coder` to encode the value accepts a _hexable_ object as a parameter. This is an object that contains the function `toHexString()`.  So it would be converted to a `b256` by `arrayify`, and then returned as a byte array.

Ethers v6 has much stricter function signatures and does not do such magic.`getBytesCopy` takes `BytesLike` values (essentially a hex string or byte array), therefore we must pass that to the function. Otherwise it cannot be encoded by the `B256Coder` which is now using `getBytesCopy` to return a byte array.

I personally think if we are expecting a `b256` address as defined by the ABI, then that is what should be passed to the function. Rather than doing some augmentation on our side. Unless this is a decision to aid dev ex?

With type inference we would not be able to pass in an `Address` instance when we are expecting a `b256` anyways (without additional magic).

packages/fuel-gauge/src/token-test-contract.test.ts
```diff
@@ -141,4 +145,18 @@ describe('TokenTestContract', () => {
...
+    const addressParameter = {
+      value: userWallet.address,
+    };
+    const assetId = BaseAssetId;
+
+    await expectToThrowFuelError(
+      () => token.functions.transfer_coins_to_output(addressParameter, assetId, 50).call(),
+      new FuelError(ErrorCode.ENCODE_ERROR, 'Invalid b256.')
+    );
+  });
```
danielbate, line 161: New test case to detail that we can no longer pass in an `Address` type intended to represent a `b256` address.

packages/interfaces/src/index.ts
```diff
@@ -14,6 +14,8 @@ export type B256Address = string;
 
 export type B256AddressEvm = `0x000000000000000000000000${string}`;
 
+export type Bytes = Uint8Array;
```
danielbate, line 17: Ethers v6 no longer exports a `Bytes` type, only `BytesLike` which is a `Uint8Array | string`, so I have introduced this for our own usage.

packages/mnemonic/src/mnemonic.ts
```diff
@@ -22,8 +28,8 @@ import {
 // "Bitcoin seed"
 const MasterSecret = toUtf8Bytes('Bitcoin seed');
 // 4 byte: version bytes (mainnet: 0x0488B21E public, 0x0488ADE4 private; testnet: 0x043587CF public, 0x04358394 private)
-const MainnetPRV = 0x0488ade4;
-const TestnetPRV = 0x04358394;
+const MainnetPRV = '0x0488ade4';
```
danielbate, line 31: Stricter function signature, must be a hex string.

packages/providers/src/provider.ts
```diff
@@ -475,10 +474,12 @@ export default class Provider {
...
-    return Promise.resolve({
-      name: 'fuelv2',
-      chainId: 0xdeadbeef,
-    });
+    const {
+      name,
+      consensusParameters: { chainId },
+    } = await this.getChain();
+    const network = new Network(name, chainId.toNumber());
+    return Promise.resolve(network);
```
danielbate, line 482: `Network` was introduced in v6 so I just fixed this function.

packages/signer/src/signer.ts
```diff
@@ -74,7 +74,7 @@ class Signer {
     // add recoveryParam to first s byte
     s[0] |= (signature.recoveryParam || 0) << 7;
 
-    return hexlify(concat([r, s]));
+    return concat([r, s]);
```
danielbate, line 77: `concat` returns a hex string so now the `hexlify` call is redundant.

packages/fuels/src/index.ts
danielbate, line 1: Exporting `getBytesCopy` to `arrayify` as to not introduce external breaking changes for this function call, however the function signature has changed.

danielbate (commented):
Some comments detailing implemetnation decisions.

packages/fuel-gauge/src/advanced-logging.test.ts
Dhaiwat10: I agree with you. Let's make sure we add a small note for this in the release notes though.

Dhaiwat10 (dismissed):

Dhaiwat10:
Excellent work @danielbate 🪄

Dhaiwat10 (dismissed):

nedsalk (approved):

arboleya (dismissed):
Very meticulous. Great stuff!

There are conflicts to fix, though.

arboleya (approved):

PR #1256, opened 2023-09-14
feat: implement GraphQL subscriptions support

### LINKED ISSUE #301: Update Provider Transaction Response waitForResult to poll for updates on SubmittedStatus
Per #203 there is a todo within [@fuel-ts/providers](https://github.com/FuelLabs/fuels-ts/blob/master/packages/providers/src/transaction-response/transaction-response.ts#L107-L109) to implement GraphQL subscription or some basic polling feature.

### LINKED ISSUE #782: Implement transaction subscriptions
Currently, the `fuel-core` implements a solution to improve transaction status monitoring via GraphQL Subscriptions. This new approach enables us to improve the current "expanding backoff" approach on `TransactionReponse.waitFor...`, where we query transactions on a timely basis before the status change to use subscriptions and avoid multiples request.

Note: We could use two strategies, `expanding backoff` and `subscribe`, to work on config bases giving the option to developers to pass wich strategy to be used. The default should be `subscribe`.

nedsalk:
This PR introduces support for subscriptions to our nodes and replaces `TransactionResponse`'s waiting mechanisms with subscriptions instead of polling.

**Note: nodejs and pnpm versions have been updated in this PR as well. Please run `nvm install` and `nvm use` after this PR is merged.**

closes #782,
closes #301


.eslintrc.js
```diff
@@ -25,6 +25,12 @@ module.exports = {
     // Disable error on devDependencies importing since this isn't a TS library
     'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
     'no-await-in-loop': 0,
+    'no-restricted-syntax': [
+      'off',
+      {
+        selector: 'ForOfStatement',
```
nedsalk, line 31: This is the relevant syntax:
```ts
  for await (const { statusChange } of this.provider.subscriptions.statusChange({
      transactionId: this.id,
    }))
...
```
Airbnb doesn't like it (see [this](https://github.com/airbnb/javascript/issues/1271) and [this](https://github.com/airbnb/javascript/issues/1271)), but it's valid JS syntax that's also [supported](https://the-guild.dev/graphql/sse/recipes#with-asynciterator) by graphql-sse.

apps/demo-nextjs/package.json
```diff
@@ -18,6 +18,6 @@
     "next": "13.4.4",
     "react-dom": "18.2.0",
     "react": "18.2.0",
-    "typescript": "^4.8.4"
+    "typescript": "^5.1.0"
```
nedsalk, line 21: The build was failing because I've removed also removed the TS version from `providers/package.json` and thus defaulted it to the repo version which is ~5.1.6.

packages/providers/test/provider.test.ts
```diff
@@ -651,4 +651,48 @@ describe('Provider', () => {
...
+    const { error } = await safeExec(() => provider.produceBlocks(10));
+
+    expect(error).toMatchObject({
+      type: 'aborted',
+      message: 'The user aborted a request.',
+    });
+  });
+
+  // Fails because the library creates its own AbortController
+  it.skip('throws AbortError on timeout when calling a subscription', async () => {
```
nedsalk, line 672: I tried implementing a timeout for subscriptions as suggested [here](https://github.com/FuelLabs/fuels-ts/blob/5310e8b5226a74058b85a6f1ca3671b6b5b5c5b7/packages/providers/fuel-core-schema.graphql#L778), but the library doesn't give a straightforward way to do so. I tried and tried, but in the end didn't succeed. I'll open up an issue for it after this gets merged. 
However, the timeout works for standard operations (non-subscriptions), as is shown in the test above.

package.json
```diff
@@ -90,7 +90,8 @@
   },
   "pnpm": {
     "overrides": {
-      "cross-fetch": "4.0.0"
+      "cross-fetch": "4.0.0",
+      "node-fetch": "^2.6.13"
```
nedsalk, line 94: I specified it here so that it overrides `node-fetch` wherever it's used, because I've introduced the same version into `providers/package.json` so that its fetch function can be used as the default one.

packages/providers/test/provider.test.ts
```diff
@@ -177,43 +201,30 @@ describe('Provider', () => {
...
 
-  it('can change the provider url of the current instance', () => {
+  it('can change the provider url of the current instance', async () => {
     const providerUrl1 = 'http://127.0.0.1:4000/graphql';
     const providerUrl2 = 'http://127.0.0.1:8080/graphql';
-    const provider = new Provider(providerUrl1);
-    const spyGraphQLClient = jest.spyOn(GraphQL, 'GraphQLClient');
+
+    const provider = new Provider(providerUrl1, {
+      fetch: (url: string, options: FetchRequestOptions) =>
```
nedsalk, line 209: The mock was giving me issues because `GraphQLClient` internally now receives a custom `fetch` function and the test was failing because of reference issues related to mocking and not because a bug was introduced. This test implementation also verifies that the provider url of the graphql client changed, but does it via the public API and not by assuming how the `GraphQLClient` is constructed.

apps/demo-nextjs/package.json
```diff
@@ -18,6 +18,6 @@
     "next": "13.4.4",
     "react-dom": "18.2.0",
     "react": "18.2.0",
-    "typescript": "^4.8.4"
+    "typescript": "~5.1.6"
```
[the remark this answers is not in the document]
nedsalk: This wasn't the issue.

packages/providers/package.json
```diff
@@ -38,6 +38,8 @@
     "graphql": "^16.6.0",
     "graphql-request": "^5.0.0",
     "graphql-tag": "^2.12.6",
+    "graphql-sse": "^2.2.1",
+    "cross-fetch": "^4.0.0",
```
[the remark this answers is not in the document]
nedsalk: It is being used under-the-hood in graphql-request so I opted to used it here as well. However, we're now providing a custom fetch function to graphql-request anyways, so I'll take a look at your proposal.

packages/providers/codegen.json
```diff
@@ -28,6 +28,37 @@
...
+          "TransactionId": "string",
+          "AssetId": "string",
+          "ContractId": "string",
+          "Salt": "string",
+          "Nonce": "string"
+        },
+        "useTypeImports": true,
+        "nonOptionalTypename": true,
+        "typesPrefix": "Gql"
+      }
```
Dhaiwat10, line 61: Curious why you created separate files just for the subscriptions?

packages/providers/src/provider.ts
```diff
@@ -580,7 +676,7 @@ export default class Provider {
       filter: { owner: owner.toB256(), assetId: assetId && hexlify(assetId) },
     });
 
-    const coins = result.coins.edges!.map((edge) => edge!.node!);
+    const coins = result.coins.edges.map((edge) => edge.node);
```
Dhaiwat10, line 679: Are these changes unrelated?

Dhaiwat10 (commented):
Looking good, just a couple of questions

packages/providers/codegen.json
nedsalk: I separated them because they're constructed differently, use different libraries, etc. However, I've got an idea on how to put them into one file and I'll investigate it.

packages/providers/src/provider.ts
nedsalk: I was getting warnings that these no-null-assertions are unnecessary so I removed them.

packages/providers/package.json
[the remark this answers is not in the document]
nedsalk: The new implementation is using node's fetch.

packages/providers/codegen.json
nedsalk: New implementation has no need for the `subscriptions.graphql` file.

.github/actions/ci-setup/action.yaml
```diff
@@ -2,7 +2,7 @@ name: "CI setup"
 inputs:
   node-version:
     description: "Node version"
-    default: 18.14.1
+    default: 18.18.0
```
nedsalk, line 5: In order to use node's fetch I had to update to a newer version which had a bugfix that I needed.

package.json
```diff
@@ -5,10 +5,10 @@
   "author": "Fuel Labs <contact@fuel.sh> (https://fuel.network/)",
   "private": true,
   "engines": {
-    "node": ">= 18.14.1",
-    "pnpm": ">= 8.1.1"
+    "node": ">= 18.18.0",
+    "pnpm": ">= 8.6.1"
```
nedsalk, line 9: The CI's default pnpm version is `8.6.1`, so I updated this one as well.

Dhaiwat10 (approved):

.eslintrc.js
arboleya: Is there an alternative that wouldn't require disabling `eslint` rules?

PR #1271, opened 2023-09-20
refactor: Purging `GAS_PRICE_FACTOR` and `GAS_PER_BYTE` constants

Torres-ssf:
- Close: #784

### DIFF
--- a/packages/program/src/functions/base-invocation-scope.ts
+++ b/packages/program/src/functions/base-invocation-scope.ts
@@ -351,7 +351,6 @@ export class BaseInvocationScope<TReturn = any> {
 
   getProvider(): Provider {
     const provider = <Provider>this.program.provider;
-    assert(provider, 'Provider is required!');
 
     return provider;
   }
--- a/packages/program/src/functions/base-invocation-scope.ts
+++ b/packages/program/src/functions/base-invocation-scope.ts
@@ -115,12 +115,14 @@ export class BaseInvocationScope<TReturn = any> {
    * @returns An array of required coin quantities.
    */
   protected getRequiredCoins(): Array<CoinQuantity> {
+    const { gasPriceFactor } = (<Provider>this.program.provider).getGasConfig();
+
     const assets = this.calls
       .map((call) => ({
         assetId: String(call.assetId),
         amount: bn(call.amount || 0),
       }))
-      .concat(this.transactionRequest.calculateFee())
+      .concat(this.transactionRequest.calculateFee(gasPriceFactor))
       .filter(({ assetId, amount }) => assetId && !bn(amount).isZero());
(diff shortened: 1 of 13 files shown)


packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -75,7 +75,7 @@ export class BaseInvocationScope<TReturn = any> {
    */
   protected get calls() {
     const script = getContractCallScript(this.functionInvocationScopes);
-    const provider = this.program.provider as Provider;
+    const provider = this.getProvider();
```
LuizAsFight, line 78: here is missing the validation ??
```

    assert(provider, 'Wallet or Provider is required!');
``` 

if provider is really mandatory for this class, getProvider could centralize the assert validation and remove the possibility of any `getProvider()` usage returning empty

LuizAsFight (changes requested):

arboleya (dismissed):

packages/program/src/functions/base-invocation-scope.ts
Torres-ssf: @LuizAsFight I am not 100% sure why these `Provider` validations are in place, or if they are needed. 

The way I see it, `this.program.provider` will never be undefined because `program` is an instance of `AbstractProgram`, meaning that this will be either a `Script` or `Contract` instance. 

After taking a look at these class constructors I've noticed that it is possible to instantiate a `Contract` without an `Account` but never without a `Provider`. To instantiate a `Script` the `Account` is required, therefore, the `Provider` instance will be always present.

My understanding is; `this.program.account` can be undefined, but `this.program.provider` will never be. 

Not sure if these assumptions are correct.

arboleya: Agreed, this validation seems redundant and not required.

On top of that, we should **_not_** use `assert` anymore, as everything is currently based on `FuelError`.

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -348,4 +348,11 @@ export class BaseInvocationScope<TReturn = any> {
 
     return result;
   }
+
+  getProvider(): Provider {
+    const provider = <Provider>this.program.provider;
+    assert(provider, 'Provider is required!');
```
arboleya, line 354: See my notes [here](https://github.com/FuelLabs/fuels-ts/pull/1271#discussion_r1331501286).

```suggestion
```

arboleya (changes requested):

arboleya (approved):

Dhaiwat10 (approved):

LuizAsFight (approved):

PR #1287, opened 2023-09-22
feat: throw/warn on mismatch of client version and supported version

### LINKED ISSUE #1283: Check the provider's version and throw an error (warning?) if the version is old and unsupported
Mirror of https://github.com/FuelLabs/fuels-rs/issues/1138.

nedsalk:
Closes #1283 .

Throws when major and minor versions mismatch, and warns when patch versions mismatch.
After we go live, we can change the implementation to warn on minor version mismatch. Currently, our minor versions contain breaking changes so we should throw.


packages/providers/src/provider.ts
```diff
@@ -327,12 +327,34 @@ export default class Provider {
...
 
+    Provider.verifyClientIsSupported(nodeInfo);
+
     return {
       chain,
       nodeInfo,
     };
   }
 
+  private static verifyClientIsSupported(nodeInfo: NodeInfo) {
```
arboleya, line 338: Please consider:
 1. Moving this utility to `versions/src/lib/ensureFuelCoreSupport.ts` or similar
 1. Refactoring it to use the `semver` package instead of manual string splitting _([ref](https://github.com/FuelLabs/fuels-ts/blob/bd2b7b6e73ecf85fdca46183f661f0bef7edb872/packages/versions/src/lib/compareUserVersions.ts#L1))_

Then you can call `ensureFuelCoreSupport(nodeInfo.nodeVersion)` from `fetchChainAndNodeInfo()` and voilà.

The idea is to keep all version-related validations in a single place, using the same approach.

arboleya (changes requested):
This is unrelated to your changes, but it'd be nice if we could fix the `Forbidden non-null assertion` [warnings](https://github.com/FuelLabs/fuels-ts/actions/runs/6276463325).

<details>
<summary>screenshot</summary>

warnings

<details>

packages/providers/src/provider.ts
nedsalk: @arboleya I've moved the semver-related logic to a helper function (`versions/src/lib/getDifferenceToUserFuelCoreVersion`) to have all the semver-based versioning logic in one place, but I've left the error-throwing logic in `Provider.ensureClientVersionIsSupported` because that's currently the only place where this logic is needed and moving it to a helper function outside of the package would bring in abstraction just for the sake of abstraction. I'd have to then test both `ensureFuelCoreSupport` for the error-throwing logic as well as ensure that the `Provider` throws on those same conditions, but there wouldn't be much benefit to it because the _only_ place `ensureFuelCoreSupport` would be used currently is the `Provider.ensureClientVersionIsSupported` method. If need arises, we can always move refactor and move the logic into `packages/versions` and then reuse it. But currently, yagni.

arboleya: That's okay, but the check is not against the user versions.

Perhaps something like this:

```ts
function checkNodeVersionCompatibility(nodeVersion: string) {
  // ...
  return {
    isPatchSupported: boolean, // bool looks cleaner than strings
    isMinorSupported: boolean,
    isMajorSupported: boolean,
  }
}
```

nedsalk:
@arboleya 
> This is unrelated to your changes, but it'd be nice if we could fix the Forbidden non-null assertion [warnings](https://github.com/FuelLabs/fuels-ts/actions/runs/6276463325).

These warnings should be sorted out repo-wide by changing linting rules to treat them as errors. IMO that should be a separate PR.

packages/versions/src/lib/checkFuelCoreVersionCompatibility.ts
```diff
@@ -0,0 +1,27 @@
...
+  const supportedMinor = semver.minor(FUEL_CORE);
+  const supportedPatch = semver.patch(FUEL_CORE);
+
+  return {
+    userVersion: FUEL_CORE,
+    isMajorSupported: major === supportedMajor,
+    isMinorSupported: minor === supportedMinor,
+    isPatchSupported: patch === supportedPatch,
+  };
+}
```
arboleya, line 27: As I mentioned, there is no `userVersion` at play in this scheme.

These nomenclatures can be _very_ confusing, so here's another suggestion.

> _Also, explicitly returned values can be omitted for conciseness._

```suggestion
export function checkFuelCoreVersionCompatibility(networkVersion: string) {
  const { FUEL_CORE: supportedVersion } = getSupportedVersions();

  const networkMajor = semver.major(networkVersion);
  const networkMinor = semver.minor(networkVersion);
  const networkPatch = semver.patch(networkVersion);

  const supportedMajor = semver.major(supportedVersion);
  const supportedMinor = semver.minor(supportedVersion);
  const supportedPatch = semver.patch(supportedVersion);

  return {
    supportedVersion,
    isMajorSupported: networkMajor === supportedMajor,
    isMinorSupported: networkMinor === supportedMinor,
    isPatchSupported: networkPatch === supportedPatch,
  };
}
```

packages/versions/src/lib/checkFuelCoreVersionCompatibility.test.ts
```diff
@@ -0,0 +1,79 @@
...
+      const { isMajorSupported, isMinorSupported, isPatchSupported, userVersion } =
+        checkFuelCoreVersionCompatibility(version);
+
+      expect(isMajorSupported).toEqual(major);
+      expect(isMinorSupported).toEqual(minor);
+      expect(isPatchSupported).toEqual(patch);
+
+      expect(userVersion).toEqual(v);
+    }
+  );
```
arboleya, line 78: This sort of `it.each([], ...)` abstraction in tests is sketchy and will backfire most of the time.

It might falsely give the impression of a DRY or more advanced code, but it is harder to read and becomes a pain once something fails. Line numbers will be lost, and the only way to locate the problem will be by string-matching things _visually_, because all assertions are fixed and point to the same place.

Additionally, I added an `afterAll` hook to ensure no mocked methods are left behind, plus minor touchups on nomenclatures, and finally removed the same version being mocked on `FUELS` and `FORC` properties to remove any ambiguities since these two are not part of this topic.


```suggestion
import { checkFuelCoreVersionCompatibility } from './checkFuelCoreVersionCompatibility';
import * as getSupportedVersionsMod from './getSupportedVersions';

describe('getDifferenceToUserFuelCoreVersion', () => {
  afterAll(() => jest.resetAllMocks());

  it('should validate all possible version mismatches', () => {
    const supportedVersion = '0.1.2';

    jest.spyOn(getSupportedVersionsMod, 'getSupportedVersions').mockImplementation(() => ({
      FUELS: '1', // not under test
      FORC: '1', // not under test
      FUEL_CORE: supportedVersion,
    }));

    expect(checkFuelCoreVersionCompatibility('1.1.2')).toEqual({
      isMajorSupported: false,
      isMinorSupported: true,
      isPatchSupported: true,
      supportedVersion,
    });

    expect(checkFuelCoreVersionCompatibility('1.2.2')).toEqual({
      isMajorSupported: false,
      isMinorSupported: false,
      isPatchSupported: true,
      supportedVersion,
    });

    expect(checkFuelCoreVersionCompatibility('1.1.3')).toEqual({
      isMajorSupported: false,
      isMinorSupported: true,
      isPatchSupported: false,
      supportedVersion,
    });

    expect(checkFuelCoreVersionCompatibility('0.2.2')).toEqual({
      isMajorSupported: true,
      isMinorSupported: false,
      isPatchSupported: true,
      supportedVersion,
    });

    expect(checkFuelCoreVersionCompatibility('0.2.3')).toEqual({
      isMajorSupported: true,
      isMinorSupported: false,
      isPatchSupported: false,
      supportedVersion,
    });

    expect(checkFuelCoreVersionCompatibility('0.1.3')).toEqual({
      isMajorSupported: true,
      isMinorSupported: true,
      isPatchSupported: false,
      supportedVersion,
    });

    expect(checkFuelCoreVersionCompatibility('0.1.2')).toEqual({
      isMajorSupported: true,
      isMinorSupported: true,
      isPatchSupported: true,
      supportedVersion,
    });
  });
});
```

arboleya (changes requested):

arboleya:
> These warnings should be sorted out repo-wide [..]


@nedsalk Nevermind, no need for "repo-wide" orphan issues.

Thank you for [fixing](https://github.com/FuelLabs/fuels-ts/pull/1256/files#diff-d71f8caaa8aeaca69608b48c286611694ae8572868e090b96c7cd4fa6402c475L686-L981) them on your other PR.

packages/versions/src/lib/checkFuelCoreVersionCompatibility.test.ts
```diff
@@ -0,0 +1,65 @@
+import { checkFuelCoreVersionCompatibility } from './checkFuelCoreVersionCompatibility';
+import * as getSupportedVersionsMod from './getSupportedVersions';
+
+describe('getDifferenceToUserFuelCoreVersion', () => {
+  afterAll(() => jest.resetAllMocks());
```
arboleya, line 5: Apologies, I made a mistake on this one. 😓 

```suggestion
  afterAll(() => jest.restoreAllMocks());
```

arboleya (changes requested):
Thanks again. Unfortunately, I made a mistake in my last suggestion.

Also, you should re-request reviews instead of dismissing them as stale.

There are two ways of doing so.

<details>
<summary>screenshots</summary>

re1

re2
</details>

arboleya (dismissed):
Thank you. ⚡

Dhaiwat10:
@nedsalk looks like this PR is ready to go but the `test` check is failing

nedsalk:
@Dhaiwat10 yes, the mocks failed after some changes got introduced. I have a fix in mind but have to focus on more urgent stuff currently.

Torres-ssf (approved):

Dhaiwat10 (approved):

PR #1294, opened 2023-09-25
refactor: rename Provider method `switchUrl` to `connect`

Torres-ssf:
This PR:  

- renames the Provider method `switchUrl` to `connect`.
- ensures new `chain` and `node` data is always fetched when instantiating a new Provider or when updating the Provider url ( `connect`


packages/providers/test/provider.test.ts
```diff
@@ -708,17 +720,90 @@ describe('Provider', () => {
...
+    expect(gasConfig.gasPerByte).toBeDefined();
+    expect(gasConfig.gasPriceFactor).toBeDefined();
+    expect(gasConfig.maxGasPerPredicate).toBeDefined();
+    expect(gasConfig.maxGasPerTx).toBeDefined();
+    expect(gasConfig.minGasPrice).toBeDefined();
+  });
 
-    expect(provider1.url).toEqual(FUEL_NETWORK_URL);
-    expect(provider2.url).toEqual(FUEL_NETWORK_URL);
+  it('should throws when using getChain or getNode and without cached data', async () => {
```
nedsalk, line 788: Given that `getChain` and `getNode` are instance methods that are only accessible via a `Provider` instance that is only ever instantiated via `Provider.create`, it seems to me that the setup of this test is a forced example which should never happen. @Dhaiwat10 is it really necessary for `getChain` and `getNode` to throw these errors?

packages/providers/src/provider.ts
```diff
@@ -326,33 +326,23 @@ export default class Provider {
...
-      Provider.nodeInfoCache[this.url] = nodeInfo;
-    }
+    const chain = await this.fetchChain();
+    const nodeInfo = await this.fetchNode();
 
-    if (!chain) {
-      chain = await this.fetchChain();
-      Provider.chainInfoCache[this.url] = chain;
-    }
+    Provider.chainInfoCache[this.url] = chain;
```
nedsalk, line 344: Rather than setting the cache here, it's more appropriate to set the relevant caches in the `fetchChain`/`fetchNode` methods. Currently as they are, when somebody calls `fetchChain` or `fetchNode` independently, they might get new data that differs from the cached data, but the cache won't be updated.

packages/providers/test/provider.test.ts
```diff
@@ -24,6 +24,11 @@ afterEach(() => {
   jest.restoreAllMocks();
 });
 
+const resetCache = () => {
+  Provider.chainInfoCache = {};
+  Provider.nodeInfoCache = {};
```
nedsalk, line 29: I've opened up #1295. If you could kill two birds with one stone and close that issue here as well - great. If not, no problem.

arboleya:
@LuizAsFight @luizstacio FWIW, I was the one who suggested renaming the `connect` method to `switchUrl` during my reviews with @Dhaiwat10 — _blame should be given where the blame is due_. ✋ 

The reasoning against the `connect` terminology is that the provider doesn't seem to _connect_ to anything, so much so that we have no `disconnect` method. Unlike a Database/Socket connection that manages connection pools, TTL, handshakes, etc., it only serves as a bridge/proxy for fetching remote data. This is my understanding, and if we draw inspiration from standard Wallet dropdowns, an improved choice could be `selectNetwork(url: string)`.

I have zero attachment to this, so feel free to do as you please.

packages/providers/test/provider.test.ts
arboleya: @luizstacio @LuizAsFight In my first review with @ Dhaiwat10, I remember there was a method to refresh/invalidate the cache, but it was later removed in favor of two different methods.

I have never implemented cache in this fashion as well:
 - `getXYZ` — get XYZ from the cache
 - `fetchXYZ` — fetch XYZ remotely and update the cache

We usually have a single method that will transparently decide if it gets stuff from the cache or remotely. Then, we may also provide a way for resetting, refreshing, or invalidating the cache _(whatever floats your boat)_.

arboleya (approved):
Approving this as-is, because I know it's been an external request.

However, I left my reasoning behind some points that [I believe] could be improved.

packages/providers/src/provider.ts
LuizAsFight: I agree @nedsalk , can do one of this 2 things:

1- remove `fetchChain` and `fetchNode` (they are new and will break almost no one). do a single query to get both chain and node info, then populate both caches in this method `fetchChainAndNodeInfo`.

OR

2- remove `fetchChainAndNodeInfo` and keep each caching separated as @nedsalk suggested.

packages/providers/test/provider.test.ts
luizstacio: When dealing with cache is a good way to have cache keys, and invalidation actions instead of multiple methods.

It would be better if we could invalidate specific parts of the cache. Or having options that can be provided to all methods, fetchXYX(...params, options: { invalidate: bool }).

These are just some ideas but not a specific way, on how to do it.

LuizAsFight: @arboleya you may not like the fashion and do it in other way, as long as it supports what we need. the code delivered before didn't support.

having a single method that will decide, will need to be async, which is not acceptable in our use case. that's why I go a bit different way that will make more sense for our scenario of memoization.

`get/fetch` is the most intuitive way I imagined

nedsalk: There seems to be more depth to this issue than I initially imagined when writing the comment. I'd move the conversation to #1295. @LuizAsFight could you please share your use cases (with code references) on that issue so that we get the full picture?

packages/providers/src/provider.ts
nedsalk: There is a third option: make `fetchChain` and `fetchNode` cache the data, and have `fetchChainAndNodeInfo` just be a wrapper around calls to both methods. This wouldn't break anyone and would be quite easy to implement. I'll request changes for this part.

nedsalk (changes requested):
Requested changes based on [this comment](https://github.com/FuelLabs/fuels-ts/pull/1294#discussion_r1338124577). Other than that, LGTM.

arboleya (approved):

nedsalk (approved):

PR #1297, opened 2023-09-26
feat(Provider): add "getMessageStatus" method

LuizAsFight:
this is a followup PR from #1284  as it was not providing a way for users to consume the new query


packages/providers/src/operations.graphql
```diff
@@ -446,6 +446,12 @@ query getMessageProof(
   }
 }
 
+query getMessageStatus($nonce: Nonce!) {
+  messageStatus(nonce: $nonce) {
+    state
+  }
+}
+
```
arboleya, line 454: Quick question — shouldn't this have been generated automatically by this script?

https://github.com/FuelLabs/fuels-ts/blob/10f54b873b05211a9240ba550a3ea33b639d471e/packages/providers/package.json#L21

arboleya (approved):
The good son returns home. 🙂

packages/providers/src/operations.graphql
LuizAsFight: no, the `build-operations` will build your `operations.graphql` file and create the `__generated__/operations.ts`file 

but all the queries in `operations.graphql` are written by us

arboleya: Oh, right! Thank you.

Issue #1281, opened 2023-09-21
Run integration tests against a live node


digorithm:
Mirror of https://github.com/FuelLabs/fuels-rs/issues/1135.

PR #1302, opened 2023-09-29
fix: check tx status before awaiting tx status change in `waitForResult`

### LINKED ISSUE #1301: Transaction response: waitForResult() to fetch id before awaiting status change
The following snippet iterates over IDs and waits for a status change.

There has been an issue, which is suspected to originate from here, which causes the call to hang.
It is believed that the ID may have changed status before this iteration occurs which means we wait on ID to change indefinitely resulting in the call hanging.

This could be changed to fetch the ID and check the status to see if it has changed at the start. If it has not then we wait for a status change in the loop otherwise we continue.

https://github.com/FuelLabs/fuels-ts/blob/30076e6fd544b2f1f2784366add9149d837cdd2f/packages/providers/src/transaction-response/transaction-response.ts#L200-L217

Dhaiwat10:
This PR adds a fix that ensures that the transaction has not already been completed before tuning into the `statusChange` subscription to listen for any status changes. This is a critical bug since this could cause the code to freeze.

Closes #1301

### DIFF
--- a/packages/providers/src/transaction-response/transaction-response.ts
+++ b/packages/providers/src/transaction-response/transaction-response.ts
@@ -200,10 +200,22 @@ export class TransactionResponse {
   async waitForResult<TTransactionType = void>(
     contractsAbiMap?: AbiMap
   ): Promise<TransactionResult<TTransactionType>> {
-    for await (const { statusChange } of this.provider.operations.statusChange({
+    let transactionCompleted = false;
+
+    // Check if the transaction is already completed
+    const { transaction } = await this.provider.operations.getTransaction({
       transactionId: this.id,
-    })) {
-      if (statusChange.__typename !== 'SubmittedStatus') break;
+    });
+    if (transaction?.status?.type !== 'SubmittedStatus') {
+      transactionCompleted = true;
+    }
+
+    if (!transactionCompleted) {
+      for await (const { statusChange } of this.provider.operations.statusChange({
+        transactionId: this.id,
+      })) {
+        if (statusChange.__typename !== 'SubmittedStatus') break;
+      }
     }
 
     const transactionSummary = await this.getTransactionSummary<TTransactionType>(contractsAbiMap);
--- a/.changeset/silver-clouds-lick.md
+++ b/.changeset/silver-clouds-lick.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/providers": patch
+---
+
+Fix a bug in `tx.waitForResult()` where the code would freeze in case a tx was already finalized by the time `waitForResult` was called.


Dhaiwat10:
@luizstacio @Braqzen let me know if this is the fix that you guys were looking for, and if I missed anything

Braqzen (approved):
Unsure about the inner workings / how flags change but at the high level it seems reasonable to me.
If not currently submitted then Tx status has already changed to completed ergo do not iterate on that ID in the next loop.

Should there be a test case for this?

nedsalk (changes requested):
There should be a test case for this because all of the original subscriptions PR's tests passed and this is a new issue. @luizstacio wrote a [possible implementation of one](https://github.com/FuelLabs/fuels-ts/issues/1301#issuecomment-1739419372).

packages/providers/src/transaction-response/transaction-response.ts
```diff
@@ -200,10 +200,22 @@ export class TransactionResponse {
...
+    if (transaction?.status?.type !== 'SubmittedStatus') {
+      transactionCompleted = true;
+    }
+
+    if (!transactionCompleted) {
+      for await (const { statusChange } of this.provider.operations.statusChange({
+        transactionId: this.id,
+      })) {
+        if (statusChange.__typename !== 'SubmittedStatus') break;
+      }
```
Torres-ssf, line 218: I've set a local node that takes 12 seconds to process transactions and placed a log inside this `for` loop. 

It seems `statusChange` does not have the `__typename` property, therefore is always `undefined`.

The `statusChange` is being logged this way:

```ts
{ type: 'SubmittedStatus', time: '4611686020123395412' }
```

and after changing statuses

```ts
{
  type: 'SuccessStatus',
  block: {
    id: '0x03565ff5aec3c976d23ba6d87f63cf586de47b14b3aff8115e85ad19f0008579'
  },
  time: '4611686020123395418',
  programState: { returnType: 'RETURN', data: '0x0000000000000000' }
}
```
I believe this loop always `breaks` when reaching this `if` statement.

Dhaiwat10:
@Torres-ssf @Braqzen thanks for the feedback! Just an update: after a discussion yesterday, @nedsalk is currently investigating this issue and taking care of this PR since he implemented the gql subscriptions.

packages/providers/src/transaction-response/transaction-response.ts
nedsalk: Great find, @Torres-ssf! I thought due to all tests passing that the functionality was correct, but it seems like the bug in the loop didn't make any difference because the transactions in the tests were always successful (and speedy - key problem). I pushed a bugfix in #1304 for this as well. #1281 would've probably shown this as well. Good that you caught it earlier.

Although this is a great find, I don't think this is the root cause of this issue, as I've explained [here](https://github.com/FuelLabs/fuels-ts/issues/1301#issuecomment-1741185766).

nedsalk:
Closing this PR based on [this explanation](https://github.com/FuelLabs/fuels-ts/issues/1301#issuecomment-1743427692) as the issue isn't on the TS SDK side. As for @Torres-ssf's [comment](https://github.com/FuelLabs/fuels-ts/pull/1302#discussion_r1341645141), the bug fix for it is in #1304.

PR #1303, opened 2023-09-29
fix: make `nodeInfoCache` and `chainInfoCache` private (`Provider`)

### LINKED ISSUE #1295: Make `Provider.chainInfoCache` and `Provider.nodeInfoCache` private
Caching is an implementation detail that shouldn't be accessible to / modifiable by end users. They should use the `getChain` and `getNode` methods only.

This would also allow us to implement cache invalidation logic, e.g.:
- user uses `getChain` to get chain info (method gets it from the cache because the cache is valid)
- 5 minutes pass by, the cache becomes invalid
- user uses `getChain` to get chain info (method refetches chain info, updates the cache and returns the value)

If we were to allow the user direct access to the cache, they could circumvent cache invalidation logic and get invalid values.

Dhaiwat10:
Closes #1295

### DIFF
--- a/packages/providers/src/provider.ts
+++ b/packages/providers/src/provider.ts
@@ -247,7 +247,7 @@ export default class Provider {
           });
   }
 
-  static resetChainAndNodeCaches() {
+  static clearChainAndNodeCaches() {
     Provider.nodeInfoCache = {};
     Provider.chainInfoCache = {};
   }
--- a/packages/providers/src/provider.ts
+++ b/packages/providers/src/provider.ts
@@ -247,8 +247,13 @@ export default class Provider {
           });
   }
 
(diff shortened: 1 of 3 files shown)


packages/providers/src/provider.ts
```diff
@@ -247,8 +247,13 @@ export default class Provider {
           });
   }
 
-  static chainInfoCache: ChainInfoCache = {};
-  static nodeInfoCache: NodeInfoCache = {};
+  static resetChainAndNodeCaches() {
```
nedsalk, line 250: `clearChainAndNodeCaches` might be a better name because `reset` might imply a previous initial state to which one can reset, whereas the implementation fully clears out the cache.

nedsalk (approved):

danielbate (approved):

PR #1305, opened 2023-10-01
fix: launchNode will create a GENESIS_KEY if not already set

dmihal:
If a developer uses `launchNode` or `launchNodeAndGetWallets ` without the `GENESIS_KEY` being set, it will currently give the error `not enough coins to fit the target`, since the internal "genesis" wallet is randomly generated.

This PR will automatically set the `GENESIS_KEY` env variable to ensure that the nodes can launch successfully without this variable set.


danielbate (dismissed):

Dhaiwat10:
@camsjams @danielbate can you guys re-approve? I improved the assertions :)

Torres-ssf (approved):

danielbate (approved):

Issue #1306, opened 2023-10-02
Incorrect Transaction Fee Calculation


Torres-ssf:
After conducting tests on the `beta-4`, it has been validated that the transaction fee is not being accurately calculated by the SDK. 

There is a noticeable discrepancy between the transaction fee and the owner's balance before and after executing a transaction; the difference is not in alignment with the total amount of the base asset deducted from the owner’s account.

Upon further investigation into the VM, I've noticed that the calculation of the transaction fee always takes chargeable bytes into account. The related code segment can be found [here](https://github.com/FuelLabs/fuel-vm/blob/a1390910c1d4ffb3e4f305d29a681c9589163580/fuel-tx/src/transaction/fee.rs#L76-L86).

However, on the SDK right now we only consider chargeable bytes for transactions with the type [CREATE](https://github.com/FuelLabs/fuels-ts/blob/30076e6fd544b2f1f2784366add9149d837cdd2f/packages/providers/src/utils/fee.ts#L108-L115)

PR #1307, opened 2023-10-02
fix!: transaction fee calculation

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/1306


packages/providers/src/utils/fee.test.ts
```diff
@@ -48,21 +59,21 @@ describe(__filename, () => {
...
-      const expectedFee = bn(
-        Math.ceil(gasUsed.toNumber() / gasPriceFactor.toNumber()) * gasPrice.toNumber()
-      );
-
+      const expectedFee = 1200;
       const expectedGasUsed = gasUsed;
 
-      expect(result.fee).toStrictEqual(expectedFee);
-      expect(result.gasUsed).toStrictEqual(expectedGasUsed);
+      expect(result.fee.toNumber()).toStrictEqual(expectedFee);
```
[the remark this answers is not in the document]
Torres-ssf: @camsjams this is the final expected fee value. The value is a lot higher because I've updated the mocks and I am using real transaction bytes and real transaction witnesses as well

luizstacio (dismissed):

Torres-ssf:
blocked by https://github.com/FuelLabs/fuels-ts/pull/1340

packages/fuel-gauge/src/fee.test.ts
```diff
@@ -0,0 +1,269 @@
...
+  FUEL_NETWORK_URL,
+  Predicate,
+  Provider,
+  ScriptTransactionRequest,
+  Wallet,
+  bn,
+} from 'fuels';
+import { join } from 'path';
+
+import callTestAbi from '../fixtures/forc-projects/call-test-contract/out/debug/call-test-abi.json';
```
arboleya, line 16: Shouldn't we use [`getForcProject`](https://github.com/FuelLabs/fuels-ts/blob/af2500bdbefbbf7236558426d55d061186867167/packages/utils/src/test-utils/getForcProject.ts#L47) to get all forc-related stuff in this file?

packages/providers/src/provider.ts
```diff
@@ -648,12 +654,13 @@ export default class Provider {
    */
   async getTransactionCost(
     transactionRequestLike: TransactionRequestLike,
-    tolerance: number = 0.2
+    _tolerance: number = 0.2
```
arboleya, line 657: Why are we keeping this if it [apparently] became useless?

arboleya (commented):
I will take another look, but I'm leaving some things I noticed at first glance.

Also, I wonder if `fee` isn't too much of a generic filename. 🤔 

Would `calculateFee` or `calculate-fee` be more descriptive?

packages/providers/src/provider.ts
Torres-ssf: @arboleya Good point. I am going to remove this param

Dhaiwat10 (dismissed):

packages/utils/src/test-utils/validateValueWithinRange.ts
```diff
@@ -0,0 +1,8 @@
+export const validateValueWithinRange = (params: { value: number; min: number; max: number }) => {
```
arboleya, line 1: I believe `expectToBeInRange` fits better here, considering this is meant to be an _assertion_ utility.

Tests can be written fluently:

```ts
  import { expectToBeInRange } from '...';

  it('should return true when value is within the range', () => {
    expectToBeInRange({ value: 7, min: 5, max: 10 });
  });
```

```suggestion
export const expectToBeInRange = (params: { value: number; min: number; max: number }) => {
```

arboleya (changes requested):

packages/utils/src/test-utils/validateValueWithinRange.ts
Torres-ssf: @arboleya makes sense

packages/fuel-gauge/src/predicate/predicate-with-contract.test.ts
```diff
@@ -126,37 +129,50 @@ describe('Predicate', () => {
...
+      const expectedFinalPredicateBalance =
+        initialPredicateBalance + amountToPredicate - amountToReceiver - predicateTxFee.toNumber();
+
+      expect(
+        expectToBeInRange({
+          value: expectedFinalPredicateBalance,
+          min: remainingPredicateBalance - 1,
+          max: remainingPredicateBalance + 1,
+        })
+      ).toBeTruthy();
```
arboleya, line 175: Ref. 1

```suggestion
      expectToBeInRange({
        value: expectedFinalPredicateBalance,
        min: remainingPredicateBalance - 1,
        max: remainingPredicateBalance + 1,
      });
```

packages/fuel-gauge/src/predicate/predicate-with-script.test.ts
```diff
@@ -68,23 +69,39 @@ describe('Predicate', () => {
...
-      expect(remainingPredicateBalance).toEqual(
-        amountToPredicate + initialPredicateBalance - amountToReceiver
-      );
+      expect(
+        expectToBeInRange({
+          value: finalReceiverBalance,
+          min: expectedReceiverBalance - 1,
+          max: expectedReceiverBalance + 1,
+        })
+      ).toBeTruthy();
```
arboleya, line 93: Ref. 2

```suggestion
    expectToBeInRange({
      value: finalReceiverBalance,
      min: expectedReceiverBalance - 1,
      max: expectedReceiverBalance + 1,
    });
```

packages/utils/src/test-utils/expectToBeInRange.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    expect(result).toBe(true);
+  });
+
+  it('should return true when value is equal to the maximum', () => {
+    const result = expectToBeInRange({
+      value: 10,
+      min: 5,
+      max: 10,
+    });
+    expect(result).toBe(true);
```
arboleya, line 39: Ref. 3

```suggestion
    expectToBeInRange({ value: 10, min: 5, max: 10 });
```

arboleya (commented):
Just giving the yak a little shave, feel free to ignore it ofc. 👍

Issue #1320, opened 2023-10-06
Move `estimateTxDependencies` method from `Provider` class to `Account` class


Torres-ssf:
The method `estimateTxDependencies` should be moved from the `Provider` class to the `Account` class.

Motivation:

- User-Defined Implementations: Users can now craft their versions of `estimateTxDependencies` to suit their specific requirements, offering flexibility and control over transaction request modifications.

- Avoiding Provider Class Modification: This change ensures that transaction requests are not modified within the `Provider` class.

nedsalk:
> User-Defined Implementations: Users can now craft their versions of estimateTxDependencies to suit their specific requirements, offering flexibility and control over transaction request modifications.

I don't understand why we'd want to give the user control over _estimating_ the costs of a transaction. The costs aren't something the user has control over; rather, the node is the source of truth for a transaction's costs. The user does have control over a `TransactionRequest`, its inputs and outputs, used gas, etc., and they can set that as they please, but if they send a transaction which doesn't cover the costs that the node sets, the node will reject it.

Am I missing something?

Torres-ssf:
@nedsalk,`estimateTxDependencies` was not created with the purpose of getting the transaction cost but rather to identify and add missing `Contract` inputs and outputs, and `Variable` outputs by executing a series of dry run calls and analyzing the outcome receipts.

This means that this method will always mutate the transaction request.

Later we made it to also estimate the gas used by predicates and to update the predicate `predicateGasUsed` inputs as well. I believe this logic was placed there because it was a convenient place, since `estimateTxDependencies` was being called before executing a transaction and also modifying the transaction request.

Having said that, even though `estimateTxDependencies` can prevent some transaction requests from reverting, it has a downside in that; it can cause excessive dry run calls.

Moving `estimateTxDependencies` to the `Account` was a suggestion from @luizstacio. 

If I am not mistaken, one of the motivations is that the user should have better control if he always wants to execute this method before submitting a transaction.

luizstacio:
@Torres-ssf  I think we should keep `estimateTxDependencies ` at the Provider class, but we should not use it on the submit, call, and methods inside the provider, removing changes on the transaction when calling the provider methods that should not change the tx.

Torres-ssf:
Closing this

> ref: https://github.com/FuelLabs/fuels-ts/pull/1321#issuecomment-1766252268

PR #1321, opened 2023-10-06
refactor!: move `estimateTxDependencies` to account

Torres-ssf:
- Close: #1320


nedsalk:
In the description of the issue as a motivation for the issue you wrote,
> - Avoiding Provider Class Modification: This change ensures that transaction requests are not modified within the Provider class.

Instead of modifying the `TransactionRequest` inside of `Provider`, waht about making `provider.estimateTxDependencies` return `Promise<TransactionCost>`, [similar to the rust-sdk](https://github.com/FuelLabs/fuels-rs/blob/f51e76fa3dbd03a1db5a7e18f03e53fd9eb4c38e/packages/fuels-accounts/src/provider.rs#L556)? We could then use that data wherever necessary. It would require some refactoring to move the logic from `Provider.getTransactionCost` to `Provider.estimateTxDependencies`, but in the end we'd have _only one_ method for getting the estimated costs of a transaction and, as a "unified experience" bonus, it would be the same one the rust-sdk .This way the `Provider` would serve as a source of truth for the costs of a transaction, a role which it arguably should be the one fulfilling.

What do you think?

danielbate (commented):
If one of the motivations for this is to allow consumers to have their own implementation for estimating tx dependencies, it would be nice for this to include a section in the docs and test case for it?

Torres-ssf:
Closing this because I've misinterpreted the issue pointed out by @luizstacio [here](https://github.com/FuelLabs/fuels-ts/issues/1320)

PR #1323, opened 2023-10-07
fix: remove verbose log

### LINKED ISSUE #1322: Fuel Client version check is too verbose

0.20.5

### Steps to Reproduce

1. connect to a test net with a new version such as 0.20.6
2. See `console.warn`:  `unsupported-fuel-client-version The patch versions of the client and sdk differ. Fuel client version: 0.20.6, Supported version: 0.20.5`

### Expected Behavior

This is a patch version, we shouldn't make log noise for this level imo.

This looked ok when first passed in Code Review, but in practice, its too noisy

### Actual Behavior

Log is made each time code connects to the Node

camsjams:
Fixes #1322


danielbate (approved):

nedsalk (changes requested):
IMO it should be noisy so that the user updates their package to the latest version, even if it's a patch version. Sorry, but I'll block this until @digorithm gives his thoughts, because this PR removes patch mismatch warnings altogether.

In the meantime, could you share with us the noisy logs and how they look?

Torres-ssf (approved):

digorithm:
> IMO it should be noisy so that the user updates their package to the latest version, even if it's a patch version. Sorry, but I'll block this until @digorithm gives his thoughts, because this PR removes patch mismatch warnings altogether.
> 
> In the meantime, could you share with us the noisy logs and how they look?

Thanks @nedsalk, this is a thoughtful remark. However, patch versions in all repos maintained by Fuel Labs are supposed to be backward compatible, meaning: no breaking changes whatsoever. So a change from `0.x.1` to `0.x.2` should never cause any issue. Given that that's a thing, it doesn't make much sense for us to be noisy about it.

digorithm (approved):

arboleya (approved):

nedsalk (approved):

Issue #1293, opened 2023-09-25
Implement timeout for subscriptions


nedsalk:
Currently timeouts work for mutations/queries, but not for subscriptions. This is a follow-up issue from [this comment](https://github.com/FuelLabs/fuels-ts/pull/1256/files/a2c48e82cfef6a2d05a7c4fa3f0f2adbf0d8f008#r1325854912).

PR #1332, opened 2023-10-10
fix: flaky timeout test

nedsalk:
I fixed the flaky timeout test when calling a query, and skipped the subscriptions one due to a [known issue](https://github.com/FuelLabs/fuels-ts/issues/1293). The tests were _sometimes_ failing (see [run 1](https://github.com/FuelLabs/fuels-ts/actions/runs/6432941619) and [run 2](https://github.com/FuelLabs/fuels-ts/actions/runs/6393425865)]. 
The only reason I see that they were sometimes failing is due to `AbortSignal.timeout` not changing the signal's `aborted` field to `true` before the request is sent and a response is received.


nedsalk:
Will be made obsolete by reverting of the commit where the tests were added

Issue #1339, opened 2023-10-12
Enhance Testing by Mimicking beta-4 Chain Config


Torres-ssf:
To improve the reliability and relevance of our tests, it's important that we emulate the chain configuration utilized by beta-4.

Aligning our test environment closely with the configurations of beta-4 will:

- Ensure a more authentic and applicable testing scenario, close to real-world use cases.
- Mitigate potential inconsistencies and discrepancies during test runs.

PR #1340, opened 2023-10-12
test: mimic beta 4 chain config 

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/1339

## Some details about this PR

The old chain config accepts a min gas price equal to `0`. This results in transactions having no fee cost at all ( transaction fee equals 0 ), and allowing us to submit transactions without setting the `gasLimit` property ( which results in `gasLimit`  0 ). 

Mimicking beta 4 and setting the minimum gas price to `1`, resulted in being required to set the `gasPrice` in transaction requests on all our tests, and an actual fee is being charged to each transaction.

Because of that, I had to refactor many tests to: 

1. Set the `gasPrice` property on all transaction requests.
2. increase the funding amount of wallets being used in tests.

Since the fee calculation is wrong, I had to increase the test wallets funding by huge amounts, to ensure all tests would continue to pass.

I had also to skip 2 tests that were dependent on precise fee calculations that are going to be unskipped on https://github.com/FuelLabs/fuels-ts/pull/1307

The changes made here are required for the tests that are going to be written on https://github.com/FuelLabs/fuels-ts/pull/1307, to ensure the SDK is properly handling fee calculations


packages/fuel-gauge/src/payable-annotation.test.ts
```diff
@@ -15,11 +16,17 @@ const setupContract = createSetupConfig({
...
+
+beforeAll(async () => {
+  contract = await setupContract();
+  ({ minGasPrice: gasPrice } = contract.provider.getGasConfig());
+});
 
+test('allow sending coins to payable functions', async () => {
   // This should not fail because the function is payable
-  expect(
+  await expect(
```
arboleya, line 29: Oof. These are fun to catch. 😰 

How could we get an `eslint` warning for missing `await`s?

Another test in the same file asserts an `async` method; shouldn't this be like that?

```suggestion
  await expect(async () =>
```

.fuel-core/configs/chainConfig.json
```diff
@@ -273,35 +273,37 @@
     "max_inputs": 255,
     "max_outputs": 255,
     "max_witnesses": 255,
-    "max_gas_per_tx": 500000000,
+    "max_gas_per_tx": 10000000,
```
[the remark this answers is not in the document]
Torres-ssf: @arboleya Good catch, I think we should mirror this in all places

packages/fuel-gauge/src/payable-annotation.test.ts
Torres-ssf: @arboleya It seems that the proper way to handle this assertion is without `async () =>`.

Jest complains that a promise was expected but it received a function instead.

Screenshot from 2023-10-13 10-09-26

arboleya (approved):
Awesome! 👏 ⚡

nedsalk (approved):

PR #1352, opened 2023-10-16
fix: rearrange provider tx param error code checks

### LINKED ISSUE #1349: GAS_PRICE_TOO_LOW vs GAS_LIMIT_TOO_LOW - Incorrect error being thrown
https://github.com/FuelLabs/fuels-ts/blob/f6fb12b7fbc5c4ef4ed0b7471ae6749468b84a96/packages/providers/src/provider.ts#L479C4-L489C6

The error being thrown is switched upside down here

danielbate:
Closes #1349


arboleya (approved):
How come you added a new test, and the overall coverage decreased?

I'm assuming this is an inaccurate report from the CI action.

packages/providers/test/provider.test.ts
```diff
@@ -830,4 +830,62 @@ describe('Provider', () => {
...
+        code: ErrorCode.GAS_LIMIT_TOO_LOW,
+        message: `Gas limit '${gasLimit}' is lower than the required: '${gasUsed}'.`,
+      }
+    );
+
+    expect(txCostSpy).toHaveBeenCalled();
+    expect(estimateTxSpy).toHaveBeenCalled();
+  });
+
+  it.only('throws when gas price is lower than min tx gas price', async () => {
```
arboleya, line 863:
```suggestion
  it('throws when gas price is lower than min tx gas price', async () => {
```

arboleya (changes requested):
I found the reason for the coverage decrease.

packages/providers/test/provider.test.ts
danielbate: Doh

arboleya (changes requested):

arboleya (approved):

digorithm (approved):

PR #1354, opened 2023-10-17
chore: lint fixes

arboleya:
These are just a couple of annoying lint issues I keep seeing from time to time.


packages/providers/src/provider.ts
```diff
@@ -702,6 +701,7 @@ export default class Provider {
       filter: { owner: owner.toB256(), assetId: assetId && hexlify(assetId) },
     });
 
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     const coins = result.coins.edges!.map((edge) => edge!.node!);
```
nedsalk, line 705:
```suggestion
    const coins = result.coins.edges.map((edge) => edge.node);
```

Ther `!` assertions are redundant because the underlying types don't have `undefined` or `null` in their type definitions.

packages/providers/src/provider.ts
```diff
@@ -970,6 +970,7 @@ export default class Provider {
       filter: { owner: owner.toB256() },
     });
 
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     const balances = result.balances.edges!.map((edge) => edge!.node!);
```
nedsalk, line 974:
```suggestion
    const balances = result.balances.edges.map((edge) => edge.node);
```

same as above

packages/providers/src/provider.ts
```diff
@@ -997,6 +998,7 @@ export default class Provider {
       owner: address.toB256(),
     });
 
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     const messages = result.messages.edges!.map((edge) => edge!.node!);
```
nedsalk, line 1002:
```suggestion
    const messages = result.messages.edges.map((edge) => edge.node);
```

same as above

nedsalk (approved):

Torres-ssf (approved):

Issue #1330, opened 2023-10-10
Scaffold CLI tool for creating new projects


arboleya:
> _Fuels CLI integration requires `fuels.config.ts` file generation_ — #1291

Create a scaffolding tool for generating ready-to-go Fullstack Fuel dApps.

It should support templates and provide a single-command frictionless entry point for newcomers.

```console
npx create fuels@latest my-fuel-dapp
npx create fuels@latest my-fuel-dapp -t <template-name>
```

Companion frameworks example:

```console
npx create fuels@latest my-fuel-dapp -t next
npx create fuels@latest my-fuel-dapp -t nuxt
npx create fuels@latest my-fuel-dapp -t sveltekit
npx create fuels@latest my-fuel-dapp -t astro
npx create fuels@latest my-fuel-dapp -t qwik
```

Dead-simple instant quickstart:

```console
npx create fuels@latest my-fuel-quickstart
```

End goal:

- Single-command, no-brainer, frictionless entry point for Fullstack dApp engineers
- Get users prototyping Fuel dApps in _seconds_
- Fast development cycle

Dhaiwat10:
Can I take up this one? @arboleya :)

arboleya:
Sure!

I'd say let's try to have a first minimal draft template structure considering the quickstart/counter that we can brainstorm over, review, and expand from there. Also, most "companion" frameworks I mentioned in the issue description have excellent interactive scaffolding tools from which we can look and draw inspiration — _if_ we need interactivity, meaning that a first version with only CLI flags like `-t, --template` may work just fine.

The output of this task should also align with #1291/`fuels.config.ts`, so you'll need to use that branch as the base branch for moving in parallel. Remember that the former PR is fresh and subject to changes, so it's wise to be flexible and keep some open ends to avoid over-refactoring as we approach the end.

Dhaiwat10:
@arboleya @FuelLabs/sdk-ts I have created the first draft for the template here if you want to have a look: https://github.com/Dhaiwat10/fuel-cli-starter-template

arboleya:
@Dhaiwat10 I wonder if keeping the whole `contracts` directory inside `src` is the best option—more [here](https://github.com/FuelLabs/fuels-ts/pull/1291#discussion_r1365662364).

arboleya:
[...]

PR #1356, opened 2023-10-17
feat!: add `launchTestNode` utility

### LINKED ISSUE #1254: Add `setupTestProvider` test utility
### Motivation

Tests in the `providers` package related to actually connecting to a node fall into two categories:
1. Those that mock the call (e.g. `provider.test.ts -> it('can getMessageProof with all data')`),
2. Those that assume a node is already running (e.g. `provider.test.ts -> it('can getBlocks')`).

Those that mock the call assume a certain response from the node (an assumption that could change from version to version), whereas those that assume a node is running will fail when run in isolation if the node is not running.

We should have a `setupTestProvider` utility (akin to the [rust-sdk's](https://github.com/FuelLabs/fuels-rs/blob/337d0eaa130dd18e9e347f83ab4fab76b3a6bd2a/packages/fuels-test-helpers/src/lib.rs#L136)) which can start a node and then return a `Provider` instance. We can use `launchNode.ts`, but we'd have to move this utility to e.g. `packages/utils` or `packages/providers`. We can also rewrite `launchNodeAndGetWallets` to internally use `setupTestProvider`, given that `packages/wallet` depends on `packages/providers`.

This would also be a step forward in the right direction to close #296, as each test would be running its own node, so they would be completely isolated from each other and no UTXO issues would pop up.

### Usage example

```ts
// setupTestProvider.ts

export async function setupTestProvider(options?: ProviderOptions): Promise<Provider> {
const { ip, port } = await launchNode(...);

return new Provider(`http://${ip}:${port}/graphql`);
}

// provider.test.ts
it('can getBlocks', async () => {
    const provider = await setupTestProvider();
    await provider.produceBlocks(10);
    const blocks = await provider.getBlocks({
      last: 10,
    });
    ...
});

 it('can cacheUtxo [numerical]', async () => {
    using provider = await setupTestProvider({
      cacheUtxo: 2500,
    });
    expect(provider.cache).toBeTruthy();
    expect(provider.cache?.ttl).toEqual(2_500);
});
```

### Possible implementations

The problem with running a node per test is how to kill a node after the test is finished? This can _probably_ be solved by using the TS 5.2 [`using` declaration](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#using-declarations-and-explicit-resource-management). We can close the node via `[Symbol.asyncDispose]`. We can add this property to the `Provider` instance returned by `setupTestProvider`.

The good thing is that `launchNode.ts` already has everything necessary for starting and closing a node, so this would just build upon that.

nedsalk:
This PR introduces the `launchTestNode` utility. The idea is for it to be _the_ way to setup tests for interacting with sway programs, both internally and for our users. It allows for full test isolation by creating a node for each test and disposing of it at the end of the test. Besides the main benefit of test isolation (which is ideal for defending against flaky tests), this tool would also speed up our test suite.
It makes use of TS 5.2 [explicit resource management](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#using-declarations-and-explicit-resource-management) to automatically kill the node it launches and provides a way to tweak wallet states in the genesis block, deploy contracts, as well as lower level settings like the provider options and node options. 

This builds on the great work done in #1204.

Usage example:

```ts
  walletConfig: { count: 2 },
  deployContracts: [ 
  {
    deployer: TestContract__factory,
    bytecode: contractBytecode
  }
 ],
});

const {
  contracts: [contract],
  wallets: [wallet1, wallet2],
} = launched;
```

This could be nicely tied into #1330 where a sample test file that's testing the contract can be added, demonstrating the preferred way of testing sway programs. We can add e.g. `vitest` as the default framework in the scaffold just so that the developer can run the tests immediately.

closes #1254


.eslintrc.js
```diff
@@ -13,7 +13,6 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'prettier',
     'plugin:eslint-comments/recommended',
-    'plugin:@typescript-eslint/recommended',
```
nedsalk, line 16: Removed because it was already defined above on line 13.

.prettierrc
```diff
@@ -8,7 +8,9 @@
         "tabWidth": 2,
         "useTabs": false,
         "singleQuote": true,
-        "bracketSpacing": true
+        "bracketSpacing": true,
+        "trailingComma": "es5",
+        "parser": "typescript"
```
nedsalk, line 13: These were the default settings of prettier v2. I set them here like this to not pollute the PR with prettier-related changes.

.vscode/settings.json
```diff
@@ -1,6 +1,6 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "prettier.prettierPath": "./node_modules/prettier",
+  "prettier.prettierPath": "",
```
nedsalk, line 3: Apparently the `vscode` plugin for prettier v3 now automatically finds the prettier package.

packages/contract/src/test-utils/test-node-launcher.ts
```diff
@@ -0,0 +1,131 @@
...
+        nodeOptions,
+      },
+      false
+    );
+
+    try {
+      const contracts = await TestNodeLauncher.deployContracts(deployContracts, wallets);
+
+      return (
+        dispose ?? true
```
nedsalk, line 66: The default is to return an object which can be disposed via `await using`. If the user passes in `dispose = false`, they won't be able to use `await using` and will have access to the `cleanup` function.

packages/providers/src/utils/test-utils/defaultChainConfig.ts
```diff
@@ -0,0 +1,172 @@
+import type { ChainConfig } from './fuel-node-interfaces';
+
+export const defaultChainConfig: ChainConfig = {
```
nedsalk, line 3: This is the beta-4 chain config

.vscode/settings.json
```diff
@@ -1,6 +1,5 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "prettier.prettierPath": "./node_modules/prettier",
```
nedsalk, line 3: Apparently the vscode plugin for prettier v3 now automatically finds the prettier package. It gave me some `TypeError` until I found this fix on their repo.

packages/fuel-gauge/src/bytes.test.ts
```diff
@@ -32,54 +19,71 @@ type Wrapper = {
...
-
 describe('Bytes Tests', () => {
-  let gasPrice: BN;
-  beforeAll(async () => {
-    const provider = await Provider.create(FUEL_NETWORK_URL);
-    ({ minGasPrice: gasPrice } = provider.getGasConfig());
-  });
-
   it('should test bytes output', async () => {
+    await using launched = await TestNodeLauncher.launch({
```
nedsalk, line 24: I modified all `bytes.test.ts` tests to use `TestNodeLauncher`. You can run this test file without `--runInBand` and without a running node with this: `pnpm jest --no-cache packages/fuel-gauge/src/bytes.test.ts` and it'll work. The goal is to use this utility in all tests and remove `--runInBand` from the `pnpm test` command.

packages/fuel-gauge/src/doc-examples.test.ts
```diff
@@ -200,44 +200,6 @@ describe('Doc Examples', () => {
     // #endregion wallet-message-signing
   });
 
-  it('can create wallets', async () => {
```
nedsalk, line 203: This test was used in the `test-wallets.md` documentation page and I removed it because I removed that page.

apps/docs/src/guide/wallets/test-wallets.md
nedsalk, line 1: I removed this file in favor of the changes done in `testing/index.md`.

packages/test-utils/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@fuel-ts/test-utils",
```
nedsalk, line 2: I [wasn't able](https://github.com/FuelLabs/fuels-ts/pull/1356/files#r1362324925) to export `TestNodeLauncher` via `fuels/test-utilities` so I created this package because I think we should have one package that serves as an "add this package to your devDependencies if you want to do anything testing-related with our sdk". I don't have a hard opinion on this; we can recommend the users to add `@fuel-ts/contract` and import `TestNodeLauncher` from there as well, although I think having a dedicated "test-utils" package makes sense.

packages/wallet/src/test-utils/launch-custom-provider-and-get-wallets.ts
```diff
@@ -0,0 +1,66 @@
...
+import type { WalletUnlocked } from '../wallets';
+
+import { WalletConfig } from './wallet-config';
+
+export interface LaunchCustomProviderAndGetWalletsOptions extends SetupTestProviderOptions {
+  /** Configures the wallets that should exist in the genesis block of a the test node. */
+  walletConfig: WalletConfig;
+}
+
+export async function launchCustomProviderAndGetWallets<
```
nedsalk, line 14: I got this name from the [rust-sdk](https://rust.fuel.network/v0.48.0/wallets/test-wallets.html#setting-up-test-wallets). I personally think a more appropriate name would be `setupTestProviderAndWallets`, but just went with their name for similarity's sake.

packages/contract/src/test-utils/test-node-launcher.test.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  test('a contract can be deployed', async () => {
+    // #region TestNodeLauncher-deploy-contract
+    await using launched = await TestNodeLauncher.launch({
+      deployContracts: [{ contractDir: pathToContractRootDir }],
+    });
+
+    const {
+      contracts: [contract],
+    } = launched;
```
nedsalk, line 69: Unfortunately, destructuring of disposables [isn't supported](https://github.com/microsoft/TypeScript/issues/55527) so it's a two-step process.

packages/wallet/src/test-utils.ts
```diff
@@ -1,3 +1,7 @@
 export * from './test-utils/generateTestWallet';
 export * from './test-utils/seedTestWallet';
 export * from './test-utils/launchNode';
+export * from './test-utils/launch-custom-provider-and-get-wallets';
+export { WalletConfig } from './test-utils/wallet-config';
+export { AssetId } from './test-utils/asset-id';
+export { launchTestNode as launchNode } from '@fuel-ts/providers/test-utils';
```
nedsalk, line 7: I reexported it here so that the PR doesn't introduce breaking changes. I'd remove it in another PR after others are made known of the `TestNodeLauncher`'s existence.

nedsalk:
@camsjams I opened up #1368 and changed this branch's target to that one until it gets merged, and then I'll change it to `master` again. I think it's okay in this case as the other PR doesn't actually introduce any new features to the code base and is just a package-updating chore, so I'll set this as "ready for review".

packages/providers/src/utils/test-utils/launchTestNode.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    // This string is logged by the client when the node has successfully started. We use it to know when to resolve.
+    const graphQLStartSubstring = 'Binding GraphQL provider to';
+
+    // Look for a specific graphql start point in the output.
+    child!.stderr.on('data', (chunk: string) => {
+      if (logger) logger(chunk);
+
+      if (chunk.indexOf(graphQLStartSubstring) !== -1) {
+        const [nodeIp, nodePort] = chunk.split(' ').at(-1)!.trim().split(':');
```
nedsalk, line 119: The fuel-core node logs on which url it's hosted, which enabled me to use port nr. `0` as the default, which gives rein to the OS to assign any port it wants to the node. This way we don't need `getPortPromise`.

packages/fuel-gauge/src/bytes.test.ts
Dhaiwat10: Am I correct in assuming that anyone using this utility will have to be using at least TS 5.2?

Dhaiwat10 (commented):
Can you maybe add a usage example in the PR description?

packages/fuel-gauge/src/bytes.test.ts
nedsalk: Yes, you're correct.

nedsalk:
Coverted to draft because I'm parallelizing tests in [another branch](https://github.com/FuelLabs/fuels-ts/tree/ns/feat/parallelize-tests) with this utility and I'm adding new features to it, fixing bugs, etc.

arboleya:
> [..] and I'm adding new features to it, fixing bugs, etc.

@nedsalk I haven't looked at this PR yet, but `+3,515 −1,066` changes seem like a **_lot_** for a single _utility_, and it makes me a bit skeptical that it will be worthwhile. Especially considering that #1310 is already running tests at light speed _without_ requiring parallelization, which usually performs very _poorly_ in CI.

Instead of investing even _more_ time in adding _more_ features to it, I'd suggest first validating the idea from a speed perspective. Unless we have substantial gains, my intuition tells me we should not move forward with it.

We should be thoughtful about where to invest time and why.

nedsalk:
> Instead of investing even more time in adding more features to it, I'd suggest first validating the idea from a speed perspective. Unless we have substantial gains, my intuition tells me we should not move forward with it.

I agree and that's what I've been working on recently. The features I'm adding in the parallelization branch are mostly related to caching, because I experienced a significant performance hit when basing myself off of #1310 and just doing the naive "create a node for every test _in the test_" approach. However, I managed to mostly mitigate that by introducing some caching mechanisms. It's still a WIP, but it seems mostly done from a feature POV. I put this PR into draft mode because I plan on moving all those features into this branch if all happens to be good and the performance hit isn't substantial.

nedsalk:
Opening it up for reviews again as #1392 was a successful experiment.

apps/docs/src/guide/testing/setting-up-a-custom-chain.md
nedsalk, line 1: I removed this file in favor of the documentation written in `testing/index.md`.

nedsalk:
Converting to draft until I fix all merge conflicts.

vite.base.config.mts
```diff
@@ -7,6 +7,7 @@ export default defineConfig({
       namedExport: false,
     }),
   ],
+  esbuild: { target: "es2022" },
```
nedsalk, line 10: This is necessary for `vitest` to run properly. The issue and workaround I found can be found [here](https://github.com/vitest-dev/vitest/issues/4183#issuecomment-1780959907).

packages/account/src/providers/provider.test.ts
```diff
@@ -206,34 +208,22 @@ describe('Provider', () => {
...
-    expect(provider.url).toBe(providerUrl1);
-    expect(await provider.getVersion()).toEqual(providerUrl1);
-
-    const spyFetchChainAndNodeInfo = vi
-      .spyOn(Provider.prototype, 'fetchChainAndNodeInfo')
-      .mockResolvedValue({
-        chain: {} as ChainInfo,
-        nodeInfo: {} as NodeInfo,
-      });
+    const { cleanup, url } = await launchNode({ port: '0' });
```
danielbate, line 213: If we have automatic resource disposal, what is `cleanup` doing here? Might be good to make reference to the functions available off `launchNode` in the docs also.

packages/account/src/providers/utils/sleep.ts
```diff
@@ -6,3 +6,16 @@ export function sleep(time: number) {
     }, time);
   });
 }
+
+export async function sleepUntilTrue(
```
danielbate, line 10: Is this only a test utility? May be better placed in a `test` directory.

packages/account/src/test-utils/asset-id.ts
danielbate, line 1: Nice idea 👍🏻

packages/account/src/test-utils/launch-custom-provider-and-get-wallets.test.ts
```diff
@@ -0,0 +1,190 @@
...
+import { WalletUnlocked } from '../wallet';
+
+import { AssetId } from './asset-id';
+import { launchCustomProviderAndGetWallets } from './launch-custom-provider-and-get-wallets';
+import { TestMessage } from './test-message';
+import { urlIsLive } from './utils';
+import { WalletConfig } from './wallet-config';
+
+/**
+ * @group node
```
danielbate, line 16: Can and should these also be browser tests? I think this can be applied to all the tests you've added.

packages/account/src/test-utils/asset-id.ts
danielbate: Maybe a follow up issue to go through and replace references to string literals in place of `AssetId.A/B`?

packages/account/src/test-utils/test-message.ts
```diff
@@ -0,0 +1,54 @@
...
+  private sender: AbstractAddress;
+  private recipient: AbstractAddress;
+  private nonce: string;
+  private amount: number | BN;
+  private data: string;
+  private da_height: string;
+
+  /**
+   *
+   */
```
danielbate, line 27: ?

packages/account/src/test-utils/test-message.test.ts
```diff
@@ -0,0 +1,48 @@
+import { Address } from '@fuel-ts/address';
+import { BN } from '@fuel-ts/math';
+import { randomBytes } from 'crypto';
+import { hexlify } from 'ethers';
```
danielbate, line 4: I appreciate these are dev dependencies, but would prefer referencing internal utilities (`@fuel-ts/utils`/`@fuel-ts/crypto`)

packages/contract/src/test-utils/test-node-launcher.test.ts
```diff
@@ -0,0 +1,254 @@
+import { Provider, sleepUntilTrue } from '@fuel-ts/account';
+import { WalletConfig } from '@fuel-ts/account/test-utils';
+import * as launchCustomProviderAndGetWalletsMod from '@fuel-ts/account/test-utils';
+import { FuelError } from '@fuel-ts/errors';
+import { expectToThrowFuelError, safeExec } from '@fuel-ts/errors/test-utils';
+import type { ChainConfig } from '@fuel-ts/utils';
+import { randomInt, randomUUID } from 'crypto';
```
danielbate, line 7: As above, please use internal crypto package.

packages/contract/src/test-utils/test-node-launcher.test.ts
```diff
@@ -0,0 +1,254 @@
...
+  return new Promise((resolve) => {
+    const req = http.request(parse(url), () => {
+      resolve(true);
+    });
+
+    req.on('error', () => resolve(false));
+
+    req.end();
+  });
+}
```
danielbate, line 52: Duplicated with `packages/account/src/test-utils/utils.ts`?

packages/utils/src/utils/fuel-node-interfaces.ts
danielbate, line 1: IMO this is better placed in `packages/utils/src/types`

packages/account/src/test-utils/launch-custom-provider-and-get-wallets.test.ts
nedsalk: They can't be browser tests because `launchNode` depends on node-specific libraries (e.g. `child_process`). I was thinking about browser tests and I think the only way for this to work with them is to spin up a server with a "generateNode" endpoint that'd return the url onto which the browser can connect for that specific test. It's way out of the scope of this PR, though :smile:

packages/account/src/test-utils/test-message.test.ts
nedsalk: Leftovers of olden days, thanks for commenting on it!

packages/account/src/test-utils/launch-custom-provider-and-get-wallets.test.ts
danielbate: Ah of course, yeah agreed that's a future nice to have

packages/account/src/providers/provider.test.ts
nedsalk: > If we have automatic resource disposal, what is cleanup doing here?

`launchNode` doesn't have automatic resource disposal, only `setupTestProvider` and above. 

This test is using `fetchChainAndNodeInfo` as a proxy for verifying that the URL did get changed and if I had used an additional `using provider2 = setupTestProvider()`, the `fetchChainAndNodeInfo` function would've been called on that other provider's creation for caching purposes and doing `await provider.connect(url)` would've skipped the call to `fetchChainAndNodeInfo`. That's why I had to explicitly use `launchNode`, so that another `Provider` isn't created.

> [..] Might be good to make reference to the functions available off `launchNode` in the docs also.

I didn't explain any of the other helpers besides `TestNodeLauncher` just so that the users aren't confused about various options that are all, in the end, covered by `TestNodeLauncher`. Everything that can be passed to `launchNode` can also be passed to `TestNodeLauncher`.

arboleya: Well, I'm unsure about the automatic disposal thing, it can be confusing.

Also, could we export `TestNodeLauncher` as a function named `launchNodeForTest(..)` instead?

I'd say it should be side-by-side with `launchNode(..)`, use common terminology, and have a similar API.

 - `launchNode.ts`
 - `launchNodeForTest.ts` _👈 new suggested name_

> _The latter as a wrapper around the former._

packages/contract/src/test-utils/test-node-launcher.test.ts
nedsalk: Yes, I moved it all into `utils/test-utils`: https://github.com/FuelLabs/fuels-ts/pull/1356/commits/ecedbf9106c83bb8c12de2503090fc80d61ab6d0

packages/account/src/providers/utils/sleep.ts
nedsalk: Moved to `@fuel-ts/utils/test-utils`: https://github.com/FuelLabs/fuels-ts/pull/1356/commits/127f7b221e65590600209e96769d3c85ce95a2e5

packages/account/src/providers/provider.test.ts
nedsalk: > Well, I'm unsure about the automatic disposal thing, it can be confusing.

I'd say that if we make it clear to the users how it's used that they'll be happy with not having to call `cleanup` manually. It's a well-known concept across multiple languages, as was mentioned in the motivation for the feature in [TS](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#using-declarations-and-explicit-resource-management):
> You might be familiar with [using declarations in C#](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/using), [with statements in Python](https://docs.python.org/3/reference/compound_stmts.html#the-with-statement), or [try-with-resource declarations in Java](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html). These are all similar to JavaScript’s new using keyword, and provide a similar explicit way to perform a "tear-down" of an object at the end of a scope.

I left the option to opt out of it by passing `false` as the second `dispose` parameter and be able to call `cleanup` manually:
```ts
const { cleanup, contracts, wallets, provider } = await TestNodeLauncher.launch({}, false);
```

> [..] Also, could we export TestNodeLauncher as a function named launchNodeForTest(..) instead?

Sure, I can do it.

> [..] I'd say it should be side-by-side with launchNode(..), use common terminology, and have a similar API.

Can't put them side-by-side because `TestNodeLauncher`/`launchNodeForTest` needs `@fuel-ts/contract` which is unavailable in `@fuel-ts/account`. The terminology and APIs are similar already, one expanding on the other.

packages/contract/src/test-utils/test-node-launcher.test.ts
nedsalk: `randomUUID` and `randomInt` aren't part of the `@fuel-ts/crypto` package currently. I'd rather have that done in another PR and leave this as it is right now.

arboleya: Should we have a `TODO` tag?

packages/account/src/providers/provider.test.ts
nedsalk: > Cool, then we should consider having automatic disposal for both, with the same API.

That I can do.

> Perhaps both launchers could be inside `@fuel-ts/utils/test-utils`. 🙂

`launchNode` is used in the `fuels` CLI so it's not only a test util, but it can be moved into `@fuel-ts/utils`. As for `TestNodeLauncher`/`launchNodeForTest`, I can't move it anywhere except higher in the dependency chain because it needs `Contract` and `ContractFactory` to function.

packages/contract/src/test-utils/test-node-launcher.test.ts
nedsalk: Created issue https://github.com/FuelLabs/fuels-ts/issues/1807 for `randomUUID`, while `randomInt` doesn't have a browser crypto equivalent. I'd leave `randomInt` here as it is because this is only a test file that's for node-based tests anyways.

Dhaiwat10:
@nedsalk looks like you've got a few conflicts to resolve

nedsalk:
@Dhaiwat10 I resolved them.

packages/account/src/providers/provider.test.ts
nedsalk: > [...] Also, could we export `TestNodeLauncher` as a function named `launchNodeForTest(..)` instead?

https://github.com/FuelLabs/fuels-ts/pull/1356/commits/d3fd97673c2ffd3538c247bde3bd2bc0e454a4bd

packages/account/src/test-utils/asset-id.ts
arboleya: - https://github.com/FuelLabs/fuels-ts/issues/1820

packages/account/src/providers/provider.test.ts
nedsalk: Moving `launchNode` to `@fuel-ts/utils` currently isn't possible because it's got a dependency on `Signer`. We'd have to remove the logic around `process.env.GENESIS_SECRET` to make it happen. I'm actually a fan of that because this PR allows for ad hoc wallet generation, but it's IMO better to do it in a subsequent PR.

As for making `launchNode` disposable, after thinking about it a bit, I'm against it more than I am for it. It'd only be used in `setupTestProvider` and `autoStartFuelCore`, and in both places it'd be used as not disposable because of additional logic around it after the node is launched. After we merge this PR, it'll be removed from all tests as well, so all references will be only in the two aforementioned functions. IMO it'd be too much ceremony in the utility for no substantial gain.

danielbate: We would need `process.env.GENESIS_SECRET` to set the key for live environments, taken from the vault.

nedsalk: > We would need process.env.GENESIS_SECRET to set the key for live environments, taken from the vault.

But that's something not related to `launchNode`. The live environment tests won't be using the helpers this PR introduces.

Dhaiwat10 (changes requested):
Great work @nedsalk with this PR, but I'm not a huge fan of what the usage API looks like currently:

```ts
using launched = await TestNodeLauncher.launch({
  walletConfig: new WalletConfig({ wallets: 2 }),
  deployContracts: [
    pathToContractDir,
    { contractDir: pathToContractDir, walletIndex: 1, options: { storageSlots: [] } },
  ],
});
```

can be simplified to:

```ts
using launched = await TestNodeLauncher.launch({
  wallets: 2,
  contracts: [ { contractDir: pathToContractDir, walletIndex: 1, options: { storageSlots: [] }
});
```

I would prefer the API to be more 'flat' in general like that, just to give you an idea. I feel like it is a bit too nested the way it is, especially with extra classes like `WalletConfig`.

danielbate (commented):
Maybe, due to the size of this PR and it's opinionated implementation of a utility that will likely impact us all if we decide we should implement it and use it across our test suite, it would be good for us all to give a more general opinion on this feature. Then if we come to a consensus we can dig into the code more.

I am very keen to see this in the SDK for a couple of reasons. Standardisation of test setup across the SDK. We do a pretty decent job of this already, however inconsistencies between `fuel-gauge` and `docs-snippets` are frustrating and slow down dev. I'd also love to remove the amount of _noise_ from tests around wallet setup, funding, deployment etc. Then it is really clear what is under test. 
Defence against flaky tests and test parallelisation problems. This comes up pretty frequently. If we have a concern about parallelisation issues then that should be tested for specifically, not as a side effect. I believe this approach will give us more clarify to our testing.

I think if we go for this we should really push it in the docs. Like we do with the fuels CLI, clearly document how it can be used and it's configuration options. I would love to get consumers really defaulting to using a test node to test their dapps rather than forum posts, as this will likely become a bigger problem when we are in mainnet and users aren't using a faucet for tokens. Secondly ensuring that the API is simple, not just for us but for our consumers.

cc @FuelLabs/sdk-ts

petertonysmith94 (commented):
This has got some great stuff in it, but potentially confusing in places.

- With the disposable related generics - it's difficult to read along.
- Struggling to see that much value in `setupTestProvider` and/or `launchCustomProviderAndSetupWallets`, and believe it'll just add noise.
- With a little work this could be a really valuable addition, but not quite there yet.

Would love to talk this over with you @nedsalk 😄

packages/account/src/test-utils/index.ts
```diff
@@ -1,3 +1,6 @@
 export * from './generateTestWallet';
 export * from './seedTestWallet';
 export * from './launchNode';
+export * from './launch-custom-provider-and-get-wallets';
+export * from './wallet-config';
+export * from './setup-test-provider';
```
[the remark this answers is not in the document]
nedsalk: We have to first have a file naming convention in our codebase ;D

packages/account/src/providers/provider.test.ts
```diff
@@ -57,15 +58,15 @@ const FUEL_NETWORK_URL = 'http://127.0.0.1:4000/graphql';
  */
 describe('Provider', () => {
```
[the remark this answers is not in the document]
nedsalk: Resolved this in a discussion.

packages/utils/src/test-utils/sleep.ts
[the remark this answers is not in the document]
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1356/commits/636e5808e2aa395791c4eab20c33808c3a211090

packages/utils/src/test-utils/sleep.ts
```diff
@@ -0,0 +1,14 @@
+import { sleep } from '../utils/sleep';
+
+export async function sleepUntilTrue(
```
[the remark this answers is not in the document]
nedsalk: https://github.com/FuelLabs/fuels-ts/commit/636e5808e2aa395791c4eab20c33808c3a211090

packages/account/src/test-utils/launchNode.ts
```diff
@@ -101,6 +104,7 @@ export const launchNode = async ({
   loggingEnabled = true,
   debugEnabled = false,
   basePath,
+  chainConfig = defaultChainConfig,
```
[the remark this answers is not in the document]
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1356/commits/e0351df3569d005d8f942731ed2d5bd2b63e5541

apps/docs/src/guide/testing/index.md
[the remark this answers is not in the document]
nedsalk: IMO the ideal would be to showcase some how-to-use examples and then reference everything to its typedoc, but the typedoc isn't being generated for the utilities introduced in the PR. :thinking: Perhaps you know what's up with that?

I did improve upon the documentation and examples in https://github.com/FuelLabs/fuels-ts/pull/1356/commits/2ceabb1f21ee50f3734fbc94b5986594e0986e26.

nedsalk:
@Dhaiwat10 We discussed renaming the passed-in options:
- `walletsConfig` to `wallets`
- `deployContracts` to `contracts`

I tried it, but it looked off because the same words were being used for passed-in arguments and the destructured results:
```ts
  wallets: {
    count: 3,
  },
  contracts: [pathToContractRootDir],
});

const { contracts, wallets } = launched;
```
I'm not a fan of it even though I initially agreed. I also think that the names `walletsConfig` and `deployContracts` are appropriate because the whole `launchTestNode` utility is in the spirit of configurability.

nedsalk:
@petertonysmith94 Thanks for the feedback, it's much appreciated!

- I have removed the disposable generic stuff, thanks for the idea.
- I removed `setupTestProvider` and moved its functionality to `setupTestProviderAndWallets`. `setupTestProvider` was a remnant of the days when the `providers` package was separate from the `account` package. If we had only one `fuels` package there wouldn't be a need for `setupTestProviderAndWallets` as well and `launchTestNode` would suffice. Who knows, maybe one day.

packages/fuels/package.json
```diff
@@ -22,12 +22,20 @@
       "require": "./dist/cli.js",
       "import": "./dist/cli.mjs",
       "types": "./dist/cli.d.ts"
+    },
+    "./test-utils": {
+      "require": "./dist/test-utils.js",
+      "import": "./dist/test-utils.mjs",
+      "types": "./dist/test-utils.d.ts"
```
Torres-ssf, line 29: Why are we adding this to the umbrella package?

I was under the impression that all test utilities should be imported from `@fuel-ts/utils`

packages/contract/src/test-utils/launch-test-node.test.ts
```diff
@@ -0,0 +1,236 @@
...
+        pathToContractRootDir,
+        { contractDir: pathToContractRootDir, walletIndex: 1, options: { storageSlots: [] } },
+      ],
+    });
+
+    const {
+      contracts: [contract1, contract2],
+      wallets: [wallet1, wallet2],
+    } = launched;
+    // #endregion multiple-contracts-and-wallets
```
Torres-ssf, line 146: @nedsalk we need to place code snippets within `app/doc-snippets`

Torres-ssf (commented):
@nedsalk Good stuff 👍

Why `launchTestNode` is placed within the `contract` package instead of the `utils` package?

nedsalk:
@Torres-ssf Thanks! 

> Why `launchTestNode` is placed within the `contract` package instead of the `utils` package?

Because we need all the `Contract`-related functionality to deploy them properly and `utils` is a lower level package.

packages/account/src/test-utils/wallet-config.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  /**
+   * Number of coins (UTXOs) per asset id.
+   */
+  coinsPerAsset: number;
+
+  /**
+   * For each coin, the amount it'll contain.
+   */
+  amountPerCoin: number;
```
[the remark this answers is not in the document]
nedsalk: Yeah, I took it from a similar utility in the rust sdk.

packages/contract/src/test-utils/launch-test-node.test.ts
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1356/commits/4752fe77d7a8a42275c78f9173896f497178b07b

apps/docs/src/guide/testing/index.md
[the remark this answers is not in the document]
danielbate: Yep I do agree actually, it was overkill, those are good improvements.

> typedoc isn't being generated for the utilities introduced in the PR

So you need to add the entrypoint for typedoc for the test utils index. Currently we've only got one for the main index.

Add a new value to [this array](https://github.com/FuelLabs/fuels-ts/blob/master/packages/contract/typedoc.json) for the test util entry.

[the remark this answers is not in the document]
nedsalk: There's a problem with typedoc so I couldn't follow through with this and created two issues:
- #1897
- #1898

arboleya (commented):
Is the PR description up to date? 🤔

nedsalk:
@arboleya Thanks, it is now.

packages/utils/src/utils/types.ts
```diff
@@ -0,0 +1,188 @@
...
+  asset_id: string;
+}
+
+interface Message {
+  sender: string;
+  recipient: string;
+  nonce: string;
+  amount: string;
+  data: string;
+  da_height: string;
```
Dhaiwat10, line 13: Shouldn't we have types for these things stored somewhere already? 🤔

tsconfig.base.json
```diff
@@ -4,7 +4,7 @@
     "declaration": true,
     "declarationMap": true,
     "esModuleInterop": true,
-    "lib": ["ES2022"],
+    "lib": ["ES2022", "ESNext.disposable"],
```
Dhaiwat10, line 7: Does this change anything for our users? Will they have to make any changes to their config files?

packages/account/src/test-utils/setup-test-provider-and-wallets.ts
```diff
@@ -0,0 +1,91 @@
...
+  wallets: WalletUnlocked[];
+  provider: Provider;
+  cleanup: () => void;
+}
+
+/**
+ * Launches a test node and creates wallets for testing.
+ * If initialized with the `using` keyword, the node will be killed when it goes out of scope.
+ * If initialized with `const`, manual disposal of the node must be done via the `cleanup` function.
+ *
```
Dhaiwat10, line 47: Can you add some documentation for both of these cases with code snippets in our docs?

packages/account/src/test-utils/launchNode.test.ts
```diff
@@ -31,7 +32,7 @@ function mockSpawn(params: { shouldError: boolean } = { shouldError: false }) {
         // The `Binding GraphQL provider to` message simulates a fuel-core
         // successful startup log message, usually meaning that the node
         // is up and waiting for connections
-        fn('Binding GraphQL provider to');
+        fn('Binding GraphQL provider to 0.0.0.0:4000');
```
Dhaiwat10, line 35: Why is this bound to a specific port now?

apps/docs/.vitepress/config.ts
```diff
@@ -214,10 +214,6 @@ export default defineConfig({
                 text: 'Signing',
                 link: '/guide/wallets/signing',
               },
-              {
-                text: 'Test Wallets',
-                link: '/guide/wallets/test-wallets',
-              },
```
Dhaiwat10, line 220: I think we should still have documentation around creating _just_ test wallets. There could be cases where someone wants to create a test wallet but has their own node running.

tsconfig.base.json
nedsalk: No, they'll just have to use TS 5.2. Given that we're now already at TS 5.4, IMO it's okay.

packages/account/src/test-utils/launchNode.test.ts
nedsalk: It's because `launchNode` is now [reading](https://github.com/FuelLabs/fuels-ts/pull/1356/files#diff-b8f07cac51692505e88e065262fd373293a373ed5f6196eb13ec811e37aee136R241) from the logs to get the IP and port. Previously, the string `Binding GraphQL provider to` was used as a mock message that'd cause `launchNode` to return successfully; now we need the `0.0.0.0:4000`, otherwise the logic would fail.

We're now reading the IP and port from the logs because we're using the `0` port [here](https://github.com/FuelLabs/fuels-ts/pull/1356/files#diff-3c0413287e9234f6662e455d68cdf5c2daf63406cd32c3b392f53c6fd5cd525fR68), thereby asking the OS to assign a port to our test node.

packages/account/src/test-utils/setup-test-provider-and-wallets.ts
nedsalk: I explained the `using` keyword in the very first example and added an **IMPORTANT** mention around `const`/`let` in the same line [here](https://github.com/FuelLabs/fuels-ts/pull/1356/files#diff-af7951c01e94cdb5779ab49a5313e5cf16c9d98b71dd445f8254095eb594cef7R26).

packages/utils/src/utils/types.ts
nedsalk: We have something similar in [`resource.ts`](https://github.com/FuelLabs/fuels-ts/blob/master/packages/account/src/providers/resource.ts), but the types defined in this file are specifically related to the chain config and have different interfaces than `RawCoin`/`RawMessage` in `resource.ts`.

apps/docs/.vitepress/config.ts
nedsalk: That is the testing scenario that I'd like to avoid after introducing `launchTestNode`. I want our users to isolate their tests on the node level and, if they want to have multiple wallets, they can use the `{walletConfig: {...}}` option for it. If they don't use this utility but rather use their own node _for tests_, they'll come across the same problems our SDK came across around test isolation, one test spending UTXOs and another failing because it wanted to spend them as well, etc.

apps/docs/src/guide/testing/index.md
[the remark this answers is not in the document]
danielbate: #2082 just pushed for multiple entry points.

Dhaiwat10:
@nedsalk seems to be several conflicts here 😅

nedsalk:
@Dhaiwat10 yeah, the fuel-core upgrade broke a lot. I'm working on it right now. :sweat_smile:

nedsalk:
I'm setting this to draft until #2217 is merged because I just went through some problems migrating to `0.24.3` and that PR updates to `0.26.0` which might introduce additional breaking changes. It wouldn't make sense to merge this PR before because it might then break #2217, which is more important.

apps/docs/src/guide/testing/index.md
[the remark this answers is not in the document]
nedsalk: @danielbate I tried doing it again but now there's this problem:
image

.changeset/breezy-carrots-bow.md
```diff
@@ -0,0 +1,8 @@
+---
+"@fuel-ts/utils": patch
+"@fuel-ts/account": minor
+"@fuel-ts/contract": patch
+"fuels": patch
```
arboleya, line 5: Considering everything is bumped together, what's the utility in mixing `patch` and `minor`?

packages/account/src/test-utils/setup-test-provider-and-wallets.test.ts
```diff
@@ -0,0 +1,263 @@
...
+      const p = await Provider.create(url);
+      return p.getBlockNumber();
+    });
+
+    expect(error).toMatchObject({
+      message: 'fetch failed',
+    });
+  });
+
+  test('kills the node if provider cant connect post-launch', async () => {
```
maschad, line 44: Why the variation between using `test` and `it` here?

packages/contract/src/test-utils/launch-test-node.ts
```diff
@@ -0,0 +1,204 @@
...
+   * Pass in either the path to the contract's root directory to deploy the contract or use `DeployContractConfig` for more control.
+   */
+  deployContracts: (DeployContractConfig | string)[];
+}
+
+interface LaunchTestNodeReturn<TContracts> extends SetupTestProviderAndWalletsReturn {
+  contracts: TContracts;
+}
+function getChainSnapshot(nodeOptions: LaunchTestNodeOptions['nodeOptions']) {
+  let envChainMetadata: SnapshotConfigs['metadata'] | undefined;
```
maschad, line 50: I think we want to handle errors from `readFileSync`

```typescript
let envChainMetadata: SnapshotConfigs['metadata'] | undefined;
let chainConfig: SnapshotConfigs['chainConfig'] | undefined;
let stateConfig: SnapshotConfigs['stateConfig'] | undefined;

const defaultChainSnapshotDir = process.env.DEFAULT_CHAIN_SNAPSHOT_DIR;

if (defaultChainSnapshotDir) {
  envChainMetadata = readJsonFile<SnapshotConfigs['metadata']>(
    path.join(defaultChainSnapshotDir, 'metadata.json')
  );

  if (envChainMetadata) {
    chainConfig = readJsonFile<SnapshotConfigs['chainConfig']>(
      path.join(defaultChainSnapshotDir, envChainMetadata.chain_config)
    );

    stateConfig = readJsonFile<SnapshotConfigs['stateConfig']>(
      path.join(defaultChainSnapshotDir, envChainMetadata.table_encoding.Json.filepath)
    );
  }
}

function readJsonFile<T>(filePath: string): T | undefined {
  try {
    const fileContent = readFileSync(filePath, 'utf-8');
    return JSON.parse(fileContent) as T;
  } catch (error) {
    // Handle how you see fit
    throw new FuelError(`Failed to read JSON file: ${filePath}`);
    return undefined;
  }
}
```

maschad (changes requested):
This is really great work! Awesome job 🚀  Looking forward to it.

I left some comments

apps/docs/.vitepress/config.ts
```diff
@@ -153,10 +153,6 @@ export default defineConfig({
                 text: 'Locking and Unlocking',
                 link: '/guide/wallets/locking-and-unlocking',
               },
-              {
```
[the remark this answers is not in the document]
nedsalk: Because the `launchTestNode` documentation covers them now.

apps/docs/src/guide/testing/index.md
```diff
@@ -8,36 +8,51 @@
 
 In order to test your Sway and TS-SDK applications, you can test your code in a number of ways:
 
-1. Testing with TS-SDK: Compiling you Sway code and connecting to the methods using TS-SDK and JS testing frameworks
-2. Using `forc test` see <a :href="url" target="_blank" rel="noreferrer">the Sway docs</a> for more info
-3. Using [the Rust SDK](https://docs.fuel.network/docs/fuels-rs/testing/)
+1. Testing with the TS-SDK: Compiling you Sway code and connecting to the methods using TS-SDK and JS testing frameworks
+1. Using [`forc test`](https://docs.fuel.network/docs/forc/commands/forc%5ftest/#forc-test)
```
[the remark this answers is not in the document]
nedsalk: This numbering style gets automatically converted so we don't have to specify 1,2,3,4...

apps/docs/src/guide/testing/index.md
```diff
@@ -8,36 +8,51 @@
...
+1. Using [the Rust SDK](https://docs.fuel.network/docs/fuels-rs/testing/)
```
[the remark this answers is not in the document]
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1356#discussion_r1596440131

packages/account/src/test-utils/setup-test-provider-and-wallets.test.ts
nedsalk: There's no real reason, I'd just write what comes to mind.

packages/contract/src/test-utils/launch-test-node.ts
nedsalk: I don't see a reason to handle them in a specific manner, and the error thrown by `readFileSync` is already descriptive enough:
`Error: ENOENT: no such file or directory, open '/tmp/.fuels-ts/d2860af5-db6c-49b0-b315-dbf035bce485/metadata-not-existing.json'`.

`FuelError` should IMO be reserved for domain-specific failure modes (e.g. `ErrorCode.INVALID_MULTICALL`).

apps/docs/src/guide/testing/index.md
```diff
@@ -8,36 +8,51 @@
...
 
-```ts
-import { launchNodeAndGetWallets } from "@fuel-ts/account/test-utils";
+You can also configure wallets and deploy multiple contracts with them:
 
-const { stop, wallets, provider } = await launchNodeAndGetWallets();
+<<< @/../../docs-snippets/src/guide/testing/launching-a-test-node.test.ts#multiple-contracts-and-wallets{ts:line-numbers}
 
-// ... do your tests - deploy contracts using the wallets, fetch info from the provider, etc.
+That's a lot of options! These docs showcase some functionality; for a detailed look it's best to fiddle with the utility yourself. Every property is documented with typedoc. If something is still unclear, you found a bug or you've got a feature idea, please contact us via official channels of communication and we'll help you.
```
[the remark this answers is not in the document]
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1356/commits/152beaf6300d9259629c6e9143b982e0bdfc098b

Torres-ssf (dismissed):

apps/docs/src/guide/wallets/test-wallets.md
```diff
@@ -1,13 +0,0 @@
...
-
-## Create a single wallet
-
-<<< @/../../docs-snippets/src/guide/wallets/access.test.ts#wallets{ts:line-numbers}
-
-## Setting up multiple test wallets
-
-If you need multiple test wallets, they can be set up as follows:
-
-<<< @/../../docs-snippets/src/guide/wallets/test-wallets.test.ts#wallet-setup{ts:line-numbers}
```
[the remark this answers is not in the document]
nedsalk: What uses do you see for it that won't be superseded by `WalletConfig`? Note that ideally we want other users to start using `launchTestNode` exclusively for tests and not give mixed messages by providing two approaches.

apps/docs/src/guide/testing/setting-up-a-custom-chain.md
```diff
@@ -1,31 +0,0 @@
...
-
-You can use custom snapshots to customize things like the chain's consensus parameters or specify some initial states for the chain.
-
-Here are some examples:
-
-- [`chainConfig.json`](https://github.com/FuelLabs/fuels-ts/blob/master/.fuel-core/configs/chainConfig.json)
-<!-- - [`stateConfig.json`](https://github.com/FuelLabs/fuels-ts/blob/master/.fuel-core/configs/stateConfig.json)
-- [`metadata.json`](https://github.com/FuelLabs/fuels-ts/blob/master/.fuel-core/configs/metadata.json) -->
-
-<<< @/../../../packages/account/src/test-utils/launchNodeAndGetWallets.test.ts#launchNode-custom-config{ts:line-numbers}
```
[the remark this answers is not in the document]
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1356/commits/8939c6f57562f0553dd636f89fdac9678fd91f29

apps/docs/src/guide/wallets/test-wallets.md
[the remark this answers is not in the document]
nedsalk: That would be done the following way:
```ts
const {wallets: [genesisWallet], provider} = launched;
const wallet = WalletUnlocked.generate({provider});

// do tests on non-funded wallet

await genesisWallet.fund(wallet.address, 1234);

// do tests on funded wallet
```

`generateTestWallet` serves the purpose of having an unique, funded wallet you can work with in your tests and which doesn't experience the "UTXO already spent" issues because of other tests using it simultaneously. With `WalletConfig` you get the same thing.

Its implementation is also tied to the specific way we've set up our current test suite and guided our users to do as well (e.g. dependency on `process.env.GENESIS_SECRET`, sharing the same node between tests, etc.). That approach will be deprecated once this PR gets merged and that's the reason why I'm removing it from the docs.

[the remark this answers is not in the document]
nedsalk: > Just a thought - is it worth converting the following [test](https://github.com/FuelLabs/fuels-ts/blob/ns/feat/test-node-launcher/apps/docs-snippets/src/guide/wallets/test-wallets.test.ts) to using launchTestNode, rather than removing it entirely?

Not really, because `WalletConfig` examples cover it and there's no practical differentiation between an address that's been there since the genesis block (added via `WalletConfig`) and one that was created later (funded via `generateTestWallet`) which would necessitate translation of that linked example.

Dhaiwat10 (dismissed):
Great work @nedsalk  :)

Torres-ssf (dismissed):

danielbate (dismissed):

arboleya (approved):
I found it confusing to follow, considering how things were split into several packages.

The rationale is valid and the utility will be extremely useful, but I think it lacks more attention to detail.

I could have done a more in-depth review, but it's a good start.

arboleya (changes requested):
I messed up, hit some shortcut here, and approved it by mistake without even finishing my rationale. Doing an empty "request changes" to make up for the mess I created.

apps/docs-snippets/src/guide/testing/launching-a-test-node.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  '../../../test/fixtures/forc-projects/counter'
+);
+
+/**
+ * @group node
+ */
+describe('launching a test node', () => {
+  test('simple contract deployment', async () => {
+    // #region deploy-contract
+    using launched = await launchTestNode({
```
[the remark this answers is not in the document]
arboleya: As we spoke, this is fine, and we need to do two new things:
 1. Instruct people to configure the `tsconfig.json` to use the `using` keyword
 1. Add an alternate flow (almost identical) using the `const` keyword instead, which requires the `cleanup` method to be called

apps/docs/src/guide/testing/index.md
```diff
@@ -8,36 +8,47 @@
...
 
-You'll often want to create one or more test wallets when testing your contracts.
+To simplify testing in isolation, the SDK provides `launchTestNode`, a utility via which you can spin up a short-lived `fuel-core` node, setup a custom provider and wallets, and deploy contracts in one go.
+Here is a simple contract deployment in a test:
 
-For this, you can find two simple utilities on the account package:
+<<< @/../../docs-snippets/src/guide/testing/launching-a-test-node.test.ts#deploy-contract{ts:line-numbers}
 
-- [`@fuel-ts/account`](https://github.com/FuelLabs/fuels-ts/tree/master/packages/account#test-utilities)
+- The `launched` variable was instantiated with the [`using`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#using-declarations-and-explicit-resource-management) declaration. This is a TypeScript 5.2 feature called explicit resource management and it's used to automatically kill the node once `launched` goes out of scope.
```
arboleya, line 26: The launched explanation could precede the snippet, followed by the `tsconfig.json` config example and then the snippet.

apps/docs/src/guide/testing/index.md
```diff
@@ -8,36 +8,47 @@
...
+  - **IMPORTANT: if you instantiate it with `const`/`let`, the automatic disposal won't happen and you have to call the returned `cleanup` function yourself.**
```
arboleya, line 27: The important block should have its own section with alternative `const` usage.

apps/docs/src/guide/testing/index.md
```diff
@@ -8,36 +8,47 @@
 
 In order to test your Sway and TS-SDK applications, you can test your code in a number of ways:
 
-1. Testing with TS-SDK: Compiling you Sway code and connecting to the methods using TS-SDK and JS testing frameworks
-2. Using `forc test` see <a :href="url" target="_blank" rel="noreferrer">the Sway docs</a> for more info
-3. Using [the Rust SDK](https://docs.fuel.network/docs/fuels-rs/testing/)
+1. Testing with the TS-SDK: Compiling you Sway code and connecting to the methods using TS-SDK and JS testing frameworks
```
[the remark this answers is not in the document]
arboleya: A more broad comment is to revisit the whole page from the perspective of someone without any context. In this sense, we could:
 - Provide cross-links to contract-examples
 - Provide better continuity (`forc build`, find `out` dir path, feed it into the `launchTestNode` contracts array, and so on)

apps/docs/src/guide/testing/index.md
```diff
@@ -8,36 +8,47 @@
...
+To simplify testing in isolation, the SDK provides `launchTestNode`, a utility via which you can spin up a short-lived `fuel-core` node, setup a custom provider and wallets, and deploy contracts in one go.
+Here is a simple contract deployment in a test:
 
-For this, you can find two simple utilities on the account package:
+<<< @/../../docs-snippets/src/guide/testing/launching-a-test-node.test.ts#deploy-contract{ts:line-numbers}
 
-- [`@fuel-ts/account`](https://github.com/FuelLabs/fuels-ts/tree/master/packages/account#test-utilities)
+- The `launched` variable was instantiated with the [`using`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#using-declarations-and-explicit-resource-management) declaration. This is a TypeScript 5.2 feature called explicit resource management and it's used to automatically kill the node once `launched` goes out of scope.
+  - **IMPORTANT: if you instantiate it with `const`/`let`, the automatic disposal won't happen and you have to call the returned `cleanup` function yourself.**
+- `launchTestNode` spun up a short-lived `fuel-core` node, deployed a contract to it and returned it for testing.
```
arboleya, line 28: Expand explanations about all the parts of the snippet, like:
 - The `launched` variable
 - All parts of the deconstructed objects (what they are and how they can be used, etc.)

packages/contract/src/test-utils/launch-test-node.ts
```diff
@@ -0,0 +1,204 @@
...
+      readFileSync(path.join(dirname, 'metadata.json'), 'utf-8')
+    ) as SnapshotConfigs['metadata'];
+    const chainConfigPath = path.join(dirname, envChainMetadata.chain_config);
+    chainConfig = JSON.parse(
+      readFileSync(chainConfigPath, 'utf-8')
+    ) as SnapshotConfigs['chainConfig'];
+    const stateConfigPath = path.join(dirname, envChainMetadata.table_encoding.Json.filepath);
+    stateConfig = JSON.parse(
+      readFileSync(stateConfigPath, 'utf-8')
+    ) as SnapshotConfigs['stateConfig'];
```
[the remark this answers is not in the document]
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1356/commits/aa58458b8717b388f0debf2a01b761058f43d559

packages/account/src/test-utils/wallet-config.ts
```diff
@@ -0,0 +1,171 @@
...
+              tx_id: hexlify(randomBytes(32)),
+            });
+          }
+        });
+      });
+
+    return coins;
+  }
+
+  private static guard({
```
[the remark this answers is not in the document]
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1356/commits/fd7eb1eae2a33a5adbce9747ca1301dff340124c

packages/account/src/test-utils/wallet-config.ts
```diff
@@ -0,0 +1,171 @@
...
+        this.wallets,
+        BASE_ASSET_ID,
+        assets,
+        coinsPerAsset,
+        amountPerCoin
+      ),
+    };
+  }
+
+  apply(snapshotConfig: PartialDeep<SnapshotConfigs> | undefined): PartialDeep<SnapshotConfigs> & {
```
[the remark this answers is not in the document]
nedsalk: It applies the `WalletConfig`'s settings (coins and messages) onto the config, returning back a config with an updated genesis block state.

packages/contract/src/test-utils.ts
```diff
@@ -0,0 +1 @@
+export * from './test-utils/launch-test-node';
```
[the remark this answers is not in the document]
nedsalk: 1. We don't have a file naming convention and this file isn't put side-by-side with `launchNode.ts` so it doesn't poke out
2. Because it depends on classes like `ContractFactory` which are only available in `contract` and up.

packages/utils/src/test-utils/url-is-live.ts
```diff
@@ -0,0 +1,14 @@
+import { request } from 'http';
+import { parse } from 'url';
+
+export async function urlIsLive(url: string): Promise<boolean> {
```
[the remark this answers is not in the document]
nedsalk: Deleted this utility in https://github.com/FuelLabs/fuels-ts/pull/1356/commits/0105874eee3493060c3cd3e7da3a58c6213fa1df

packages/utils/src/test-utils/wait-until-unreachable.ts
```diff
@@ -0,0 +1,13 @@
+import { sleep } from '../utils/sleep';
+
+import { urlIsLive } from './url-is-live';
+
+export async function waitUntilUnreachable(url: string) {
```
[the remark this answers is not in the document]
nedsalk: [description](https://github.com/FuelLabs/fuels-ts/commit/0105874eee3493060c3cd3e7da3a58c6213fa1df#diff-94dfea50105ccb0600b4f5be412c5dc2031c831f63026ea680b0353aef292489R6)

packages/account/src/test-utils/wallet-config.ts
```diff
@@ -0,0 +1,171 @@
...
+  private generateWallets: () => WalletUnlocked[] = () => {
+    const generatedWallets: WalletUnlocked[] = [];
+    for (let index = 1; index <= this.options.count; index++) {
+      generatedWallets.push(new WalletUnlocked(randomBytes(32)));
+    }
+    return generatedWallets;
+  };
+
+  constructor(baseAssetId: string, config: WalletConfigOptions) {
+    const BASE_ASSET_ID = baseAssetId.startsWith('0x') ? baseAssetId : `0x${baseAssetId}`;
```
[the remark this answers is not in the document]
nedsalk: Seems like this safeguard is not needed anymore: https://github.com/FuelLabs/fuels-ts/pull/1356/commits/068544a01270222bc12c15ee6581cd098cb54f8d

tsconfig.base.json
nedsalk: I tested it with the wallet sdk and tests passed.

Dhaiwat10 (approved):

packages/contract/src/test-utils/launch-test-node.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  return wallets[config.walletIndex];
+}
+
+export async function launchTestNode<TFactories extends DeployContractConfig[]>({
+  providerOptions = {},
+  walletConfig = {},
+  nodeOptions = {},
+  // @ts-expect-error asd
```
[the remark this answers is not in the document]
nedsalk: Good catch! https://github.com/FuelLabs/fuels-ts/pull/1356/commits/9f612a22b1b128c29c8e7aa382faeec6e31fa209

arboleya (commented):
On a high level, the property `deployContracts` looks a bit weird.

Perhaps we could trim the terminologies:

```ts
launchTestNode({
  wallets: { ... },
  contracts: { ... },
  provider: { ... },
  node: { ... },
});
```


On the docs level, I've put up some suggestions here:
 - https://github.com/FuelLabs/fuels-ts/pull/2386

nedsalk:
> On a high level, the property `deployContracts` looks a bit weird.
> Perhaps we could trim the terminologies:

This was already [discussed](https://github.com/FuelLabs/fuels-ts/pull/1356#issuecomment-2003044598). In a nutshell, I'd love to have the API as you proposed it, but it just feels off having the inputs and outputs be named the same:
```ts
  wallets: { ... },
  contracts: [ ... ],
  provider: { ... },
  node: { ... }
});

const {
  wallets: [wallet],
  contracts: [contract],
  provider,
} = launched;
```
Glancing over the snippet presented like this brings unease to me :sweat_smile:. I realize that I might be alone in this, though, and would love input from the team. @FuelLabs/sdk-ts Please thumbs-up if you want me to change it to the proposed and thumbs-down if you want it to stay the way it is.

arboleya:
If nothing else, we should at least standardize the terminologies.

```ts
  walletsConfigs: { ... },
  contractsConfigs: [ ... ],
  providerOptions: { ... },
  nodeOptions: { ... }
});
```

nedsalk:
@arboleya I like that, makes sense. I made the changes in https://github.com/FuelLabs/fuels-ts/pull/1356/commits/fd4f2c5cc1f3cbd5b914de19d304592c7cb7624a.

maschad (approved):

Torres-ssf (dismissed):

Dhaiwat10 (dismissed):

Dhaiwat10 (approved):

maschad (approved):

Torres-ssf (approved):

arboleya (approved):

danielbate (approved):
:shipit:

apps/docs/src/guide/testing/index.md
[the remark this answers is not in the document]
danielbate: Resolved the above in #2937

Issue #1316, opened 2023-10-04
Upgrade ethers `v5` -> `v6`


danielbate:
Ethers `v6` came out earlier in the year so we should upgrade to future proof and benefit from the additional optimisations that come from the version upgrade, such as ES module support that will aid our efforts to support browser testing (pre-requisite for Vitest implementation #813). `v6` has also cleaned up a lot of the code and brought stricter function signatures so seems to have been developed with defensive programming in mind.

danielbate:
Upgrades:
- [x] `sha`
- [x] `pbkdf2`
- [x] `network`
- [x] `basex`
- [x] bytes (`BytesLike`)
- [x] `hexSlide`
- [ ] hexlify

Remove:
- [x] `abi`
- [x] `bignumber`
- [x] `properties`
- [x] `strings`
- [x] `logger` - our own errors package should take precedence (take care of via #1278)

Augment:
- [ ] `arrayify`
- [ ] `concat` - the function signature has been made stricter here and now only returns a `string`, we should look to implement our own bytes concatenation function also

PR #1365, opened 2023-10-20
fix: re-add `concat` function that accepts `BytesLike`

LuizAsFight:
`concat` function was purged in #1316 , and now it's exposed as `concatBytes`.

this PR aims to include the `concat` function again but in our own implementation

benefits:
- improve flexibility as `concatBytes` don't accept `BytesLike`.

problem of current implementation:
- the user needs to manually parse the arguments first
- unnecessary overhead to figure out how to convert, sdk could facilitate 

Before:
```
export function getContractTokenId(
  contractId: `0x${string}`,
  subId = ZeroBytes32
) {
  return hash(concat([byteContract, byteSubId]));
}
```

Now:
```
export function getContractTokenId(
  contractId: `0x${string}`,
  subId = ZeroBytes32
) {
  const byteContract = arrayify(contractId);
  const byteSubId = arrayify(subId);
  return hash(concatBytes([byteContract, byteSubId]));
}

```


LuizAsFight:
to overcome this problem in frontend I'm doing the calculations 

if someone in @FuelLabs/sdk-ts  can pick this one up to finish / include tests / any other thing in sdk process, would be good 🤓

otherwise should be ready for review
cc @arboleya @Torres-ssf

arboleya:
@LuizAsFight Right and we'll need additional tests covering this new method.

@danielbate I believe this has to do with the last things you did with the ethers upgrade, yes?

arboleya (dismissed):

danielbate:
I've gone through and just cleaned up usages of `concatBytes` in place of `concat` as it is closer to that of the previous implementation. I've also added a couple tests to the umbrella package so that the `ethers` re-xported functions aren't missed off in future work.

Previously we had been exporting the entire contents of `@ethersproject/bytes` which included `concat` within `fuels.` Not possible with `ethers` v6 as there were many function/classname collisions with our own packages. I maintained `hexlify` and `arrayify` in the upgrade as they were referenced by our documentation and used by integration tests, hence `concat` was purged however it's new API would not have been suitable.

danielbate (approved):

Torres-ssf (approved):

PR #1367, opened 2023-10-21
Update README.md

papaperez1:
update spelling


arboleya:
You created a tyop instead of correcting one.

PR #1368, opened 2023-10-24
chore: update eslint, prettier, typescript, typedoc

nedsalk:
This PR is mergable as-is and is for the purpose of splitting relevant and irrelevant changes made in #1356, as requested [here](https://github.com/FuelLabs/fuels-ts/pull/1356#pullrequestreview-1694017610).

It also updates `jest` and `ts-jest` just because that's necessary for #1356 to work, but it shouldn't interfere with #1310 as the updates are for minor versions.


apps/docs/package.json
```diff
@@ -25,7 +25,7 @@
     "markdown-it": "^13.0.1",
     "nodemon": "^2.0.22",
     "replace": "^1.2.2",
-    "typedoc": "^0.24.8",
+    "typedoc": "^0.25.1",
```
Torres-ssf, line 28: Should we also add to the description that we are updating `typedoc` ?

.eslintrc.js
```diff
@@ -13,7 +13,6 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'prettier',
     'plugin:eslint-comments/recommended',
-    'plugin:@typescript-eslint/recommended',
```
Dhaiwat10, line 16: Why did you remove this plugin?

Dhaiwat10 (approved):

jest.env.ts
```diff
@@ -21,7 +21,7 @@ import faucets from './.fuel-core/configs/faucets.json';
  * See: https://jestjs.io/docs/environment-variables#jest_worker_id
  */
 const getPrivateKeyForCurrentWorker = () => {
-  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+
```
arboleya, line 24: So you decided to allow _no-non-null_ assertions freely now?

.eslintrc.js
nedsalk: It is already defined above on line 13

jest.env.ts
nedsalk: No, it seems like the default changed from version to version. I set it back to warning again.

arboleya: Gotcha. To be clear, I have no hard feelings about it.

These `!` assertions must be manually set _anyway_; we could just as well turn off this rule. 👍

arboleya (approved):

PR #1333, opened 2023-10-10
fix: revert graphql subscriptions PR

nedsalk:
This PR reverts the changes done in #1256 due to #1298.

closes #1298.

After it is merged, I'll reopen  #301, #782.

PR #1374, opened 2023-10-25
feat: GraphQL subscriptions

### LINKED ISSUE #301: Update Provider Transaction Response waitForResult to poll for updates on SubmittedStatus

nedsalk:
This PR introduces GraphQL subscriptions without a dependency on any external library by using the [Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) and [async generators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncGenerator). The native [`EventSource`](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) unfortunately couldn't cut it because it's limited only to GET requests. Alternatives to `EventSource` like [eventsource](https://github.com/EventSource/eventsource), [SSE.js](https://github.com/mpetazzoni/sse.js), [fetch-event-source](https://github.com/Azure/fetch-event-source) have their own problems (only GET, slow/non-existant maintenance, missing typescript declarations, lack of nodejs support...), so I opted to just use the native APIs. The core logic is in [`fuel-graphql-subscriber.ts`](https://github.com/FuelLabs/fuels-ts/pull/1374/files#diff-ecd228ccff1188ae0d7975398ab20d89d85b10fc5f6bd3edb9e9902b1b31cb54) and fits in less than 100 lines of code.

Also, for documentation purposes, we already tried graphql-sse and [reverted it](https://github.com/FuelLabs/fuels-ts/pull/1333).

closes #301 #782 #1293.


packages/providers/src/fuel-graphql-subscriber.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const subscriptionsStreamReader = response
+    .body!.pipeThrough(new FuelSubscriptionStream())
+    .getReader();
+
+  for (;;) {
+    const { value, done } = await subscriptionsStreamReader.read();
+    if (value instanceof FuelError) throw value;
+    yield value;
```
nedsalk, line 72: Because it's an iterator, the fetch above won't run every time a new iteration occurs. Instead, once `yield` is hit, the next iteration will continue from the previously-hit `yield`, which means that it'll continue looping until `done` is `true`.

.nvmrc
```diff
@@ -1 +1 @@
-18.14.1
+18.18.2
```
nedsalk, line 1: This update is necessary because I switched to using native `fetch` and some bugfixes happened on it between v18.14.1 and v18.18.2.

package.json
```diff
@@ -5,7 +5,7 @@
   "author": "Fuel Labs <contact@fuel.sh> (https://fuel.network/)",
   "private": true,
   "engines": {
-    "node": "^18.14.1",
+    "node": "^18.18.2",
```
arboleya, line 8: This needs to be updated in all `package.json` files to [ensure compatibility](https://github.com/FuelLabs/fuels-ts/pull/1338) for all packages.

packages/providers/package.json
```diff
@@ -49,10 +49,9 @@
     "@fuel-ts/utils": "workspace:*",
     "@graphql-codegen/cli": "^2.13.7",
     "@graphql-codegen/typescript": "^2.8.0",
-    "@graphql-codegen/typescript-graphql-request": "^4.5.7",
     "@graphql-codegen/typescript-operations": "^2.5.5",
-    "@types/ramda": "^0.29.3",
+    "@graphql-codegen/typescript-generic-sdk": "^3.1.0",
```
arboleya, line 53: Am I right to assume that our tests/CI would still pass even if someone deleted these `codegen-*` plugins?

I say this because there is no reference to them in any Typescript file.

I wonder if we should try validating the GraphQL generation steps.

Have you thought about it?

arboleya (dismissed):
I just did a quick pass, but I still need to review it.

I also noticed a couple unused variables and imports in:
 - [packages/providers/test/provider.test.ts](https://github.com/FuelLabs/fuels-ts/pull/1374/files#diff-e2b89d46cef35a962ab8274426f26ddfa470f08ffaa8726895240a44d258b46f)

package.json
nedsalk: I created the #1399 PR to have this as part of the CI.

packages/providers/package.json
arboleya: Never mind, I just checked it, and the build will properly fail for missing plugins.

```
@fuel-ts/providers:prebuild: [FAILED] plugin_helpers_1.DetailedError is not a constructor
```

package.json
nedsalk: As far as the comment goes, I resolved it in [9c08646](https://github.com/FuelLabs/fuels-ts/pull/1374/commits/9c086f63f24350f4d2d6dbf456a650ec4b86ec5f).

arboleya: Thank you.

packages/providers/src/fuel-graphql-subscriber.ts
nedsalk: Because a `while` loop would require an external variable to break it:

```ts
let doneStreaming = false;

while(!doneStreaming) {
    const { value, done } = await subscriptionStreamReader.read();
    if (value instanceof FuelError) throw value;
    yield value;
    doneStreaming = done;
}
```

I like breaking an infinite loop with `break` more in cases where the additional variable is created only for the purpose of breaking the loop and doesn't bring any additional context with it.

packages/providers/src/fuel-graphql-subscriber.ts
```diff
@@ -0,0 +1,75 @@
...
+        if (text.startsWith('data:')) {
+          const { data, errors } = JSON.parse(text.split('data:')[1]);
+          if (Array.isArray(errors)) {
+            this.readableStreamController.enqueue(
+              new FuelError(
+                FuelError.CODES.INVALID_REQUEST,
+                errors.map((err) => err.message).join('\n\n')
+              )
+            );
+          } else this.readableStreamController.enqueue(data);
```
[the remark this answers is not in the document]
arboleya: Yep, me too.

I'd vote for a conventional `else { .. }` block to keep consistency with the rest of the codebase.

packages/providers/src/fuel-graphql-subscriber.ts
arboleya: Hmm, I have to agree here as well.

A more conventional syntax feels proper and keeps style consistency.

```ts
while (true) {
    const { value, done } = await subscriptionStreamReader.read();

    if (value instanceof FuelError) {
      throw value;
    }

    yield value;

    if (done) {
      break;
    }
}
```

nedsalk: Ha! I didn't even consider breaking a while loop with ...`break`, lol.

packages/providers/src/fuel-graphql-subscriber.ts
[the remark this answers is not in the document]
nedsalk: The codebase itself is not consistent on this. Setting `curly: ['error', 'all']` in `.eslintrc.js` gives 36 errors as of the current master. This is because our eslint config extends `airbnb-base` where they define [`curly: ['error', 'multiline']`](https://github.com/airbnb/javascript/blob/11ab37144b7f846f04f64a29b5beb6e00d74e84b/packages/eslint-config-airbnb-base/rules/best-practices.js#L31).

nedsalk:
Setting it as draft until the `beta-5` is merged, but I'd appreciate additional reviews if there are any as it's a finished piece of work.

packages/fuel-gauge/src/transaction-response.test.ts
```diff
@@ -75,25 +83,33 @@ describe('TransactionSummary', () => {
   });
 
   it('should ensure waitForResult always waits for the transaction to be processed', async () => {
-    const destination = Wallet.generate({
-      provider,
+    const { cleanup, ip, port } = await launchNode({
+      args: ['--poa-interval-period', '10s'],
     });
```
Dhaiwat10, line 88: Why are we starting a new node just for this test?

Dhaiwat10 (dismissed):

packages/providers/src/fuel-graphql-subscriber.ts
```diff
@@ -0,0 +1,81 @@
...
+  const subscriptionStreamReader = response
+    .body!.pipeThrough(new FuelSubscriptionStream())
+    .getReader();
+
+  while (true) {
+    const { value, done } = await subscriptionStreamReader.read();
+    if (value instanceof FuelError) {
+      throw value;
+    }
+    yield value;
```
arboleya, line 76: I wonder if we should go the generator's way for all this.

packages/providers/src/transaction-response/transaction-response.ts
```diff
@@ -204,14 +203,16 @@ export class TransactionResponse {
...
-    if (this.gqlTransaction?.status?.type === 'SubmittedStatus') {
-      await this.sleepBasedOnAttempts(++this.resultAttempts);
-
-      return this.waitForResult<TTransactionType>(contractsAbiMap);
+    for await (const { statusChange } of this.provider.operations.statusChange({
+      transactionId: this.id,
+    })) {
+      if (statusChange.type !== 'SubmittedStatus') {
+        break;
+      }
```
arboleya, line 211: This whole `for await` doesn't feel nice IMO.

Could we do it differently and improve readability somehow?

packages/providers/test/provider.test.ts
```diff
@@ -939,6 +929,82 @@ describe('Provider', () => {
     expect(estimateTxSpy).toHaveBeenCalled();
   });
 
+  it('An invalid subscription request throws a FuelError and does not hold the test runner (closes all handles)', async () => {
+    const provider = await Provider.create(FUEL_NETWORK_URL);
+
+    await expectToThrowFuelError(
+      async () => {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
arboleya, line 937: You should [prefix](https://github.com/FuelLabs/fuels-ts/blob/2a7ad0984e7f1a3822f07857784f67e4760f2ccf/.eslintrc.js#L55-L61) unused-but-required variables with `_` instead of disabling the rule.

arboleya (commented):
Did a first pass and left some notes — still need to go more in depth. 🫡

packages/fuel-gauge/src/transaction-response.test.ts
```diff
@@ -75,25 +83,33 @@ describe('TransactionSummary', () => {
   });
 
   it('should ensure waitForResult always waits for the transaction to be processed', async () => {
-    const destination = Wallet.generate({
-      provider,
+    const { cleanup, ip, port } = await launchNode({
+      args: ['--poa-interval-period', '10s'],
```
Torres-ssf, line 87: How many transactions are being executed in this node? 

If takes 10 seconds to process a transaction and supposedly the test executes 3 transactions, this means that this test will take at least 30 seconds to finish.

packages/providers/src/fuel-graphql-subscriber.ts
```diff
@@ -0,0 +1,81 @@
...
+    headers: {
+      'Content-Type': 'application/json',
+      Accept: 'text/event-stream',
+    },
+  });
+
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const subscriptionStreamReader = response
+    .body!.pipeThrough(new FuelSubscriptionStream())
+    .getReader();
```
Torres-ssf, line 69: Is it safe to do this non-null assertion here?

packages/providers/src/operations.graphql
```diff
@@ -4,6 +4,31 @@
 # generate `operations.ts` from this file.
 
 # Fragments
+
+fragment transactionStatusFragment on TransactionStatus {
+  type: __typename
+  ... on SubmittedStatus {
+    time
+  }
+  ... on SuccessStatus {
```
Voxelot, line 13: If it's of any use, you can now get all the receipts on either the Success or Failure status

Torres-ssf (changes requested):

packages/fuel-gauge/src/transaction-response.test.ts
nedsalk: To verify that the subscription will wait for 10 seconds until the transaction changes status.

packages/providers/src/fuel-graphql-subscriber.ts
nedsalk: Yes. The node always returns a non-null response on a subscription.

packages/providers/src/operations.graphql
nedsalk: Thanks for the info! I created an issue for that: #1530.

packages/providers/src/provider.ts
```diff
@@ -382,14 +409,31 @@ export default class Provider {
...
+        ? fuelGraphQLSubscriber({
+            url: this.url,
+            query,
+            fetchFn: (url, requestInit) =>
+              fetchFn(url as string, requestInit as FetchRequestOptions, this.options),
+            variables: vars as Record<string, unknown>,
+          })
+        : gqlClient.request(query, vars);
+    });
   }
```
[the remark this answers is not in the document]
nedsalk: I prefer it more in the way I wrote it. I'm not really a fan of putting callbacks into variables in cases like these because you have to specify the types of the parameters to conform with `getOperationsSdk`, whereas when you define the callback directly in it, you get the type inference implicitly.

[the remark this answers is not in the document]
nedsalk: I prefer it more as it is currently. I'm not really a fan of putting callbacks into variables in cases like these because you have to specify the types of the parameters to conform with `getOperationsSdk`, whereas when you define the callback directly in it, you get the type inference implicitly.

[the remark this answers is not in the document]
arboleya: @nedsalk If you don't specify, all types should be inferred just the same.

@Torres-ssf NIT'ing in this direction, I'd even go as far as formatting this:

```ts
const opDefinition = query.definitions.find((x) => x.kind === 'OperationDefinition') as { operation: string };
const isSubscription = opDefinition?.operation === 'subscription';

 if (isSubscription) {
  ...
```

I usually favor readability. We read code way more than we write it, and increased readability **_now_** usually means maintainability speed in the **_future_**. The opposite can be inversely proportional.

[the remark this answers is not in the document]
Torres-ssf: @arboleya agreed.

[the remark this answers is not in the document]
nedsalk: It seems like preference, as I find both implementations equally readable. I implemented your suggestion.

packages/fuel-gauge/src/transaction-response.test.ts
nedsalk: I changed it to 5s. I'd still like to have it big enough to actually test the subscription behavior properly.

packages/providers/test/provider.test.ts
nedsalk: The rule you referenced is `no-unused-vars`, whereas this is a problem with `@typescript-eslint/naming-convention`. I tried prefixing it and the linter wouldn't allow it due to that rule. I then tried writing my own rule for it to allow it, but it doesn't seem like a quick fix and focusing too much on it would be out of the scope of this issue. I created #1352 for that purpose.

nedsalk: Prefixing it with an underscore causes an issue with `@typescript-eslint/naming-convention`. I tried writing my own rule for it to allow it for a quick win, but it caused linting issues across the codebase :D. It doesn't seem like a quick fix so I created #1352. I did, however, prefix it and the rule that's now being ignored is `@typescript-eslint/naming-convention`.

nedsalk: Prefixing it with an underscore causes an issue with `@typescript-eslint/naming-convention`. I tried writing my own rule for it to allow it for a quick win, but it caused linting issues across the codebase :D. It doesn't seem like a quick fix so I created #1532. I did, however, prefix it and the rule that's now being ignored is `@typescript-eslint/naming-convention`.

arboleya: Good catch, I speed-read it and didn't notice it wasn't a _function argument_. In this case, please delete the issue you created and remove the `varsIgnorePattern` from the rule, leaving only `argsIgnorePattern`. We should not allow unused _const_ declarations.

Back to your code, I will repeat myself and say I feel this `for await` looks off. If that wasn't enough, the fact you're running around lint rules to make it pass is another indication of something unusual, which I usually flag as undesirable.

packages/providers/src/fuel-graphql-subscriber.ts
nedsalk: I don't see an alternative that would be this elegant. Even the graphql generator is generating `AsyncIterable<T>` types for subscriptions.

packages/fuel-gauge/src/transaction-response.test.ts
arboleya: Would it be possible to spy on the surroundings of the execution and ensure functionality without using `--poa`?

nedsalk: This is the only way to be 100% certain that a transaction who's processing is taking longer on the _node_ will be properly awaited. IMO this needs to be an integration test, nothing less.

packages/providers/src/transaction-response/transaction-response.ts
nedsalk: `for await...of` loops implicitly call the `generator.next()` method on every iteration, which yields back the next result. On break, they call the `generator.return()` method. The `generator.return()` call can be omitted for our generator as we don't do anything in it. A different implementation thus can be:
```ts
    const generator = this.provider.operations.statusChange({
      transactionId: this.id,
    });

    while (true) {
      const {
        value: { statusChange },
      } = await generator.next();
      if (statusChange.type !== 'SubmittedStatus') {
        break;
      }
    }
```

This doesn't make much sense to be used when we have `for await...of` built for the specific purpose of asynchronous iteration and it's calling the `.next()` and `.return()` methods for us.

This is the most elegant solution I could think of because it's using two language constructs: an [`async generator function`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function*) and the accompanying  [`for await...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of) loop. I don't know of anything else that would better fit the bill. If somebody wants to investigate further and propose another solution - I'd gladly review it.

nedsalk: The only way that I see this being nicer is to change our generated graphql types to enable a loop without the destructuring like this:
```ts
    for await (const statusChange of this.provider.operations.statusChange({
      transactionId: this.id,
    })) {
      if (statusChange.type !== 'SubmittedStatus') {
        break;
      }
    }
```

I can do that for subscriptions in this PR because nobody is using them. I wouldn't touch queries and mutations on `provider.operations` in the same manner, however, as it'd be a breaking change wherever our consumers are using the graphql sdk directly.

packages/providers/test/provider.test.ts
```diff
@@ -939,6 +929,82 @@ describe('Provider', () => {
...
+    expect(provider.options.timeout).toBeUndefined();
+  });
+
+  it('throws TimeoutError on timeout when calling an operation', async () => {
+    const timeout = 500;
+    const provider = await Provider.create(FUEL_NETWORK_URL, { timeout });
+    jest
+      .spyOn(global, 'fetch')
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore TS is throwing error when test is run, but not in IDE
```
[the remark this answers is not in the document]
nedsalk: Fixed! :)

packages/providers/src/transaction-response/transaction-response.ts
arboleya: I was looking for the first code snippet, thanks.

You can safely revert aa3e80f.

packages/providers/src/transaction-response/transaction-response.ts
```diff
@@ -133,7 +125,14 @@ export class TransactionResponse {
     });
 
     if (!response.transaction) {
-      await this.sleepBasedOnAttempts(++this.fetchAttempts);
+      for await (const statusChange of this.provider.operations.statusChange({
+        transactionId: this.id,
+      })) {
```
arboleya, line 130: This can be done to all occurrences to simplify the reading, reinforce the usage, and add context.

```suggestion
      const subscription this.provider.operations.statusChange({ transactionId: this.id });
      for await (const statusChange of subscription) {
```

packages/providers/test/provider.test.ts
```diff
@@ -939,6 +931,82 @@ describe('Provider', () => {
...
+          fetch(args[0] as RequestInfo | URL, args[1] as RequestInit | undefined)
+        )
+      );
+
+    const { error } = await safeExec(async () => {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      for await (const iterator of provider.operations.statusChange({
+        transactionId: 'doesnt matter, will be aborted',
+      })) {
+        // shouldn't be reached
```
arboleya, line 1001: Better to use this than to disable the rule.

```suggestion
        expect(iterator).toBeTruthy(); // shouldn't be reached
```

packages/fuel-gauge/src/transaction-response.test.ts
arboleya: There is no way we can accommodate a test with a 25s timeout. 😅 

Let's try thinking of other ways. We need it to be faster, without using `--poa`.

Requests to the node are always async, delays shouldn't be necessary to validate the behavior.

nedsalk: I lowered to `--poa-interval-period 3s` and the test takes 11s to complete on my machine. I removed the 25s timeout. That's acceptable in my books, especially for such a critical feature.
This is the _only_ test in our test suite that's **fully** verifying that our integration with the fuel node's server-sent events functionality works in cases where the transaction takes longer to process and its status isn't a success or failure right away. This is a real-world scenario that is to be expected; a node might have multiple transactions on it at one moment and delay the execution some.

To test this without the node, we'd have to mock `fetch` to return a `Response` object who's body would be a [`ReadableStream`-like object implementing a custom async iterator](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#async_iteration) that would:
1. Return the appropriate JSON with `TransactionStatus.SubmittedStatus`,
2. Sleep for a bit?
3. Return the appropriate JSON with `TransactionStatus.SuccessStatus`. 

This would probably work, but it wouldn't protect us from false positives due to possible breaking changes from one `fuel-core` version to another. IMO the benefits of a full integration test in this case outweigh the cost of a relatively slow test, especially because it's the only test of this kind in the code base. 

With all that said - if you still feel strongly about it, I'll implement the mocking.

.eslintrc.js
```diff
@@ -21,6 +21,12 @@ module.exports = {
     },
   },
   rules: {
+    'no-restricted-syntax': [
+      'off',
+      {
+        selector: 'ForOfStatement',
```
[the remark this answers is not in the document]
nedsalk: Based on your [suggestion](https://github.com/FuelLabs/fuels-ts/pull/1374#discussion_r1429128715), do you still hold this opinion?

packages/providers/src/fuel-graphql-subscriber.ts
nedsalk: Resolving based on [this](https://github.com/FuelLabs/fuels-ts/pull/1374#discussion_r1428112104) resolved discussion.

Torres-ssf (approved):
LGTM 🚀 

@nedsalk Why not adapt `TransactionResponse` class to make it possible to use the subscription `submitAndAwait` instead of just `submit` mutation when submitting a Transaction?

https://github.com/FuelLabs/fuels-ts/blob/1690004e856e08121c0464dddfee49517dbd0f62/packages/providers/src/provider.ts#L540-L544

.eslintrc.js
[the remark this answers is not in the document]
arboleya: One can't have it all.

packages/fuel-gauge/src/transaction-response.test.ts
arboleya: Much better!

arboleya:
We can have it as a secondary option, like `sendTransactionAndAwait()` or similar.

packages/providers/test/provider.test.ts
```diff
@@ -939,6 +931,82 @@ describe('Provider', () => {
...
+
+    await expectToThrowFuelError(
+      async () => {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        for await (const value of provider.operations.statusChange({
+          transactionId: 'invalid transaction id',
+        })) {
+          //
+        }
+      },
```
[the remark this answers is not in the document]
nedsalk: It seems like I fixed it and then reverted it...

nedsalk:
@arboleya 
> ... and there's the `sendTransactionAndAwait` thing.

This PR is just to introduce the subscription mechanism and verify that it works on `TransactionResponse` by replacing the previous polling mechanism with the new server-sent events one. Introducing `submitAndAwait` properly for the user to benefit from it would require some additional refactoring, thinking about the API on the level above the `Provider` (e.g. how to best provide it on `FunctionInvocationScope`, `Account` methods like `transfer`, etc.), which IMO deserves to be its own PR.

arboleya:
@nedsalk I thought it'd be easy; maybe I got something wrong.

Could you please elaborate on the rationale in a new issue and link it here?

nedsalk:
@arboleya 
> Could you please elaborate on the rationale in a new issue and link it here?

Here's the issue: #1563.

arboleya:
Thank you.

Torres-ssf:
> @Torres-ssf
> 
> > Why not adapt `TransactionResponse` class to make it possible to use the subscription `submitAndAwait` instead of just `submit` mutation when submitting a Transaction?
> 
> The `TransactionResponse` class isn't concerned with _submitting_ a transaction; it's concern is to get a submitted transaction, so I'm not sure why and how we'd adapt it for those mutations.
> 
> If you were thinking about modifying the `Provider.sendTransaction` method to use `submitAndAwait` as the default, that wouldn't be appropriate because a transaction can take an indefinite amount of time to execute on the node and using `submitAndAwait` as the default would block our users from executing anything on their side until the node executes the transaction. We'd have to create a new workflow for those who want to both submit _and_ await the execution of a transaction in one go.

Sorry @nedsalk I don't follow, can you help me by giving more context here?

My initial thought was to modify the `Provider.sendTransaction` to use `submitAndAwait` as I was under the impression that the process was not going to stop and wait till the transaction was processed as `submitAndAwait ` is also a subscription, just like the one being used in this PR `statusChange`.

Maybe I am missing something here.

Torres-ssf (approved):

arboleya (approved):

PR #1383, opened 2023-10-26
fix: update calculate fee outputs

Torres-ssf:
Made `CalculateTransactionFee` to return `minFee` and `maxFee`. 

These values are relevant when calculating the amount of funds a Transaction Request needs before being submitted to the chain

**EDIT**

On [1307](https://github.com/FuelLabs/fuels-ts/pull/1307), I did not return the `maxFee` because I was under the impression that it was not relevant, since the actual fee deducted from the user's wallet was always the `minFee`.

However, I've recently become aware that the [VM always uses](https://github.com/FuelLabs/fuel-vm/blob/94677aa1cfa5a150368c4820a8a757baac479e1f/fuel-vm/src/checked_transaction/balances.rs#L73-L78) the `maxFee` value to estimate if the transaction has enough balances to cover the fee.

In other words, we need `maxFee` in order to properly fund a transaction before submitting it to the chain.

### DIFF
--- a/.changeset/strong-garlics-prove.md
+++ b/.changeset/strong-garlics-prove.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/providers": minor
+---
+
+made calculateTransactionFee to return minFee and maxFee
--- a/packages/providers/src/provider.ts
+++ b/packages/providers/src/provider.ts
@@ -678,7 +678,7 @@ export default class Provider {
 
     const gasUsed = getGasUsedFromReceipts(receipts);
 
-    const { fee } = calculateTransactionFee({
+    const { minFee } = calculateTransactionFee({
       gasPrice,
       gasPerByte,
       gasPriceFactor,
@@ -691,7 +691,7 @@ export default class Provider {
       minGasPrice,
       gasPrice,
       gasUsed,
-      fee,
+      fee: minFee,
     };
   }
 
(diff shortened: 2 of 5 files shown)


arboleya (approved):
The code looks ok, but I must admit, I didn't follow the change.

For example, why is this right and the previous one wrong?

I'm assuming we had no errors before, and both seem to work ok.

> The terminology `feeFromGasUsed` also feels weird, but maybe it's me out of context.

Torres-ssf:
> The code looks ok, but I must admit, I didn't follow the change.
> 
> For example, why is this right and the previous one wrong?
> 
> I'm assuming we had no errors before, and both seem to work ok.
> 
> > The terminology `feeFromGasUsed` also feels weird, but maybe it's me out of context.

@arboleya apologies for the lack of a deeper context here.

On [1307](https://github.com/FuelLabs/fuels-ts/pull/1307), I did not return the `maxFee` because I was under the impression that it was not relevant, since the actual fee deducted from the user's wallet was always the `minFee`.

However, I've recently become aware that the [VM always uses](https://github.com/FuelLabs/fuel-vm/blob/94677aa1cfa5a150368c4820a8a757baac479e1f/fuel-vm/src/checked_transaction/balances.rs#L73-L78) the `maxFee` value to estimate if the transaction has enough balances to cover the fee.

In other words, we need `maxFee` in order to properly fund a transaction before submitting it to the chain.

The value from `feeFromGasUsed` is related to the part of the total fee that will always be charged depending on the amount of gas used.

danielbate (approved):

PR #1388, opened 2023-10-31
fix: provider connect method accept options

### LINKED ISSUE #1396: Accept `ProviderOptions` via `Provider.connect`
Currently, it is possible to pass in some `ProviderOptions` for customization via `Provider.create`:
```ts
const provider = await Provider.create(url, { fetch: customFetch });
```
but not via `Provider.connect`
```ts
await provider.connect(newNetworkUrl);
```

This should be possible.

LuizAsFight:
I can create a `Provider` passing custom client options with 
```
const provider = await Provider.create(url, { fetch: customFetch });
```

but when connecting to a new network url, we cannot pass the options anymore
```
await provider.connect(newNetworkUrl);
```

---------------
this PR aims to
- allow `connect` method to receive `options`

Closes #1396


packages/providers/src/provider.ts
```diff
@@ -310,9 +310,9 @@ export default class Provider {
   /**
    * Updates the URL for the provider and fetches the consensus parameters for the new URL, if needed.
    */
-  async connect(url: string) {
+  async connect(url: string, options?: ProviderOptions) {
     this.url = url;
-    this.operations = this.createOperations(url);
+    this.operations = this.createOperations(url, options);
     await this.fetchChainAndNodeInfo();
   }
```
luizstacio, line 317: Should we save the options when we create the instance and in this case just forward the options?

luizstacio (changes requested):

packages/providers/src/provider.ts
LuizAsFight: I think it doesn't make sense as different networks can have different fetch options

luizstacio: In this case, you would create a completely new provider... we can also make it optional and use `options || this.options`.

Dhaiwat10:
Thanks @LuizAsFight, I'll get this one over the line

packages/providers/src/provider.ts
```diff
@@ -243,7 +243,7 @@ export default class Provider {
    * @param options - Additional options for the provider
    * @hidden
    */
-  protected constructor(
```
Dhaiwat10, line 246: @LuizAsFight I'm curious - why did you remove the `protected` modifier from the constructor here?

packages/providers/src/provider.ts
Dhaiwat10: I think saving the options makes sense

packages/providers/src/provider.ts
arboleya: Yes, I'm assuming it was unintentional.

That should be kept.

packages/providers/src/provider.ts
arboleya: Yep, and also giving the option to override it on the `connect` method:

```ts
this.operations = this.createOperations(url, options ?? this.options);
```

packages/providers/src/provider.ts
LuizAsFight: yep can remove it @Dhaiwat10

arboleya (approved):

Torres-ssf (approved):

PR #1391, opened 2023-11-02
test: run only fuel-gauge **not parallelized**

nedsalk:


nedsalk:
I only needed it for the [test time](https://github.com/FuelLabs/fuels-ts/actions/runs/6730866971/job/18294606143?pr=1391#step:4:3222).

PR #1392, opened 2023-11-02
test: run only fuel-gauge **parallelized**

nedsalk:
The current situation on #1310:
Local tests went from174s to 24s on @danielbate's machine. The latest successful CI on his that PR had the tests finish in [327s](https://github.com/FuelLabs/fuels-ts/actions/runs/6735982301/job/18310563161#step:4:3419), which is ~150s faster than jest (e.g. see [this run](https://github.com/FuelLabs/fuels-ts/actions/runs/6811609558/job/18522299045#step:9:305)).

As far as this branch goes:
After having all the tests that used the node now run their own node, running the tests on @danielbate's machine took ~25s.
Those tests took around [270s](https://github.com/FuelLabs/fuels-ts/actions/runs/6802429219/job/18495506273#step:5:4052) on the CI.

However, when I set the isolate property of vitest to false, the tests took **10s** on @danielbate's machine and [~75s](https://github.com/FuelLabs/fuels-ts/actions/runs/6813519232/job/18528294918#step:5:4248) on the CI.

Setting the isolate flag doesn't come without a cost, however. The [documentation](https://vitest.dev/guide/features.html#threads) states:
"Vitest also isolates each file's environment so env mutations in one file don't affect others. Isolation can be disabled by passing --no-isolate to the CLI (trading correctness for run performance)."
Due to this, some mocks broke. The tests using `TestNodeLauncher` and the other additional helpers however weren't affected, as they're just running the implementation and don't depend on the test runner's environment like mocking. for example.

@danielbate suggested a great idea that we could split our tests: those that depend on vitest's environment (e.g. need mocks like some unit tests) and those that don't (like feature, integration, e2e tests). We can then set `isolate` on one test config to `false`, and in another to `true` and have the best of both worlds.


nedsalk:
This PR served its purpose of determining the proposed test parallelization approach's effect on CI time. I'll close it for the time being and either reopen it later or use its changes in a gradual approach to parallelizing tests package by package (more likely).

PR #1419, opened 2023-11-09
feat: add `deploy` flag to fuels cli `build` command

### LINKED ISSUE #1418: Add `deploy` flag to Fuels CLI `build` command
Before being able to `build` frontend applications, users must run `fuels build` to generate all required files.

However, the `contract-ids.json` file is only created when contracts are deployed via the `deploy` command.

That's why we need this flag, to make it straightforward to have routines like:

```json
{
  "scripts": {
    "prebuild": "fuels build --depoy",
    "build": "next build",
  }
}
```

> This was a requirement for https://github.com/FuelLabs/fuels-ts/pull/1408

arboleya:
- Close #1418

### DIFF
--- a/apps/docs/src/guide/cli/commands.md
+++ b/apps/docs/src/guide/cli/commands.md
@@ -51,11 +51,33 @@ In a nutshell:
 ## `fuels build`
 
 ```console
+npx fuels help build
+```
+
+```
+Options:
+  -p, --path <path>  Path to project root (default: "/Users/anderson/Code/fuel/fuels-ts/apps/docs")
+  -d, --deploy       Deploy contracts after build (auto-starts a `fuel-core` node if needed)
+  -h, --help         Display help
+```
+
+Examples:
+
+```console
 npx fuels build
 ```
 
 1.  Build all Sway programs under your `workspace` using `forc` <sup>[1](#commands-for-wrapped-utiltities)</sup>
 1.  Generate types for them using `fuels-typegen` <sup>[2](#typegen)</sup>
+
+```console
+npx fuels build --deploy
+```
+
+Using the `--deploy` flag will aditionally:
+
+1. Auto-start a short-lived `fuel-core` node if _needed_ ([docs](./config-file.md#autostartfuelcore))
+1. Runs and `deploy` and quit
 
 ## `fuels deploy`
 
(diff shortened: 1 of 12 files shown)


apps/docs/src/guide/cli/commands.md
```diff
@@ -50,13 +50,35 @@ In a nutshell:
...
 1.  Generate types for them using `fuels-typegen` <sup>[2](#typegen)</sup>
 
+```console
+npx fuels build --deploy
+```
+
+Using the `--deploy` flag will aditionally:
+
+1. Auto-start a short-lived `fuel-core` node if _needed_ ([docs](./config-file.md#autostartfuelcore))
+1. Runs and `deploy` and quit
```
nedsalk, line 80:
```suggestion
2. Run `deploy` and kill the node

This is useful when working with contracts, because a contract's id is generated on deployment.
```

apps/docs/src/guide/cli/commands.md
```diff
@@ -50,13 +50,35 @@ In a nutshell:
...
 ## `fuels build`
 
+```console
+npx fuels help build
+```
+
+```
+Options:
+  -p, --path <path>  Path to project root (default: "/Users/anderson/Code/fuel/fuels-ts/apps/docs")
+  -d, --deploy       Deploy contracts after build (auto-starts a `fuel-core` node if needed)
```
nedsalk, line 60: Just an observation:

It would be nice to have some way to use the text in 
```ts
.option('-d, --deploy', 'Deploy contracts after build (auto-starts a `fuel-core` node if needed)')
```
here as well, instead of copy-pasting.

arboleya: It would, not only here but in a lot of doc pages.

I don't know how that would be possible.

Probably not worth the trouble.

nedsalk (approved):

apps/docs/src/guide/cli/commands.md
nedsalk: Maybe the usage of regions, something like:
```ts
const fuelsBuildOptions: Record<string, { option: string; description: string }> = {
  path: {
    option:
      // #region fuels-build-options-path-option
      '-p, --path <path>',
    // #endregion
    // #region fuels-build-options-path-description
    description: 'Path to project root',
    // #endregion
  },
  deploy: {
    // #region fuels-build-options-deploy-option
    option: '-d, --deploy',
    // #endregion
    // #region fuels-build-options-deploy-description
    description: 'Deploy contracts after build (auto-starts a `fuel-core` node if needed)',
    // #endregion
  },
};
```

nedsalk (approved):

Dhaiwat10 (approved):

PR #1426, opened 2023-11-10
feat: improve ABI Coders `decode` validation

### LINKED ISSUE #1353: [Mirror] TOB-FUEL-32: ABI decoder is crashing on certain inputs
This is a mirror of https://github.com/FuelLabs/fuels-rs/issues/1108. 

Opening an issue to investigate these same decoding bugs from the Rust SDK in the Typescript as well. 

Trail of Bits (the company doing some audits in our stack) hasn't gone through the TS SDK codebase yet, but some of these problems could also be here; It's worth investigating those proactively.

danielbate:
Now validates for overflow/underflow and empty bytes, and adds a max capacity check for arrays and vecs.

Closes #1353


packages/abi-coder/src/coders/byte.ts
```diff
@@ -54,9 +54,18 @@ export class ByteCoder extends Coder<number[], Uint8Array> {
...
 
   decode(data: Uint8Array, offset: number): [Uint8Array, number] {
+    if (data.length < BASE_VECTOR_OFFSET) {
+      this.throwError(ErrorCode.DECODE_ERROR, 'Invalid byte data size.');
+    }
+
     const len = data.slice(16, 24);
     const length = bn(new U64Coder().decode(len, 0)[0]).toNumber();
-    const byteData = data.slice(BASE_VECTOR_OFFSET, BASE_VECTOR_OFFSET + length * 8);
+    const byteDataLength = length;
```
danielbate, line 63: @camsjams pretty sure we don't want to be multiplying by `WORD_SIZE` here. When length is initially encoded it is just the array length. Would have been fine as we were slicing from the correct point, but to a point larger than the size of the byte array. And as heap types go at the end of the bytecode didn't make a difference to integration tests. But potential for overflow issues.

danielbate:
Thanks for your reviews @camsjams, will address them all in due course! Going to convert to draft for now tho as lots of overlapping changes with #1435 and that takes prescedence.

packages/providers/src/transaction-summary/assemble-transaction-summary.test.ts
```diff
@@ -47,7 +48,28 @@ describe('TransactionSummary', () => {
     gasCosts = provider.getChain().gasCosts;
   });
 
+  beforeEach(() => {
+    vi.resetAllMocks();
+  });
+
+  const mockCalculateTransactionFee = () => {
```
danielbate, line 55: Added this mock as the initial mocked bytecode was invalid. Wasn't a problem prior as the decoding would just return 0 and the test would pass. However with the additional validation, invalid bytes for a type will `throw`.

danielbate:
Have merged `rc/salamander` so this work is blocked by #1495

arboleya:
@danielbate Shouldn't this PR be against `beetle`? Why did you pull from `salamander`?

I'm afraid we [still] can't have work on a given RC depending on work from another RC.

danielbate:
`rc/salamander` was included for debugging with Vitest. However as there is enough time, I shall target this against `salamander` instead of `beetle`, removing the need to revert that work.

packages/abi-coder/src/coders/boolean.ts
```diff
@@ -1,6 +1,8 @@
 import { ErrorCode } from '@fuel-ts/errors';
 import { bn, toBytes } from '@fuel-ts/math';
 
+import { WORD_SIZE } from '../constants';
```
arboleya, line 4: `WORD_SIZE` is not in use.
```suggestion
```

packages/abi-coder/src/coders/number.ts
```diff
@@ -1,6 +1,8 @@
 import { ErrorCode } from '@fuel-ts/errors';
 import { toNumber, toBytes } from '@fuel-ts/math';
 
+import { WORD_SIZE } from '../constants';
```
arboleya, line 4: `WORD_SIZE` is not in use.
```suggestion
```

arboleya (changes requested):
Great stuff! Did a quick first pass.

arboleya (approved):
Meticulous job! 🎯😌

.changeset/giant-pumpkins-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/abi-coder": minor
+---
+
+Improve decode valiadtion of ABI Coders
```
nedsalk, line 5:
```suggestion
Improve decode validation of ABI Coders
```

packages/abi-coder/src/coders/array.ts
```diff
@@ -34,6 +35,10 @@ export class ArrayCoder<TCoder extends Coder> extends Coder<
   }
 
   decode(data: Uint8Array, offset: number): [DecodedValueOf<TCoder>, number] {
+    if (data.length < this.encodedLength || data.length > MAX_BYTES) {
+      this.throwError(ErrorCode.DECODE_ERROR, `Invalid array data size.`);
```
nedsalk, line 39: I don't see much merit in the existence of `this.throwError(...)` because internally it's just a wrapper around `new FuelError`. It's one one-liner replacing another one-liner.

However, what could be done to make this more meaningful is to change `throwError`:
```ts
  throwError(type: "encode" | "decode", message: string): never {
    throw new FuelError(type === "encode" ? ErrorCode.ENCODE_ERROR : DECODE_ERROR, message);
  }
```

This suggestion is only to make the `throwError` method have a reason to exist. I'd also be fine with removing it totally and replacing its invocations with `new FuelError(...)`.

danielbate: I think my preference is to get rid of the function, it does seem unnecessary.

nedsalk (approved):

arboleya (approved):

Issue #1331, opened 2023-10-10
Inaccurate Offset Calculation for a Decoded Transaction witnesses


Torres-ssf:
There appears to be an inconsistency in decoding the Transaction `witnesses` data. 

The `witnesses` data is the last part of an encoded transaction. Because of this, an accurate offset calculation can be done through the following simple expression:

```ts
const witnessesLength = transaction.witnesses.reduce((total, w) => total + w.dataLength + 8, 0)

const witnessesOffset = transactionBytesLength - witnessesLength
```

However, while debugging `fuel-core` locally and running tests with many different transactions, I've noticed that the offset calculation is not always accurate. While the calculation is correct in some instances, there are discrepancies ranging from 4 to 20 in others. Could this be an indication that something is off when decoding the transaction `witnesses`?

## Steps to Reproduce
1. Run `fuel-core` locally in debug mode.
2. Within `fuel-core` external libraries look for the `fuel-vm` file: `fuel-tx/src/transaction/fee.rs`
3. Add breakpoints on function `checked_from_values` and watch the value of `metered_bytes` variable.
4. Submit different types of transactions: simple transfers, contract deployments, coin mintings, contract calls, and multi-contract calls.

nedsalk:
I tried calculating the `witnessesOffset` value in the way you presented above inside of `calculateTxChargeableBytes` and all the `packages/fuel-gauge/src/fee.test.ts` tests passed. Does that mean that this is not an issue anymore or are the `fee.test.ts` tests not testing all possible scenarios?

Torres-ssf:
@nedsalk Not sure what you mean by tests not testing all possible scenarios. I have added tests for all sorts of transactions.

The `packages/fuel-gauge/src/fee.test.ts` tests are meant to validate that the fee is being properly handled. Which has nothing to do with this corner case.

The error above does not happen on every execution. I couldn't determine why.

nedsalk:
> Not sure what you mean by tests not testing all possible scenarios. I have added tests for all sorts of transactions.
> 
> The packages/fuel-gauge/src/fee.test.ts tests are meant to validate that the fee is being properly handled. Which has nothing to do with this corner case. 

Sorry, the second reproduction step you mentioned _(2. Within fuel-core external libraries look for the fuel-vm file: fuel-tx/src/transaction/fee.rs)_ and me stumbling upon this issue while investigating #1428 swayed me into thinking that the `fee.test.ts` covers this as well :sweat_smile:. 

[...]

PR #1428, opened 2023-11-11
fix: `calculateTxChargeableBytes` when transaction has no witnesses

nedsalk:
The chargeable bytes of a transaction are all transaction bytes except the witnesses part of the transaction.
`calculateTxChargeableBytes` was calculating properly when witnesses were present, but not when there were none: it returned `0` instead of `transactionBytes.length`. `transactionBytes.length` is the correct value because, without witnesses, _all_ the bytes of a transaction are chargeable.


nedsalk:
> @camsjams
>  Changes make sense, do you think we need more tests around this?

Not really, I think these are enough as far as the functionality goes. 
On a related note, there is also #1331 which will refactor out `Witness.offset` and these tests will be affected again.

packages/providers/src/utils/calculate-fee.test.ts
```diff
@@ -120,33 +120,34 @@ describe(__filename, () => {
...
-          dataLength: 64,
-          data: '0x3d8943b87436d54d7f1282ed9b01d38633f8003a37676276bbaf0560390d21f826b5f6e9ea85b1bca435af2bec4982436c20c1294876025f78aaa0143867eb7f',
-          offset,
-        },
-      ];
 
       const chargeableBytesFee = calculateTxChargeableBytes({
         transactionBytes,
-        transactionWitnesses,
+        transactionWitnesses: [],
```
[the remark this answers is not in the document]
nedsalk: It happens :man_shrugging:

packages/providers/src/utils/calculate-fee.ts
```diff
@@ -31,9 +31,9 @@ export interface CalculateTxChargeableBytesParams {
 
 /** @hidden */
 export const calculateTxChargeableBytes = (params: CalculateTxChargeableBytesParams): BN => {
-  const { transactionWitnesses } = params;
+  const { transactionWitnesses, transactionBytes } = params;
 
-  const txChargeableBytes = bn(transactionWitnesses?.[0]?.offset || 0);
+  const txChargeableBytes = bn(transactionWitnesses?.[0]?.offset || transactionBytes.length);
```
Torres-ssf, line 36: Good catch.

Torres-ssf (approved):

arboleya (approved):

packages/providers/src/utils/calculate-fee.ts
nedsalk: Thanks!

PR #1429, opened 2023-11-11
fix: `dataLength` of witness should be `u16`, not `u32`

nedsalk:
As per the [specs](https://github.com/FuelLabs/fuel-specs/blob/master/src/tx-format/witness.md), `dataLength` of a witness is of type `uint16`, not `uint32`.


arboleya (approved):
🕵️‍♂️

PR #1454, opened 2023-11-25
fix: ensure hex values are prefixed or fail

### LINKED ISSUE #1148: Ensure that expected hex data is always prefixed
Follow-up from #1093:

We should ensure that in all the places where we are expecting the input to be a hex value, it is correctly prefixed. We should not be converting loose hex values into hex prefixed strings (this can cause serious errors when applied to non-hex values incorrectly).

camsjams:
Closes #1148 

Storage slot json files do not prefix hex values:
```json
[
  {
    "key": "f383b0ce51358be57daa3b725fe44acdb2d880604e367199080b4379c41bb6ed",
    "value": "0000000000000000000000000000000000000000000000000000000000000000"
  }
]
```

There is an issue worth pursuing separately, but I think its fine as is since storage-slots are presumed to be generated by `forc` and can therefore be trusted as a hex value.

### DIFF
--- a/.changeset/selfish-rocks-tap.md
+++ b/.changeset/selfish-rocks-tap.md
@@ -1,0 +1,5 @@
+---
+"fuels": patch
+---
+
+Updated npm keywords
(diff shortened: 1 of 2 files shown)


arboleya (approved):

danielbate (approved):
Adding a testcase for the reduced coverage on `utils.ts` would be nice to have, but I will leave up to you 👍🏻

PR #1407, opened 2023-11-06
chore: make `SDK` conform with `Tx` policies

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/1436

# Requirements

- Update rust to version 1.73 — `rustup update`

> This is necessary to build the latest `forc` from source.

# Status

```
Test Suites: 3 failed, 249 passed, 252 total
Tests:       4 failed, 18 skipped, 1436 passed, 1458 total
```

# Failing tests

- [x] `apps/docs-snippets/src/guide/contracts/call-parameters.test.ts` _(fee-related)_

<br />

- [x] `packages/fuel-gauge/src/bytes.test.ts` _(predicate args using heap-types)_
- [x] `packages/fuel-gauge/src/contract.test.ts` _(fee-related)_
- [x] `packages/fuel-gauge/src/coverage-contract.test.ts` _(flaky, two tests using same wallet pvt key)_
- [x] `packages/fuel-gauge/src/fee.test.ts` _(fee-related)_
- [x] `packages/fuel-gauge/src/predicate/predicate-arguments.test.ts` _(predicate args using heap-types)_
- [x] `packages/fuel-gauge/src/predicate/predicate-invalidations.test.ts` _(fee-related)_
- [x] `packages/fuel-gauge/src/predicate/predicate-with-contract.test.ts` _(fee-related)_
- [x] `packages/fuel-gauge/src/predicate/predicate-with-script.test.ts` _(fee-related)_
- [x] `packages/fuel-gauge/src/raw-slice.test.ts` _(predicate args using heap-types)_
- [x] `packages/fuel-gauge/src/script-main-args.test.ts` _(fee-related)_
- [x] `packages/fuel-gauge/src/std-lib-string.test.ts` _(predicate args using heap-types)_
- [x] `packages/fuel-gauge/src/vector-types.test.ts` _(predicate args using heap-types)_

<br />

- [x] `packages/providers/test/provider.test.ts` _(pending version string)_

<br />

- [x] `packages/script/src/script.test.ts` _(args encoding problem)_

<br />

- [x] `packages/wallet/src/account.test.ts` _(fee-related)_
- [x] `packages/wallet/src/transfer.test.ts` _(fee-related)_

<br />

- [x] `packages/versions/scripts/rewriteVersions.test.ts` _(pending version string)_
- [x] `packages/versions/src/lib/getBuiltinVersions.test.ts` _(pending version string)_

PR #1455, opened 2023-11-25
chore: adjust new predicate offset

camsjams:
Trying new angle on some of the Predicate changes in https://github.com/FuelLabs/fuels-ts/pull/1407

### Goals
- keep logic in `packages/transactions/src/coders/transaction.ts` more general
- abstract functionality to read policy meta in `packages/providers/src/transaction-request/transaction-request.ts`
- maintain domain of `packages/predicate/src/predicate.ts` to handle offset work and Predicate data handling, as well as using pre-existing `predicateData` property

### Tests pass the same

Base Branch:
image

This Branch:
image

### DIFF
--- a/packages/predicate/src/predicate.ts
+++ b/packages/predicate/src/predicate.ts
@@ -76,7 +76,7 @@ export class Predicate<ARGS extends InputValue[]> extends Account implements Abs
         // eslint-disable-next-line no-param-reassign
         input.predicate = this.bytes;
         // eslint-disable-next-line no-param-reassign
-        input.predicateData = this.#getPredicateData(policies.length * WORD_SIZE);
+        input.predicateData = this.#getPredicateData(policies.length);
       }
     });
 
@@ -117,7 +117,7 @@ export class Predicate<ARGS extends InputValue[]> extends Account implements Abs
     return this;
   }
 
-  #getPredicateData(txLength: number): Uint8Array {
+  #getPredicateData(policiesLength: number): Uint8Array {
     if (!this.predicateArgs.length) {
       return new Uint8Array();
     }
@@ -134,7 +134,7 @@ export class Predicate<ARGS extends InputValue[]> extends Account implements Abs
       INPUT_COIN_FIXED_SIZE +
       WORD_SIZE +
       paddedCode.byteLength +
-      txLength;
+      policiesLength * WORD_SIZE;
 
     return mainFn?.encodeArguments(this.predicateArgs, OFFSET) || new Uint8Array();
   }


packages/predicate/src/predicate.ts
```diff
@@ -68,14 +69,14 @@ export class Predicate<ARGS extends InputValue[]> extends Account implements Abs
...
+
     request.inputs?.forEach((input) => {
       if (input.type === InputType.Coin && hexlify(input.owner) === this.address.toB256()) {
         // eslint-disable-next-line no-param-reassign
         input.predicate = this.bytes;
         // eslint-disable-next-line no-param-reassign
-        input.predicateData = this.predicateData;
-        // eslint-disable-next-line no-param-reassign
-        input.getPredicateData = this.getPredicateData;
+        input.predicateData = this.#getPredicateData(policies.length * WORD_SIZE);
```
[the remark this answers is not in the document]
arboleya: I agree, and I recommend separating concerns with extra explicitness and moving the calculation part.

It reads better together in summing up the whole TX, and we avoid introducing the `txLength` variable name later on, which may sound a bit generic.

```suggestion
        input.predicateData = this.#getPredicateData(policies.length);
```

packages/predicate/src/predicate.ts
```diff
@@ -111,18 +112,31 @@ export class Predicate<ARGS extends InputValue[]> extends Account implements Abs
    * @returns The Predicate instance with updated predicate data.
    */
   setData<T extends ARGS>(...args: T) {
+    this.predicateArgs = args;
+
+    return this;
+  }
+
+  #getPredicateData(txLength: number): Uint8Array {
```
arboleya, line 120: Following the previous suggestion.

```suggestion
  #getPredicateData(policiesLength: number): Uint8Array {
```

packages/predicate/src/predicate.ts
```diff
@@ -111,18 +112,31 @@ export class Predicate<ARGS extends InputValue[]> extends Account implements Abs
...
-      mainFn?.encodeArguments(args, SEMI_OFFSET + txLength) || new Uint8Array();
-
-    return this;
+    const OFFSET =
+      VM_TX_MEMORY +
+      SCRIPT_FIXED_SIZE +
+      INPUT_COIN_FIXED_SIZE +
+      WORD_SIZE +
+      paddedCode.byteLength +
+      txLength;
```
arboleya, line 137: Following previous suggestions.

```suggestion
      (policiesLength * WORD_SIZE);
```

arboleya (approved):
TBH, we started the change in this direction in a huddle, and then the callback approach was suggested and tested. It stuck because it worked, and we were in a rush.

I do prefer this one though, by far.

Left a minor suggestion.

Great stuff! 🚀

Issue #1448, opened 2023-11-23
Add CI check for the `authors` field for Sway contracts


Dhaiwat10:
A common requested change in our PRs is changing the `authors` field to `FuelLabs` from our personal IDs because when you run `forc init`, it always uses your personal ID for that field.

We should add a CI check for this. More context: https://github.com/FuelLabs/fuels-ts/pull/1403#discussion_r1387726172

PR #1456, opened 2023-11-27
chore: ensure all sway projects follow the `authors` naming convention

### LINKED ISSUE #1447: Update the `authors` field in Sway contracts throughout the repo
The `authors` field should be equal to `authors = ["Fuel Labs <contact@fuel.sh>"]` everywhere for consistency.

nedsalk:
closes #1447, #1448

### DIFF
--- a/.changeset/many-carrots-draw.md
+++ b/.changeset/many-carrots-draw.md
@@ -1,0 +1,5 @@
+---
+"fuels": patch
+---
+
+fix: make `startFuelCore` package manager agnostic
--- a/packages/fuels/package.json
+++ b/packages/fuels/package.json
@@ -76,6 +76,7 @@
     "handlebars": "^4.7.7",
     "joycon": "^3.1.1",
     "lodash.camelcase": "^4.3.0",
+    "npm-which": "^3.0.1",
     "portfinder": "^1.0.32",
     "rimraf": "^3.0.2",
     "toml": "^3.0.0",
--- a/packages/fuels/src/cli/commands/dev/startFuelCore.ts
+++ b/packages/fuels/src/cli/commands/dev/startFuelCore.ts
@@ -6,11 +6,13 @@ import { getPortPromise } from 'portfinder';
 import treeKill from 'tree-kill';
 
 import type { FuelsConfig } from '../../types';
-import { findPackageRoot } from '../../utils/findPackageRoot';
 import { getBinarySource } from '../../utils/getBinarySource';
 import { error, log, loggingConfig } from '../../utils/logger';
 
 import { defaultChainConfig, defaultConsensusKey } from './defaultChainConfig';
+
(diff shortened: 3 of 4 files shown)


scripts/forc-check.sh
```diff
@@ -5,14 +5,27 @@ main_dir=$(pwd)
...
+        cd "$main_dir" || exit
+        continue
+    fi
+
+    authors=$(cat Forc.toml | grep authors)
+    if [ "$authors" != "authors = [\"Fuel Labs <contact@fuel.sh>\"]" ]; then
+        echo -e authors field should be: $RED"authors = [\"Fuel Labs <contact@fuel.sh>\"]"$NC but is $RED"$authors"$NC
+        ERRORED=1
+    fi
+    cd "$main_dir" || exit
```
arboleya, line 28: Maybe something like this could also work:

```sh
    # not tested / pseudo-code to illustrate

    author="authors = [\"Fuel Labs <contact@fuel.sh>\")]"
    is_invalid=$(grep -LE "/(\[project\])|(${author})/g" $i)

    if [ ! -z "${is_invalid}" ]; then
      echo "Authors field should be: $RED${author}"
      echo $i
      exit 1
    fi
```


Taken from [here](https://github.com/FuelLabs/fuels-ts/blob/cbd1459d08437de75da129d1cd46c30a1f33d675/scripts/tests-validate.sh#L1-L12).

scripts/forc-format.sh
```diff
@@ -5,7 +5,28 @@ main_dir=$(pwd)
...
+    new_content=""
+    if [ "$authors_line" = "" ]; then
+        new_content=$(awk 'NR == 2 {print "authors = [\"Fuel Labs <contact@fuel.sh>\"]"} 1' Forc.toml)
+    else
+        new_content=$(awk -v line="$authors_line" 'NR == line {$0="authors = [\"Fuel Labs <contact@fuel.sh>\"]"} 1' Forc.toml)
+    fi
+
+    echo "$new_content" >Forc.toml
+
+    cd "$main_dir" || exit
```
arboleya, line 31: Here, I think we could use an in-place search/replace using `sed`.


```shell
    # pseudo-code again..

    author="authors = [\"Fuel Labs <contact@fuel.sh>\")]"
    sed -i.bkp -E "s/^authors\s*=.+/${author}/g" ${i}
    rm ${i}/.bkp
```

nedsalk: `sed` is now used.

scripts/forc-check.sh
nedsalk: I stayed with the original solution because of the `authors` checking happening only on `[project]` files, which is IMO easier to understand off the bat with a split `if` statement for it. I did clean it up a bit, though.

arboleya (approved):

danielbate (approved):

Issue #1465, opened 2023-11-27
Consolidate Tx Preparation for all program types


danielbate:
Given improvements to DX for predicates and scripts:
- #1458
- #1457 

We will then have a similar DX flow regarding transaction preparation across all program types. However we should then look to DRY both our source code and documentation where possible, to provide a consistent approach to transaction preparation regardless of program type, rather than being specific to it.

arboleya:
Closing in favor of:
 - https://github.com/FuelLabs/fuels-ts/issues/1457

Issue #1451, opened 2023-11-25
Implement `getTransactionId` for `transfers` on the `Predicate` class


arboleya:
Given this example:

```ts
// Prepare the transaction.
const tx = await state.predicate
    .transfer(recipient, value, types.BaseAssetId, {
        gasPrice: 1,
        gasLimit: 3_500_000,
});

// Wait for result.
const result = await tx.waitForResult();
```

One might need to get the transaction ID before sending the transaction:

```ts
// Getting chain Id
const chainId = provider.getChainId();

// Get tx ID
const txId = hashTransaction(txRequest, chainId); // <— like this

// Sending the tx to the chain
const tx = await myWallet.sendTransaction(txRequest);

// Waiting for the tx to be processed
const res = await tx.waitForResult();
```

The problem is that the `transfer` method creates a `ScriptTransactionRequest`, `funds` it, and `sends` in the [same call](https://github.com/FuelLabs/fuels-ts/blob/ddafc22dcee410806ac0ac75204adc79c535d04c/packages/wallet/src/account.ts#L255-L257).

To intercept the request and compute the transaction ID before it's sent, we must fragment this chain-of-action and expose the request instance in isolation via a method utility or something similar, enabling one to modify it before `submitting` the transaction and even `getTransactionId()` before calling `submit()`.

From this:

```ts
predicate.transfer(...);
```

To something like this:

```ts
// first iteration
const txID = predicate.getTransactionId();
predicate.transfer(...);
```

And then this:

```ts
// second iteration
const transfer = predicate.transfer(...);
const txID = transfer.getTransactionId()

transfer.submit(...);
```

> [!WARNING]
>
> Remember to roll this update in two phases:
> 
>  1. Add method `getTransferTransactionId()` and keep `transfer()` as is, but with a deprecation notice
>     - No breaking changes here; [annotate](https://stackoverflow.com/a/62991642) _deprecations_.
> 
>  1. Implement `.transfer().getTransactionId().submit()`
>     - Possible breaking changes; may apply _deprecations_.

arboleya:
Either way, instead of getting the TX ID like this:

```ts
const transaction = contract.functions.xyz();
const request = await transaction.getTransactionRequest();
const txID = hashTransaction(request, 0);
```

We could have something more straightforward:

```ts
const transaction = contract.functions.xyz();
const txID = transaction.getTransactionId();
```

This would eliminate the need to use `hashTransaction` manually.

[...]

PR #1466, opened 2023-11-27
feat: document tx preparation for a `Script` and implement txId helper

### LINKED ISSUE #1458: Improve DX for `Script` class
It should follow the same DX we have on `Contract`, supporting things like:

 - `txParams`
 - `callParams`
 - `getTransactionRequest`
 - Etc.

This increases flexibility and ease of use, giving users complete control of everything.

Related:
 - https://github.com/FuelLabs/fuels-ts/issues/1457

danielbate:
Closes #1458

### DIFF
--- a/apps/docs/src/guide/scripts/preparing-a-script.md
+++ b/apps/docs/src/guide/scripts/preparing-a-script.md
@@ -1,0 +1,5 @@
+# Preparing a Script Transaction
+
+Akin to Contracts, we can configure the [call parameters](../contracts/call-parameters.md) and [transaction parameters](../contracts/transaction-parameters.md) for Scripts. As well as retrieving the entire transaction request or transaction ID; prior to submission.
+
+<<< @/../../docs-snippets/src/guide/scripts/script-with-configurable.test.ts#preparing-scripts{ts:line-numbers}
--- a/packages/program/src/functions/base-invocation-scope.ts
+++ b/packages/program/src/functions/base-invocation-scope.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import type { InputValue } from '@fuel-ts/abi-coder';
 import { ErrorCode, FuelError } from '@fuel-ts/errors';
+import { hashTransaction } from '@fuel-ts/hasher';
 import type { AbstractContract, AbstractProgram } from '@fuel-ts/interfaces';
 import type { BN } from '@fuel-ts/math';
 import { bn, toNumber } from '@fuel-ts/math';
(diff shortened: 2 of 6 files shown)


apps/docs/src/guide/scripts/preparing-a-script.md
```diff
@@ -0,0 +1,5 @@
+# Preparing a Script Transaction
+
+Akin to Contracts, we can configure the [call parameters](../contracts/call-parameters.md) and [transaction parameters](../contracts/transaction-parameters.md) for Scripts. As well as retrieving the entire transaction request or transaction ID; prior to submission.
```
nedsalk, line 3:
```suggestion
Akin to Contracts, we can configure the [call parameters](../contracts/call-parameters.md) and [transaction parameters](../contracts/transaction-parameters.md) for Scripts, as well as retrieve the entire transaction request or transaction ID prior to submission.
```

nedsalk (approved):

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -372,4 +373,17 @@ export class BaseInvocationScope<TReturn = any> {
...
     return provider;
   }
+
+  /**
+   * Obtains the ID of a transaction.
+   * 
+   * @param chainId - the chainId to use to hash the transaction with
+   * @returns the ID of the transaction.
+   */
+  async getTransactionId(chainId?: number): Promise<string> {
```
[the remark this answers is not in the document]
danielbate: I'd opted to allow passing it in, as in most of my testcases I'd already retrieved the `chainId`, which is unlikely to change. So it seemed like we could save a few function calls if it had already been retrieved. If it is taking away from the real purpose of the function, I can remove it?

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -372,4 +373,17 @@ export class BaseInvocationScope<TReturn = any> {
...
+   *
+   * @param chainId - the chainId to use to hash the transaction with
+   * @returns the ID of the transaction.
+   */
+  async getTransactionId(chainId?: number): Promise<string> {
+    const chainIdToHash = chainId ?? (await this.getProvider().getChainId());
+
+    const transactionRequest = await this.getTransactionRequest();
+    return hashTransaction(transactionRequest, chainIdToHash);
+  }
```
Torres-ssf, line 388: Maybe it would be nice to have this implemented on `BaseTransactionRequest`, this way it would also be possible to get the `TxID` when using custom transactions created with `new ScriptTransactionRequest` or `new CreateTransactionRequest`. 

Then in `BaseInvocationScope` we could simply call `transactionRequest.getTransactionId(chainId)`.

Just an idea, if you feel that does not make sense, we can skip that.

packages/program/src/functions/base-invocation-scope.ts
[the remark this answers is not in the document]
Torres-ssf: Ok, that makes sense

Torres-ssf (approved):
:rocket:

packages/program/src/functions/base-invocation-scope.ts
danielbate: This is a good idea, going to reference #1465 as I would like to consolidation tx preparation there, and @Dhaiwat10 is also working #1451 which I believe brings a txId function to `ScriptTransactionRequest`.

Issue #1385, opened 2023-10-27
Contracts calls are not working when returning String in a Struct or Tuple


pedronauck:
### What version of fuels-ts are you using?

0.65.0

### Steps to Reproduce

Hey guys, I was trying to run a small project yesterday and got into some issues related to decoding contract calls. What's happening is that it is not possible to return a Struct or a tuple that contains a String or Bytes inside. It seems it worked before 0.58.0, but things started to break since [this PR](https://github.com/FuelLabs/fuels-ts/pull/1277).

### Expected Behavior

This is the example of code in Sway 

![CleanShot 2023-10-27 at 15 24 29](https://github.com/FuelLabs/fuels-ts/assets/2029172/c1c6dfa9-9bde-4fac-b010-38cefdd9549c)
![CleanShot 2023-10-27 at 15 24 39](https://github.com/FuelLabs/fuels-ts/assets/2029172/b226c12a-ee06-4fad-9786-671036e5d089)

This is the example of usage on the SDK:
![CleanShot 2023-10-27 at 15 56 31](https://github.com/FuelLabs/fuels-ts/assets/2029172/8519cf0d-3118-410a-b64a-3ae6000421a0)

### Actual Behavior

And that's the error I'm facing right now:

![CleanShot 2023-10-27 at 15 24 04@2x](https://github.com/FuelLabs/fuels-ts/assets/2029172/a8bc18ff-e6ad-4d4d-aec4-5121838f9d95)

Seems that the problem is related to this part of the code:
https://github.com/FuelLabs/fuels-ts/blob/master/packages/abi-coder/src/coders/stdString.ts#L52-L58

I don't know, but I guess this `toNumber()` error when tryting to convert from `bn` will break other decoders as well.

Torres-ssf:
This is currently blocked:
ref: https://github.com/FuelLabs/fuels-ts/pull/1468#issuecomment-1831818401

Torres-ssf:
@danielbate I believe this problem was solved already, right?

danielbate:
Encoding wise we support all these types. This is blocked because on Sway side you can't use [heap types](https://github.com/FuelLabs/sway/issues/5110) in storage maps. Once unblocked Sway side, I expect this to just _work_.

maschad:
Just to add some more context on why this is blocked as per @IGI-111 

Using slices in encoding still being worked on as apart of https://github.com/FuelLabs/sway/issues/5110 as @danielbate mentioned

[...]

PR #1468, opened 2023-11-29
feat: add typegen support for string slice

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/1469


arboleya:
@Torres-ssf Do you think this could be related to https://github.com/FuelLabs/fuels-ts/issues/1385? (cc @nedsalk)

Torres-ssf:
> @Torres-ssf Do you think this could be related to #1385? (cc @nedsalk)

@arboleya I am not completely sure but it seems the problem @pedronauck is having is related to the type `Std String`.

We currently already have some support for this type. But it seems its use inside Structs or Vectors are not supported by Sway yet. 

@hal3e told me that at the Rust SDK this type is only supported inside Enums.

I reckon @pedronauck was not having this issue in older versions of the SDK because it was previously being treated as `Bytes`

hal3e:
FYI we have this issue on the rust sdk side: https://github.com/FuelLabs/fuels-rs/issues/1158 It is currently blocked.

arboleya (approved):

danielbate (approved):

Issue #1078, opened 2023-06-28
Retry mechanism


digorithm:
We should support an optional, configurable parameter to enable retries when sending a transaction to the node. This mechanism's design and API are part of this work; nothing in mind yet!

arboleya:
Great idea; transient failures should indeed be accounted for. We could have a simple option like this (below) and have an internal retry pattern strategy with exponential back-off.

```ts
// if nothing is specified, we could have a default maxRetries=5
contract.functions.xyz(params).call({ maxRetries: 5 });
```

digorithm:
Nope, not yet. This API looks good to me!

nedsalk:
What about a global fuels.config.ts file in addition to @arboleya's proposal?

```js
{
 maxRetries: 5,
 txParams: {
  gasPrice: 2
 },
 predicates: {
  maxRetries: 2,
  txParams: {
   gasPrice: 1
  }
 }
}
```

digorithm:
That's a lot of complexity for very little gain. The JS ecosystem already suffers from *way* too many config files, I'd rather not contribute to that, haha. 

And by very little gain I mean: gas price and many other transaction related configurations are mostly configured per call, usually being dynamically changed very quickly; often gas price being queried from external sources before a call happens. It isn't something that's set one time and the user rarely touches it. Retries does fit in this behaviour, though. But adding a whole new config file for just this value isn't justifiable.

nedsalk:
Haha, yeah, makes sense. However, having to specify retries on a per-call basis might be cumbersome. A per-program config object passed via the constructor would alleviate some of that pain, so I think it's worth considering.
```ts
const myContract = new Contract(theAbi, ...whateverElse...,
 {
 maxRetries: 5, 
 onCallFail: async () => await notifyTheWhiteHouse()
 })
```
The `onCallFail` is just an example of how easily extendable this config object would be.

PR #1474, opened 2023-12-02
feat: retry mechanism

nedsalk:
Add ability to define retry mechanisms that apply to every call the `Provider` makes towards the node.

Note: this branch merges into #1374 currently as that is its base branch, because I needed that branch's `fetch` functionality. It's reviewable as-is, so if anybody has got any comments - they're welcomed. I'll put it as draft until that base branch is merged into `master`.


Torres-ssf:
@nedsalk Is this related to the open issue https://github.com/FuelLabs/fuels-ts/issues/1078?

Should we reference it on this PR?

packages/providers/test/retry.test.ts
```diff
@@ -0,0 +1,138 @@
+import { safeExec } from '@fuel-ts/errors/test-utils';
+
+import Provider from '../src/provider';
+
+// TODO: Figure out a way to import this constant from `@fuel-ts/wallet/configs`
```
danielbate, line 5: Is this because of cyclic deps?

packages/providers/src/call-retrier.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  switch (options.backoff) {
+    case 'linear':
+      return options.baseDuration * attempt;
+    case 'exponential':
+      return options.baseDuration * (2 ^ (attempt - 1));
+    case 'fixed':
+      return options.baseDuration;
+    default:
+      throw new Error();
```
danielbate, line 33: A default strategy might be nice here to reduce the implementation on the consumer.

packages/providers/test/retry.test.ts
nedsalk: Yes.

packages/providers/src/call-retrier.ts
arboleya: That's cool but too complicated—steps for simplifying:
 1. Internalize the `baseDuration`
 1. Use `exponential` as the default
 1. Remove everything else

Before:

```ts
const retryOptions = { maxAttempts: 5, baseDuration: 150, backoff: 'fixed' };
const provider = await Provider.create(FUEL_NETWORK_URL, { retryOptions });
```

After:

```ts
const provider = await Provider.create(FUEL_NETWORK_URL, { maxRetries: 5 });
```

packages/providers/src/provider.ts
```diff
@@ -279,17 +282,23 @@ export default class Provider {
...
+        : (url: string, request: FetchRequestOptions) =>
+            fetch(url, {
+              ...request,
+              signal:
+                options.timeout !== undefined ? AbortSignal.timeout(options.timeout) : undefined,
+            });
+
+    type FetchParams = Parameters<NonNullable<ProviderOptions['fetch']>>;
+
+    return (...args: FetchParams) => retrier(() => fetchFn(...args), options.retryOptions);
```
arboleya, line 301: These points make this code chunk unnecessarily unreadable.
 1. Lots of lambdas
 1. Ternary conditions
 1. Anonymous returns

packages/providers/test/retry.test.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  test('throws if last attempt fails', async () => {
+    const retryOptions = { maxAttempts, baseDuration: duration, backoff: 'fixed' as const };
+
+    const provider = await Provider.create(FUEL_NETWORK_URL, { retryOptions });
+
+    const fetchSpy = jest
+      .spyOn(global, 'fetch')
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore TS is throwing error when test is run, but not in IDE
```
arboleya, line 120: You must tell `providers/tsconfig.json` to consider your `test` directory.

```json
{
  "extends": "../../tsconfig.base.json",
  "compilerOptions": {
    "outDir": "./dist"
  },
  "include": ["src", "test"]
}

```

arboleya (changes requested):
Please, stop using `ts-ignore` and `eslint-ignore`—you use them way _**too much**_.

Instead, try working around your code strategy and make it fit.

Save these two "jokers" for really rare and critical cases.

> I will remind you of this on every PR from now on.

packages/providers/test/retry.test.ts
arboleya: It's always helpful to have the reason within the comment.

packages/providers/test/retry.test.ts
nedsalk: This was a copy-paste linting ignore that I said to myself "I'll fix it later"... but the later never came. :D

packages/providers/src/call-retrier.ts
nedsalk: > 1. Internalize the baseDuration

I didn't quite get this part. Did you mean to not allow the user specify a `baseDuration`?

> 2. Use exponential as the default

But still allow the other two?

> 3. Remove everything else

I don't see anything else besides `baseDuration` and `backoff` and `maxAttempts` (to be renamed into `maxRetries`). What did you mean by this?

packages/providers/src/provider.ts
nedsalk: True that. I refactored it a bit. How does it look now?

packages/providers/src/call-retrier.ts
nedsalk: I did the following:
- Renamed `maxAttempts` into `maxRetries`,
- Made `baseDuration` optional with a default of 150ms,
- Made `backoff` optional with the default being `exponential`

arboleya: I believe you got into the flow and overengineered it a bit.

1. Yes.
2. No, only `maxRetries`
3. Everything else = `baseDuration`, `backoff`

arboleya: As we just spoke, although I think it's a bit over, and that's ok.

We can revisit things as we go if necessary.

packages/providers/src/provider.ts
```diff
@@ -279,17 +282,22 @@ export default class Provider {
...
+        return options.fetch(...args);
+      }
+
+      const url = args[0];
+      const request = args[1];
+      return fetch(url, {
+        ...request,
+        signal: options.timeout !== undefined ? AbortSignal.timeout(options.timeout) : undefined,
+      });
+    }, options.retryOptions);
```
arboleya, line 300: Suggestion on top of https://github.com/FuelLabs/fuels-ts/pull/1474#discussion_r1434329966.

```suggestion
    const { retryOptions, timeout } = options;

    return retrier((...args) => {
      if (options.fetch) {
        return options.fetch(...args);
      }

      const url = args[0];
      const request = args[1];
      const signal = timeout ? AbortSignal.timeout(timeout) : undefined;

      return fetch(url, { ...request, signal });
    }, retryOptions);
```

packages/providers/src/provider.ts
arboleya: Much better, thanks.

I've left an extra NIT/suggestion [here](https://github.com/FuelLabs/fuels-ts/pull/1474#discussion_r1435262592) for your appreciation.

packages/providers/src/provider.ts
```diff
@@ -218,6 +220,7 @@ export type ProviderOptions = {
   ) => Promise<Response>;
   timeout?: number;
   cacheUtxo?: number;
+  retryOptions?: RetryOptions;
```
arboleya, line 223: I just noticed we forgot to add docs around this.

It's a matter of wrapping some tests in code regions and mentioning them in the guide.

nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1474/commits/42aa40ef433b7a0eb524063ceb4befc0bd8b497e

arboleya (approved):

Torres-ssf (approved):

PR #1476, opened 2023-12-04
feat: fix `AssetID` encoding and improve support

### LINKED ISSUE #1475: Fix `AssetID` encoding and improve support for the type
Motivated by this [forum post](https://forum.fuel.network/t/getting-invalid-struct-assetid-error-in-ts-sdk/3812), we are currently incorrectly encoding an `AssetId`. Alike an `EvmAddress` it is a wrapped `b256` address so we can handle it as a struct and delete our `AssetId` specific encoding.

We should also look to improve our documentation on this given it is a sway standard type.

danielbate:
Closes #1475


arboleya (dismissed):

Torres-ssf (dismissed):
LGTM

danielbate:
The [forum post](https://forum.fuel.network/t/getting-invalid-struct-assetid-error-in-ts-sdk/3812) that motivated this post has now been resolved, however they now have a new issue regarding the contract itself, therefore we know that the `AssetId` encoding issue has been resolved and we should deal with the new problem outside of this PR.

PR #1477, opened 2023-12-05
fix: make `startFuelCore` package manager agnostic

Dhaiwat10:
This PR fixes a bug where the initial approach used by `startFuelCore` to locate the `fuels-core` binaries was only working for `pnpm`. This is now package manger agnostic with this PR.

### DIFF
--- a/packages/fuels/src/cli/commands/dev/startFuelCore.ts
+++ b/packages/fuels/src/cli/commands/dev/startFuelCore.ts
@@ -6,11 +6,13 @@ import { getPortPromise } from 'portfinder';
 import treeKill from 'tree-kill';
 
 import type { FuelsConfig } from '../../types';
-import { findPackageRoot } from '../../utils/findPackageRoot';
 import { getBinarySource } from '../../utils/getBinarySource';
 import { error, log, loggingConfig } from '../../utils/logger';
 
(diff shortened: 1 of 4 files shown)


nedsalk (approved):
LGTM.

packages/fuels/src/cli/commands/dev/startFuelCore.ts
```diff
@@ -73,8 +75,8 @@ export const startFuelCore = async (config: FuelsConfig): Promise<FuelCoreNode>
   ].flat();
 
   return new Promise((resolve, reject) => {
-    const pkgRootDir = findPackageRoot();
-    const builtInFuelsCorePath = join(pkgRootDir, 'node_modules', '.bin', 'fuels-core');
+    // This line finds the path to the built-in fuels-core binary
```
arboleya, line 78: Should we remove the `findPackageRoot` utility?

Or is it being used elsewhere?

arboleya (approved):

Torres-ssf (approved):

packages/fuels/src/cli/commands/dev/startFuelCore.ts
Dhaiwat10: It is being used in other files

PR #1485, opened 2023-12-07
feat: add method to hash TX on `TransactionRequest` classes

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/1489
- Close: https://github.com/FuelLabs/fuels-ts/issues/1451


arboleya (dismissed):

packages/providers/src/transaction-request/create-transaction-request.ts
```diff
@@ -95,6 +96,10 @@ export class CreateTransactionRequest extends BaseTransactionRequest {
     );
   }
 
+  getTransactionId(chainId: number): string {
```
danielbate, line 99: Doc block would be beneficial here as we gen API docs for this class.

apps/docs/src/guide/predicates/send-and-spend-funds-from-predicates.md
```diff
@@ -38,6 +38,14 @@ Note the method transfer has two parameters: the recipient's address and the int
 
 Once the predicate resolves with a return value `true` based on its predefined condition, our predicate successfully spends its funds by means of a transfer to a desired wallet.
 
+---
+
+In a similar approach, you can use the `createTransfer` method, which returns a `ScriptTransactionRequest`. Then, we can submit this transaction request by calling the `sendTransaction` method.
+
+This can be useful if you need the transaction ID before actually submitting it to the node.
```
danielbate, line 45: Wondering whether we need some docs here or _somewhere_ around how tx ids are formed, and the risk of it changing if the request is changed?

danielbate (dismissed):
Left a couple of comments but will let you decide on priority

arboleya (commented):
Should we consider merging this after `beta-5`?

Just wondering if some change could cascade into issues for the FE.

Torres-ssf:
> Should we consider merging this after `beta-5`?
> 
> Just wondering if some change could cascade into issues for the FE.

We can definitely postpone this one. 

But the changes made here should not cause any breaking changes.

arboleya (approved):

danielbate (approved):

PR #1493, opened 2023-12-08
docs: fix invalid `context` imports in `doc-examples`

### LINKED ISSUE #1355: Dysfunctional code on Docs
In [this page](https://fuellabs.github.io/fuels-ts/guide/providers/querying-the-chain.html#get-all-coins-from-an-address), the third line of the first code block uses a variable that has never been imported:

```ts
```

This is the source:

https://github.com/FuelLabs/fuels-ts/blob/b3b59e273334cba9d01b849ab7599ac11becaf98/packages/fuel-gauge/src/doc-examples.test.ts#L267-L275

The fix is simple:

```ts
// #context import { Provider, FUEL_NETWORK_URL } from 'fuels';
```

But I wonder:
 - Don't these `#context` bits defeat the purpose of having unit-tested docs?
 - Shouldn't these `doc-examples.test.ts` be placed in the `doc-snippets`?

danielbate:
Closes #1355


arboleya (approved):
Thanks for fixing this.

Just copying my two questions from:
 - #1355 

 1. Don't these `#context` bits defeat the purpose of having unit-tested docs?
 1. Shouldn't these `doc-examples.test.ts` be placed in the doc-snippets?

In short, I think all code-snippets should be placed inside the `docs-snippets` package to avoid confusion.

Also, it would be good to have a way to bulletproof the use of `#context` to avoid breaking the snippets — is that possible? (cc @Torres-ssf)

Torres-ssf (approved):

Torres-ssf:
> Thanks for fixing this.
> 
> Just copying my two questions from:
> 
> * [Dysfunctional code on Provider's docs #1355](https://github.com/FuelLabs/fuels-ts/issues/1355)
> 
> 1. Don't these `#context` bits defeat the purpose of having unit-tested docs?
> 2. Shouldn't these `doc-examples.test.ts` be placed in the doc-snippets?
> 
> In short, I think all code-snippets should be placed inside the `docs-snippets` package to avoid confusion.
> 
> Also, it would be good to have a way to bulletproof the use of `#context` to avoid breaking the snippets — is that possible? (cc @Torres-ssf)

@arboleya Since these `#context` are comments inside our code, it is hard to bulletproof them. Unless we find another approach to using them

danielbate:
1. Agreed, we probably need to change approach completely to how these are implemented. Not 100% sure what the solution looks right now, I did some digging in the past but did not find something that was both unit testable and documentable. I'll investigate again _at some point_ but lower priority than some of my other work. @FuelLabs/devrel may have some ideas?
2. Agreed, have raised #1505

Issue #1504, opened 2023-12-11
Update chain config for Fuel's CLI


arboleya:
While we don't [unify](https://github.com/FuelLabs/fuels-ts/issues/1373) all chain configs, we must keep them all in sync.

This PR updated the configs:
 - https://github.com/FuelLabs/fuels-ts/pull/1437

However, we must update the one inside the Fuel's CLI to reflect the latest changes.

PR #1513, opened 2023-12-13
fix: replace the outdated `--manual_blocks_enabled` flag with `--debug` (fuel-core)

Dhaiwat10:
Along with the chain config updates in #1504, we also have to update this one flag that has been renamed in the latest client version.


arboleya (commented):
Is this required for `beta-5`?

If yes, it should be against `master` _(only until `beta-5` release)_.

Dhaiwat10:
@arboleya I believe so, even though it only affects local development - it is a change caused by the latest beta-5-compliant client version. Let me clean up the commit history since changing the base has added some unwanted changes

Dhaiwat10:
Ready now.

arboleya (approved):

Torres-ssf (approved):

PR #1574, opened 2023-12-20
chore: upgrade `tsx`

danielbate:
I recently ran into a [CI error](https://github.com/FuelLabs/fuels-ts/actions/runs/7278702468/job/19833461331#step:6:119) in #1310. It can be resolved by upgrading our version of `tsx` (or downgrade node version). Resolving here as out of scope of that work.


arboleya (approved):

package.json
```diff
@@ -10,7 +10,6 @@
   },
   "packageManager": "pnpm@8.9.0",
   "scripts": {
-    "dan": "tsx ./scripts/release-unpublish.ts",
```
danielbate, line 13: Oops 😬

danielbate: Removing as it shouldn't have been there, my b

Dhaiwat10 (approved):

PR #1544, opened 2023-12-19
chore: update nodejs to v20

nedsalk:

PR #1575, opened 2023-12-20
chore: add node version test matrix

danielbate:
As we will support node `v20` in #1544, we should also provide test support for a range of versions. This PR provides a way for testing multiple node versions, however this PR will only introduce a single version. As additional versions are supported, they can be tested by adding to the `versions` array.


package.json
```diff
@@ -5,7 +5,7 @@
   "author": "Fuel Labs <contact@fuel.sh> (https://fuel.network/)",
   "private": true,
   "engines": {
-    "node": "^18.18.2",
+    "node": ">=18 <21",
```
arboleya, line 8: Shouldn't it be like this?

```suggestion
    "node": ">=18.18.2 <21",
```

danielbate: Absolutely

package.json
```diff
@@ -5,7 +5,7 @@
   "author": "Fuel Labs <contact@fuel.sh> (https://fuel.network/)",
   "private": true,
   "engines": {
-    "node": "^18.18.2",
+    "node": ">=18.18.2 <21",
```
arboleya, line 8: Just rolling back.

```suggestion
    "node": "^18.18.2",
```

.github/workflows/test.yaml
```diff
@@ -19,18 +19,21 @@ jobs:
 
   environments:
     runs-on: buildjet-4vcpu-ubuntu-2204
-    name: ${{ matrix.env }}
+    name: ${{ matrix.env }} - v${{ matrix.version }}
     strategy:
       fail-fast: false
       matrix:
         env: [node]
+        version: [18.18.2, 20]
```
arboleya, line 27: @danielbate I think we should only consider `18.18.2`.

@nedsalk I think you can add the `20` variation as part of your PR since that is the one that adds `v20` support.

```suggestion
        version: [18.18.2]
```

danielbate:
Will fail until #1579 is merged

arboleya (approved):

Torres-ssf (approved):

PR #1581, opened 2023-12-21
chore: fix temp test workflow

danielbate:
Previous fix in #1579 did not fix the workflow completely.
This removes any _hacky_ conditional checks, however will misreport coverage until `rc/salamander` has been merged to `master` and we can remove the temporary step all together.


danielbate:
Closing as will resolve in #1582 due to varied failing CIs, so must be resolved in the same PR

PR #1583, opened 2023-12-22
feat: accepting addresses as `string`

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/748

### DIFF
--- a/apps/docs-snippets/src/guide/cookbook/transferring-assets.test.ts
+++ b/apps/docs-snippets/src/guide/cookbook/transferring-assets.test.ts
@@ -79,7 +79,7 @@ describe(__filename, () => {
 
     // #region transferring-assets-2
     const transactionRequest = await sender.createTransfer(
-      destination.address,
+      destination.address.toB256(),
       amountToTransfer,
       assetId,
       txParams
--- a/apps/docs-snippets/src/guide/cookbook/transferring-assets.test.ts
+++ b/apps/docs-snippets/src/guide/cookbook/transferring-assets.test.ts
@@ -79,7 +79,7 @@ describe(__filename, () => {
 
     // #region transferring-assets-2
     const transactionRequest = await sender.createTransfer(
-      destination.address.toB256(),
+      destination.address,
       amountToTransfer,
       assetId,
       txParams
--- a/packages/address/src/address.ts
+++ b/packages/address/src/address.ts
@@ -199,6 +199,26 @@ export default class Address extends AbstractAddress {
   }
 
   /**
+   * Takes an ambiguous BECH-32 string or address and creates an `Address`
+   *
+   * @returns a new `Address` instance
+   */
+  static fromAddressOrBech32String(address: string | AbstractAddress): AbstractAddress {
+    if (typeof address === 'string') {
+      if (isBech32(address)) {
+        return new Address(address as Bech32Address);
+      }
+
+      throw new FuelError(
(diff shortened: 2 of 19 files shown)


packages/address/src/address.ts
```diff
@@ -198,6 +198,26 @@ export default class Address extends AbstractAddress {
...
+      }
+
+      throw new FuelError(
+        FuelError.CODES.INVALID_BECH32_ADDRESS,
+        `Invalid BECH-32 Address: ${address}.`
+      );
+    } else {
+      return address;
+    }
+  }
```
nedsalk, line 219:
```suggestion
  static fromAddressOrBech32String(address: string | AbstractAddress): AbstractAddress {
    if (typeof address !== "string") {
      return address;
    }

    if (!isBech32(address)) {
      throw new FuelError(
        FuelError.CODES.INVALID_BECH32_ADDRESS,
        `Invalid BECH-32 Address: ${address}.`
      );
    }

    return new Address(address as Bech32Address);
  }
```

The suggestion is to reduce nesting, make use of type narrowing, and fail fast.

Torres-ssf: @nedsalk This is better, but I have removed this method entirely since I have discussed it with @arboleya and we agreed that it is better not to enforce the use of the BECH-32 now.

apps/docs-snippets/src/guide/cookbook/transferring-assets.test.ts
```diff
@@ -79,7 +79,7 @@ describe(__filename, () => {
 
     // #region transferring-assets-2
     const transactionRequest = await sender.createTransfer(
-      destination.address,
+      destination.address.toB256(),
```
danielbate, line 82: Given we want to store enforcing `bech32` in the future, and we are using `fromDynamicInput` anyways, reckon it is better to start using `bech32` for address examples?

packages/providers/src/provider.ts
```diff
@@ -843,16 +843,17 @@ export default class Provider {
   }
 
   async getResourcesForTransaction(
-    owner: AbstractAddress,
+    owner: string | AbstractAddress,
```
danielbate, line 846: Could a `AddressLike` type be better here as a single point of entry?

danielbate (commented):
Couple thoughts so far. Made 2 comments but the sentiment is carried across all places where similar has been implemented in this PR.

arboleya (dismissed):

packages/providers/src/provider.ts
arboleya: It won't matter after this:
 - https://github.com/FuelLabs/fuels-ts/issues/1584

apps/docs-snippets/src/guide/cookbook/transferring-assets.test.ts
arboleya: I wonder if it's worth updating _all_ docs now or have a new issue for that. 🤔

Torres-ssf: @danielbate I've changed a DOC test snippet by mistake, I wanted to add some cases for every one of the Account-affected methods just to validate that this is working.

@arboleya good question. I don't know if we need to add a complete page for it, but maybe place this info or a use-case in someplace. 

Also, this change will already be present on the Docs-generated API.

arboleya (approved):

danielbate (approved):

PR #1585, opened 2023-12-22
chore: Enforce node v20 for our repo

nedsalk:
This is only enforcing it for us when developing in our repo. The packages are still supporting v18 and v20.


nedsalk:
Closed because we have some github actions that are using v18 and aren't compatible with this change.

PR #1589, opened 2023-12-22
chore: set changeset base to `rc/salamander`

danielbate:
At the moment all of our changesets are based against `master`. This is no longer valid for our `rc` workflow as we are still tracking against `master`. We should set the branch manually for now.


arboleya (commented):
We can't merge this because it'd break everything outside this RC.

We'll have to do it manually until we figure out a better way.

A better way could be:
 - Detecting the base branch automatically
 - Informing the base branch via CLI arguments _(not versioned)_
 - Something else that isn't added to GIT

danielbate:
This PR will be super-ceded by a more permanent approach:

> Detecting the base branch automatically
> Informing the base branch via CLI arguments (not versioned)
> Something else that isn't added to GIT

PR #1595, opened 2023-12-29
feat: handle non-successful transaction statuses

nedsalk:
#1374 introduced subscriptions to transaction status changes, but it didn't handle the various states a transaction could change into. This PR introduces that.


nedsalk:
Closing because this is related to #1588 which needs to be fleshed out a bit more. The receipts should be properly decoded and be put into the error message.

PR #1597, opened 2023-12-31
feat: remove graphql-request dependency

### LINKED ISSUE #1593: Remove `semver` dependency
This issue is based on the discussion #1592.

nedsalk:
This PR replaces the `graphql-request` and `graphql-tag` dependencies with a simple `fetch` implementation.

This change reduces the final bundle size by 15.68 kB.

> [!NOTE]
> - Incoming from https://github.com/FuelLabs/fuels-ts/discussions/1592


packages/providers/src/fuel-graphql-request.ts
```diff
@@ -0,0 +1,48 @@
...
+import { print } from 'graphql';
+
+import { fuelGraphQLSubscriber } from './fuel-graphql-subscriber';
+
+function isSubscription(query: DocumentNode) {
+  const opDefinition = query.definitions.find((x) => x.kind === 'OperationDefinition') as {
+    operation: string;
+  };
+  return opDefinition?.operation === 'subscription';
+}
```
arboleya, line 11: No hard-casting.

```suggestion
function isSubscription({ definitions: defs }: DocumentNode) {
  const opDef = defs.find((op) => op.kind === 'OperationDefinition');

  if (opDef && 'operation' in opDef) {
    return opDef?.operation === 'subscription';
  }

  return false;
}
```

packages/providers/src/fuel-graphql-request.ts
```diff
@@ -0,0 +1,48 @@
+import type { DocumentNode } from 'graphql';
```
arboleya, line 1: For easy reference in the future.

```suggestion
/* 

  Replaces `graphql-request` and `graphql-tag` dependencies with a simple
  `fetch` implementation, reducing the final bundle size by 15.68 kB.

  More:
    https://github.com/FuelLabs/fuels-ts/discussions/1592#discussioncomment-8003515

*/

import type { DocumentNode } from 'graphql';
```

packages/providers/test/provider.test.ts
```diff
@@ -56,7 +56,8 @@ const getCustomFetch =
     }
   ) => {
     const graphqlRequest = JSON.parse(options.body);
-    const { operationName } = graphqlRequest;
+    const { query } = graphqlRequest;
+    const operationName = query.split('{')[0].split(' ')[1].split('(')[0]; // e.g. query getMessageProof(...) {...} -> getMessageProof
```
arboleya, line 60: This must be improved.

arboleya (changes requested):
It'd be great to have independent tests guaranteeing the functionality in isolation.

 - `fuel-graphql-request.test.ts`
 - `fuel-graphql-subscriber.test.ts`

Dhaiwat10 (dismissed):

packages/providers/test/provider.test.ts
nedsalk: [acfa244](https://github.com/FuelLabs/fuels-ts/pull/1597/commits/acfa2449d11bea7699da1de189856ba9498ca912)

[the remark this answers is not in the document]
nedsalk: [acfa244](https://github.com/FuelLabs/fuels-ts/pull/1597/commits/acfa2449d11bea7699da1de189856ba9498ca912)

nedsalk:
@arboleya 
> It'd be great to have independent tests guaranteeing the functionality in isolation.

I don't know what to test in isolation here that isn't tested somewhere else. Our whole code base from the `Provider` up depends on this code and if something is wrong almost all tests will turn red. One could arguably make the case that this is the most tested piece of code we have :D.

packages/providers/test/provider.test.ts
arboleya: I think you didn't get it—_we can't outsource this problem_.

This info should be easily accessible by external users, as it was before.

arboleya:
I am very much against this mindset of _"something else is already testing this new code I'm adding"_.

packages/providers/test/provider.test.ts
nedsalk: I'm still not getting it. Are you referring to the `graphqlRequest` variable that doesn't have `operationName` directly on it or to the way I'm getting the `operationName` variable?

arboleya: You shouldn't modify existing tests to fit your [new] implementation if it was supposed to be a drop-in replacement.

You are outsourcing the burden of figuring out how to extract `operationName` to external users:

```ts
const { query } = graphqlRequest;
const operationName = query.split('{')[0].split(' ')[1].split('(')[0]; // e.g. query getMessageProof(...) {...} -> getMessageProof
```

It should be just like it was before:

```ts
const { operationName } = graphqlRequest;
```

nedsalk:
If this was a case of 
> "I've got 1-2 tests that are testing what I'm adding implicitly and it isn't clear from them that they're testing what I'm adding" 

then I'd be against it too, but this is core fetching logic that's tested in hundreds of tests that are using the `Provider` class. Subscriptions, queries and mutations are tested thoroughly throughout our code base. Each of those tests, in the end, is hitting the `Provider.operations` property which is directly using the `fuelGraphqlRequest` function. If the function was either sending a bad request to the node or improperly parsing the received response, tests would fail everywhere and one could easily figure out that the fetching logic is wrong. As I was replacing the `graphql-request` dependency I was having red tests everywhere until I got to this implementation. Remove half a line from the code and a test somewhere will fail.

I honestly don't know what additional test you'd like to see here that you think would give us additional certainty that this is working correctly. Could you please give me some examples?

packages/providers/test/provider.test.ts
nedsalk: Aha, I understand. The `graphql-request` library was adding an `operationName` property to the request body alongside the `query` and `variables` properties. The `query` and `variables` are the important ones, however. After I replaced it, tests that were using this `getCustomFetch` helper started to fail because this `operationName` property was missing from the body. IMO this property is unnecessary and there's no need for us to actually add it to each one of our requests as it doesn't add any production value and we can make this helper happy with the implementation I presented in the commit above. 

After this additional context, if you still feel that the `operationName` field should be there - it's easy to reintroduce it.

nedsalk:
@arboleya I added tests.

packages/providers/src/fuel-graphql-request.ts
```diff
@@ -0,0 +1,154 @@
...
+        ...request.headers,
+        Accept: 'text/event-stream',
+      },
+    });
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    this.stream = response.body!.getReader();
+  }
+
+  private async readStream(): Promise<IteratorResult<unknown, unknown>> {
+    const { value, done } = await this.stream.read();
```
Torres-ssf, line 77: Operations on streams can fail due to unexpected reasons ( network issues ).

I believe we need to handle any potential error here:

```ts
    try {
      const { value, done } = await this.stream.read();

      const parsed = FuelGraphqlSubscriber.parseBytesStream(value);

      if (parsed === undefined && !done) {
        return this.readStream();
      }

      return { value: parsed, done };
    } catch(err) {
      // Handling the error
    }
```

packages/providers/src/fuel-graphql-request.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    return { value, done };
+  }
+
+  /**
+   * Gets called when `break` is called in a `for-await-of` loop.
+   */
+  async return(): Promise<IteratorResult<unknown, undefined>> {
+    await this.stream.cancel();
```
Torres-ssf, line 106: Just like `this.stream.read`, `this.stream.cancel()` and `this.stream.releaseLock()` can fail for unexpected reasons. 

Will be safer to add a `Try/Catch` block here and properly handle any potential error.

packages/providers/src/fuel-graphql-request.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  private async readStream(): Promise<IteratorResult<unknown, unknown>> {
+    const { value, done } = await this.stream.read();
+
+    const parsed = FuelGraphqlSubscriber.parseBytesStream(value);
+
+    if (parsed === undefined && !done) {
+      // we recursively wait for the next message until it's a proper gql response or the stream is done i.e. {value: undefined, done: true}
+      return this.readStream();
+    }
```
Torres-ssf, line 84: Are we sure that this recursion cannot cause a stack overflow? Like in cases where the wait time turns out to be very long?

Maybe a loop-based approach will be safer here.

packages/providers/src/fuel-graphql-request.ts
```diff
@@ -0,0 +1,154 @@
...
+    return { done: true, value: undefined };
+  }
+
+  [Symbol.asyncIterator](): AsyncIterator<unknown, unknown, undefined> {
+    return this;
+  }
+}
+
+async function handleQueryAndMutation(fetchFn: typeof fetch, url: string, request: RequestInit) {
+  const response = await fetchFn(url, request);
```
Torres-ssf, line 117: I fear that `fetchFn` can throw directly instead of returning an `error` within the response, does that make sense?

packages/providers/src/fuel-graphql-request.ts
```diff
@@ -0,0 +1,154 @@
...
+    private options: {
+      url: string;
+      request: RequestInit;
+      fetchFn: typeof fetch;
+    }
+  ) {}
+
+  private async setStream() {
+    const { url, request, fetchFn } = this.options;
+    const response = await fetchFn(`${url}-sub`, {
```
Torres-ssf, line 65: Does this URL modification assume that the URL was given in a very specific and unique format? 

Can there be cases where this URL can be given in a way that will make this fetch result in an error?

packages/providers/src/fuel-graphql-request.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    const text = this.textDecoder.decode(bytes);
+
+    // https://github.com/FuelLabs/fuel-core/blob/e1e631902f762081d2124d9c457ddfe13ac366dc/crates/fuel-core/src/graphql_api/service.rs#L247
+    if (text === 'keep-alive-text') {
+      return undefined;
+    }
+
+    const { data, errors } = JSON.parse(text.split('data:')[1]);
```
Torres-ssf, line 43: Is it guaranteed that the value of `text.split('data:')[1]` will always be parseable? 

Can this value be `undefined` or any other value that will make `JSON.parse` throws?

If so, we must add a `Try/Catch` block here.

Torres-ssf (changes requested):

packages/providers/src/fuel-graphql-request.ts
nedsalk: It can throw and that'll bubble up to the user. We can't defend against that. This code is only handling the case of a successful response from the node that's got errors in it in.

packages/providers/src/fuel-graphql-request.ts
nedsalk: The `-sub` prefix is added by the fuel node [see here](https://github.com/FuelLabs/fuel-core/blob/e1e631902f762081d2124d9c457ddfe13ac366dc/crates/fuel-core/src/graphql_api/service.rs#L191) so it can't be modified. The fuel node serves subscriptions over `/graphql-sub`.

packages/providers/src/fuel-graphql-request.ts
nedsalk: Those errors aren't part of our domain and we'd have to bubble them up to the consumer in the `catch` block anyways, so it's best to leave them as they are and not obfuscate them with custom error handling. It's better for us to to get an error report from an user in this case where the error possibly contains environment-related information than if we got an error message we ourselves created and perhaps obfuscated some crucial environment-related info.

packages/providers/src/fuel-graphql-request.ts
nedsalk: The thought process in https://github.com/FuelLabs/fuels-ts/pull/1597#discussion_r1453626098 applies here as well, but with the added caveat that the call to `this.stream.read()` will only happen if the connection was already established. I just tried locally spinning up a node, subscribing to it and then killing it while the subscription connection is still active and I got a `TypeError: fetch failed` error soon after, which means that node throws errors when the underlying stream source is abruptly closed due to a network failure. As stated in the linked comment, I prefer not obfuscating the errors in any way.

packages/fuel-gauge/src/transaction-response.test.ts
```diff
@@ -87,7 +87,17 @@ describe('TransactionSummary', () => {
...
+       * This is set to so long in order to test keep-alive message handling as well.
+       * Keep-alive messages are sent every 15s.
+       * It is very important to test this because the keep-alive messages are not sent in the same format as the other messages
+       * and it is reasonable to expect subscriptions lasting more than 15 seconds.
+       * We need a proper integration test for this
+       * because if the keep-alive message changed in any way between fuel-core versions and we missed it,
+       * all our subscriptions would break.
+       * We need at least one long test to ensure that the keep-alive messages are handled correctly.
+       * */
+      args: ['--poa-interval-period', '17sec'],
```
nedsalk, line 100: I updated this test and explained it above, but I'm not really sure if this is worthwhile doing because why would they change the keep-alive message?

packages/providers/src/fuel-graphql-request.ts
nedsalk: @Torres-ssf and I had a discussion where we questioned does calling `ReadableStream.cancel()` actually cancel the underlying connection made with `fetch`? The answer can be found in the example shown on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/cancel#examples). In short, the connection is indeed cancelled.

Torres-ssf (commented):
@FuelLabs/sdk-ts After going through this PR solution, I can validate that it is working as intended and accomplishes what it's supposed to do. The implementation is both valid and well-structured.

I do agree with the advantages of having our own solution, particularly in terms of reducing our package size, but I have a few considerations to share.

After reviewing the `graphql-request` library, I've noticed that there are some aspects of their implementation that are not fully addressed in our custom solution. These include:

1 - Comprehensive Error Handling: The `graphql-request` library provides a robust mechanism for handling various types of errors, which can be crucial in a production environment. While I understand that in this proposed solution, any thrown error will be caught elsewhere, there is a clear benefit to having specific errors properly handled and managed within the context of the library itself.

2 - Subscription Management: `graphql-request` effectively manages multiple subscriptions, which is important for resource management.

3 - Community Support and Maintenance: By utilizing a well-established library like `graphql-request`, we benefit from ongoing community support and regular updates. This significantly reduces the maintenance burden on our team and ensures that we stay aligned with industry standards and best practices.

While the custom solution works well, I believe it's important to weigh these factors.
 
We should consider the potential risks and missing features against the benefits of a reduced package size. 

Unforeseen errors and edge cases could introduce instability and require additional development and maintenance efforts, potentially offsetting the initial benefits.

I suggest we carefully consider the long-term implications of this change. It might be more beneficial to continue using `graphql-request`, ensuring that we have a robust, well-supported solution for our GraphQL needs. 

However, I'm open to further discussion and insights you might have on this matter.

nedsalk:
>1 - Comprehensive Error Handling: The `graphql-request` library provides a robust mechanism for handling various types of errors, which can be crucial in a production environment.

I went through their source code and didn't find anything comprehensive around error handling. I only found them providing the ability to [configure an error policy](https://github.com/jasonkuhrt/graphql-request?tab=readme-ov-file#errorpolicy), which is something we don't have a need for because we want to throw all errors. 

> While I understand that in this proposed solution, any thrown error will be caught elsewhere, there is a clear benefit to having specific errors properly handled and managed within the context of the library itself.

The only specific error they have is [`ClientError`](https://github.com/jasonkuhrt/graphql-request/blob/main/src/helpers/types.ts#L51), meaning they just differentiate between two types of errors:
- Those that are returned by the client in the response body's `errors` property (handled by their `ClientError`),
- all other errors that are just re-thrown

They don't provide any comprehensive error handling because that's not their purpose as they're a general-purpose library. It is up to their consumers to do a `try/catch` on their `ClientError`, read the `message` or [response](https://github.com/jasonkuhrt/graphql-request/blob/main/src/helpers/types.ts#L52) field, and then differentiate between specific errors by implementing custom error handling via some string matching. This PR isn't passing the error-handling responsibility away to someone else, it's doing the exact same amount of error handling as does `graphql-request`.

> 2 - Subscription Management: `graphql-request` effectively manages multiple subscriptions, which is important for resource management.

`graphql-request` only supports subscriptions over web sockets, which isn't applicable to us as our fuel node implements subscriptions over SSE.

> 3 - Community Support and Maintenance: By utilizing a well-established library like `graphql-request`, we benefit from ongoing community support and regular updates. This significantly reduces the maintenance burden on our team and ensures that we stay aligned with industry standards and best practices.

This library has been in the SDK's dependencies since the beginning and its only usage was for creating a proper fetch request and parsing the response correctly - something that was replaced by 21 lines of code ([create a request object](https://github.com/FuelLabs/fuels-ts/blob/45a77dc6af341810742659a27132923c4676055a/packages/providers/src/fuel-graphql-request.ts#L140) and [send and parse the response](https://github.com/FuelLabs/fuels-ts/blob/45a77dc6af341810742659a27132923c4676055a/packages/providers/src/fuel-graphql-request.ts#L119)). We don't need the ceremonies of a library to do a simple fetch, however well the library is maintained. The maintenance burden of a `fetch` call is minimal and IMO less than keeping up with a library's changes in this simple context. Any industry standard improvements upon `fetch` will benefit both us and `graphql-request` the same.

> I suggest we carefully consider the long-term implications of this change. It might be more beneficial to continue using `graphql-request`, ensuring that we have a robust, well-supported solution for our GraphQL needs.

IMO two-ish years of usage in the SDK that only amounted to the use case of fetching is enough to say, "We don't need this library if we can replace it with a simple fetch". Fetching has been our only need met by `graphql-request`. IMO `fetch` should have been the default here since the beginning and a library like `graphql-request` should've been considered only if some additional functionality was needed. 

As far as the future goes, in the potential case where we need some additional functionality we can always take inspiration from them or even return to them if we see that it's too complex to maintain. As for the current situation, the size reduction and easier grokking of `fetch` compared to whatever other library IMO is enough to go through with this PR.

Dhaiwat10:
My two cents on this PR:

If everything is working as intended like @Torres-ssf mentioned, I think we should go ahead with this PR and get it merged. 

I did read through the points that Sergio mentioned in his comment, too. I think the Nedim's explanation in his response is fair enough. I don't see a downgrade in terms of quality/reliability or an increase in burden maintenance-wise if we go ahead with Nedim's solution. On top of that, this PR would decrease the bundle size - which is _crucial_.

arboleya:
All points make sense, but I'm mainly concerned with unknown unknowns and the possible introduction of unforeseen problems, especially in preparation for the subsequent releases, as I'm not sure the risk outweighs the benefits.

danielbate:
I agree with the fact that `graphql-request` was potentially overkill at the time it was implemented here. As it was implemented at the inception of the SDK, it was likely it was chosen to obtain functionality with minimal overhead. Now we have the time to question these decisions and I am glad we are doing so. I believe a `fetch` would suffice, if we needed additional functionality that we don't want to maintain, at that point we'd bring in a library. The _using a sledgehammer to crack a nut_ analogy comes to mind here.

My main problem here is that right now stability should be our top priority, alongside delivering main net features. Although package size is a big consideration, I would rather deliver a stable package that we have complete confidence in than one that is leaner in size, but has unknowns. At this point in time, if something even has 1% unknowns it is worth not merging at the moment. Especially, as we already have a working solution. I'd be in favour of closing the PR.

nedsalk:
Closing this PR because most of the team is not in favor of it at the moment.

nedsalk:
I wouldn't delete this PR's branch because it's a finished piece of work that might come handy in the future.

Issue #1542, opened 2023-12-18
Migrate `bn.js` to `bignumber`


arboleya:
Suggestions were made to replace one library with the other.

Ideally, we should have a pros and cons list before deciding.

> [!NOTE]
> - Incoming from https://github.com/FuelLabs/fuels-ts/discussions/1537

nedsalk:
Based on [this discussion](https://github.com/ethers-io/ethers.js/discussions/2710) and ethers v6 using native bigint in their maths "package" ([see here](https://github.com/ethers-io/ethers.js/blob/6ee1a5f8bb38ec31fa84c00aae7f091e1d3d6837/src.ts/utils/maths.ts#L14)), it seems like ethers got rid of big-number libraries altogether and went completely in favor of native bigint.

LuizAsFight:
just to add more context: #280 #468 

it's interesting to take a look at those to understand why `bn` was chosen at first place and what problems it solves

nedsalk:
Related to the `JSON.stringify`, we could fix it by defining our own `BigInt.prototype.toJSON` (based on [this comment](https://github.com/GoogleChromeLabs/jsbi/issues/30#issuecomment-953187833)):
![image](https://github.com/FuelLabs/fuels-ts/assets/17943891/00d4a732-9cb9-4fca-8bcd-bef7e01c5a07)

It would solve the issue across the board. Although messing with the prototype is risky as it can cause collisions ... what's the chance that two libraries would be adding this same method? Also, as the comment I linked to above mentioned, [MDN also recommends it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#use_within_json), although in the image I'm serializing into hex, not a stringified number.

LuizAsFight:
[...]

Issue #1028, opened 2023-05-30
Improve `changelog` template


arboleya:
A few notes:
 - The version is mentioned only once, at the top
 - Changes are grouped by patch|minor|major
 - Empty groups can be hidden (like _major_, in the example below)
 - For each commit, there's a list of the packages affected by it

----

# Release - 0.43.1

## Patch:

  - Support EVM Address type, by [@danielbate](https://github.com/danielbate) (See [#995](https://github.com/FuelLabs/fuels-ts/pull/995))
    - @fuel-ts/address
    - @fuel-ts/interfaces
    - @fuel-ts/abi-coder
  - Another Patch Message.. [@arboleya](https://github.com/arboleya) (See [#NNN](https://github.com/FuelLabs/fuels-ts/pull/NNN))
    - @fuel-ts/abc
    - @fuel-ts/xyz

## Minor:

  - Something else.. [@Dhaiwat10](https://github.com/Dhaiwat10) (See [#NNN](https://github.com/FuelLabs/fuels-ts/pull/NNN))
    - @fuel-ts/xyz

## Breaking Changes:
  - Something else.. [@Dhaiwat10](https://github.com/Dhaiwat10) (See [#NNN](https://github.com/FuelLabs/fuels-ts/pull/NNN))
    - @fuel-ts/xyz

> Here we'd repeat only the breaking ones.
----

_Originally posted by @arboleya in https://github.com/FuelLabs/fuels-ts/issues/1022#issuecomment-1567505857_

Dhaiwat10:
I'll take care of this along with #970

arboleya:
Related:
 - https://github.com/FuelLabs/fuels-ts/issues/1347

arboleya:
We _really_ need to devise a solution to these [noisy changelogs](https://github.com/FuelLabs/fuels-ts/releases/tag/v0.73.0).

arboleya:
Related:
 - https://github.com/FuelLabs/fuels-ts/issues/970

nedsalk:
Here's a bit of a different take but in the same spirit of reducing noise:
- There are no groupings by patch|minor|major because the title of the changelog already contains what type of update it is (RELEASE - 0.77.0 is a minor update),
- All the affected packages of the release are listed once under the title for completion's sake,
- Changes are grouped into breaking and non-breaking and each change is shown only once,
- Changes are sub-grouped into features, fixes, refactors, chores, and docs,
[...]

Issue #970, opened 2023-05-03
Craft a solution for a 'centralized' changelog for the `fuels` package


Dhaiwat10:
Currently, the [`CHANGELOG.md` at the root of our project](https://github.com/FuelLabs/fuels-ts/blob/master/CHANGELOG.md) is a symlink that points to the `CHANGELOG.md` file inside of the `fuels` package. Apart from the fact that GitHub does not handle symlinks properly, and instead shows this:
<img width="650" alt="image" src="https://user-images.githubusercontent.com/39617427/235912171-50889ea3-ec24-4579-a7d5-ff7d29986ced.png">
... the `CHANGELOG.md` file inside the `fuels` package also does not contain _all_ the changes that are incorporated in our codebases. It only includes information about version bumps to other packages.

Ideally, we should have a `CHANGELOG.md` file at the root of the repo where we append any and all the changes we add to any of the child packages' changelogs.

The motivation behind this is to make it easy for SDK users to browse through our version history and see what changes happened when.

arboleya:
I agree, but I also like the idea of a consolidated Changelog.

Perhaps a combination of all could involve:
 1. Improve the [Changelog template](https://github.com/FuelLabs/fuels-ts/issues/416) (patch changes seems useless)
 3. Add a link to the `Releases` page at the very top of it
 3. Have a consolidated `<repoRoot>/CHANGELOG.md` with the all releases combined (probably auto-[generated](https://github.com/FuelLabs/fuels-ts/issues/416#issuecomment-1201618820))

arboleya:
Related:
 - https://github.com/FuelLabs/fuels-ts/issues/1347

arboleya:
This should be automatically resolved by:
 - https://github.com/FuelLabs/fuels-ts/issues/2493

PR #1601, opened 2024-01-01
feat: replace `elliptic` with `@noble/curves`

### LINKED ISSUE #1600: Replace `elliptic` with a tree-shakeable dependency
This issue is based on the discussion #1592.

nedsalk:
This PR replaces the `elliptic` dependency with [`@noble/curves`](https://github.com/paulmillr/noble-curves). Our test suite is covering the `Signer` class extensively directly and indirectly so it served as a great indicator of correct replacement.

This change reduces the final bundle size by  37.14 kB.

Closes #1600.


.changeset/beauty-die-joy.md
```diff
@@ -0,0 +1 @@
+TODO: delete me
```
arboleya, line 1: This file needs to be deleted.

arboleya (changes requested):

packages/signer/src/signer.ts
```diff
@@ -107,12 +90,12 @@ class Signer {
     // remove recoveryParam from s first byte
     s[0] &= 0x7f;
 
-    const publicKey = getCurve()
-      .recoverPubKey(getBytesCopy(data), { r, s }, recoveryParam)
-      .encode('array', false)
-      .slice(1);
+    const sig = new secp256k1.Signature(BigInt(hexlify(r)), BigInt(hexlify(s))).addRecoveryBit(
```
danielbate, line 93: Interesting seeing another package opting for `BigInt` - #1542

danielbate (approved):

.changeset/warm-poems-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/signer": minor
+---
+
+Stopped exporting secp256k1 curve
```
arboleya, line 5: Isn't this off?

```suggestion
Replacing `elliptic` with `@noble/curves`
```

arboleya (approved):
Nice drop-in replacement!

.changeset/warm-poems-swim.md
nedsalk: No, because [here](https://github.com/FuelLabs/fuels-ts/pull/1601/files#diff-76473a2f7b3391ac4ca49a7c64262b5520fa1eccc77fd003589fe3a23348d9bfL18) I stopped exporting the `getCurve` method. Replacing `elliptic` with `@noble/curves` was a purely internal change that I don't think should be mentioned in the changeset because nothing changed which the end user of `@fuel-ts/signer` would notice (besides `getCurve` not being exported anymore which I mentioned).

I stopped exporting `getCurve` because exporting the full secp256k1 curve from any package isn't the job of `@fuel-ts/signer`. We're just the consumers of an implementation of the curve. If somebody else wants this same curve, they themselves can put `elliptic`, `@noble/curves` or something third in their dependencies and use it as they wish.

packages/signer/src/signer.ts
nedsalk: Good catch! Thanks for linking it to the issue.

.changeset/warm-poems-swim.md
arboleya: Can we use multiline text in these changelogs?

```
 - Replacing `elliptic` with `@noble/curves`
 - Stopped exporting `getCurve()` / secp256k1
```

Trivial details, but something we definitely [need](https://github.com/FuelLabs/fuels-ts/issues/1028) to [improve](https://github.com/FuelLabs/fuels-ts/issues/970).

nedsalk: Sure, I can do that. Conceptually, I was guided by the idea that changesets should contain only the descriptions for changes that are observable to the end user.

arboleya (approved):

.changeset/warm-poems-swim.md
arboleya: I hope the formatting is not affected when rendering the final consolidated changelog.

We will see. :)

danielbate (approved):

PR #1602, opened 2024-01-02
chore!: Share single chainConfig and review node-related utilities

### LINKED ISSUE #1373: Share single `chainConfig` and review node-related utilities
The idea is to remove duplicated configs and _implementations_, eliminating code repetition.

 1. Unify all `chainConfig.json` files <sup>_([why?](https://github.com/FuelLabs/fuels-ts/pull/1291/files#r1361916496))_</sup>
    - Consider using [these](https://github.com/FuelLabs/fuels-portal/blob/main/docker/fuel-core/accounts.json) default accounts <sup>_([why?](https://github.com/FuelLabs/fuels-ts/pull/1291/files/2d37a955b89252f413211a8aaa9430c10ffc9ace#r1365491981))_</sup>
 1. Unify "launch node" utilities, remove overlapping features
    - [`fuels/src/cli/commands/dev/startFuelCore.ts`](https://github.com/FuelLabs/fuels-ts/blob/c422007b0399f97f33a782fb837ffaffb8b7070b/packages/fuels/src/cli/commands/dev/startFuelCore.ts#L38)
    - [`wallet/src/test-utils/launchNode.ts`](https://github.com/FuelLabs/fuels-ts/blob/cfb06e88dba97792b7bc5f4e048c9b9373fb3c03/packages/wallet/src/test-utils/launchNode.ts#L46)
    - ~[`providers/src/utils/test-utils/launchTestNode.ts`](https://github.com/FuelLabs/fuels-ts/blob/2a70e7d4972718a61425a52d124dc359e6f8f621/packages/providers/src/utils/test-utils/launchTestNode.ts#L42)~ (file doesn't exist anymore)

Dhaiwat10:
Closes #1373 

- `startFuelCore` now re-uses `launchNode` instead of having its own node-launching logic
- `@fuel-ts/utils` now exports a `defaultChainConfig` and a `defaultConsensusKey` which is used everywhere in the source code.
- The `chainConfig.json` file inside the `.fuel-core` folder at the root also uses the same chain config. The `run-node` script has been modified to copy over the contents of the chain config file from the utils package.

# Breaking Changes

- Multiple fuel-core config-related options have been removed from `LaunchNodeOptions`:
  - `chainConfigPath`
  - `consensusKey`
  - `useInMemoryDb`
  - `poaInstant`

 - The only way to pass in these config values now is through the `args` property, i.e.:
  ```ts
    const { cleanup, ip, port } = await launchNode({
      args: ['--poa-interval-period', '750ms', '--poa-instant', 'false'],
    });
  ```


Dhaiwat10:
@arboleya yep that makes sense if `autoStartFuelCore` is the only function that's used externally. I thought these two functions served two different purposes. Is `autoStartFuelCore` just a wrapper?

arboleya:
@Dhaiwat10 Yes, it's sort of a wrapper.

packages/fuels/src/cli/commands/dev/autoStartFuelCore.test.ts
```diff
@@ -1,39 +1,101 @@
...
+    if (existsSync(chainConfig)) {
+      rmSync(chainConfig);
+    }
+  });
 
-    return { startFuelCore };
+  function mockLaunchNode() {
+    vi.spyOn(testUtilsMod, 'launchNode').mockImplementation((..._: unknown[]) =>
+      Promise.resolve({ cleanup: () => {}, ip: '0.0.0.0', port: '4000' })
+    );
```
arboleya, line 35: It would be better to return a named variable with the spy instance to keep consistency with other mocking strategies across the codebase. Also, if you don't need to modify the implementation but rather return something else, you can use `mockReturnValue`.

```suggestion
    const cleanup = Promise.resolve({ cleanup: () => {}, ip: '0.0.0.0', port: '4000' });
    const launchNode = vi.spyOn(testUtilsMod, 'launchNode').mockReturnValue(cleanup);
    return { launchNode };
```

Then, on the test cases using it:


```ts
const { launchNode } = mockLaunchNode();

// ...

expect(launchNode).toHaveBeenCalledTimes(0);
```

packages/fuels/src/cli/commands/dev/autoStartFuelCore.ts
```diff
@@ -1,15 +1,86 @@
...
+  };
+};
+
+export const killNode = (params: KillNodeParams) => () => {
+  const { core, state, killFn } = params;
+  if (core.pid && !state.isDead) {
+    state.isDead = true;
+    killFn(Number(core.pid));
+  }
+};
```
arboleya, line 35: We no longer need this method since you already do it inside the `launchNode`, no?

https://github.com/FuelLabs/fuels-ts/blob/2154aae9bf8149df3df05e2e9379ffd025f39ec8/packages/wallet/src/test-utils/launchNode.ts#L64

packages/fuels/src/cli/commands/dev/autoStartFuelCore.ts
```diff
@@ -1,15 +1,86 @@
...
+
+export const createTempChainConfig = (coreDir: string) => {
+  const chainConfigPath = join(coreDir, 'chainConfig.json');
+  const chainConfigJson = JSON.stringify(defaultChainConfig, null, 2);
+  mkdirSync(dirname(chainConfigPath), { recursive: true });
+  writeFileSync(chainConfigPath, chainConfigJson);
+  return chainConfigPath;
+};
```
arboleya, line 43:
```suggestion
```

Same thing here ☝️; this functionality seems a duplicate of what's done inside the `launchNode`:

https://github.com/FuelLabs/fuels-ts/blob/2154aae9bf8149df3df05e2e9379ffd025f39ec8/packages/wallet/src/test-utils/launchNode.ts#L167

Dhaiwat10:
@arboleya looks like the failing tests are unrelated and flaky. Please feel free to re-review

packages/fuel-gauge/src/transaction-response.test.ts
```diff
@@ -88,6 +88,7 @@ describe('TransactionSummary', () => {
   it('should ensure waitForResult always waits for the transaction to be processed', async () => {
     const { cleanup, ip, port } = await launchNode({
       args: ['--poa-interval-period', '750ms'],
+      poaInstant: false,
```
arboleya, line 91: Why was this necessary?

It seems unrelated to what's been done.

packages/fuels/src/cli/commands/dev/autoStartFuelCore.ts
```diff
@@ -1,14 +1,65 @@
...
+      ip: bindIp,
+      port: port.toString(),
+      loggingEnabled: loggingConfig.isLoggingEnabled,
+      debugEnabled: loggingConfig.isDebugEnabled,
+      useInMemoryDb: false,
+      basePath: config.basePath,
+      useSystemFuelCore: !config.useBuiltinFuelCore,
+    });
+
+    const fuelCoreNode = {
```
arboleya, line 52:
```suggestion
    fuelCore = {
```

packages/fuels/src/cli/commands/dev/autoStartFuelCore.ts
```diff
@@ -1,14 +1,65 @@
...
+      bindIp,
+      accessIp,
+      port,
+      providerUrl,
+      chainConfigPath,
+      killChildProcess: cleanup,
+    };
+
+    fuelCore = fuelCoreNode;
```
arboleya, line 61:
```suggestion

```

packages/fuels/test/utils/mockAutoStartFuelCore.ts
```diff
@@ -16,7 +15,7 @@ export const mockStartFuelCore = (): {
     port: 4000,
     providerUrl: `http://127.0.0.1:4000/graphql`,
     killChildProcess,
-    chainConfig: '/some/path/chainConfig.json',
+    chainConfigPath: '/some/path/chainConfig.json',
```
arboleya, line 18: We could declare this variable with a `FuelCoreNode` type annotation for increased clarity.

```ts
const fuelCore: FuelCoreNode = {
```

Also, since the original method is `async`, you should return a `Promise.resolve()`:

```ts
const autoStartFuelCore = vi
    .spyOn(autoStartFuelCoreMod, 'autoStartFuelCore')
    .mockResolvedValue(Promise.resolve(fuelCore));
```

packages/wallet/src/test-utils/launchNode.test.ts
```diff
@@ -1,81 +1,192 @@
...
+  const spawn = vi
+    .spyOn(childProcessMod, 'spawn')
+    .mockImplementation(
+      (..._: unknown[]) => innerMocks as unknown as ChildProcessWithoutNullStreams
+    );
+
+  return { spawn, innerMocks };
+}
+
+const defaultLaunchNodeConfig = {
```
arboleya, line 60: This could have a type annotation as well:

> _You'll have to export `LaunchNodeOptions`, which I also think is a good idea._

```ts
const defaultLaunchNodeConfig: Partial<LaunchNodeOptions> = {
```

packages/wallet/src/test-utils/launchNode.test.ts
```diff
@@ -1,81 +1,192 @@
...
+    stderr: {
+      pipe: vi.fn(),
+      on: vi.fn(stderrOn),
+      removeAllListeners: vi.fn(),
+    },
+    stdout: {
+      pipe: vi.fn(),
+      removeAllListeners: vi.fn(),
+    },
+  };
```
arboleya, line 49: Nit I.

```suggestion
  } as unknown as ChildProcessWithoutNullStreams;;
```

packages/wallet/src/test-utils/launchNode.test.ts
```diff
@@ -1,81 +1,192 @@
...
+
+  const spawn = vi
+    .spyOn(childProcessMod, 'spawn')
+    .mockImplementation(
+      (..._: unknown[]) => innerMocks as unknown as ChildProcessWithoutNullStreams
+    );
```
arboleya, line 55: Nit II.

```suggestion
    .mockReturnValue(innerMocks)
```

packages/wallet/src/test-utils/launchNode.test.ts
```diff
@@ -1,81 +1,192 @@
...
+    let child = {
+      pid: undefined,
+      stdout: {
+        removeAllListeners: () => {},
+      },
+      stderr: {
+        removeAllListeners: () => {},
+      },
+    } as ChildProcessWithoutNullStreams;
+    killNode({
```
arboleya, line 153: Nit / line-break.

```suggestion

    killNode({
```

packages/wallet/src/test-utils/launchNode.test.ts
```diff
@@ -1,81 +1,192 @@
...
+      child,
+      configPath: '',
+      killFn,
+      state,
     });
+    expect(killFn).toHaveBeenCalledTimes(0);
```
arboleya, line 159: Nit / line-break.
```suggestion

    expect(killFn).toHaveBeenCalledTimes(0);
```

packages/wallet/src/test-utils/launchNode.test.ts
```diff
@@ -1,81 +1,192 @@
...
+    child = {
+      pid: 1,
+      stdout: {
+        removeAllListeners: () => {},
+      },
+      stderr: {
+        removeAllListeners: () => {},
+      },
+    } as ChildProcessWithoutNullStreams;
+    killNode({
```
arboleya, line 172: Nit / line-break.

```suggestion

    killNode({
```

packages/wallet/src/test-utils/launchNode.test.ts
```diff
@@ -1,81 +1,192 @@
...
+      child,
+      configPath: '',
+      killFn,
+      state,
     });
+    expect(killFn).toHaveBeenCalledTimes(0);
```
arboleya, line 178: Nit / line-break.

```suggestion

    expect(killFn).toHaveBeenCalledTimes(0);
```

packages/wallet/src/test-utils/launchNode.test.ts
```diff
@@ -1,81 +1,192 @@
...
-        expect(wallet).toBeInstanceOf(WalletUnlocked);
-      });
-
-      expect(process.env.GENESIS_SECRET).toBeDefined();
-      expect(process.env.GENESIS_SECRET).not.toEqual(GENESIS_SECRET);
-      expect(process.env.GENESIS_SECRET).toHaveLength(66);
-      stop();
+    // should kill
+    state.isDead = false;
+    killNode({
```
arboleya, line 183: Nit / line-break.

```suggestion

    killNode({
```

packages/wallet/src/test-utils/launchNode.test.ts
```diff
@@ -1,81 +1,192 @@
...
+      child,
+      configPath: '',
+      killFn,
+      state,
     });
+    expect(killFn).toHaveBeenCalledTimes(1);
```
arboleya, line 189: Nit / line-break.

```suggestion

    expect(killFn).toHaveBeenCalledTimes(1);
```

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -2,11 +2,12 @@ import { BaseAssetId } from '@fuel-ts/address/configs';
 import { toHex } from '@fuel-ts/math';
 import { Provider } from '@fuel-ts/providers';
 import { Signer } from '@fuel-ts/signer';
+import { defaultChainConfig, defaultConsensusKey } from '@fuel-ts/utils';
 import type { ChildProcessWithoutNullStreams } from 'child_process';
 import { spawn } from 'child_process';
 import { randomUUID } from 'crypto';
 import { hexlify } from 'ethers';
-import fsSync from 'fs';
+import fsSync, { existsSync } from 'fs';
```
arboleya, line 10: We can DRY this import:

> _You'll need to replace `fsSync.existsSync` and `fsSync.mkdirSync` usages below._

```suggestion
import { existsSync, mkdirSync } from 'fs';
```

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -102,13 +132,13 @@ export const launchNode = async ({
 
     let chainConfigPathToUse = chainConfigPath;
 
-    const tempDirPath = path.join(os.tmpdir(), '.fuels-ts', randomUUID());
+    const tempDirPath = path.join(basePath || os.tmpdir(), '.fuels', basePath ? '' : randomUUID());
```
arboleya, line 135: Nit.

```suggestion
    const prefix = basePath || os.tmpdir();
    const suffix = basePath ? '' : randomUUID();
    const tempDirPath = path.join(prefix, '.fuels', suffix);
```

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -140,20 +170,34 @@ export const launchNode = async ({
...
+      command,
+      [
+        'run',
+        ['--ip', ipToUse],
+        ['--port', portToUse],
+        useInMemoryDb ? ['--db-type', 'in-memory'] : ['--db-path', tempDirPath],
+        ['--min-gas-price', '0'],
+        poaInstant ? ['--poa-instant', 'true'] : [],
+        ['--consensus-key', consensusKey],
+        ['--chain', chainConfigPathToUse as string],
```
arboleya, line 183: To avoid using this `chainConfigPathToUse as string` cast multiple times, you could:

```ts
let chainConfigPathToUse: string;

if (chainConfigPath) {
  chainConfigPathToUse = chainConfigPath;
} else {
  // ...
  chainConfigPathToUse = tempChainConfigFilePath
}
```

packages/wallet/src/test-utils/launchNodeAndGetWallets.test.ts
```diff
@@ -0,0 +1,92 @@
...
+      process.env.GENESIS_SECRET = GENESIS_SECRET;
+    });
+
+    test('launchNodeAndGetWallets - empty config', async () => {
+      const { stop, provider, wallets } = await launchNodeAndGetWallets({
+        launchNodeOptions: {
+          loggingEnabled: false,
+        },
+      });
+      expect(provider).toBeInstanceOf(Provider);
```
arboleya, line 80: Nit.

```suggestion

      expect(provider).toBeInstanceOf(Provider);
```

packages/wallet/src/test-utils/launchNodeAndGetWallets.test.ts
```diff
@@ -0,0 +1,92 @@
...
+      expect(wallets.length).toBe(10);
+      wallets.forEach((wallet) => {
```
arboleya, line 82: Nit.

```suggestion

      wallets.forEach((wallet) => {
```

packages/wallet/src/test-utils/launchNodeAndGetWallets.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        expect(wallet).toBeInstanceOf(WalletUnlocked);
+      });
+
+      expect(process.env.GENESIS_SECRET).toBeDefined();
+      expect(process.env.GENESIS_SECRET).not.toEqual(GENESIS_SECRET);
+      expect(process.env.GENESIS_SECRET).toHaveLength(66);
+      stop();
```
arboleya, line 89:
```suggestion

      stop();
```

arboleya (changes requested):
Left a couple of notes and a bunch of nits—_feel free to select the ones you like._

Finally, I'd just say you could use all fs-* `sync` methods. 

Currently, there is a mix of `sync` and `async`.

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -27,12 +28,18 @@ type LaunchNodeOptions = {
   port?: string;
   args?: string[];
   useSystemFuelCore?: boolean;
+  useInMemoryDb?: boolean;
+  loggingEnabled?: boolean;
+  debugEnabled?: boolean;
+  basePath?: string;
+  poaInstant?: boolean;
```
Torres-ssf, line 35: It is great to have these options as flags so they can be easily set like: `{ poaInstant: true }`. My concern is that many of them can be also set by using the `args` flag. 

My concern is that someone can use the `args` parameters to set options that are already being set with these other flags ( `useInMemoryDb`, `debugEnabled`, ...). 

What would happen if the flag `poaInstant` is set to `true` and within the `args` we have `['--poa-interval-period', '750ms']` ?

Not sure what would be the best approach here.

Just raising this concern to validate if it makes sense.

arboleya: True. If possible, it may be best to have only one way to set these.

Otherwise, we should at least validate conflicting settings.

Dhaiwat10: @Torres-ssf @arboleya great point. I decided against adding any extra validation because `fuel-core` handles this for us. If the user passes in any invalid options, they will get a readable errort that tells them that x flag is invalid and cannot be passed alongside y, z, etc. 

We should also note that this validation logic isn't exactly straightforward and changes frequently as new options are added/renamed. If we were to add extra validation in the TS SDK for this, we would be just duplicating this code from the client and we would always have to keep up with this validation logic. I don't think this is ideal.

Dhaiwat10: @arboleya I do agree that having only one way to set these things could be the way. In that case, I would favour only keeping the `args` property, and removing all the other params to maximize flexibility.

arboleya: It makes sense; having only `args` seems enough. 👍

packages/fuel-gauge/src/transaction-response.test.ts
Dhaiwat10: @arboleya I think this is because `poaInstant` is `true` by default now, and it wasn't so earlier. `poaInstant` has to be `false` for us to use `--poa-interval-period`. Anyways, this should be solved when I remove all the extra params in favour of just `args`.

Dhaiwat10: When I say that `poaInstant` was `false` earlier, I mean the internal 'inferred' value of the flag. I think it just wasn't present in the list of the default args that we are passing in to the spawn command, but it is there now.

packages/fuels/test/utils/mockAutoStartFuelCore.ts
Dhaiwat10: @arboleya I think it already takes into account the fact that the original method is `async` and _resolves_ the Promise with `fuelCore`. I tried adding in your suggestion and it gave me a type error saying the same.

Dhaiwat10:
@arboleya @Torres-ssf thanks for the in-depth reviews, guys. I have made the adjustments and also marked the PR with a breaking change since we removed multiple params from the `launchNode` function.

Dhaiwat10:
Should I update the changeset as well? 🤔

packages/wallet/src/test-utils/launchNode.ts
```diff
@@ -43,6 +65,17 @@ export type KillNodeParams = {
...
+const findBinPath = (binCommandName: string) => {
+  let binPath = npmWhich.sync(binCommandName);
+
+  if (!existsSync(binPath)) {
+    // The user might be using bun, which has a different structure for binaries inside node_modules
+    binPath = path.join('node_modules', '.bin', binCommandName);
+  }
+
+  return binPath;
+};
```
arboleya, line 77: We should deduplicate this `findBinPath` and use the one from `launchNode.ts.`

https://github.com/FuelLabs/fuels-ts/blob/9923e0a2fa02b0b1b8ac8e77bed0be42565f31f9/packages/wallet/src/test-utils/launchNode.ts#L68-L77

arboleya (changes requested):

arboleya (approved):

Torres-ssf (approved):

packages/wallet/src/test-utils/launchNode.ts
arboleya: It seems like we hit this exact configuration conflict [here](https://github.com/FuelLabs/fuels-ts/actions/runs/7571055346/job/20617760404?pr=1602#step:9:372).

Dhaiwat10: @arboleya ah, must have crept in after we merged the latest changes from the rc branch. Should be fixed now

arboleya (approved):

Torres-ssf (approved):

PR #1603, opened 2024-01-03
fix: getResourcesToSpend should default to 1 when amount < 1

nedsalk:
_Note: I set it to merge into `master` so that the fix is available asap._

As you can see e.g. [here](https://github.com/calldelegation/multisig-predicate/blob/61bd920a923d7f8087094e08bbcd0856029eb0f3/frontend/src/components/create.tsx#L106), a fraction was passed into `getResourcesToSpend`. This fraction gets rounded to zero in our big-number library, which led to unexpected behavior (no resources were fetched because the amount was 0). Passing a >1 fraction succeeds because it gets rounded to 1.


packages/wallet/src/account.test.ts
```diff
@@ -103,6 +103,22 @@ describe('Account', () => {
     // #endregion Message-getResourcesToSpend
   });
 
+  it('getResourcesToSpend should work with <1 amount', async () => {
+    // #region Message-getResourcesToSpend
```
danielbate, line 107: Seems we now have 2 regions called `Message-getResourcesToSpend`, potentially copied from the above test?

packages/wallet/src/account.test.ts
```diff
@@ -103,6 +103,22 @@ describe('Account', () => {
...
+      provider
+    );
+    const resourcesToSpend = await account.getResourcesToSpend([
+      {
+        amount: 0.9,
+        assetId: '0x0101010101010101010101010101010101010101010101010101010101010101',
+      },
+    ]);
+    expect(resourcesToSpend[0].amount.gte(1)).toBeTruthy();
+    // #endregion Message-getResourcesToSpend
```
danielbate, line 119: As above.

packages/wallet/src/account.test.ts
nedsalk: Good catch! Indeed, it was a copy-paste.

packages/providers/src/coin-quantity.ts
```diff
@@ -24,9 +24,10 @@ export const coinQuantityfy = (coinQuantityLike: CoinQuantityLike): CoinQuantity
     max = coinQuantityLike.max ?? undefined;
   }
 
+  const bnAmount = bn(amount);
   return {
     assetId: hexlify(assetId),
-    amount: bn(amount),
+    amount: bnAmount.toNumber() > 0 ? bnAmount : bn(1),
```
arboleya, line 30: According to the changelog [message](https://github.com/FuelLabs/fuels-ts/pull/1603/files#diff-7e2912a79a5d3caf3f2ac8e34f4decd617c7805f6a4e01cdaa41749a284cf6e4) and your [test](https://github.com/FuelLabs/fuels-ts/pull/1603/files#diff-f755dbc4776c2abee9a5e41e77e122fbf3eecaf6aa5dec65e211fee5b1858a53R113), this should be like this:

```suggestion
    amount: bnAmount.toNumber() < 1 ? bn(1) : bnAmount,
```

arboleya (changes requested):

packages/providers/src/coin-quantity.ts
arboleya: In the [test](https://github.com/FuelLabs/fuels-ts/pull/1603/files#diff-f755dbc4776c2abee9a5e41e77e122fbf3eecaf6aa5dec65e211fee5b1858a53R113), remember to assert it became `1` instead of a generic [`truthy`](https://github.com/FuelLabs/fuels-ts/pull/1603/files#diff-f755dbc4776c2abee9a5e41e77e122fbf3eecaf6aa5dec65e211fee5b1858a53R117) value.

nedsalk: Related to the test, the amount of a given resource is determined by the chain config of the node so I can't assert a certain value, just that it's greater than or equal to 1.

arboleya: You can validate if `coinQuantityfy()` explicitly turns `amount: 0` into `amount: 1`.

This will test your code addition without margin for false positives.

arboleya (commented):
Why can't we use fractions for this?

I wonder if this could become a problem.

> cc @FuelLabs/sdk-ts

Torres-ssf:
> Why can't we use fractions for this?
> 
> I wonder if this could become a problem.
> 
> > cc @FuelLabs/sdk-ts

@arboleya I believe fractions are not used because of the context of blockchain asset amounts, where floats/fractions are avoided.

The default behavior of the `bn` helper is:
```ts
bn(0.5)
// BN(0)
```

If that could become a problem, well, that is a good question. 

I am presuming that `1` will always be the smallest amount of the base asset (ETH) or any given asset. 

If that is not the case, then most likely something is already wrong and we need to fix it.

@luizstacio @LuizAsFight, do you have any insights on this?

packages/providers/src/coin-quantity.ts
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1603/commits/1c44eeca2e5e41a599d3c34d1d4659d849063591

packages/providers/src/coin-quantityfy.test.ts
```diff
@@ -0,0 +1,6 @@
+import { coinQuantityfy } from './coin-quantity';
+
+test('coinQuantityfy returns 1 when input is <1', () => {
+  expect(coinQuantityfy([0]).amount.toNumber()).toEqual(1);
+  expect(coinQuantityfy([0.9]).amount.toNumber()).toEqual(1);
+});
```
Torres-ssf, line 6: Great addition with this unit test! :+1:

I believe we can make this change to maintain consistency with the majority of our test suites.

```suggestion
describe('coinQuantityfy', () => {
  it('should returns 1 when input is < 1', () => {
    expect(coinQuantityfy([0]).amount.toNumber()).toEqual(1);
    expect(coinQuantityfy([0.9]).amount.toNumber()).toEqual(1);
  });
});
```

Torres-ssf (changes requested):

arboleya (approved):
Looking good!

Please, remind us to go over [this](https://github.com/FuelLabs/fuels-ts/pull/1603#issuecomment-1875363089) during the next sync.

Torres-ssf (dismissed):

arboleya (approved):

arboleya (approved):

Torres-ssf (approved):

luizstacio:
> > Why can't we use fractions for this?
> > I wonder if this could become a problem.
> > > cc @FuelLabs/sdk-ts
> 
> @arboleya I believe fractions are not used because of the context of blockchain asset amounts, where floats/fractions are avoided.
> 
> The default behavior of the `bn` helper is:
> 
> ```ts
> bn(0.5)
> // BN(0)
> ```
> 
> If that could become a problem, well, that is a good question.
> 
> I am presuming that `1` will always be the smallest amount of the base asset (ETH) or any given asset.
> 
> If that is not the case, then most likely something is already wrong and we need to fix it.
> 
> @luizstacio @LuizAsFight, do you have any insights on this?

Yes BN does not work with float and ignore any decimals, this is a expected behavior when working with integers that is the case for BNs.

Another thing looking at the code: `amount: bnAmount.toNumber() < 1 ? bn(1) : bnAmount,` this could lead to a issue as a BN can be bigger them a Integer in JS can support throwing a error it would be safer to evaluate using `bn(amount).isZero()`.

LuizAsFight:
> > Why can't we use fractions for this?
> > I wonder if this could become a problem.
> > > cc @FuelLabs/sdk-ts
> 
> @arboleya I believe fractions are not used because of the context of blockchain asset amounts, where floats/fractions are avoided.
> 
> The default behavior of the `bn` helper is:
> 
> ```ts
> bn(0.5)
> // BN(0)
> ```
> 
> If that could become a problem, well, that is a good question.
> 
> I am presuming that `1` will always be the smallest amount of the base asset (ETH) or any given asset.
> 
> If that is not the case, then most likely something is already wrong and we need to fix it.
> 
> @luizstacio @LuizAsFight, do you have any insights on this?

I think the default behavior is correct, should never round up


```
bn() = 0
bn(0.5) = 0
bn(0.8) = 0
```

if the user wants the float, should go through `bn.parseUnits("0.5")`

PR #1495, opened 2023-12-08
rc: salamander

arboleya:
- [x] #1310
 - [x] #1374
 - [x] #1403
 - [x] #1426
 - [x] #1445
 - [x] #1467
 - [x] #1474
 - [x] #1496
 - [x] #1497
 - [x] #1520
 - [x] #1529
 - [x] #1544
 - [x] #1559
 - [x] #1566
 - [x] #1567
 - [x] #1574
 - [x] #1575
 - [x] #1579
 - [x] #1580
 - [x] #1582
 - [x] #1583
 - [x] #1586
 - [x] #1590
 - [x] #1591
 - [x] #1594
 - [x] #1601
 - [x] #1602
 - [x] #1614
 - [x] #1615
 - [x] #1619
 - [x] #1629
 - [x] #1630
 - [x] #1637
 - [x] #1640
 - [x] #1642
 - [x] #1648
 - [x] #1654
 - [x] #1657
 - [x] #1658
 - [x] #1659
 - [x] #1660
 - [x] #1661
 - [x] #1676
 - [x] #1679

PR #1431, opened 2023-11-13
feat: implement browser testing

danielbate:
- Closes #284
 - Closes #765

PR #1608, opened 2024-01-04
feat: remove `uuid` dependency in favor of `crypto.randomUUID()`

nedsalk:
This is a quick win that follows what the [`uuid` library](https://www.npmjs.com/package/uuid) suggests:

> Note Only interested in creating a version 4 UUID? You might be able to use [crypto.randomUUID()](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID), eliminating the need to install this library.

### DIFF
--- a/packages/wallet/src/keystore-wallet.ts
+++ b/packages/wallet/src/keystore-wallet.ts
@@ -11,7 +11,6 @@ import {
 import { ErrorCode, FuelError } from '@fuel-ts/errors';
 import type { AbstractAddress } from '@fuel-ts/interfaces';
 import { hexlify } from 'ethers';
-import { v4 as uuidv4 } from 'uuid';
 
 export type KeystoreWallet = {
   id: string;
@@ -90,7 +89,7 @@ export async function encryptKeystoreWallet(
 
   // Construct keystore.
   const keystore: KeystoreWallet = {
-    id: uuidv4(),
+    id: crypto.randomUUID(),
     version: 3,
     address: removeHexPrefix(ownerAddress.toHexString()),
     crypto: {
--- a/packages/wallet/package.json
+++ b/packages/wallet/package.json
@@ -62,10 +62,6 @@
     "@fuel-ts/errors": "workspace:*",
     "ethers": "^6.7.1",
     "portfinder": "^1.0.32",
-    "tree-kill": "^1.2.2",
-    "uuid": "^9.0.0"
-  },
-  "devDependencies": {
-    "@types/uuid": "^9.0.1"
+    "tree-kill": "^1.2.2"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1227,13 +1227,6 @@ importers:
       tree-kill:
         specifier: ^1.2.2
         version: 1.2.2
-      uuid:
-        specifier: ^9.0.0
-        version: 9.0.0
-    devDependencies:
-      '@types/uuid':
-        specifier: ^9.0.1
-        version: 9.0.1
 
   packages/wallet-manager:
     dependencies:
@@ -9172,10 +9165,6 @@ packages:
 
   /@types/unist@2.0.6:
     resolution: {integrity: sha512-PBjIUxZHOuj0R15/xuwJYjFi+KZdNFrehocChv4g5hu6aFroHue8m0lBP0POdK2nKzbw0cgV1mws8+V/JAcEkQ==}
-    dev: true
-
-  /@types/uuid@9.0.1:
-    resolution: {integrity: sha512-rFT3ak0/2trgvp4yYZo5iKFEPsET7vKydKF+VRCxlQ9bpheehyAJH89dAkaLEq/j/RZXJIqcgsmPJKUP1Z28HA==}
     dev: true
 
   /@types/web-bluetooth@0.0.16:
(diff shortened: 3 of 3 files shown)


packages/wallet/src/keystore-wallet.ts
```diff
@@ -90,7 +89,7 @@ export async function encryptKeystoreWallet(
 
   // Construct keystore.
   const keystore: KeystoreWallet = {
-    id: uuidv4(),
+    id: crypto.randomUUID(),
```
arboleya: @nedsalk Did you test it on a browser? It might be ok as it is, but we must be sure. Before #1431 goes in, I'm afraid browser tests must be done manually.

@Torres-ssf Would it make sense to export this method from the [`crypto`](https://github.com/FuelLabs/fuels-ts/tree/master/packages/crypto/src) package?

nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1608/commits/2cb31dc4fc4956174a840ccde45fd4bdc08f44d6

I exported it from `crypto` in the commit above. I tried out `crypto.randomUUID()` in my console and it worked. Don't see why it wouldn't work the same way the other functions that depend on the browser's `crypto` API work.

nedsalk: A thing to note is that the CI [failed](https://github.com/FuelLabs/fuels-ts/actions/runs/7410773675/job/20163803909#step:9:449) on node v18, but passed on node v20. Perhaps the code can be simplified when we drop support for v18 in the far future.

.changeset/new-rats-return.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
Torres-ssf, line 2: I believe the changeset should not be empty since we are adding a new exportable helper on the `crypto` package and made changes on the `wallet` package. 

Therefore we want to bump versions on these 2 packages.

Torres-ssf (changes requested):
It seems `crypto.randomUUID()` is only [supportable](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID#browser_compatibility) by Node 19. 

So if we are going to make this change, we need to require Node >= 19 on the main `package.json`.

packages/wallet/src/keystore-wallet.ts
Torres-ssf: Not sure if I am tripping but taking a look at the [Browser compatibility](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID#browser_compatibility) section for `ramdomUUID` I've just noticed that the Crypto API and all its methods/helpers are all supported in different versions among different browser.

I am raising this concern because I am not sure if we need to improve the browser validations [here](https://github.com/FuelLabs/fuels-ts/blob/master/packages/crypto/src/browser/crypto.ts) to ensure that `subtle` is present on Crypto and, from now on, `randomUUID`

nedsalk:
> It seems crypto.randomUUID() is only [supportable](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID#browser_compatibility) by Node 19.

After I integrated `randomUUID` into our `crypto` package (https://github.com/FuelLabs/fuels-ts/pull/1608/commits/2cb31dc4fc4956174a840ccde45fd4bdc08f44d6), our Node v18 test suite [passed](https://github.com/FuelLabs/fuels-ts/actions/runs/7410971821/job/20164425490?pr=1608). That list is perhaps referring to `crypto` being available globally without the need for `import crypto from 'crypto'`, which would be in line with [my comment](https://github.com/FuelLabs/fuels-ts/pull/1608#discussion_r1441859829).

packages/wallet/src/keystore-wallet.ts
nedsalk: Hmmm, the browser support for `randomUUID` started from July 2021 to March 2022, depending on the browser, whereas the other `crypto` utilities have been available for quite some time. I'm not sure if we should go this route anymore.

@arboleya, you've been developing for the interwebs for ages :D, what do you think? Is ~2 years of existing support acceptable? If it's raising alarms for you, let's drop the PR.

arboleya: @Torres-ssf Good point. It's safe to validate everything in the [crypto.ts](https://github.com/FuelLabs/fuels-ts/blob/master/packages/crypto/src/browser/crypto.ts) file—`subtle`, `randomUUID`, etc.

@nedsalk By addressing Sergio's concerns, I believe we'll be good to go.

nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1608/commits/6f6b856a22581e2e3acb6c6a84555fa66c74d6c6

arboleya (commented):
This looks great! Could you please add some tests?

This `crypto` package is unique in providing two different builds for `node` and `browser`, so the tests are a bit different and may be slightly off, potentially depending on things from:
 - #1431

Just want to say that this package caused problems in the past and needs special attention. I know it directly relates to #1431 and indirectly to #1610.

> _You can reach out to @Torres-ssf and @danielbate if you need insights._

----

@Torres-ssf @danielbate @nedsalk Guys, the final aim is to have the `crypto` package completely unit-tested in the 💯<sup>s</sup> when testing in both `node` and `browser` envs.

.changeset/new-rats-return.md
arboleya: These packages are published independently.

We can't consider things as "internal" in this way.

The usual is to bump everything that will get a new published version.

nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1608/commits/732d8f9ff2d86e42aa0de35db519e9eb258e01a2

nedsalk:
@arboleya
> This looks great! Could you please add some tests?

https://github.com/FuelLabs/fuels-ts/pull/1608/commits/7b18afc366b9553025ecf5c9664956994fc900ca

I had to put every test into a separate file because putting them into one file caused them to share the same environment and stubbing `globalThis.crypto` once with `vi.stubGlobal("crypto", whatever)` caused it to override the `globalThis.crypto` for _every_ test, even when calling `vi.unstubAllGlobals()` after each test. This caused tests to fail because different `crypto` stubs are necessary for each test to work properly.

arboleya:
Also, if the gain in terms of bundle size will be marginal, we could also drop this PR altogether.

nedsalk:
@arboleya Yeah, let's just drop it for the time being. We can revisit it after #1431 is merged. Maybe then the tests will be more straightforward to implement.

arboleya:
@nedsalk Alright. Please remember to remove this PR's mention from:
 - https://github.com/FuelLabs/fuels-ts/pull/1495

Issue #1609, opened 2024-01-04
Refactor build for `CDNJS` Integration


Torres-ssf:
Our current build of the fuels package generates files that successfully work with frameworks like React and Next.js. 

However, it encounters compatibility issues when used directly in a browser through the `<script>` tag in plain HTML. This limitation holds back its functionality in simple web browser environments.

We need to update the build process to produce a JavaScript file that is compatible with the basic browser `<script>` tag. This change is essential for enabling the effective use of the `fuels` package through Content Delivery Networks ([CDNJS](https://cdnjs.com/)), since the [current](https://cdnjs.com/libraries/fuels) version is not working.

After finishing with this we need to create a new Issue and PR on [cdnjs](https://github.com/cdnjs/packages) repository to properly adjust the implementation that was made [here](https://github.com/cdnjs/packages/pull/1707). 

> Reference for context:
- Add fuels to CDNJS issue: https://github.com/FuelLabs/fuels-ts/issues/1397
- Closed issue on CDNJS repo: https://github.com/cdnjs/packages/issues/1699
- Closed PR on CDNJS repo: https://github.com/cdnjs/packages/pull/1707

PR #1610, opened 2024-01-04
chore: refactor build for `CDNJS` Integration

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/1609


packages/fuels/vite.config.js
```diff
@@ -0,0 +1,18 @@
...
+    emptyOutDir: false,
+    lib: {
+      entry: './src/index.ts',
+      name: 'browser',
+      formats: ['es'],
+      fileName: 'browser',
+    },
+  },
+  define: {
+    'process.env.NODE_ENV': JSON.stringify('production'),
```
danielbate, line 16:
```suggestion
    'process.env.NODE_ENV': 'production',
```
Unnecessary use of `stringify` here?

.changeset/pretty-goats-matter.md
danielbate, line 1: Does this need a changeset as it's just build related for dev dependencies.

arboleya (dismissed):

packages/fuels/vite.config.js
Torres-ssf: If I remove the `JSON.Stringfy` an error is shown:

Screenshot from 2024-01-05 08-17-38

Screenshot from 2024-01-05 08-19-00

arboleya: Can/should we go for `vite.config.ts`? (typescript)

Does it make any difference?

.changeset/pretty-goats-matter.md
arboleya: I don't think so because the package needs re-publishing.

packages/fuels/vite.config.js
Torres-ssf: With `.ts` we can have type checking and autocompletion

arboleya: Is the `JSON.stringify()` still needed when using `.ts`?

Torres-ssf: Yes, I've just checked it :laughing:

danielbate: 🤯

.changeset/pretty-goats-matter.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
arboleya, line 2: Ops, I think you didn't see [this comment](https://github.com/FuelLabs/fuels-ts/pull/1610#discussion_r1442762417).

See if it makes sense.

.changeset/pretty-goats-matter.md
Torres-ssf: That was my motivation in the first place. 

But to be honest, this changeset is very confusing to me.

arboleya: Yes, that's because of all the abstraction.

One easy way to think is:
 - Does the package need re-publishing?

If yes, the changeset can't be empty.

arboleya (dismissed):

arboleya (approved):

danielbate (approved):

PR #1613, opened 2024-01-05
fix: include a `.gitignore` file in the copied over template from `create-fuels`

### LINKED ISSUE #1612: Include a `.gitignore` file in the `create-fuels` template
As of now, the `create-fuels` CLI does not copy over a `.gitignore` file along with the rest of the template files, even though it is there in the source code for the template.

Dhaiwat10:
npm doesn't include `.gitignore` files in package binaries once published to npm. This is why the source code of our templates - once publishsed to npm - didn't have a .gitignore file with it.

To go around this, I have added an extra file called `gitignore` (without the . at the start) so that npm includes it in the bundle. This file is then renamed to `.gitignore` by the CLI tool when it extracts the template.

I have also kept the `.gitignore` since the template folder is built out during our monorepo's build and we want the resulting build artifacts ignored.

This is not the fanciest solution but it is the one I've seen most commonly used by scaffolding CLI projects, even by `create-next-app`.

Closes #1612


packages/create-fuels/src/cli.ts
```diff
@@ -64,6 +64,7 @@ export const runScaffoldCli = async (
   await mkdir(projectPath);
 
   await cp(join(__dirname, '../templates/nextjs'), projectPath, { recursive: true });
+  await rename(join(projectPath, 'gitignore'), join(projectPath, '.gitignore'));
```
arboleya, line 67: Did you consider keeping the previous filename and copy the file over?

```suggestion
  await cp(join(projectPath, '.gitignore'), join(projectPath, '.gitignore'));
```

Dhaiwat10: We cannot do that since this file will not be present in the bundle

Dhaiwat10: Only the `gitignore` (without the .) will be present

arboleya:
Looking good! It just needs to be against `master`.

PR #1615, opened 2024-01-05
feat: use `submitAndAwait` graphql endpoint

### LINKED ISSUE #1563: Introduce `submitAndAwait` graphql endpoint support
Submitting a transaction and getting its results is currently a two-step process:
1. Submit the transaction,
2. Get its results via the `TransactionResponse` class

This is the default workflow and should remain the default one because a transaction can take an indefinite amount of time to execute on the node and using `submitAndAwait` as the default would block our users from executing anything on their side until the node executes the transaction.

However, we should allow our users to submit _and_ await the results of the transaction execution in one call if they wish to do so. The main benefit of this single call over splitting it into two calls (submit and and subscribe to `statusChange`) is that the transaction might fail or be squeezed out in-between the submitting and subscribing, which would result in the subscription call never getting an update and thus never ending. This would be because the node immediately deletes a failed and squeezed-out transaction and us requesting that deleted transaction would hang to infinity.

A possible implementation of this would be to expand the current `ProviderSendTxParams` type:
```ts
export type ProviderSendTxParams = EstimateTransactionParams & {
  awaitExecution: boolean;
}

export class Provider {
// ...
  async sendTransaction(..., params: ProviderSendTxParams) {
  // ...
    if (params.awaitExecution) {
    const subscription = await this.operations.submitAndAwait({encodedTransaction});
    for await (const statusAndReceipts of subscription) {
      if(statusAndReceipts.type === "SubmittedStatus") continue;
      const transactionId = hashTransaction(encodedTransaction);
      return TransactionResponse(transactionId, statusAndReceipts);
    }
  }
  const { submit: { id: transactionId } } = await this.operations.submit({ encodedTransaction });
  return TransactionResponse(transactionId, this);
  }
}
```

These `ProviderSendTxParams` can then be forwarded via `FunctionInvocationScope.call()`, `Account.transfer()` and the like so the user can set the desired behavior through a flag, e.g.: 
```ts
const response: TransactionResponse = await my_contract
  .functions
  .my_fn()
  .call(..., {awaitExecution: true});
```

nedsalk:
This PR introduces the `submitAndAwait` graphql endpoint. A user can now send a transaction and wait for its execution in one go before receiving a `TransactionResponse` back.

closes #1563


packages/providers/src/provider.ts
```diff
@@ -595,12 +603,27 @@ export default class Provider {
...
 
+    const encodedTransaction = hexlify(transactionRequest.toTransactionBytes());
+
+    if (awaitExecution) {
+      const subscription = this.operations.submitAndAwait({ encodedTransaction });
+      for await (const { submitAndAwait } of subscription) {
+        if (submitAndAwait.type === 'SuccessStatus') {
+          break;
+        }
+      }
```
Torres-ssf, line 614: What happens if the response type is different from `SuccessStatus` ?

packages/fuel-gauge/src/await-execution.test.ts
```diff
@@ -0,0 +1,114 @@
...
+  });
+  test('calling contracts with awaitExecution works', async () => {
+    const contract = await getSetupContract('coverage-contract')();
+    const sendTransactionSpy = vi.spyOn(contract.provider, 'sendTransaction');
+    const gasPrice = contract.provider.getGasConfig().minGasPrice;
+
+    const { value } = await contract.functions
+      .echo_u8(3)
+      .txParams({ gasPrice, gasLimit: 10_000 })
+      .call({ awaitExecution: true });
```
Torres-ssf, line 54: @nedsalk What is the practical difference between using the flag `awaitExecution` in the context of a contract call?

Even when we are not passing this flag as `true` and therefore not using `submitAndAwait`, the code will wait for the TX to be processed the same way.

nedsalk: The practical difference is the number of calls towards the node:
- If you pass in `awaitExecution: false` or leave it `undefined`, the transaction will be submitted in one call and then awaited in another call via the [statusChange subscription in `TransactionResponse`](https://github.com/FuelLabs/fuels-ts/blob/78e5e405ef442b47032943196f67c3c0db125c92/packages/providers/src/transaction-response/transaction-response.ts#L212).
- If you pass in `awaitExecution: true`, the transaction will be submitted _and_ awaited in the **same** call.
- 
You touched upon a point I wanted to raise in our sync: given that the default behavior of contract function/script calls currently is to await the execution of the transaction, we can using this new `submitAndAwait` endpoint by default for them to reduce network calls until we flesh out what the actual default should be and how to signal that to the end user.

packages/providers/src/provider.ts
nedsalk: Good question. By default, the fuel node closes the connection when it returns a non-`SubmittedStatus` after executing/squeezing out the transaction. This, in turn, causes the `for-await-of` loop to close as well.

The code itself currently doesn't handle any non-successful transactions. In theory, we should throw errors to the user if a non-`SuccessStatus` status is returned. In practice, however, this needs to be fleshed out and centralized because currently there are multiple places where statuses/failure receipts are handled ([1](https://github.com/FuelLabs/fuels-ts/blob/78e5e405ef442b47032943196f67c3c0db125c92/packages/providers/src/transaction-response/transaction-response.ts#L240), [2](https://github.com/FuelLabs/fuels-ts/blob/78e5e405ef442b47032943196f67c3c0db125c92/packages/program/src/errors.ts#L34), and possibly more). I'm mentioning failure receipts here alongside different statuses because of #1588. I already tried handling `FailureStatus` and `SqueezedOutStatus` but realized that it deserves a different PR because it's not trivial given the codebase handles them already in other places via the `TransactionResponse`'s transaction summary.

packages/fuel-gauge/src/await-execution.test.ts
Torres-ssf: @nedsalk I agree with you, I think we should use `submitAndAwait` by default when executing `call` on the `BaseInvocationScope`.

nedsalk: @Torres-ssf  Implemented in https://github.com/FuelLabs/fuels-ts/commit/8e68217e693dfe888e9dda5bba3a732776e5e1e6.

packages/providers/src/provider.ts
nedsalk: Okay, so it was weird that the `for-await-of` wasn't ending because, as it says [here](https://github.com/FuelLabs/fuels-ts/blob/d3a0cffb29d371218375df5fb26d2b75982d33f2/packages/providers/fuel-core-schema.graphql#L1003), the connection does close after a non-`SubmittedStatus` is sent. I even remember seeing `done` be set to `true` [here](https://github.com/FuelLabs/fuels-ts/blob/ad5923d40fdcfcd8b8e7f4473449a7b28981d715/packages/providers/src/fuel-graphql-subscriber.ts#L77), where `done` is the indicator that the stream ended, so I assumed everything worked. Anyways, after I moved from piping `ReadableStream` to using `for-await-of` internally in `fuel-graphql-subscriber.ts` (https://github.com/FuelLabs/fuels-ts/pull/1615/commits/bb9e117dd47968b7775f82cf2cb884454f718900), the stream closes properly and as expected. This was a bug that got fixed by that commit.

packages/fuel-gauge/src/await-execution.test.ts
danielbate: Given we are now using the subscription by default now, why do we still need to pass the `awaitExecution` config?

apps/docs-snippets/src/utils.ts
```diff
@@ -46,10 +46,7 @@ export const getTestWallet = async (seedQuantities?: CoinQuantityLike[]) => {
   await genesisWallet.fund(request, requiredQuantities, minFee);
 
   // execute the transaction, transferring resources to the test wallet
-  const response = await genesisWallet.sendTransaction(request);
-
-  // wait for the transaction to be confirmed
-  await response.wait();
+  await genesisWallet.sendTransaction(request, { awaitExecution: true });
```
danielbate, line 49: `awaitExecution` feels like an implementation detail that is not important to the consumer, and only really to the client. Is the plan to have this completely replace the current submit and retry approach?

packages/fuel-gauge/src/await-execution.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * @group node
+ */
+describe('await-execution', () => {
+  test('awaiting execution of a transaction on the provider works', async () => {
+    const { cleanup, ip, port } = await launchNode({
+      args: ['--poa-interval-period', '400ms'],
+    });
+    const nodeProvider = await Provider.create(`http://${ip}:${port}/graphql`);
```
danielbate, line 21: Nice 👍🏻

packages/providers/src/provider.ts
```diff
@@ -595,12 +603,27 @@ export default class Provider {
       );
     }
 
+    const encodedTransaction = hexlify(transactionRequest.toTransactionBytes());
+
+    if (awaitExecution) {
+      const subscription = this.operations.submitAndAwait({ encodedTransaction });
+      for await (const { submitAndAwait } of subscription) {
+        if (submitAndAwait.type !== 'SubmittedStatus') {
```
danielbate, line 611: Could change the signature of `waitForStatusChange` and reuse that here?

danielbate (commented):
Looking good. Just wondering if something like `subscribeToResult` is more correct given the current behaviour is to `await` anyway, however the underlying implementation just retries.

packages/providers/src/provider.ts
nedsalk: I'd rather keep it like it is in this PR and leave the consolidation of logic to the PR that corresponds to [this previous comment](https://github.com/FuelLabs/fuels-ts/pull/1615#discussion_r1445794664).

packages/providers/src/provider.ts
nedsalk: Unfortunately I had to revert the above commit in https://github.com/FuelLabs/fuels-ts/pull/1615/commits/01267c0373007bfba09382f9c3331c4e0568ac10 because `for-await-of` is not supported for `ReadableStream` in most browsers ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#browser_compatibility), see the `@@asyncIterator` row). However, I added a fix to at least close the stream intentionally in https://github.com/FuelLabs/fuels-ts/pull/1615/commits/9b3a05706df334115a23e45ed74333824da12a41 by refactoring from an async generator function to crafting our own async iterator. I assume the `ReadableStream` was closed in the previous implementation as well via garbage collection, but now we're being explicit about it. The `return` method, in which the stream is closed, is called when a `break` is hit inside a `for-await-of` loop.

I created #1628 to further investigate this issue. It isn't a blocker though, because we're breaking the `for-await-of` loops ourselves and aren't relying on connections closing for any behavior.

packages/fuel-gauge/src/await-execution.test.ts
nedsalk: @danielbate I covered it as it was a consumer of `provider.sendTransaction` where `awaitExecution` in the end resolves. Your question is valid and I now think that this PR shouldn't cover the `BaseInvocationScope.call` method. There were some discussions around what should be returned to the user when using `submitAndAwait` vs using the regular `submit` endpoint because one awaits until there are actual results for of the execution vs the other that returns a `TransactionResponse` of a transaction that could be in `SubmittedStatus` as well. In the end, I now believe that this `BaseInvocationScope.call` use case should be holistically covered by another PR.

apps/docs-snippets/src/utils.ts
nedsalk: No, we want to allow the user to be able to await the execution on the node but also to just submit the transaction and call `TransactionResponse.waitForResult` whenever they want. This PR is only introducing `submitAndAwait` and applying it to some places as default (e.g. `BaseInvocationScope.call`) and allowing it to be used in other places (e.g. `Account.transfer`), keeping in mind to not introduce any breaking change.
There are other discussions in circulation ([the comments on this issue](https://github.com/FuelLabs/fuels-ts/pull/1615#discussion_r1447587623), #1588) that will undoubtedly lead to some restructuring and more clarity when to use which feature. Now we're gaining a quick win by utilizing the existence of this endpoint.

packages/fuel-gauge/src/await-execution.test.ts
danielbate: Gotcha, agreed.

danielbate (commented):
Should we be adding some documentation for this?

danielbate:
This PR is also missing a changeset.

packages/providers/src/provider.ts
nedsalk: Phew, the whole issue is fixed in https://github.com/FuelLabs/fuels-ts/pull/1615/commits/7ff5d1637871f6dc5ebf93cc771c998eed668227. The custom stream transformer `FuelSubscriptionStream` had an issue. After replacing it with the `parseBytesStream` everything works.

Torres-ssf: @nedsalk Sorry but I am having a hard time understanding why this problem happened in the first place and how you managed to solve it. Can you help me understand?

nedsalk: When using GraphQL subscriptions, we are using the [Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) to read the streams of data being sent from the fuel node via an SSE connection. The fuel node guarantees that it'll close this SSE connection after the status of a transaction changes from `SubmittedStatus` to any other. On our side, the indicator that the connection closed which would, in turn, close our stream was being lost inside of the `FuelSubscriptionStream` class that was transforming the stream of data being sent from the fuel node. This caused the [loop that was expecting data from the server](https://github.com/FuelLabs/fuels-ts/blob/ad5923d40fdcfcd8b8e7f4473449a7b28981d715/packages/providers/src/fuel-graphql-subscriber.ts#L71) to loop to infinity because `done` (the indicator of the stream ending) was never `true`. This slipped through the cracks because we'd always manually close the whole async generator with a `break` of the `for-await-of` loops that are consuming the generator.

Everything now works after I removed this `FuelSubscriptionStream` transformer and opted for a simple function that wasn't transforming the actual stream read from `response.body` but just parsing its bytes. After the connection is closed by the fuel node, that stream returns `{done: true, value: undefined}` and the loop inside of the async generator will break on the `if(done) break` condition and the generator will finish.

packages/fuel-gauge/src/await-execution.test.ts
nedsalk: After contemplating a bit more on this, I decided to not introduce this feature on the account methods like `transfer` and `withdrawToBaseLayer` as well because it kind of suffers from the same DX dilemmas as `BaseInvocationScope.call`. I'd boil this PR down to a quick win for contract calls until I further investigate this DX rabbit hole and formulate a proper proposal.

packages/fuel-gauge/src/await-execution.test.ts
```diff
@@ -0,0 +1,97 @@
...
+      }
+      // { awaitExecution: true }
+    );
+
+    expect(sendTransactionSpy).toHaveBeenCalledTimes(1);
+    const awaitExecutionArg = sendTransactionSpy.mock.calls[0][1];
+    expect(awaitExecutionArg).toMatchObject({ awaitExecution: true });
+  });
+
+  test.skip('withdrawToBaseLayer works with awaitExecution', async () => {
```
Torres-ssf, line 72: Why are we skipping this test?

packages/fuel-gauge/src/await-execution.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    );
+
+    const response = await nodeProvider.sendTransaction(transfer, { awaitExecution: true });
+
+    expect(response.gqlTransaction?.status?.type).toBe('SuccessStatus');
+
+    cleanup();
+  });
+
+  test.skip('transferring funds with awaitExecution works', async () => {
```
Torres-ssf, line 45: Why are we skipping this test?

packages/fuel-gauge/src/await-execution.test.ts
nedsalk: I skipped it after [this comment](https://github.com/FuelLabs/fuels-ts/pull/1615#discussion_r1448828304) but left it there for reference.

packages/fuel-gauge/src/await-execution.test.ts
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1615#discussion_r1450445767

packages/providers/src/provider.ts
nedsalk: I reverted this bugfix (https://github.com/FuelLabs/fuels-ts/pull/1615/commits/ca6b64331e050052815c85da0efaf24f3f854dd6) and moved it to #1597 because the bug presented itself there as well when I was writing tests for the subscriptions and that PR is more appropriate. I don't want to have two PRs that are solving the same issue open at the same time.

danielbate (approved):

Torres-ssf (approved):

PR #1470, opened 2023-11-29
docs: standardize shell commands in the docs

richardgreg:
Removed `yarn`, ensured all examples have two variations and tabularized examples.

close #1459

PR #1629, opened 2024-01-10
fix: do not generate a coverage diff without coverage artifact

danielbate:
If we have not retrieved a coverage artefact, we should not generate a coverage diff, as this will cause CI to [fail](https://github.com/FuelLabs/fuels-ts/actions/runs/7475211588/job/20343181063).


.github/workflows/test.yaml
```diff
@@ -113,6 +113,7 @@ jobs:
           path: coverage-master
 
       - name: Generate Coverage Diff
+        if: ${{ (steps.findPr.outputs.number) }}
```
arboleya, line 116: AFAIU, this conditional checks whether the action is running from inside a PR and not `master` or any other branch. However, the error you [reported](https://github.com/FuelLabs/fuels-ts/actions/runs/7475211588/job/20343181063) ran inside of a PR, no?

How is this going to fix that?

danielbate: Correct, however `steps.findPr.outputs.number` is `undefined` for external contributors. Therefore we have not been generating a coverage diff for external PRs. Previously this [was not a problem](https://github.com/FuelLabs/fuels-ts/pull/1470), as our previous CI action would succeed, and just not comment on the PR. However now it fails as the new CI requires a base coverage artefact. This will put us back to previous behaviour.

I will also look into why `steps.findPr.outputs.number` is misreporting and come up with a more complete fix, however this will resolve currently failing CI.

arboleya: Oof, that's weird.

Could we replace this action with something like [this](https://stackoverflow.com/questions/59077079/how-to-get-pull-request-number-within-github-actions-workflow)?

danielbate: Isn't it just. The only issue with that approach is that it won't work for `push` events, such as one to `master` or `rc`. However a hybrid using both should cover both all scenarios we need.

arboleya: Gotcha. A few items below the voted answer you'll find this [snippet](https://stackoverflow.com/a/75420270) that works for `push`, but it looks pretty hacky.

I'll leave this with you. 😌

arboleya (approved):

Torres-ssf (approved):

PR #1630, opened 2024-01-10
feat: implement browser compatibility testing

danielbate:
Partially implementing #284, this PR includes infrastructure for browser compatibility testing.

- Adds browser based vitest configuration and dependencies
- Tagged applicable browser tests
- Script for hardlnking dependencies
- Extended scripts for coverage merging and reporting
- Added internal browser based environment test
- Implemented in CI

> Currently tests using `fs` (such as `fuel-gauge`) are not yet supported, this is detailed further in [this comment](https://github.com/FuelLabs/fuels-ts/issues/284#issuecomment-1887561250), but a fix will be included in a follow up PR.


.github/workflows/test.yaml
```diff
@@ -19,12 +19,15 @@ jobs:
 
   environments:
     runs-on: ubuntu-latest
-    name: ${{ matrix.env }} - v${{ matrix.version }}
+    name: ${{ matrix.env }}${{ matrix.env == 'node' && '@' || ''}}${{ matrix.env == 'node' && matrix.version || ''}}
```
danielbate, line 22: Looks ugly, but gives `node@version` or `browser`

arboleya: Ouch. 😅

arboleya, line 22:
```suggestion
    # name will be node@<version> or browser
    name: ${{ matrix.env }}${{ matrix.env == 'node' && '@' || ''}}${{ matrix.env == 'node' && matrix.version || ''}}
```

.github/workflows/test.yaml
```diff
@@ -19,12 +19,15 @@ jobs:
...
     strategy:
       fail-fast: false
       matrix:
-        env: [node]
+        env: [node, browser]
         version: [18.18.2, 20]
+        exclude:
+          - env: browser
+            version: 18.18.2
```
arboleya, line 30: Tricky.

.github/workflows/test.yaml
```diff
@@ -126,8 +129,8 @@ jobs:
           mode: recreate
           create_if_not_exists: true
 
-  live-tests:
-    name: Run Live Tests
+  test-live:
+    name: live
```
arboleya, line 133: Should we keep semantics here?

```suggestion
  test-e2e:
    name: e2e
```

arboleya (commented):
The idea was to run the same `crypto` tests everywhere.

Why did you split duplications of them between browser/node?

> cc @Torres-ssf

.github/workflows/test.yaml
danielbate: Good idea 👍🏻

.github/workflows/test.yaml
danielbate: Could have gone for for a matrix that uses an array of objects if we prefer and conditionally check for the presence of a versions key. However [the docs](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#excluding-matrix-configurations) seems to prefer `include`/`include` keyword usage rather than using conditionals. I am indifferent.

arboleya: How would it look with an array of objects?

danielbate:
@arboleya 

> Why did you split duplications of them between browser/node?

So the issue lies in how we import the code under test into our tests, or potentially more how we want to test our environment specific tests. Previously we had a single feature test file doing 2 environment based imports:

```ts
import { encrypt, decrypt } from '../../src/index'; // node
import { encrypt, decrypt } from '../../src/index.browser'; // browser
``` 

We have a `throw` condition in the node export file to check that `crypto` is accessible on `globalThis`:

https://github.com/FuelLabs/fuels-ts/blob/9b079ab0df770a82ed5d54a4a42e5b3f965594c6/packages/crypto/src/browser/crypto.ts#L8

In a node environment, `crypto` has to be imported and is not accessible on `globalThis`. So the import from `../../src/index.browser` will throw in a node environment.

This can be rectified by importing from the package and letting the code bundler decide which code to use.

```ts
import { encrypt, decrypt } from 'fuels';

/**
   * @group node
   * @group browser
   ........
```

cc @Torres-ssf

danielbate:
I had massively overcomplicated the problem around splitting up the `crypto` tests, thankyou to @arboleya for the fresh perspective and idea 🙌🏻  [This commit](https://github.com/FuelLabs/fuels-ts/pull/1630/commits/13a15266e0070f7583c981128ab7c487ddcf9f86) has resolved the `crypto` test grouping problem.

scripts/tests-browser.ts
```diff
@@ -0,0 +1,75 @@
...
+      setTimeout(() => {
+        symlinkDeps();
+        execSync('pnpm install --no-frozen-lockfile', { stdio: 'inherit' });
+      }, 10000);
+    } else {
+      executeBrowserTest();
+    }
+  } catch (err) {
+    symlinkDeps();
+    console.error(err);
```
arboleya, line 72: The console calls are raising lint warnings.

It's safe to disable the rule for the entire script.

lint

.github/workflows/test.yaml
danielbate: [Like this](https://github.com/FuelLabs/fuels-ts/pull/1630/commits/d918fadc857c4156e49746c4c75c93e98581eb1f)

scripts/tests-browser.ts
danielbate: [Rule disabled in script](https://github.com/FuelLabs/fuels-ts/pull/1630/commits/e965a62e43182f0745a05bbf853ab55ccee9dd9b) and added #1645

scripts/tests-browser.ts
```diff
@@ -0,0 +1,76 @@
...
+const executeBrowserTest = () =>
+  execSync(
+    'vitest --run --coverage --config vite.browser.config.mts $(scripts/tests-find.sh --browser)',
+    { stdio: 'inherit' }
+  );
+
+(() => {
+  try {
+    if (packagesSupportingBrowserTesting.length) {
+      hardlinkDeps();
```
arboleya, line 61: Why is this needed?

Can't `vite` work seamlessly with `pnpm` workspace?

danielbate: Cast your mind back to our early Vitest browser testing. We ran into the following error for certain tests: `Class extends value undefined is not a constructor or null`. Online didn't give much away, however I assumed it was to do with browser bundling due to the discrepancy with the `node` environment, and polyfill errors were _similar_. Hardlinking the dependencies resolved the problem for our `browser` tests.

However, I had not ran the suite with workspace linking since Vitest had released a major version of the browser testing package, and it seems to now be [working](https://github.com/FuelLabs/fuels-ts/pull/1630/commits/45b6a3e65e47efa2609fae4a91eca2a483b96aba). They had made a lot of changes to the browser package regarding both configuration and bundling in the major release, so potentially something in here resolved it. 🎉

Torres-ssf (approved):

scripts/tests-browser.ts
arboleya: Oh yeah, I do remember that weird error!

I'm glad that doesn't happen anymore 🎉

arboleya (approved):

PR #1640, opened 2024-01-12
chore: remove `ethers` dependency from `utils`

### LINKED ISSUE #1638: Remove `ethers` dependency from the `utils` package
Since our move to Vitest in #1310, although the API is very close to Jest, there are differences in how Jest approaches mocking. One area is that we can no longer mock dependencies that are used by test setup files. Such as our dependency on `ethers` to `hexlify` a test worker ID to retrieve a private key, to access a faucet to be used in tests.

We now see the following error - [source](https://github.com/FuelLabs/fuels-ts/actions/runs/7493242747/job/20398872527?pr=1495#step:9:385):
> Error: [vitest] Cannot mock "ethers" because it is already loaded. Did you import it in a setup file?

Since we have introduced our own `hexlify` function in #1401, we can use this instead. However we will then need to remove the `ethers` dependency from the `utils` package. The main awkwardness for this is the `BytesLike` type that is used across the SDK. However we should look to introduce our own to across the board reduce our dependency on `ethers`.

We have also [pinned our vitest version](https://github.com/FuelLabs/fuels-ts/pull/1637) to counteract this error, so currently we risk being out of date.

danielbate:
Closes #1638 

- Removes `ethers` dependency from `utils` by using internal types and functions
- Removes `ethers` dependency from workspace by using `utils` in `vite.env.ts`
- Resolves mocking `ethers` issue in tests - [source](https://github.com/FuelLabs/fuels-ts/actions/runs/7493242747/job/20398872527?pr=1495#step:9:385)
- Allows us to unpin `vitest` and upgrade


arboleya (approved):

Dhaiwat10 (approved):

arboleya (approved):
Clean. 🧹

PR #1649, opened 2024-01-15
chore: implement turbo CI cache

danielbate:
- Adds a new GitHub cache for turbo


.github/actions/test-setup/action.yaml
```diff
@@ -15,7 +15,15 @@ runs:
         path: ~/.local/share/pnpm/store/v4
         key: ${{ runner.os }}-node-${{ inputs.node-version }}-pnpm-${{ inputs.pnpm-version }}-lock-${{ hashFiles('**/pnpm-lock.yaml') }}
         restore-keys: |
-          ${{ runner.os }}-pnpm
+          ${{ runner.os }}-node-
```
danielbate, line 18: This `restore-key` would have been incorrect, had resolved here but can move if out of scope

package.json
```diff
@@ -12,7 +12,7 @@
   "scripts": {
     "dan": "tsx ./scripts/release-unpublish.ts",
     "dev": "nodemon --config nodemon.config.json -x 'pnpm build:packages'",
-    "build": "turbo run build",
+    "build": "turbo run build --cache-dir=.turbo",
```
Torres-ssf, line 15: I experienced several cache invalidation issues with Turbo Repo while working on Ubuntu. 

Do you think these issues might be solved with the changes introduced in this pull request?

Torres-ssf (dismissed):

arboleya (approved):
Does this mean a single build would be shared among all subsequent workflows?

Meaning, the first would build the project, and the rest would be cache-hit.

danielbate:
@arboleya correct, the cache key uses the `sha` so subsequent builds will hit the cache. Useful for our multiple stage `test` CI.

package.json
arboleya: @danielbate Not sure if it is related, but something [broken](https://github.com/FuelLabs/fuels-ts/actions/runs/7556364202/job/20573212449?pr=1649).

danielbate: Hmm it should still be caching from `node_modules` by default, this just makes the path more easily accessible from CI. Let me know if this does solve it for you.

danielbate: On second thoughts, we should not be using partial restores here. Unless anyone can think of a reason to? I believe that is why the build was failing. Have resolved it by removing `restore-keys`. Worked the first time round as it was the first cache run. We can see it working in a dependent stage [here](https://github.com/FuelLabs/fuels-ts/actions/runs/7560466639/job/20586820316#step:3:153).

.github/actions/ci-setup/action.yaml
```diff
@@ -14,8 +14,12 @@ runs:
       with:
         path: ~/.local/share/pnpm/store/v4
         key: ${{ runner.os }}-node-${{ inputs.node-version }}-pnpm-${{ inputs.pnpm-version }}-lock-${{ hashFiles('**/pnpm-lock.yaml') }}
-        restore-keys: |
-          ${{ runner.os }}-pnpm
```
danielbate, line 18: Removed the `restore-key`. It was incorrect anyway as the above restore key stars with ${{ runner.os }}-node so was redundant. And we should not be restoring the cache from a different run, if versions have changed.

arboleya (approved):

Torres-ssf (approved):

Issue #1633, opened 2024-01-11
Add new workflow step around breaking changes


arboleya:
We need to add some safeguards around breaking changes to reduce problem-cascading to dependent parts of the stack.

This can be done with some simple steps:

1. Check if the PR has to breaking changes
2. If it does, requires a review from the `FuelLabs/frontend`
3. Checks must fail until someone from the `FuelLabs/frontend` team approves it

PR #1650, opened 2024-01-15
chore: add workflow for FE conditional review

Torres-ssf:
This PR introduces new CI configurations designed to optimize our review process for breaking changes. 

It does that by involving the `frontend` team in all PRs marked with breaking changes.

When a PR is marked as a breaking change ( indicated by a `!` in the PR title e.g. `feat!: add xyz` ), the workflow will add the `frontend` team to the list of reviewers. 

This ensures that the `frontend` team is promptly involved in reviewing significant changes.

The workflow is configured to fail if no review has been provided by any member of the frontend team. This check will continue to fail until a `frontend` team member has reviewed the PR, ensuring that breaking changes do not get merged without their knowledge.

- Closes: https://github.com/FuelLabs/fuels-ts/issues/1633


danielbate (changes requested):
testing workflow

danielbate (approved):

arboleya (approved):
It does so much with so little!

Very clean and concise.

Great stuff! 🔥 


> _If we consider change requests events in the future, perhaps we can start using something like below or similar — this is especially the case for dealing with complex JSON responses:
> ```
> run tsx ./script/fe-conditional-review.ts
> ```

PR #1654, opened 2024-01-16
fix: flaky retry test

nedsalk:
Increased `expectedWaitTime` tolerance to 50. This is still in the acceptable range because `baseDuration` is 150, so anything off in the retry algorithm would be registered.


arboleya (approved):

arboleya:
Aren't these assertions redundant?

https://github.com/FuelLabs/fuels-ts/blob/7e7dae9cbac01bc1293fb26e16b8343b664a3eb9/packages/providers/test/retry.test.ts#L88-L92

https://github.com/FuelLabs/fuels-ts/blob/7e7dae9cbac01bc1293fb26e16b8343b664a3eb9/packages/providers/test/retry.test.ts#L115-L119

arboleya:
Also, shouldn't we be testing this on @browser as well?

https://github.com/FuelLabs/fuels-ts/blob/7e7dae9cbac01bc1293fb26e16b8343b664a3eb9/packages/providers/test/retry.test.ts#L32-L34

> cc @danielbate

nedsalk:
@arboleya Ha, I reorganized them into redundancy while organizing the tests. I fixed it in https://github.com/FuelLabs/fuels-ts/pull/1654/commits/d0da15f9f4a73aab1e994a6a470dad010e33f06d.

The idea is to verify that the mock repeats the retries the set amount of times, after which it hits the node again and returns the same name it returned before the mocks were set.

packages/providers/test/retry.test.ts
```diff
@@ -52,7 +52,7 @@ describe('Retries correctly', () => {
     // this might be something related to the event loop and how it schedules setTimeouts
     // expectedWaitTime minus 5ms seems like reasonable to allow
     expect(waitTime).toBeGreaterThanOrEqual(expectedWaitTime - 5);
-    expect(waitTime).toBeLessThanOrEqual(expectedWaitTime + 15);
+    expect(waitTime).toBeLessThanOrEqual(expectedWaitTime + 50);
```
arboleya, line 55: Isn't this still prone to flakiness?

Timeouts and small ms-drifts are fragile, as things can always run faster/slower.

nedsalk: I agree and that's the reason why we were having flakiness, but I don't have an idea how to test it in another manner. I'd be happy to discuss another take on it. We can also leave it with 50ms and see if the flakiness returns and then consider it further.

packages/providers/src/utils/auto-retry-fetch.test.ts
```diff
@@ -0,0 +1,111 @@
+import { safeExec } from '@fuel-ts/errors/test-utils';
+
+import type { FetchRequestOptions } from '../provider';
+
+import type { RetryOptions } from './auto-retry-fetch';
+import { autoRetryFetch, getWaitDelay } from './auto-retry-fetch';
+
+/**
+ * @group node
```
arboleya, line 9: Forgot this—_probably worth doing_.

> cc @danielbate 

```suggestion
 * @group node
 * @group browser
```

packages/providers/src/utils/auto-retry-fetch.ts
```diff
@@ -0,0 +1,78 @@
...
+   */
+  backoff?: Backoff;
+  /**
+   * Starting delay for backoff strategy. Default is 150ms.
+   */
+  baseDelay?: number;
+};
+
+/**
+ * Calculate the wait duration for the next retry attempt
```
arboleya, line 26: Tyop, my bad.

```suggestion
 * Calculate the delay for the next retry attempt
```

arboleya (changes requested):

packages/providers/src/utils/auto-retry-fetch.ts
```diff
@@ -0,0 +1,78 @@
...
+  retryAttemptNum: number = 0
+): NonNullable<ProviderOptions['fetch']> {
+  if (options === undefined) {
+    return fetchFn;
+  }
+
+  return async (...args) => {
+    try {
+      return await fetchFn(...args);
+    } catch (_error: unknown) {
```
nedsalk, line 64: @arboleya I merged this but didn't realize that you didn't actually add the `ECONNREFUSED` handling back https://github.com/FuelLabs/fuels-ts/pull/1662#discussion_r1455721427. I think it's better to have it like it was previously until we figure out all the use cases where we want to retry instead of retrying on _every_ error. This implementation would unnecessarily repeat rejected bad requests and delay feedback to the user.

arboleya: Yes, maybe I wasn't clear enough, but I left it for you. 👍 

That's what I mean by:
 - ["[...] ensure you didn't forget to test anything after re-adding the `ECONNREFUSED` conditional"](https://github.com/FuelLabs/fuels-ts/pull/1662#discussion_r1456024812)

packages/providers/src/utils/auto-retry-fetch.test.ts
```diff
@@ -0,0 +1,129 @@
...
+
+describe('autoRetryFetch', () => {
+  const url = 'http://anythibng.com';
+  const fetchOptions: FetchRequestOptions = { method: 'POST', headers: {}, body: '' };
+
+  const maxRetries = 5;
+  const baseDelay = 1;
+  const retryOptions: RetryOptions = { maxRetries, baseDelay, backoff: 'fixed' };
+  const econnRefusedError = new Error();
+  econnRefusedError.cause = { code: 'ECONNREFUSED' };
```
arboleya, line 66: Nit.

```suggestion

  const econnRefusedError = new Error();
  econnRefusedError.cause = { code: 'ECONNREFUSED' };
```

packages/providers/src/utils/auto-retry-fetch.test.ts
```diff
@@ -0,0 +1,129 @@
...
+  it('should retry until maxRetries and fail', async () => {
+    const fn = vi.fn(() => {
+      throw econnRefusedError;
+    });
+
+    const autoRetry = autoRetryFetch(fn, retryOptions);
+
+    const { error, result } = await safeExec(async () => autoRetry(url, fetchOptions, {}));
+
+    expect(fn).toHaveBeenCalledTimes(1 + maxRetries); // the 1 is for the first try that's not counted as a retry
```
arboleya, line 91: Nit.

```suggestion
    expect(fn).toHaveBeenCalledTimes(1 + maxRetries); // 1st call is not a retry
```

packages/providers/src/utils/auto-retry-fetch.test.ts
```diff
@@ -0,0 +1,129 @@
...
+        throw econnRefusedError;
+      }
+      return Promise.resolve(new Response());
+    });
+
+    const autoRetry = autoRetryFetch(fn, retryOptions);
+
+    const { error, result } = await safeExec(async () => autoRetry(url, fetchOptions, {}));
+
+    expect(fn).toHaveBeenCalledTimes(1 + maxRetries); // the 1 is for the first try that's not counted as a retry
```
arboleya, line 111: Nit.

```suggestion
    expect(fn).toHaveBeenCalledTimes(1 + maxRetries); // 1st call is not a retry
```

packages/providers/src/utils/auto-retry-fetch.ts
```diff
@@ -0,0 +1,82 @@
...
+   * Starting delay for backoff strategy. Default is 150ms.
+   */
+  baseDelay?: number;
+};
+
+/**
+ * Calculate the delay for the next retry attempt
+ * @param options - Retry options configuration
+ * @param retryAttemptNum - 1-based, tells which retry attempt is this
+ * @returns Next wait duration
```
arboleya, line 29: Tyop.

```suggestion
 * @returns Next wait delay
```

packages/providers/test/auto-retry-fetch.test.ts
```diff
@@ -0,0 +1,32 @@
+import Provider from '../src/provider';
+import type { RetryOptions } from '../src/utils/auto-retry-fetch';
+import * as autoRetryFetchMod from '../src/utils/auto-retry-fetch';
+
+// TODO: Figure out a way to import this constant from `@fuel-ts/wallet/configs`
+const FUEL_NETWORK_URL = 'http://127.0.0.1:4000/graphql';
+
+/**
+ * @group node
```
arboleya, line 9: Should we also run this one on browser env?

```suggestion
 * @group node
 * @group browser
```

arboleya (approved):
Great stuff! 🚀 

I left a couple of optional nits, one tyop, one question, one todo.

packages/providers/src/utils/auto-retry-fetch.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  return async (...args) => {
+    try {
+      return await fetchFn(...args);
+    } catch (_error: unknown) {
+      const error = _error as Error & { cause?: { code: string } };
+      if (error.cause?.code !== 'ECONNREFUSED') {
+        throw error;
+      }
```
arboleya, line 68: Another one.

```suggestion

      /**
       * So far, we are auto-retrying only for `ECONNREFUSED` error.
       * TODO: Investigate if we should consider more errors.
       */
      if (error.cause?.code !== 'ECONNREFUSED') {
        throw error;
      }
```

packages/providers/test/auto-retry-fetch.test.ts
nedsalk: I'd add it because it can't hurt, although I don't see the clear benefit in running these types of tests in the browser as well because they are just function calls without any dependencies to anything.

nedsalk: We got a [failing test](https://github.com/FuelLabs/fuels-ts/actions/runs/7567069619/job/20605548932?pr=1654#step:9:90) because of this. @danielbate Do you have some insight into this?

arboleya: Interesting. Does it fail locally for you? It worked for me.

The only gotcha is the process takes a while to exit.

nedsalk: Running `pnpm test:browser` failed locally for me as well, with the same error.

nedsalk: Removed tag: https://github.com/FuelLabs/fuels-ts/pull/1654/commits/f59e20fc54ce2aa1e612e1cf7b83fe67ca3eae16

arboleya (approved):

Torres-ssf (approved):

PR #1659, opened 2024-01-17
chore: fix string replace in `rc` ci

danielbate:
Fixed [this](https://github.com/FuelLabs/fuels-ts/actions/runs/7548552848/job/20550861298) as string replacement should be like [this](https://github.com/FuelLabs/fuels-ts/actions/runs/7543914422/job/20536003398)


nedsalk (approved):

Torres-ssf (approved):

arboleya:
@danielbate It looks like things [broke](https://github.com/FuelLabs/fuels-ts/pull/1495/commits/1b0bb20d0bdb380ae40e0004d2774255df434c29) again [here](https://github.com/FuelLabs/fuels-ts/actions/runs/7556059402/job/20572290954).

If I may suggest something, perhaps we should go back to the [`if: false`](https://github.com/FuelLabs/fuels-ts/blob/master/.github/workflows/pr-release.yaml#L11) routine for auto-deploying this PR.

I say this because we'll move away from using RCs anyway, and it may be a waste of time to keep trying to fix this.

arboleya:
@Torres-ssf That doesn't mean you wouldn't have to implement something similar on the breaking-changes endeavor.

However, I believe that would be simpler since there will be no intermediary PRs.

danielbate:
@arboleya Agreed, just frustrating. Looks as though because both the `pull_request` and `push` events triggered the workflow, one had not finished publishing and therefore failed.

arboleya:
@danielbate Btw, it looks like the browser-testing PR also [failed](https://github.com/FuelLabs/fuels-ts/pull/1495/commits/268214814138ecb9cf4c90f8c6f194842c5e56c8) after we merged it.

danielbate:
Oh man, `if:false` it is

arboleya:
@danielbate Feel free to add another PR to remove this workflow then.

After that, I can turn the auto-release on/off directly on the `rc/salamander` branch.

arboleya:
Actually, the problem is that the `if: false` doesn't follow the same naming pattern and will break the FE integrations. 🤔

danielbate:
I think it still has merit individually releasing `@rc-salamander` tags, but will change the trigger to `pull request` only with an `rc/` check #1660

PR #1662, opened 2024-01-17
fix: auto-retry fetch calls, take #2

arboleya:
1. Renamed file/methods for more clarity
2. Moved the utility inside the `utils` directory
3. Unit-tested everything w/ 💯 coverage
4. Kept a very small integration test to ensure the method is called


packages/providers/src/call-retrier.ts
```diff
@@ -1,70 +0,0 @@
...
-
-  return async (...args) => {
-    try {
-      return await fetchFn(...args);
-    } catch (e: unknown) {
-      const error = e as Error & { cause?: { code?: string } };
-
-      if (error.cause?.code !== 'ECONNREFUSED') {
-        throw e;
-      }
```
arboleya, line 56: @nedsalk Note that I removed this `ECONNREFUSED` conditional.

Maybe I'm missing something, but I'd consider this an error that could _also_ be retried.

nedsalk: The idea was to retry _only_ on failure to connect to the fuel node. There is no point in retrying on an error like "not enough coins...".

arboleya: I see, thanks for clarifying.

In this case, we must add new test cases to cover this.

How did you narrow this error down to only `ECONNREFUSED`?

Could there be other valid cases that are not accounted for? For example, what if the user gets a rate-limit response or something else?

nedsalk: The discussions at the time were around not being able to connect so I just implemented that. I didn't think about other ways that the connection might be refused. I think it's best to create a discussion/issue around that.

arboleya: No worries, my main intention was to remove flakiness, which seems out of the way now. We can go like this now and revisit it later as necessary. Your implementation is almost as before, except for very tiny fixes, improvements, and docs. The big chunk of changes was by concentrating on unit-tests.

Note how it became easier to fully stress-test things in a unit fashion and keep an eye on the code coverage report to ensure you didn't forget to test anything after re-adding the `ECONNREFUSED` conditional.

```shell
pnpm test:filter packages/providers/src/utils/auto-retry-fetch.test.ts
open coverage/environments/node/index.html
```

coverage

packages/providers/src/utils/auto-retry-fetch.ts
```diff
@@ -0,0 +1,78 @@
...
+  const duration = options.baseDelay ?? 150;
+
+  switch (options.backoff) {
+    case 'linear':
+      return duration * retryAttemptNum;
+    case 'fixed':
+      return duration;
+    case 'exponential':
+    default:
+      return 2 ** (retryAttemptNum - 1) * duration;
```
arboleya, line 41: I forgot to mention that `**` is the same as `Math.pow`, while `^` is a [bitwise XOR](https://www.w3schools.com/js/js_bitwise.asp) operator.

nedsalk: Yeah, math brain fart from my side.

Issue #1550, opened 2023-12-19
Lack of import and constructor examples	


arboleya:
Each type/object should show me exactly where the module is, how to import it and how to construct it. There was a big lack of this for each type in the type section, which left me having to go read code to figure it out.

> [!NOTE]
> - Incoming from https://github.com/FuelLabs/fuels-ts/discussions/1545

arboleya:
Completed via https://github.com/FuelLabs/fuels-ts/pull/1665

PR #1665, opened 2024-01-18
docs: validate imports added to code snippets

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/1666

This PR adds a new way to generate imports within code snippets by using the `#import` flag:

```ts
// #import { AssetId };

const assetId: AssetId = {
  value: Bits256,
};
```

This final result will be:
```ts
import { AssetId } from 'fuels';

const assetId: AssetId = {
  value: Bits256,
};
```


danielbate:
🚀

apps/docs-snippets/src/guide/cookbook/transferring-assets.test.ts
```diff
@@ -26,7 +26,7 @@ describe(__filename, () => {
 
   it('should successfully transfer asset to another account', async () => {
     // #region transferring-assets-1
-    // #context import { Wallet, BN, BaseAssetId } from 'fuels';
+    // #addImport: Wallet, BN, BaseAssetId
```
arboleya, line 29: Is there any particular reason why you went with `#addImport a, b, c`?

I think a more standardized syntax would be more intuitive:

```ts
// #import { Wallet, BN, BaseAssetId };
```

Torres-ssf: @arboleya No particular reason.

But yeah, the suggested approach seems more intuitive to me as well.

packages/fuel-gauge/src/doc-examples.test.ts
```diff
@@ -344,7 +342,7 @@ describe('Doc Examples', () => {
 
   it('can create a predicate', async () => {
     // #region predicate-basic
```
petertonysmith94, line 344: The `predicate-basic` context isn't used in the docs.
- Suggest to add reference in the docs to `predicate-basic`

packages/script/src/script.test.ts
```diff
@@ -68,7 +68,7 @@ const callScript = async <TData, TResult>(
 };
 
 // #region script-init
-// #context import { Script, AbiCoder, getBytesCopy } from 'fuels';
+// #import { ScriptRequest, getBytesCopy }
```
petertonysmith94, line 71: Missing `Interface` import

packages/wallet/src/wallet-unlocked.test.ts
```diff
@@ -50,7 +50,7 @@ describe('WalletUnlocked', () => {
 
   it('Sign a transaction using wallet instance', async () => {
     // #region wallet-transaction-signing
-    // #context import { WalletUnlocked, hashMessage, Signer} from 'fuels';
+    // #import { WalletUnlocked, Signer }
```
petertonysmith94, line 53: Missing `Provider` import

petertonysmith94 (commented):
Just had a quick look through 👍🏼 
A couple a small issues 😄

packages/script/src/script.test.ts
Torres-ssf: This will be addressed on another PR.
> ref: https://github.com/FuelLabs/fuels-ts/issues/1550

danielbate (commented):
Nice work, especially enjoyed the test cases in `extractImports.test.ts`. 🧙🏻 

I'll admit it wasn't easy reading through `extractImports.ts`, I don't necessarily think it is something to do with how you've written it, but this kind of file parsing and line matching is never pleasant. Potentially a nit on my side, but I think it could benefit by making it more verbose, some docs on the regexes and docs for some of the real specific code. As an example:
```ts
if (ignoreMatches && ignoreMatches[1]) {
        const importsToIgnore = ignoreMatches[1].split(',').map((item) => item.trim());
...
```
What is the condition we are checking for here and what is being split, to return in what format?

But the test cases made it a very easy to understand, maybe more functions to make it more independently testable in places? Your work in #1668 is a good example of this.

apps/docs-snippets/src/guide/contracts/contract-balance.test.ts
```diff
@@ -15,8 +15,7 @@ describe(__filename, () => {
 
   it('should successfully get a contract balance', async () => {
     // #region contract-balance-3
-    // #context import type { AssetId } from 'fuels';
-    // #context import { Wallet, BN, BaseAssetId } from 'fuels';
+    // #import { AssetId, Wallet, BN, BaseAssetId }
```
arboleya, line 18: Just reinforcing what I said already, unless there's a reason behind it, I suggest keeping these imports with an ending semi-colon to keep the standards and avoid variations and unnecessary questions.

```
// #import { AssetId, Wallet, BN, BaseAssetId };
```

apps/docs/.vitepress/plugins/utils/extractImports.test.ts
```diff
@@ -0,0 +1,282 @@
...
+
+describe('extractImports', () => {
+  describe('parseIgnoreImportFlags', () => {
+    it('should return an empty set for no ignore flags', () => {
+      const lines = ['import { something } from "somewhere";', 'const x = 5;'];
+      expect(parseIgnoreImportFlags(lines)).toEqual(new Set());
+    });
+
+    it('should correctly parse a single ignore flag', () => {
+      const lines = ['// #ignore { testImport }'];
```
arboleya, line 23: I couldn't find any snipped using this `#ignore` variation.

I suggest adding clear docs around `#import` and `#ignore` on the PR description.

arboleya (changes requested):

apps/docs/.vitepress/plugins/utils/extractImports.test.ts
Torres-ssf: @arboleya I've removed the `#ignore` import functionality since it was not being used.

arboleya (approved):

Dhaiwat10 (approved):
Looks good!

- In the PR description, can you also mention the fact that this PR adds _validation_ that checks for the validity of the imports in our docs snippets?
- The FE conditional review workflow seems to be broken on this PR

PR #1667, opened 2024-01-19
fix: bin installation from git

arboleya:
# 1) Boken Commands

A couple of commands were broken, like:

```ts
import fs from 'fs/promises';

fs.rmSync(...);
```

But there's no `<cmd>Sync()` methods in `fs/promises`.

I also used the opportunity to standardize a couple of things.

# 2) Problematic `tar` package

I had problems with the TAR package, saying the `tar` files were invalid.

Since we won't support Windows soon, I removed them and rolled back the original commands.

### DIFF
--- a/.changeset/beige-actors-march.md
+++ b/.changeset/beige-actors-march.md
@@ -1,0 +1,6 @@
+---
+"@fuel-ts/forc": patch
+"@fuel-ts/fuel-core": patch
+---
+
+Fixing installation from git branches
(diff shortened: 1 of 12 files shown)


Torres-ssf (approved):

Torres-ssf (approved):

petertonysmith94 (commented):
Won't approve but looks good 💯

danielbate (approved):
:shipit:

Issue #1672, opened 2024-01-22
Encoding / Decoding revamp


arboleya:
The Sway compiler is going through a long-discussed revamp of its encoder/decoder scheme, eventually leading to an intermediary representation that both the SDK and the compiler will target.

This will act as a topic issue to encompass everything necessary:

Must have:
 - [x] https://github.com/FuelLabs/fuels-ts/issues/1671
 - [x] https://github.com/FuelLabs/fuels-ts/issues/1739
 - [x] https://github.com/FuelLabs/fuels-ts/issues/1749

----

Issues from other repos:
 - Specs: https://github.com/FuelLabs/fuel-specs/pull/541
 - Sway: https://github.com/FuelLabs/sway/issues/4769
 - Rust SDK: https://github.com/FuelLabs/fuels-rs/issues/1246

----

Desirable items will be addressed as follow-ups:
 - https://github.com/FuelLabs/fuels-ts/issues/1187
 - Configurable constants (currently [unsupported](https://github.com/FuelLabs/sway/issues/5727) in sway)
 - Introduce config driven coding and reduce touch points - #1795 
 - Functional rewrite for `abi-coder` - #1795 
 - Merge the `forc-project` directories - #2005

danielbate:
The process I am using for this work is as follows:

- [ ] Introduce `v1` decode, two version support and encode `RawSlice` as `u8` - #1673 
- [ ] Introduce `v1` encode and multiple version support
- [ ] Support in contracts for #1739 
- [ ] Support in scripts and predicates for #1749
- [ ] Reduce coder touch points and be more config driven
- [ ] Introduce width checking for #1187

arboleya:
Yes, we can. The desired items can be follow-ups.

PR #1673, opened 2024-01-22
feat!: new encoding scheme for logs and encode `RawSlice` as `u8`

### LINKED ISSUE #1671: Support new encode/decode for `__log`
New encoding/decoding was implemented for `__log` calls, as per:
 - https://github.com/FuelLabs/sway/pull/5306

We need to start the work around the TS SDK to support it.

danielbate:
Closes #1671

This PR adds support for sway's experimental encoding in logs. This is demonstrated via a new test suite, which is a subset of `fuel-gauge`, which builds another `forc` workspace using the experimental flag (`--experimental-new-encoding`).

There are opportunities for refactoring here, however that shall be detailed and implemented as part of #1672 given this PR is already over 2K lines.

**Breaking Change**

`RawSlice` data encoded as unpadded `u8`, also implemented by the [RS SDK](https://github.com/FuelLabs/fuels-rs/pull/1259). Currently we are encoding it as `u64` and doing additional arithmetic on `encode` and `decode` to compute the byte length. A `RawSlice` indicates a slice of bytes so we should treat it as so.

This poses a problem either on `v0` or `v1`, one will have to compromise to the other as `v1` assumes the data as `u8` so we'd have to do additional arithmetic to get the correct `len`, rather than decoding it from the byte data. [I have opted](https://github.com/FuelLabs/fuels-rs/pull/1259) to introduce the breaking change in `v0` as it brings us closer to `v1`.

**Type Support**

- [x] `u8` 
- [x] `u16`
- [x] `u32`
- [x] `u64`
- [x] `b256`
- [x] `b512`
- [x] Array
- [x] String Array (`str[n]`)
- [x] Option
- [x] Raw Slice
- [x] Tuple
- [x] Struct
- [x] Enum
- [x] String Dynamic (`struct String`)
- [x] Vec
- [x] Bytes

> :warning: String Slices are currently not supported in the TS SDK by encoding `v0`. This is possible in `v1`, however requires contract calls to be encoded as well, so the `value` can be encoded/decoded, and thus the `logs`. I have added a `StrSlice` coder using the specs and partially tested, but I have left it unimplemented within `AbiCoder`. Therefore it is currently blocked by [this Sway issue](https://github.com/FuelLabs/sway/issues/5512). 

**Further changes**
- [x] Experimental integration test suite in parallel
- [x] Use `ABIEncode` implementations from `sway-std-lib`
- [x] Unit tests for `v1` coders
- [x] TOB validation changes - #1426

**Blockers**
- [x] Forc `0.49.2` upgrade - #1707
- [x] Get encoding version from ABI - https://github.com/FuelLabs/sway/pull/5481


packages/abi-coder/src/coders/v1/enum.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+import { U64_MAX } from '../../../test/utils/constants';
+import { U64Coder } from '../v0/u64';
+
+import { BooleanCoder } from './boolean';
+import { EnumCoder } from './enum';
+
+/**
+ * @group node
+ * @group browser
```
Torres-ssf, line 13: Why do we need to run this on the browser?

Torres-ssf (approved):
LGTM 🚀

packages/abi-coder/src/coders/v1/enum.test.ts
danielbate: Good question. For me there isn't necessarily a need but if it can, and it is used in a browser environment (unlike CLI tools etc), then we should test it in a browser env.

arboleya: Yes, the default rationale should be to test everything in the browser.

Unless there's a reason not to, which is the case for CLI tools.

nedsalk (approved):
LGTM!

danielbate:
Sorry @arboleya, I enabled auto merge because of the 2 SDK reviews, but also on the basis that a FE review would be the third reviewer.

> cc @Torres-ssf

danielbate:
Any concerns can also be picked up in #1672

Torres-ssf:
Yeah, it shouldn't be possible to merge this one without the review from someone in the FE team 🤔 .

arboleya:
@danielbate Perhaps we should avoid using auto-merge in considerably big PRs to gather the maximum reviews possible.

@Torres-ssf Yes, I will take a look at it later on. Maybe I need to change something in the repo settings, but I'm unsure.

Torres-ssf:
@Dhaiwat10 I believe that was the problem. The check didn't pass but the PR got merged anyway.

arboleya:
@Torres-ssf Yes. That's what I'll double check in the repo settings.

Issue #1518, opened 2023-12-14
Account Restructure


arboleya:
Following the whole re-structuring process, this issue will deal with the Accounts alone:
 - https://github.com/FuelLabs/fuels-ts/discussions/1517

PR #1675, opened 2024-01-22
chore!: restructure `Account` related packages

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/1518


packages/wallet/src/account.test.ts
```diff
@@ -478,4 +483,235 @@ describe('Account', () => {
...
+      gasLimit: 10_000,
+    });
+    await response.wait();
+
+    const senderBalances = await sender.getBalances();
+    const receiverBalances = await receiver.getBalances();
+
+    expect(senderBalances).toEqual([{ assetId: BaseAssetId, amount: bn(499921) }]);
+    expect(receiverBalances).toEqual([{ assetId: BaseAssetId, amount: bn(1) }]);
+  });
```
arboleya, line 502: Are you adding _new_ tests? Ideally, this restructuring would only move things around and update paths.

Also, I know it's early to review this, but I strongly suggest we use the `account` nomenclature instead of `wallet` for the umbrella package.

Torres-ssf: @arboleya I've just moved tests from one test suite to another. Both test suites were related to the account methods.

arboleya: Oh, I see; you didn't move the file itself, but combined two in one. 👍

arboleya (commented):
We could use the fact that this PR is a breaking change to address the moving of the `providers` package, avoiding a second breaking change shot after. I'd gladly look at the ASM issues if you push the broken code. 😃

packages/account/src/utils/formatTransferToContractScriptData.ts
```diff
@@ -4,7 +4,35 @@ import { BN, type BigNumberish } from '@fuel-ts/math';
 import * as asm from '@fuels/vm-asm';
 import { getBytesCopy, type BytesLike } from 'ethers';
 
-export const composeScriptForTransferringToContract = async () => {
+interface IAssembleTransferToContractScriptParams {
+  hexelifiedContractId: B256Address;
```
nedsalk, line 8: This should be `hexlifiedContractId` but I can't make a typo fix suggestion as it spans multiple files.

nedsalk (changes requested):
LGTM! The [typo](https://github.com/FuelLabs/fuels-ts/pull/1675#discussion_r1477901551) is the only issue I found.

danielbate (dismissed):
🪓

arboleya (approved):
This PR just shows how packages' over-fragmentation bloats everything up!

Look at the defrag aftermath! 🙌 🚀 

bloat

> [!NOTE]
> I suspect we could add a `accounts/src/wallet` directory to contain all the _wallet_-realted files that are currently in the `accounts/src` root.

nedsalk (approved):

danielbate (approved):

Issue #1505, opened 2023-12-11
Migrate snippets from `fuel-gauge`


danielbate:
We currently have quite a few tests that references documentation from `fuel-gauge`. These scenarios are at times improperly formed in order to be more verbose and illustrate techniques rather than validate the end to end functionality of the SDK. As that is the intended usage of `fuel-gauge`, we should instead move `fuel-gauge/src/doc-examples.test.ts` to appropriate places in `apps/docs-snippets`. The documentation references will also need to be altered as the links will have been broken.

petertonysmith94:
@danielbate pls

danielbate:
Sorry meant to assign you @petertonysmith94

petertonysmith94:
Notes:
- [`vector-4`](https://github.com/FuelLabs/fuels-ts/blob/master/apps/docs-snippets/src/guide/types/vector.test.ts#L46)
Doesn't show a return value in the documentation `// expect value` would be beneficial.

mvares:
@petertonysmith94, should only the code-snippets be changed? (that it's with `// #region`)

PR #1684, opened 2024-01-24
docs: consolidate wallet docs

petertonysmith94:
relates: #1505 

### Changes made
- Moved _most_ wallet-based tests from `fuel-gauge` over `docs-snippets`.


packages/fuel-gauge/src/doc-examples.test.ts
```diff
@@ -152,196 +141,6 @@ describe('Doc Examples', () => {
     expect(arrayify(assetId)).toEqual(arrayify(Address.fromB256(assetId).toB256()));
   });
 
-  test('it can work with wallets', async () => {
```
Torres-ssf, line 155: @petertonysmith94, we do want to move all code snippets to the doc-snippets directory when it is possible. 

However, we cannot remove these e2e test cases from here, as they were most likely added for a specific reason. 

When moving these code snippets, we need to write new, specific tests that validate only the required steps to ensure that the code snippet is working. 

Therefore, in many cases, when migrating a code snippet, many test cases can be simplified.

Torres-ssf (changes requested):
@petertonysmith94 Thanks for helping us with this. We really appreciate it 👍

packages/fuel-gauge/src/doc-examples.test.ts
petertonysmith94: Hey @Torres-ssf, that all sounds reasonable!

I will reinstate the `fuel-gauge` package tests now.

In terms of test simplification, did you have any pointers/guidelines for this process?

From what is there now, I assume that will be generally keeping small, organised, with descriptive test file-name, and having single purpose. Happy to continue with this for now

Torres-ssf: @petertonysmith94 The conventions that we use for code snippet tests are:

1 - Placing them inside `apps/docs-snippets`

2 - Naming the test file with the same name as the doc page.

3 - Design the test case based on what the code snippets are intended to show.

4- If the code snippet is from a Sway project, it needs to come from one of the projects within `app/docs-snippets/test/fixtures/forc-projects/` or a new one created on this dir.

For instance, on the [Address section](https://fuellabs.github.io/fuels-ts/guide/types/address.html#from-a-public-key), [this specific](https://github.com/FuelLabs/fuels-ts/blob/master/apps/docs-snippets/src/guide/types/address.test.ts#L20-L28) snippet was created to show how to instantiate a wallet from a public key.

petertonysmith94:
Closed in favour of #1885

PR #1691, opened 2024-01-27
feat: add more program options to `create-fuels` CLI

### LINKED ISSUE #1562: Add more program options to the `create-fuels` CLI tool
The default template of the `create-fuels` CLI only includes and is configured for one Sway contract. We should add more flags that let the users pick which Sway programs they want included like mentioned here: https://github.com/FuelLabs/fuels-ts/pull/1408#discussion_r1430698905

```
--contract (default)
--predicate
--script
--all
```

Dhaiwat10:
Closes #1562 

This PR adds a multiselect choice in the `create-fuels` CLI that asks users which Sway programs they want included in the template like so:

https://github.com/FuelLabs/fuels-ts/assets/39617427/e5a765b2-c9b4-445d-a633-02cfd8c90500


packages/create-fuels/package.json
```diff
@@ -19,10 +19,12 @@
     "prepublishOnly": "cp -r ../../templates . && tsx ./scripts/rewriteTemplatePackageJson.ts"
   },
   "dependencies": {
+    "@iarna/toml": "^2.2.5",
     "chalk": "4",
     "commander": "^9.4.1",
     "fuels": "workspace:*",
-    "prompts": "^2.4.2"
+    "prompts": "^2.4.2",
+    "toml": "^3.0.0"
```
Dhaiwat10, line 27: Oops just noticed this unused dependency. Will push a fix today

templates/nextjs/sway-programs/script/src/main.sw
```diff
@@ -0,0 +1,5 @@
+script;
+
+fn main() -> bool {
+    return true;
+}
```
Dhaiwat10, line 5: Maybe this starter code could be better 🤔

arboleya (commented):
Two things:

 1. I got this error when trying to start the generated app:
error
 1. Although generating samples for everything is a good start, I think the app should somehow interact with the generated `Predicate` and `Script` files, just as it does with the `Contract`. For example, it could call the `Script` with some parameter, which could then execute the `Contract` in the event of the `Predicate` evaluating to true _(based on the `Script` received parameter)_. Something like that.

templates/nextjs/sway-programs/script/src/main.sw
nedsalk: Yeah, this is the same as the simplest predicate. Perhaps something like this just to differentiate:
```rs
fn main(number_to_increment: u8) -> u8 {
    return number_to_increment + 1;
}
```

packages/create-fuels/src/cli.ts
```diff
@@ -13,13 +14,49 @@ const log = (...data: unknown[]) => {
...
+    if (member === 'predicate' && !programsToInclude.predicate) {
+      return false;
+    }
+    if (member === 'contract' && !programsToInclude.contract) {
+      return false;
+    }
+    if (member === 'script' && !programsToInclude.script) {
+      return false;
+    }
+    return true;
```
nedsalk, line 40:
```suggestion
    switch (member) {
      case 'predicate':
        return programsToInclude.predicate;
      case 'contract':
        return programsToInclude.contract;
      case 'script':
        return programsToInclude.script;
      default:
        return true;
    }
```

.changeset/unlucky-eggs-design.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-fuels": minor
+---
+
+Add more program optoins to the `create-fuels` CLI. Users can now get a multiselect choice to choose out of contract, script & predicate.
```
nedsalk, line 5:
```suggestion
Add more program options to the `create-fuels` CLI. Users can now get a multiselect choice to choose out of contract, script & predicate.
```

packages/create-fuels/src/cli.ts
```diff
@@ -61,19 +98,55 @@ export const runScaffoldCli = async (
...
+  // delete the programs that are not to be included
+  if (!programsToInclude.contract) {
+    rmSync(join(projectPath, 'sway-programs/contract'), { recursive: true });
+  }
+  if (!programsToInclude.predicate) {
+    rmSync(join(projectPath, 'sway-programs/predicate'), { recursive: true });
+  }
+  if (!programsToInclude.script) {
+    rmSync(join(projectPath, 'sway-programs/script'), { recursive: true });
+  }
```
nedsalk, line 134:
```suggestion
  await cp(join(__dirname, '../templates/nextjs'), projectPath, {
    recursive: true,
    filter: (_source, destination) => {
      switch (destination) {
        case join(projectPath, 'sway-programs/contract'):
          return programsToInclude.contract;
        case join(projectPath, 'sway-programs/script'):
          return programsToInclude.script;
        case join(projectPath, 'sway-programs/predicate'):
          return programsToInclude.predicate;
        default:
          return true;
      }
    },
  });
  await rename(join(projectPath, 'gitignore'), join(projectPath, '.gitignore'));
```

You can skip copying the folders instead of copying and then deleting them. Note that I didn't try out the code above. I still wanted to create a suggestion for the purpose of showcasing the `filter` in the full context.

packages/create-fuels/src/cli.ts
arboleya: If line 26 changes from this:

```ts
members: string[];
```

To this:

```ts
members: ('predicate' | 'contract' | 'script')[];
```

Could we get away with just this?

```ts
parsed.workspace.members = parsed.workspace.members.filter((m) => programsToInclude[m]);
```

nedsalk: Yeah, I was thinking about that too but didn't suggest it just because if any of the property names of `ProgramsToInclude` changes or if the members in `Forc.toml` change the code will break. I'm fine with that suggestion as well, though, as I feel this concern of mine isn't that big of a deal.

nedsalk:
@Dhaiwat10 I believe the `cli.ts` code could use some restructuring, but it was hard to put it into a suggestion due to the size of the suggested change and it's not a deal breaker so I'll just put the idea like this and you use it or just ignore it:
```ts
async function promptForProjectPath() {
  const res = await prompts({
    type: 'text',
    name: 'projectName',
    message: 'What is the name of your project?',
    initial: 'my-fuel-project',
  });

  return res.projectName as string;
}

async function promptForPackageManager() {
  const packageManagerInput = await prompts({
    type: 'select',
    name: 'packageManager',
    message: 'Select a package manager',
    choices: [
      { title: 'pnpm', value: 'pnpm' },
      { title: 'npm', value: 'npm' },
    ],
    initial: 0,
  });
  return packageManagerInput.packageManager as string;
}

async function promptForProgramsToInclude() {
  const programsToIncludeInput = await prompts({
    type: 'multiselect',
    name: 'programsToInclude',
    message: 'Which Sway programs do you want?',
    choices: [
      { title: 'Contract', value: 'contract', selected: true },
      { title: 'Predicate', value: 'predicate' },
      { title: 'Script', value: 'script' },
    ],
    instructions: false,
  });
  return {
    contract: programsToIncludeInput.programsToInclude.includes('contract'),
    predicate: programsToIncludeInput.programsToInclude.includes('predicate'),
    script: programsToIncludeInput.programsToInclude.includes('script'),
  };
}

export const runScaffoldCli = async (
  explicitProjectPath?: string,
  explicitPackageManger?: string,
  shouldInstallDeps = true,
  explicitProgramsToInclude?: ProgramsToInclude
) => {
  const program = new Command(packageJson.name).version(packageJson.version);

  const projectPath = explicitProjectPath || (await promptForProjectPath());
  if (existsSync(projectPath)) {
    // throw and exit
    chalk.red(`A folder already exists at ${projectPath}. Please choose a different project name.`);
    process.exit(1);
  }

  if (!projectPath) {
    console.log(
      '\nPlease specify the project directory:\n' +
        `  ${chalk.cyan(program.name())} ${chalk.green('<project-directory>')}\n`
    );
    process.exit(1);
  }
  const packageManager = explicitPackageManger || (await promptForPackageManager());

  const programsToInclude: ProgramsToInclude =
    explicitProgramsToInclude || (await promptForProgramsToInclude());

// ... rest of the code
}
```

packages/create-fuels/src/cli.ts
arboleya: I'm not sure if this could make it more resilient.

```ts
parsed.workspace.members = parsed.workspace.members.filter((m) => programsToInclude[m] ?? true);
```

nedsalk: On second thought, we can safely go with
```ts
parsed.workspace.members = parsed.workspace.members.filter((m) => programsToInclude[m]);
```
if the tests cover it properly.  They should be covering it in one way or another anyways. I added a comment related to these tests [here](https://github.com/FuelLabs/fuels-ts/pull/1691/files#r1478031625).

arboleya: Code-wise, the first version looks just fine; it is a trivial change that may not even deserve this much attention.

packages/create-fuels/src/cli.ts
Dhaiwat10: @nedsalk let me explain the rationale here: I am selectively deleting the programs I don't need because the 'template' has to have all three programs since we copy the _whole_ template folder at once. And I don't think it is a good idea to have three different templates for the three different program types. The current design also easily enables the users to choose multiple program types.

packages/create-fuels/test/cli.test.ts
```diff
@@ -3,34 +3,47 @@ import { join } from 'path';
...
-
-  // check if the template was extracted to the test-project directory.
-  // compare the templates/nextjs folder with the test-project folder recursively
-  const originalTemplateFiles = await fs.readdir(join(__dirname, '../templates/nextjs'));
-
-  // remove the gitignore file from the array because it is not copied to the test-project folder
-  const gitIgnoreIndex = originalTemplateFiles.indexOf('gitignore');
-  if (gitIgnoreIndex > -1) {
-    originalTemplateFiles.splice(gitIgnoreIndex, 1);
+test.each(possibleProgramsToInclude)(
```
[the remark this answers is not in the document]
Dhaiwat10: It is making sure that the resulting folder contains the correct list of directories, subdirs and files.

templates/nextjs/src/pages/predicate.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const [predicateBalance, setPredicateBalance] = useState<BN>();
+
+  const [pin, setPin] = useState<string>();
+
+  useEffect(() => {
+    (async () => {
+      const provider = await Provider.create("http://127.0.0.1:4000/graphql");
+
+      // 0x1 is the private key of one of the fauceted accounts on your local Fuel node
+      const wallet = Wallet.fromPrivateKey("0x01", provider);
```
arboleya, line 24: I wonder if this is clear enough or how to add more clarity around default faucets.

Perhaps we could generate projects with a `chainConfig.json` by default to reduce the abstractions, making things more explicit for users to customize. It could be side-by-side with the `fuels.config.ts` or something.

arboleya (changes requested):
Good stuff! I noticed a few things we could improve; some may be out of scope for this PR, but I will leave the judgment to you.

packages/create-fuels/src/cli.ts
```diff
@@ -13,58 +13,133 @@ const log = (...data: unknown[]) => {
   process.stdout.write(`${data.join(' ')}\n`);
 };
 
+type ProgramsToInclude = {
+  contract: boolean;
+  predicate: boolean;
+  script: boolean;
+};
```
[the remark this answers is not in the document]
Dhaiwat10: @arboleya I would love that, but I would prefer if we kept both the methods. That's what most scaffolding tools like `create-next-app` do. You can either pass in a config via the command line, or via an interactive prompt. I think I can tackle this in a separate PR.

packages/create-fuels/src/cli.ts
```diff
@@ -13,58 +13,133 @@ const log = (...data: unknown[]) => {
...
+  new Command(packageJson.name).version(packageJson.version);
 
+  const projectPath = explicitProjectPath || (await promptForProjectPath());
   if (existsSync(projectPath)) {
-    // throw and exit
-    chalk.red(`A folder already exists at ${projectPath}. Please choose a different project name.`);
-    process.exit(1);
+    throw new Error(
+      `A folder already exists at ${projectPath}. Please choose a different project name.`
+    );
```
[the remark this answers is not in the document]
Dhaiwat10: This code takes care of it now:
https://github.com/FuelLabs/fuels-ts/blob/dp/scaffold-cli-program-options/packages/create-fuels/src/bin.ts#L9-L12

templates/nextjs/src/pages/index.tsx
```diff
@@ -1,39 +1,43 @@
...
+          contractId,
+          wallet,
+        );
+        setContract(testContract);
+        const { value } = await testContract.functions
+          .get_count()
+          .txParams({
+            gasPrice: 1,
+            gasLimit: 10_000,
+          })
```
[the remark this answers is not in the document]
Dhaiwat10: Was just waiting for that PR to get merged. Will fix this right away!

templates/nextjs/src/pages/predicate.tsx
Dhaiwat10: Interesting. Or maybe instead of the whole chain config, we could start with adding some of these options to the `fuels.config.ts` file itself?

Dhaiwat10: Maybe we can turn this into a discussion

packages/create-fuels/src/cli.ts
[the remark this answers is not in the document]
arboleya: Yes, that's what I meant with `return program.projectName ?? await prompts(...)`. 👍 

 - https://github.com/FuelLabs/fuels-ts/issues/1732

templates/nextjs/src/pages/predicate.tsx
arboleya: The `fuels.config.ts` already have the [chainConfig](https://fuellabs.github.io/fuels-ts/guide/cli/config-file.html#chainconfig) property, and it's the path for the file. 

I'd say we should keep `chainConfig.json` as a JSON file for consistency and portability.

Users can use it for spinning up `fuel-core` nodes using their binaries instead of the [built-in](https://fuellabs.github.io/fuels-ts/guide/cli/built-in-binaries.html#built-in-binaries) ones.

arboleya: @Dhaiwat10 Feel free to open the discussion with your proposal.

nedsalk (approved):

Issue #1463, opened 2023-11-27
Re-incorporate parts from the Wallet SDK


arboleya:
There is a refactoring going on the `fuels-wallet` repo regarding the SDK package:
 - https://github.com/FuelLabs/fuels-wallet/tree/sdk-v2/packages/sdk

The idea of the issue is to re-incorporate that part in this repo for a better separation of concerns.

It's part of a bigger RFC that @luizstacio started here:
 - https://github.com/FuelLabs/fuels-wallet/wiki/Fuel-Wallet-Connectors

#1518 should super-cede this task by first restructuring the `Account` package to then incorporate the above repo.

> [!NOTE]
> - Incoming from https://github.com/FuelLabs/fuels-ts/discussions/1537

PR #1699, opened 2024-02-01
feat: wallet connector

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/1463


Torres-ssf:
Blocked by: https://github.com/FuelLabs/fuels-ts/pull/1675

luizstacio (changes requested):
We should remove the connectors implementation from the ts-sdk. Files `fuel-wallet-development-connector...`, `fuelet-wallet-connector` and `fuel-wallet-connector`...

packages/account/src/connectors/types/fuel-storage.ts
```diff
@@ -0,0 +1,10 @@
+/**
```
luizstacio, line 1: I think we already have a abstraction for storage inside the account used by the vault-manager and we could merge both.

luizstacio (changes requested):
We should remove the connectors implementation from the ts-sdk. Files `fuel-wallet-development-connector...`, `fuelet-wallet-connector` and `fuel-wallet-connector`...

.eslintrc.js
```diff
@@ -29,6 +29,9 @@ module.exports = {
     ],
     '@typescript-eslint/no-non-null-assertion': 1,
     // Disable error on devDependencies importing since this isn't a TS library
+    'no-param-reassign': ['error', { props: false }],
```
arboleya, line 32: AFAICT, this can cause unwanted issues, and we should probably consider keeping the rule on to reduce the chance of mistakes. Unless there's a reason why not to.

.github/workflows/pr-release.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
     name: "Release PR to npm"
     runs-on: ubuntu-latest
     # comment out if:false to enable release PR to npm
-    if: false
+    # if: false
```
arboleya: I guess it slipped through.

arboleya (changes requested):
Did a first quick pass and left some notes.

Nicely done! 🙌 

Two more questions:
 1. Should we add docs around this?
 2. At first sight, it felt a bit off to have `fixtures` and `test` dir inside the `accounts/src/connectors/*` as I'd expect it to be in the `accounts` package root, like `accounts/test/connectors` or something. Have you considered this option?

.github/workflows/pr-release.yaml
arboleya: My tab was probably outdated. https://github.com/FuelLabs/fuels-ts/pull/1699/commits/3344c7d7a769177dc3a65f702f9ac0cb0719bdec

luizstacio (approved):

arboleya (approved):

PR #1702, opened 2024-02-01
fix: remove problematic validation on `estimateTxDependencies`

Torres-ssf:
This PR intends to remove the problematic `if` statement: https://github.com/FuelLabs/fuels-ts/blob/d261154911ba59bc776fe35d3ff7bb3c08ab8053/packages/providers/src/provider.ts#L732-L738

After running some validations, I could validate that sometimes the code does not enter the `if` block even though the instance of `txRequest` is a `ScriptTransactionRequest`.


nedsalk (approved):

packages/providers/src/provider.ts
```diff
@@ -729,13 +730,11 @@ export default class Provider {
         return;
       }
 
-      if (txRequest instanceof ScriptTransactionRequest) {
-        txRequest.addVariableOutputs(missingOutputVariableCount);
+      txRequest.addVariableOutputs(missingOutputVariableCount);
```
arboleya, line 733: Was this conditional even necessary? Why was it in place?

I wonder if `txRequest` can be anything other than `ScriptTransactionRequest`.

Torres-ssf: @arboleya I was the one who added this `if` statement in the past. 

I did this because back then TS was flagging an error saying that, at this point, this `txRequest` could be an instance of `CreateTransactionRequest` even though [this code](https://github.com/FuelLabs/fuels-ts/blob/fdb929d515f04cb0faa959852bacda5b867125a5/packages/providers/src/provider.ts#L707-L709) was already in place. 

But now TS does not flag this as a type error anymore.

arboleya (approved):

danielbate (approved):

PR #1703, opened 2024-02-01
feat: add `@fuel-ts/utils/assets-utils` package

### LINKED ISSUE #1541: Export Asset/Networks Config Interface
Provide an interface and a JSON with Network/Assets metadata.

 - Could the JSON `live` in a CDN?
 - Should we have a `fetch` utility?

This is where it is today:
 - https://github.com/FuelLabs/fuels-npm-packs/tree/main/packages/assets

The idea is not to migrate it fully but only provide types and utilities around it.

Dhaiwat10:
Closes #1541


packages/utils/package.json
```diff
@@ -24,6 +24,11 @@
       "require": "./dist/cli-utils.js",
       "import": "./dist/cli-utils.mjs",
       "types": "./dist/cli-utils.d.ts"
+    },
+    "./assets-utils": {
+      "require": "./dist/assets-utils.js",
+      "import": "./dist/assets-utils.mjs",
+      "types": "./dist/assets-utils.d.ts"
```
arboleya, line 31: I wonder if it should be inside the new `account` package, following the `Provider` route:
 - #1675

> _cc @Torres-ssf_

packages/utils/src/assets-utils/index.ts
```diff
@@ -0,0 +1,11 @@
...
+import { resolveIconPath } from './utils/resolveIconPath';
+
+export * from './constants';
+export * from './utils/network';
+export * from './utils/resolveIconPath';
+export * from './types';
+
+export const assets = assetList;
+
+export default resolveIconPath(process.env.BASE_URL || './', assetList);
```
arboleya, line 11: I usually dislike `export default` since it can result in inconsistent nomenclatures.

.changeset/dry-paws-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/utils": minor
+---
+
+add @fuel-ts/utils/assets-utils package
```
arboleya, line 5: It looks more like a directory/script. 🤔

packages/utils/src/assets-utils/README.md
```diff
@@ -0,0 +1,102 @@
...
+`brigde` informations for those assets that support it.
+
+### Install
+
+`npm install @fuel-ts/utils/assets-utils`
+
+### Usage
+
+```ts
+import assets from "@fuel-ts/utils/assets-utils";
```
arboleya, line 15: We might not need an entry point, even _if_ we don't move `assets` inside `account`.

```ts
import { assets } from "@fuel-ts/utils";
```

packages/utils/src/assets-utils/index.ts
Dhaiwat10: Agreed, and I don't think we use it anywhere at all. I kept it there because for now I just copy pasted all the source from the old repo, but happy to make this change

.changeset/dry-paws-cry.md
Dhaiwat10: Hmm.. thought of it as a package/mini-package since I put it under a separate export path. 

Directory makes sense tho lol!

packages/utils/src/assets-utils/README.md
Dhaiwat10: Cool, will remove it

packages/utils/package.json
Dhaiwat10: @arboleya Makes sense. Do you believe I should wait for that PR to be merged, or can I work on this in parallel?

arboleya: There are things you could do ahead of time, or you could also branch off of that branch/PR.

That's entirely up to you.

.changeset/dry-paws-cry.md
arboleya: After you move it around, the description may naturally morph with it.

packages/utils/package.json
Torres-ssf: @arboleya I agree, this can be part of the `account` package.

arboleya: Maybe it could even go inside `providers`:
 - `accounts/src/providers/chains.ts`
 - `accounts/src/providers/assets/index.ts`
 - `accounts/src/providers/assets/icons/..`

LuizAsFight: why inside `providers` ? assets has nothing to do with network/account... they can be even from other chain

I may lack some context here but `@fuel-ts/assets` feels good to me

LuizAsFight (changes requested):
friendly reminder to add `index` file under `utils` directory, just like this pr -> https://github.com/FuelLabs/fuels-npm-packs/pull/41/files

packages/utils/package.json
arboleya: I agree; however, over-categorizing small chunks of code led us to have dozens of packages flattened out in the repo root without an apparent relationship between them, which created a maintenance burden we are trying to reduce by re-grouping things in [this PR](https://github.com/FuelLabs/fuels-ts/pull/1675) as an attempt to defrag the list of packages. Things might not look perfect and are subject to change, but we must start somewhere, and flexibility is required to get things moving.

Nevertheless, everything will still be available from `fuels`.

Dhaiwat10:
Closing this in favor of #1730. The new PR is based off of #1675

PR #1704, opened 2024-02-02
fix: don't compare bn with integer

nedsalk:
This is a quick fix based on [this comment](https://github.com/FuelLabs/fuels-ts/pull/1603#issuecomment-1881210134).


.changeset/grumpy-countries-perform.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/providers": minor
```
arboleya, line 2:
```suggestion
"@fuel-ts/providers": patch
```

arboleya (changes requested):

.changeset/grumpy-countries-perform.md
nedsalk: https://edwardvilga.com/wp-content/uploads/2020/11/facepalm-head-768x432.jpg

packages/providers/src/coin-quantity.ts
```diff
@@ -27,7 +27,7 @@ export const coinQuantityfy = (coinQuantityLike: CoinQuantityLike): CoinQuantity
   const bnAmount = bn(amount);
   return {
     assetId: hexlify(assetId),
-    amount: bnAmount.toNumber() < 1 ? bn(1) : bnAmount,
+    amount: bnAmount.isNeg() || bnAmount.isZero() ? bn(1) : bnAmount,
```
danielbate, line 30:
```suggestion
    amount: bnAmount.lt(1) ? bn(1) : bnAmount,
```

I can appreciate that based off [this comment](https://github.com/FuelLabs/fuels-ts/pull/1603#issuecomment-1881210134), `BN` ignores decimals so we can just check for negative / zero, however this is not obvious from the code. Wondering if using the built in `lt` function would be more succinct? Giving us both a `bn` comparison and the previous less than 1 use case.

danielbate (approved):

arboleya (approved):

PR #1707, opened 2024-02-02
chore: upgrading `forc` to `0.49.2`

arboleya:


scripts/forc-check.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 main_dir=$(pwd)
-forc_tomls=$(find . -type f -name "Forc.toml")
+forc_tomls=$(find . -type f -name "Forc.toml  -path './packages/forc/sway-repo' -prune")
```
nedsalk, line 4: Could you explain this change? Adding `echo $forc_tomls` with this change present gives back no `Forc.toml`s to check in the script.

scripts/forc-format.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 main_dir=$(pwd)
-forc_tomls=$(find . -type f -name "Forc.toml")
+forc_tomls=$(find . -type f -name "Forc.toml -path './packages/forc/sway-repo' -prune")
```
nedsalk, line 4: Same as [above](https://github.com/FuelLabs/fuels-ts/pull/1707/files#r1476795632).

Dhaiwat10 (dismissed):

scripts/forc-check.sh
arboleya: Good catch! b2e45e5f19282b9b83b342c7274c89f92e052398

scripts/forc-format.sh
arboleya: Good catch! b2e45e5f19282b9b83b342c7274c89f92e052398

scripts/forc-check.sh
nedsalk: It might be good to somehow verify that the scripts are actually running on a list of `Forc.toml`s and aren't passing the CI just because no `Forc.toml`s were found.

nedsalk (approved):

Dhaiwat10 (approved):

Issue #1706, opened 2024-02-02
Ensure dry runs request only necessary data


arboleya:
To lower bandwidth, we should ensure dry runs query only necessary data from GQL API, as they probably don't need all fields.

For example, here is one receipt payload on the sway marketplace:

![fieds](https://github.com/FuelLabs/fuels-ts/assets/26660/ef3161e6-8224-49ce-ab06-0bb517ffc9d9)

arboleya:
One idea is to add a receipt filter to the query args to just return a specific log or script.

This would require changes on `fuel-core`.

Voxelot:
There's also other data that's likely irrelevant to the ts sdk when fetching consensus params:

### latest block tx ids
![image](https://github.com/FuelLabs/fuels-ts/assets/794823/9ebba3e6-edd6-4945-b936-94182a45b2db)


### gas costs
![image](https://github.com/FuelLabs/fuels-ts/assets/794823/5e3bf615-7a37-49a5-9b78-afb52f530da8)

Voxelot:
We also don't need salt and the **full contract bytecode** on every receipt from a contract:

![image](https://github.com/FuelLabs/fuels-ts/assets/794823/e547330c-216c-448f-afe4-4381c68d0a80)

If a contract has multiple receipts, and the full contract bytecode gets logged on every receipt from that contract, the response is going to be super huge!

Issue #1697, opened 2024-01-31
Getting invalid `gasLimit` when using `getTransactionCost`


arboleya:
Incoming from this forum post:
 - https://forum.fuel.network/t/issue-with-gettransactioncost-returning-incorrect-gas-limit-for-contract-function-calls/4087
 
 [EDIT]
 It seems the same error is happening [here](https://github.com/FuelLabs/fuels-ts/pull/1708#discussion_r1479021618)

PR #1708, opened 2024-02-02
feat: automatically set defaults for `txParams` if not specified

### LINKED ISSUE #1696: Use dynamically fetched defaults for `txParams`
The idea is to improve DX by making things like gas price/limit optional and using smart defaults from the network.

Context:
 - https://github.com/FuelLabs/fuels-ts/discussions/1571

Here is the Rust PR, to serve as a reference:
 - https://github.com/FuelLabs/fuels-rs/pull/1121

Dhaiwat10:
## Summary
This PR makes sure that the SDK will now automatically set sensible defaults for `gasPrice` and `gasLimit` transaction parameters in case they are not specified by the user.

```typescript
// earlier, you would have always have to specify `txParams`. If you didn't, the transaction would fail.
await contract.functions.test_function().txParams({ gasPrice: 1, gasLimit: 10_000 }).call();

// now, you don't need to specify these values if you don't want to. This code works perfectly fine:
await contract.functions.test_function().call();
``` 

Closes #1696


Dhaiwat10:
Thinking of adding a test and some docs so marking this is a draft momentarily

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -292,8 +293,19 @@ export class BaseInvocationScope<TReturn = any> {
     assert(this.program.account, 'Wallet is required!');
 
     const transactionRequest = await this.getTransactionRequest();
+    const { maxFee, gasUsed, minGas } = await this.getTransactionCost();
```
arboleya, line 296: 1. Will this add an extra request to the node?
 1. Which data is asked?

Please note the problems we are addressing on this parallel quest:
 - https://github.com/FuelLabs/fuels-ts/issues/1706

We need to be very careful.

Dhaiwat10: @arboleya No, this won't add any _new_ requests. It just uses more data from the request we are already making as of now.

Dhaiwat10: @arboleya just confirmed. All good. No new calls to the client

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -292,8 +293,20 @@ export class BaseInvocationScope<TReturn = any> {
...
+
+    // set defaults for gasLimit and gasPrice if not specified
+    if (!specifiedGasLimit) {
+      transactionRequest.gasLimit = gasUsed;
+    }
 
-    const { maxFee } = await this.getTransactionCost();
+    if (!specifiedGasPrice) {
+      transactionRequest.gasPrice = minGasPrice;
+    }
```
arboleya, line 309: This whole block seems repeated 3x times in this file.

Should we DRY it by encapsulating it in a new function or something?

Dhaiwat10: Yessir: https://github.com/FuelLabs/fuels-ts/pull/1708/commits/8f95d7b7670bb2c16cd01da1486029dfc45697ec

danielbate (commented):
Should we remove the redundant `txParam` calls in tests?

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -388,4 +392,25 @@ export class BaseInvocationScope<TReturn = any> {
     const transactionRequest = await this.getTransactionRequest();
     return transactionRequest.getTransactionId(chainIdToHash);
   }
+
+  /**
+   * In case the gasLimit and gasPrice are *not* set by the user, this method sets some default values.
+   */
+  async setDefaultTxParams() {
```
nedsalk, line 399:
```suggestion
  private async setDefaultTxParams() {
```
We can make this private as it's an internal detail.

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -292,9 +293,10 @@ export class BaseInvocationScope<TReturn = any> {
     assert(this.program.account, 'Wallet is required!');
 
     const transactionRequest = await this.getTransactionRequest();
-
     const { maxFee } = await this.getTransactionCost();
 
+    await this.setDefaultTxParams();
+
     await this.fundWithRequiredCoins(maxFee);
```
nedsalk, line 301:
```suggestion
    await this.setDefaultTxParams();

    const { maxFee } = await this.getTransactionCost();
    await this.fundWithRequiredCoins(maxFee);

    const transactionRequest = await this.getTransactionRequest();
```

At first read I had trouble understanding the flow of logic because `transactionRequest` was initialized and only used later down the line while some additional transaction modifications were happening with calls to `this.setDefaultTxParams` and `this.fundWithRequiredCoins`. The catch is that `transactionRequest` is a pointer to `this.transactionRequest`. I think that the suggesed order of logic is more natural: modify state of transaction, get `transactionRequest`, send it.

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -352,9 +355,10 @@ export class BaseInvocationScope<TReturn = any> {
     const provider = this.getProvider();
 
     const transactionRequest = await this.getTransactionRequest();
-
     const { maxFee } = await this.getTransactionCost();
 
+    await this.setDefaultTxParams();
+
     await this.fundWithRequiredCoins(maxFee);
```
nedsalk, line 362:
```suggestion
    await this.setDefaultTxParams();

    const { maxFee } = await this.getTransactionCost();
    await this.fundWithRequiredCoins(maxFee);

    const transactionRequest = await this.getTransactionRequest();
```

Same as in `call` suggestion.

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -331,9 +333,10 @@ export class BaseInvocationScope<TReturn = any> {
     }
 
     const transactionRequest = await this.getTransactionRequest();
-
     const { maxFee } = await this.getTransactionCost();
 
+    await this.setDefaultTxParams();
+
     await this.fundWithRequiredCoins(maxFee);
```
nedsalk, line 340:
```suggestion
    await this.setDefaultTxParams();

    const { maxFee } = await this.getTransactionCost();
    await this.fundWithRequiredCoins(maxFee);

    const transactionRequest = await this.getTransactionRequest();
```

Same as in `call` suggestion.

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -388,4 +392,25 @@ export class BaseInvocationScope<TReturn = any> {
...
+    const specifiedGasPrice = this.txParameters?.gasPrice;
+
+    // set defaults for gasLimit and gasPrice if not specified
+    if (!specifiedGasLimit) {
+      transactionRequest.gasLimit = gasUsed;
+    }
+
+    if (!specifiedGasPrice) {
+      transactionRequest.gasPrice = minGasPrice;
+    }
```
nedsalk, line 414:
```suggestion
    const transactionRequest = await this.getTransactionRequest();

    const gasLimitSpecified = this.txParameters?.gasLimit || this.hasCallParamsGasLimit;
    const gasPriceSpecified = this.txParameters?.gasPrice;

    // set defaults for gasLimit and gasPrice if not specified
    if (!gasLimitSpecified) {
      const { gasUsed } = await this.getTransactionCost();
      transactionRequest.gasLimit = gasUsed;
    }

    if (!gasPriceSpecified) {
      const { minGasPrice } = this.getProvider().getGasConfig();
      transactionRequest.gasPrice = minGasPrice;
    }
```
Two things:
- Rename `specifiedGasLimit`/`specifiedGasPrice` to `gasLimitSpecified`/`gasPriceSpecified`. The initial naming sounds like it's holding a numerical value while the suggested naming implies a boolean.
- Move `gasUsed` and `minGasPrice` into the `if` blocks where they're used.

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -388,4 +392,25 @@ export class BaseInvocationScope<TReturn = any> {
...
+   * In case the gasLimit and gasPrice are *not* set by the user, this method sets some default values.
+   */
+  async setDefaultTxParams() {
+    const transactionRequest = await this.getTransactionRequest();
+    const { minGasPrice } = this.getProvider().getGasConfig();
+    const { gasUsed } = await this.getTransactionCost();
+
+    const specifiedGasLimit = this.txParameters?.gasLimit || this.hasCallParamsGasLimit;
+    const specifiedGasPrice = this.txParameters?.gasPrice;
+
```
Torres-ssf, line 406: Why are we calling `getTransactionCost` again here? This will add at least 1 additional `dryRun` and `estimatePredicates` call to the client. 

This function ( `getTransactionCost` ) is already called by the following `BaseInvocationScope` methods: `call`, `dryRun`, and `simulate`. 

We also already called `getTransactionRequest` ( which will among other things create the script code for the contract call ) within these 3 methods from `BaseInvocationScope`.

What we can do is set `setDefaultTxParams` to be private so it won't be called out of context, and then make it receive `transactionRequest`, `gasLimit`, and `gasPrice` params.

Please let me know if am missing something here.

Torres-ssf (changes requested):
I believe we should also support default values on `Account` transfer methods. 

Since we are already updating the Docs to let users aware that they do not need to inform `gasLimit` and `gasPrice` on contract calls, they might infer that the same approach is valid when `wallet.transfer...` or `wallet.transferToContract...`

packages/program/src/functions/base-invocation-scope.ts
Dhaiwat10: @Torres-ssf good catch about the account methods. Might have to do this for `deployContract` too. Anything else that comes to mind?

packages/fuel-gauge/src/auth-testing.test.ts
```diff
@@ -33,29 +33,23 @@ describe('Auth Testing', () => {
...
+    const { value } = await contractInstance.functions.is_caller_external().call();
 
     expect(value).toBeTruthy();
   });
 
   it('can check_msg_sender [with correct id]', async () => {
     const { value } = await contractInstance.functions
       .check_msg_sender({ value: wallet.address.toB256() })
-      .txParams({ gasPrice, gasLimit: 10_000 })
+      .txParams({ gasLimit: 500 })
```
Dhaiwat10, line 44: I was able to remove `txParams` from all of the tests, barring this one. 

For this call, the `gasUsed` returned by the dry-run (452) was less than the required gas (465). I tried and tried, but I wasn't able to find out why this was the case. I've hardcoded a limit here for now.

Torres-ssf: I've tried this locally and it seems this happening because the `dryRun` called by `getTransactionCost` is reverting due `FAILED_ASSERT_SIGNAL ( AssertFailed )`. 

Strangely enough, this error does not happen when submitting the real transaction. Therefore, since the TX does not revert, it ends up using more gas, so the gas estimation made with the `dryRun` is not accurate.

Not sure why this TX is reverting to a `dryRun` but not when it is submitted.

Most likely this is the same error that is happening [here](https://github.com/FuelLabs/fuels-ts/issues/1697)

Dhaiwat10: Could this be an issue on the client?

Torres-ssf: @Dhaiwat10 no. This is something wrong at our side.

arboleya: @Dhaiwat10 Can we have a TODO with a link to the issue above?

arboleya (approved):
LGTM ✅

Torres-ssf (approved):

PR #1710, opened 2024-02-02
chore: filter out receipt properties on dryrun

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/1706


.changeset/long-suns-float.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/providers": minor
```
arboleya, line 2: Is this a breaking change?

packages/providers/src/provider.ts
```diff
@@ -649,7 +650,7 @@ export default class Provider {
       encodedTransaction,
       utxoValidation: utxoValidation || false,
     });
-    const receipts = gqlReceipts.map(processGqlReceipt);
+    const receipts = (<GqlReceipt[]>gqlReceipts).map(processGqlReceipt);
```
arboleya, line 653: Why's this casting necessary?

packages/providers/test/fixtures/receipts.ts
```diff
@@ -1,7 +1,7 @@
 import { GqlReceiptType } from '../../src/__generated__/operations';
-import type { GqlReceiptFragmentFragment } from '../../src/__generated__/operations';
+import type { GqlReceipt } from '../../src/__generated__/operations';
```
arboleya, line 2: Why `GqlReceipt` instead of `GqlReceiptDryRunFragmentFragment`, following the previous implementation?

```suggestion
import type { GqlReceiptDryRunFragmentFragment } from '../../src/__generated__/operations';
```

packages/providers/src/operations.graphql
```diff
@@ -66,8 +62,6 @@ fragment receiptFragment on Receipt {
   reason
   ra
   rb
-  rc
-  rd
```
arboleya, line 70: I think you should keep `rc` and `rd` because of they're used in `assembleReceiptByType()`:
https://github.com/FuelLabs/fuels-ts/blob/cfb06e88dba97792b7bc5f4e048c9b9373fb3c03/packages/providers/src/utils/receipts.ts#L146-L147

arboleya (commented):
Good stuff! Thank you. 🙏 

By removing all the `GqlReceipt<>` castings, I could find a few typings mismatches between some `Fragment` types.

I am probably missing something, so I created this PR to investigate my suspicion; it covers all the comments I left on this PR. Please take a look.
 - https://github.com/FuelLabs/fuels-ts/pull/1712

packages/providers/src/operations.graphql
Torres-ssf: @arboleya Yeah, but this is expected because `assembleRceiptByType` was made to take into account all properties existent on receipts. 

But since these 2 values were not being used anywhere else I presumed that they could be removed.

But I reckon that me most critical ones were the contract bytecodes.

.changeset/long-suns-float.md
Torres-ssf: I don't think so because your fix took into uses the proper Receipts type. Also, all receipts properties are still included into the other queries responses.

arboleya: Should this be a `patch` then?

Torres-ssf: Makes sense

packages/providers/src/operations.graphql
arboleya: Removing these props from the GQL may be ok, but we must also purge all remaining usages.

packages/providers/src/transaction-summary/get-transaction-summary.ts
```diff
@@ -6,6 +6,7 @@ import { getBytesCopy } from 'ethers';
 import type {
   GqlGetTransactionsByOwnerQueryVariables,
   GqlPageInfo,
+  GqlReceipt,
```
arboleya, line 9:
```suggestion
```

packages/providers/src/transaction-summary/get-transaction-summary.ts
```diff
@@ -38,12 +39,11 @@ export async function getTransactionSummary<TTransactionType = void>(
...
-  const [decodedTransaction] = new TransactionCoder().decode(
-    getBytesCopy(gqlTransaction.rawPayload),
-    0
-  );
+  const { receipts: gqlReceipts, rawPayload, status } = gqlTransaction;
 
-  const receipts = gqlTransaction.receipts?.map(processGqlReceipt) || [];
+  const [decodedTransaction] = new TransactionCoder().decode(getBytesCopy(rawPayload), 0);
+
+  const receipts = (<GqlReceipt[]>gqlReceipts)?.map(processGqlReceipt) || [];
```
arboleya, line 46:
```suggestion
  const receipts = gqlReceipts?.map(processGqlReceipt);
```

packages/providers/src/transaction-summary/get-transaction-summary.ts
```diff
@@ -134,7 +134,7 @@ export async function getTransactionsSummaries(
 
     const [decodedTransaction] = new TransactionCoder().decode(getBytesCopy(rawPayload), 0);
 
-    const receipts = gqlReceipts?.map(processGqlReceipt) || [];
+    const receipts = (<GqlReceipt[]>gqlReceipts)?.map(processGqlReceipt) || [];
```
arboleya, line 137:
```suggestion
    const receipts = gqlReceipts?.map(processGqlReceipt) || [];
```

arboleya (changes requested):

packages/providers/src/transaction-summary/get-transaction-summary.ts
arboleya: https://github.com/FuelLabs/fuels-ts/pull/1710/commits/7d5eed211525bcdc0d34fc584ed60faed148b155

packages/providers/src/transaction-summary/get-transaction-summary.ts
arboleya: https://github.com/FuelLabs/fuels-ts/pull/1710/commits/7d5eed211525bcdc0d34fc584ed60faed148b155

packages/providers/src/transaction-summary/get-transaction-summary.ts
arboleya: https://github.com/FuelLabs/fuels-ts/pull/1710/commits/7d5eed211525bcdc0d34fc584ed60faed148b155

Dhaiwat10 (approved):
🔥🔥

packages/providers/src/operations.graphql
```diff
@@ -80,6 +76,34 @@ fragment receiptFragment on Receipt {
...
+  param1
+  param2
+  val
+  len
+  receiptType
+  result
+  gasUsed
+  data
+  recipient
+  contractId
```
nedsalk, line 104: This seems to be redundant given that `contract { id }` is defined above.

arboleya (commented):
@Torres-ssf Even after my suggestions, this is still a bit off. The dry-run queries will return less data than a few utility methods will require down the road.

I don't think it's worth the risk for this release, so I created another PR here:
 - https://github.com/FuelLabs/fuels-ts/pull/1715

This is safer because the removed properties are not used anywhere by any other type. I also fixed a few fragment types, and everything looks okay.

I understand we still have more properties to trim, but I'd suggest doing those in separate PRs with more time to double-check everything.

Torres-ssf:
Closing this if favor of https://github.com/FuelLabs/fuels-ts/pull/1715

PR #1713, opened 2024-02-03
chore: filter out TX props on `estimatePredicates` query

Torres-ssf:
This PR filters out most of the queried properties for the `estimatePredicates` query. 

Since we just need `predicateGasUsed` from each predicate input, there is no reason to query all TX properties.


nedsalk (approved):

danielbate (approved):

packages/fuel-gauge/src/predicate/predicate-estimations.test.ts
```diff
@@ -74,14 +79,48 @@ describe('Predicate', () => {
...
       const predicateCoinInput = <MessageTransactionRequestInput>txEstimated.inputs[0];
       expect(Number(predicateCoinInput.predicateGasUsed)).toBeGreaterThan(1);
       const predicateMessageInput = <MessageTransactionRequestInput>txEstimated.inputs[1];
       expect(Number(predicateMessageInput.predicateGasUsed)).toBeGreaterThan(1);
+      const nonPredicateMessageInput = <MessageTransactionRequestInput>txEstimated.inputs[2];
+      expect(nonPredicateMessageInput.predicateGasUsed).toBeUndefined();
+      const contractInput = <ContractTransactionRequestInput>txEstimated.inputs[3];
+      expect(contractInput.contractId).toBe(contract.contractId);
+      const nonPredicateInput = <CoinTransactionRequestInput>txEstimated.inputs[4];
+      expect(nonPredicateInput.predicateGasUsed).toBeUndefined();
```
arboleya, line 123: I haven't reviewed this yet, but I want to point out the same concerns I posted [here](https://github.com/FuelLabs/fuels-ts/pull/1710#pullrequestreview-1862291286).

In short, when we change GQL fragments, we need to be sure that all underlying related types and method utilities are updated accordingly, as selectively hard-casting things can result in incomplete data being shuffled around.

PR #1719, opened 2024-02-05
chore: remove turbo cache inputs for `docs`

### LINKED ISSUE #1717: Verify flaky `CI Setup` workflow
Sometimes, it seems to lose the cache or something like that:
 - https://github.com/FuelLabs/fuels-ts/actions/runs/7787432043/job/21234957254

danielbate:
Closes #1717 

Our build stage is dependent on auto generated code. Therefore if a cache is missed in `docs`, but not during other packages, it will fail as it attempts to reference auto generated code. 

We should treat the `docs` cache the same we do as other packages to ensure they are in sync and do not cause CI failures.


danielbate:
@Torres-ssf what was the reason for the separate Turbo config, and not inheriting the base? Or the reason for the inputs specifically?

apps/docs/turbo.json
```diff
@@ -4,8 +4,7 @@
   "pipeline": {
     "build": {
       "dependsOn": ["^build", "prebuild"],
-      "outputs": ["dist/**"],
-      "inputs": ["src/**", ".vitepress/**", "types/**"]
```
arboleya, line 8: We must invalidate the cache when things inside the `.vitepress` or `types` directory change. Otherwise, re-building the docs after changing something inside these directories won't do anything. (cache-hit)

> cc @Torres-ssf

danielbate: Good point. I'm going to close this and revisit. The more complete fix is to probably move the auto generated content outside of `src` and into `.typedoc`.

danielbate:
Closing to rethink approach.

Issue #1722, opened 2024-02-06
Remove `Provider` from `WalletManager` types


Torres-ssf:
During https://github.com/FuelLabs/fuels-ts/pull/1181 the provider became required to instantiate wallets. 

As a consequence, the provider was added to all `WalletManager` specific types because wallets could not be generated without a Provider instance.

Since we have [recently changed this](https://github.com/FuelLabs/fuels-ts/issues/1539), it makes sense now to remove the provider from all `WalletManager` specific types, as it has no use for them.

PR #1723, opened 2024-02-06
chore!: remove `Provider` from `WalletManager` types

Torres-ssf:
- Closes: https://github.com/FuelLabs/fuels-ts/issues/1722


arboleya (dismissed):
🕵️‍♂️

packages/wallet-manager/src/vaults/mnemonic-vault.ts
```diff
@@ -56,7 +51,7 @@ export class MnemonicVault implements Vault<MnemonicVaultOptions> {
     do {
       const wallet = Wallet.fromMnemonic(
         this.#secret,
-        this.provider,
+        undefined,
         this.getDerivePath(numberOfAccounts)
```
luizstacio, line 55: Maybe we should change the order of params on `fromMnemonic`?

luizstacio (changes requested):

Dhaiwat10 (approved):

packages/wallet-manager/src/vaults/mnemonic-vault.ts
arboleya: Good catch! 🎯

arboleya (approved):

danielbate (approved):

luizstacio (approved):

Issue #1397, opened 2023-11-03
Add CDN Availability


arboleya:
We're looking at having it added here:
- https://cdnjs.com/libraries?q=fuels

Then, we need to improve our installation docs — here's an example:
- https://docs.ethers.org/v5/getting-started/#getting-started--importing--web-browser

Ideally, it would be a new step in your CI pipeline.

arboleya:
Just registering what we spoke about—we may have a new step to generate an ESM/browser build.

This could be something using Rollup or Vite, which runs on a `postbuild` step on the `fuels` package.

Torres-ssf:
Blocked by: https://github.com/FuelLabs/fuels-ts/issues/1609

arboleya:
FYI — I've moved this out of `salamander`.

arboleya:
@Torres-ssf Blocking PR has been [merged](https://github.com/cdnjs/packages/pull/1743). 🎉 

Before closing this, we need to mention the CDN option in the docs.

PR #1724, opened 2024-02-06
docs: add `CDNJS` example

Torres-ssf:
- Closes: https://github.com/FuelLabs/fuels-ts/issues/1397


apps/docs/src/getting-started.md
```diff
@@ -62,6 +67,28 @@ function App() {
 export default App;
 ```
 
+## Browser
+
+For a quick test or just playing around, you can load it in your Web Apps straight from our CDN.
```
danielbate, line 72: Is it worth noting here that this is only supported from `0.73.0`?

danielbate (approved):

apps/docs/src/getting-started.md
arboleya: I thought the same, but the version is self-contained in the snippet URL.


version

apps/docs/src/getting-started.md
```diff
@@ -62,6 +67,28 @@ function App() {
 export default App;
 ```
 
+## Browser
```
arboleya, line 70:
```suggestion
## CDN Usage (browser only)
```

arboleya (commented):
We should have provided a minified version, but that can be done in another PR.
 - https://github.com/FuelLabs/fuels-ts/issues/1728

danielbate (approved):

PR #1731, opened 2024-02-09
chore: remove additional dryrun call

Torres-ssf:
This PR intends to remove a `dryRun` [call](https://github.com/FuelLabs/fuels-ts/blob/53babd507ac4ed9c775237cb6500df8e20409cd7/packages/providers/src/provider.ts#L586-L605) that happens before sending the transaction that has the only purpose of validating if the `gasLimit` or `gasPrice` is enough for the transaction to be processed.

Since these 2 validations can be done without an additional dryRun call we can safely remove it.


packages/providers/src/provider.ts
```diff
@@ -583,27 +583,6 @@ export default class Provider {
     }
     // #endregion Provider-sendTransaction
 
-    const { gasUsed, minGasPrice } = await this.getTransactionCost(transactionRequest, [], {
-      estimateTxDependencies: false,
-      estimatePredicates: false,
-    });
-
-    if (bn(minGasPrice).gt(bn(transactionRequest.gasPrice))) {
```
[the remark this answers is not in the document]
Torres-ssf: @nedsalk I've also thought about adding more private properties to the `TransactionRequest` classes to keep track of operations that happened on the TX request. I've been thinking about doing so for quite a while now. I am letting this idea cook for a little bit more.

The flow of manually assembling a `ScriptTransactionRequest` involves handling all required steps manually. This includes executing `getTrasactionCost` beforehand to have a value that will be used to later call `Accoount.fund` to fund the transaction. If this is not explicitly said in our docs, it needs to be.

I agree that improving the error message that will come from the node ( `OutOfGas`, something like this if I am not mistaken ) should be part of this PR.

arboleya (commented):
Conflicts to fix.

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -410,12 +403,24 @@ export class BaseInvocationScope<TReturn = any> {
...
 
+    const { gasLimit, gasPrice } = transactionRequest;
+
     if (!gasLimitSpecified) {
       transactionRequest.gasLimit = gasUsed;
+    } else if (gasLimit.lt(gasUsed)) {
+      throw new FuelError(
+        ErrorCode.GAS_LIMIT_TOO_LOW,
+        `Gas limit '${gasLimit}' is lower than the required: '${gasUsed}'.`
+      );
```
arboleya, line 414: I suppose it'd create more problems than solutions if we tried reusing the `validateGas` helper, right?

https://github.com/FuelLabs/fuels-ts/blob/572e85164b5c98843541e0015f077b4f11a7ba9b/packages/account/src/account.ts#L529-L534

arboleya (approved):

packages/program/src/functions/base-invocation-scope.ts
Torres-ssf: @arboleya Yeah, I've had the same though. 

We could create a helper function to validate and throw this error and place it inside the `errors` package. It makes more sense to exist there since this error is being thrown in more than one package. 

But in the end, I thought this idea was overengineering.

nedsalk (approved):
Approving this after sync with @Torres-ssf around [this comment](https://github.com/FuelLabs/fuels-ts/pull/1731#discussion_r1484161509). We'll create a separate issue around it after we flesh it out.

PR #1737, opened 2024-02-09
fix: resolve transpilation issue with `BN`

### LINKED ISSUE #1736: Transpilation issue with `BN`
Dependent on a users build config, they may be facing transpilation issues with `BN`. @K1-R1 faced this recently when working with the [EVM Wallet Connector](https://github.com/FuelLabs/EVM-Wallet-Connector). However this could impact others with a similar CRA setup. The following error is reported.

```sh
cra-dapp:build: SyntaxError: /Users/daniel/Documents/Development/Fuel/EVM-Wallet-Connector/packages/wallet-connector-evm/dist/wallet-connector-evm.js: When using '@babel/plugin-transform-parameters', it's not possible to compile `super()` in an arrow function with default or rest parameters without compiling classes.
cra-dapp:build: Please add '@babel/plugin-transform-classes' to your Babel configuration.
cra-dapp:build:   6268 | var Le = class extends Hi {
cra-dapp:build:   6269 |   constructor(t, n, r) {
cra-dapp:build: > 6270 |     var e = (...jx) => (super(...jx), U(this, "MAX_U64", "0xFFFFFFFFFFFFFFFF"), this);
cra-dapp:build:        |                         ^^^^^^^^^^^^
cra-dapp:build:   6271 |     if (Le.isBN(t)) {
cra-dapp:build:   6272 |       e(t.toArray(), n, r);
cra-dapp:build:   6273 |       return;
```

As the error suggests, this can be resolved using the above babel configuration, however I don't think we should require users to alter their build config to use the SDK.

The error is the result of the `BN::constructor`:

https://github.com/FuelLabs/fuels-ts/blob/53babd507ac4ed9c775237cb6500df8e20409cd7/packages/math/src/bn.ts#L48-L60

This transpiles to the following:

```js
  constructor(t, n, r) {
    var e = (...jx) => (super(...jx), U(this, "MAX_U64", "0xFFFFFFFFFFFFFFFF"), this);
    if (Le.isBN(t)) {
      e(t.toArray(), n, r);
      return;
    }
    if (typeof t == "string" && t.slice(0, 2) === "0x") {
      e(t.substring(2), n || "hex", r);
      return;
    }
    e(t ?? 0, n, r);
  }
```

Calling `super` inside a conditional statement leads to an infinite loop. I found a [similar fix](https://github.com/pixijs/pixijs/pull/9093/files) in another SDK that should also work for us.

danielbate:
Closes #1736

### DIFF
--- a/.changeset/cyan-mice-suffer.md
+++ b/.changeset/cyan-mice-suffer.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/math": patch
+---
+
+Resolve BN transpilation issue by restructuring constructor.
--- a/.github/workflows/pr-release.yaml
+++ b/.github/workflows/pr-release.yaml
@@ -8,7 +8,7 @@ jobs:
     name: "Release PR to npm"
     runs-on: ubuntu-latest
     # comment out if:false to enable release PR to npm
-    # if: false
+    if: false
     permissions: write-all
     steps:
       - name: Checkout
--- a/.github/workflows/pr-release.yaml
+++ b/.github/workflows/pr-release.yaml
@@ -8,7 +8,7 @@ jobs:
     name: "Release PR to npm"
     runs-on: ubuntu-latest
     # comment out if:false to enable release PR to npm
-    if: false
+    # if: false
     permissions: write-all
     steps:
       - name: Checkout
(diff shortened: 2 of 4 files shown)


arboleya (approved):
Tiny and clean. 🎯🧘‍♂️

nedsalk (approved):

PR #1744, opened 2024-02-11
chore!: upgrade `forc` to `v0.50.0`

nedsalk:
Forc builds are now in `release` mode because the default `debug` sometimes generates bytecode that is too big to deploy.


packages/utils/src/test-utils/getForcProject.ts
```diff
@@ -7,22 +7,24 @@ import { normalizeString } from '../utils/normalizeString';
 interface IGetForcProjectParams {
   projectDir: string;
   projectName: string;
+  build: 'debug' | 'release';
```
nedsalk, line 10: I added this to allow support for both `debug` and `release` builds with this utility. This might be overengineering because for our purposes `release` is enough. The train of thought is that we're exporting this utility to our users so I wanted to not break their code and they might sometimes want to compile to debug as well.

danielbate (approved):
Should this be a breaking change due to the new behaviour of `getForcProject`? Alternatively we could alter the default behaviour so a `build` parameter isn't required?

nedsalk:
@danielbate 
> Should this be a breaking change due to the new behaviour of getForcProject? 

You're right, I'll rename the PR.

> Alternatively we could alter the default behaviour so a build parameter isn't required?

It'd still be a breaking change because the build results folder would change from `out/debug` to `out/release` anyways. The way I did it with the `build` parameter is just more explicit about it.

arboleya (changes requested):
Should this be updated to `release` too?

https://github.com/FuelLabs/fuels-ts/blob/0a52951a607bf119c411a42050cc2475b5d1efa9/turbo.json#L27

nedsalk:
> Should this be updated to `release` too?
> 
> https://github.com/FuelLabs/fuels-ts/blob/0a52951a607bf119c411a42050cc2475b5d1efa9/turbo.json#L27

https://github.com/FuelLabs/fuels-ts/pull/1744/commits/0547e9b10ef5aacbc863166817c98aab963dac53

sdankel (approved):

Dhaiwat10 (approved):

arboleya (approved):

PR #1768, opened 2024-02-18
docs: add a section for numbers

### LINKED ISSUE #1555: No section around numbers
We should include an entire section around how numbers are handling, particularly big and small numbers and what the proper way to handle them are.

> [!NOTE]

Dhaiwat10:
Closes #1555


arboleya (approved):

apps/docs/src/guide/types/numbers.md
```diff
@@ -0,0 +1,29 @@
...
+2. `u16` (16-bit unsigned integer)
+3. `u32` (32-bit unsigned integer)
+4. `u64` (64-bit unsigned integer)
+5. `u256` (256-bit unsigned integer)
+
+Conveniently, all of these types are handled the same way in the TS SDK. This guide explains how to create and interact with Sway numbers while using the SDK.
+
+## Creating Numbers
+
+Just like Ethereum, when you pass in a number to a Sway program from JavaScript, you must first convert it to a `BigNumber` object. Here's how you can do that:
```
danielbate, line 15:
```suggestion
Just like Ethereum, when you pass in a number to a Sway program from JavaScript, you must first convert it to a `BigNum` object. Here's how you can do that:
```

danielbate: Just ensuring that we are differentiating between [bn.js](https://github.com/indutny/bn.js/?tab=readme-ov-file) and [bignumber.js](https://mikemcl.github.io/bignumber.js/)

danielbate (commented):
Do we think it's a good idea to introduce a `BigInt` example also, as those working with `ethers` made need assistance with conversions? 

Additionally a sub paragraph on handling and differentiating big numbers I think would be beneficial. As you only need `bn` for `u64`.

Dhaiwat10:
> Do we think it's a good idea to introduce a BigInt example also, as those working with ethers made need assistance with conversions?

@danielbate as in, how does one convert a `BigInt` into a `BN` object? Or rather, how does one directly use an ethers-compatible BigInt with our SDK?

> Additionally a sub paragraph on handling and differentiating big numbers I think would be beneficial. As you only need bn for u64.

Interesting. Did not know that. So, for the smaller types - we can directly pass in JavaScript numbers?

apps/docs/src/guide/types/numbers.md
```diff
@@ -0,0 +1,29 @@
...
+2. `u16` (16-bit unsigned integer)
+3. `u32` (32-bit unsigned integer)
+4. `u64` (64-bit unsigned integer)
+5. `u256` (256-bit unsigned integer)
+
+Conveniently, all of these types are handled the same way in the TS SDK. This guide explains how to create and interact with Sway numbers while using the SDK.
+
+## Creating Numbers
+
+Just like Ethereum, when you pass in a number to a Sway program from JavaScript, you must first convert it to a `BigNum` object. Here's how you can do that:
```
arboleya, line 15: Everything in this PR mentions `BigNumber`, except here.

When we mention `BigNum`, are we referring to this? 🤔 
 - https://www.npmjs.com/package/bignum

Dhaiwat10: This is because I accepted @danielbate's suggestion for one occurence of the term but didn't update the rest.

Now that I think about it, originally, I used the term `BigNumber` as a 'general' term referring to big numbers, not a term specific to any libraries.

Maybe what we can do is 

1. Still use `BigNumber` everywhere but make it link to the implementation for `bn` in our codebase
and 2. Show people how to convert from a `BigInt` (or whatever the compatible BigNumber variation for `ethers` is) for usage with the TS SDK - _if_ required. Need to investigate if it is required.

Thoughts? @danielbate @arboleya

danielbate: This was a change I recommended. We should refer to it as `BigNum` through, as that it is what `bn.js` [refer to it as](https://github.com/indutny/bn.js/blob/master/README.md?plain=1#L3). IMO `BigNumber` is referring to [bignumber.js](https://www.npmjs.com/package/bignumber.js/v/4.0.4).

apps/docs/src/guide/types/numbers.md
```diff
@@ -0,0 +1,43 @@
+# Numbers
+
+In Sway, there are multiple types of numbers:
+
+1. `u8` (8-bit unsigned integer)
+2. `u16` (16-bit unsigned integer)
+3. `u32` (32-bit unsigned integer)
+4. `u64` (64-bit unsigned integer)
+5. `u256` (256-bit unsigned integer)
```
danielbate, line 9: Do we even support `u256`?

danielbate (approved):
Nice one Dhai 💪🏻  Just one thought, may be irrelevant so approving.

apps/docs/src/guide/types/numbers.md
Torres-ssf, line 9: Correct me if am wrong but I believe we do not support `u256`

Torres-ssf (changes requested):

Dhaiwat10:
@Torres-ssf @danielbate thanks for the heads-up, investigating u256 support

Dhaiwat10:
@danielbate @Torres-ssf you were right lads, we dont support `u256`. I've added a notice for it in the docs. 

Should I also create an issue to add support for this type? @arboleya

arboleya:
> Should I also create an issue to add support for this type? @arboleya

Hmm, good catch. Let's also check if the Rust SDK has it.

There may be some reason it has never been added; it's worth investigating.

But yes, unless there's some strong reason behind it, feel free to create an issue. 👍

apps/docs-snippets/src/guide/types/numbers.test.ts
```diff
@@ -0,0 +1,73 @@
+import { toBigInt } from 'ethers';
```
arboleya, line 1: I believe we were striving to eliminate `ethers`; has that changed? 🤔

danielbate (approved):

arboleya (approved):
✨

apps/docs-snippets/src/guide/types/numbers.test.ts
danielbate: I saw this also, however as it is in `doc-snippets` therefore not included in the `fuels` build so will not impact tree shakeability, which was one of the reasons for us trying to eliminate it. Also one of the intentions here was to show conversions between `ethers` `BigInt` and our `BigNum`, so for me it is okay here.

Dhaiwat10: Right @danielbate, and it's just a dev dependency to demonstrate the conversion.

Issue #1755, opened 2024-02-15
Fix issues with flaky tests


arboleya:
We have a number of flakey tests in CI. The source of which seems to be the shared `fuel-core` node across test suites. Given we have merged https://github.com/FuelLabs/fuels-ts/pull/2378 we can now integrate `launchTestNode` across our testing infrastructure, then we review the flakey tests. 

E.g. I've seen this one a couple of times now:
https://github.com/FuelLabs/fuels-ts/actions/runs/7914766231/job/21605183065#step:9:505

```
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  packages/fuel-gauge/src/predicate/predicate-with-contract.test.ts > Predicate > With Contract > calls a predicate and uses proceeds for a contract call
Error: Expected value: '1299797' to be within range: '1300778-1300818'
 ❯ Module.expectToBeInRange packages/utils/src/test-utils/getForcProject.ts:118:9
     76|     inputGlobal,
     77|   };
     78| };
       |             ^
     79| 
 ❯ packages/fuel-gauge/src/predicate/predicate-with-contract.test.ts:149:7
```

Here's the list of issues related to the integration of `launchTestNode` across our suites:

- [ ] https://github.com/FuelLabs/fuels-ts/issues/1820
- [ ] https://github.com/FuelLabs/fuels-ts/issues/2081
- [ ] https://github.com/FuelLabs/fuels-ts/issues/2744
- [ ] https://github.com/FuelLabs/fuels-ts/issues/2698

arboleya:
Three more in the same run:
https://github.com/FuelLabs/fuels-ts/actions/runs/7920711419/job/21624341501?pr=1759#step:9:597

```
 FAIL  packages/account/src/providers/memory-cache.test.ts > Memory Cache > can get active [with data]
AssertionError: expected [ '0xa82b33dd9ae34fd0' ] to strictly equal [ '0xa82b33dd9ae34fd0', …(2) ]
```

```
 FAIL  packages/account/src/providers/memory-cache.test.ts > Memory Cache > can get all [with data + expired data]
AssertionError: expected [ '0xa82b33dd9ae34fd0', …(1) ] to strictly equal [ '0xa82b33dd9ae34fd0', …(3) ]
```

```
[...]

PR #1769, opened 2024-02-19
chore: only run release workflow if tests pass

### LINKED ISSUE #1718: Tweak `release` workflow to abort if tests fail
Currently, the `release` workflow happens independently.

We should make it wait for everything else to pass _before_ releasing.

Here's one occurrence of the problem:

 - https://github.com/FuelLabs/fuels-ts/actions/runs/7787432043/job/21234957254

----

errors

Dhaiwat10:
Closes #1718


.github/workflows/release.yaml
```diff
@@ -1,15 +1,17 @@
 name: "Release"
 
 on:
-  push:
-    branches:
-      - master
+  workflow_run:
+    workflows: ["Test"]
+    types: [completed]
+    branches: [master]
```
danielbate, line 7: Nit, do we need arrays here if it's just a single value?

Dhaiwat10: @danielbate I thought we did, but we don't. Changed

.github/workflows/release.yaml
```diff
@@ -1,15 +1,17 @@
 name: "Release"
 
 on:
-  push:
-    branches:
-      - master
+  workflow_run:
+    workflows: "Test"
```
arboleya, line 5: My only hesitation in merging this before fixing [these flaky tests](https://github.com/FuelLabs/fuels-ts/issues/1755) is that it can potentially halt subsequent releases _(we should do one today)_, which may be a bit paradoxical because blocking releases _is_ what we want. 

However, flaky tests are not what we're proofing against, and it wouldn't be fun to have to battle with this for this next release specifically.

Thoughts?

arboleya (approved):

.github/workflows/release.yaml
Dhaiwat10: That's a dilemma. Let's do a coin toss to decide 😆 jk I would say let's wait for the flaky tests to get resolved, and then merge this in. Let's not introduce anything that could potentially delay today's release

danielbate: I am pro us introducing this if it means we resolve issues (such as flaky tests) with a higher level of urgency, as they could be indicative of a different problem.

danielbate (approved):

.github/workflows/release.yaml
arboleya: @danielbate It's best to resolve the [flakiness](https://github.com/FuelLabs/fuels-ts/issues/1755) with a higher level of urgency _without_ necessarily halting our release process, which can create an additional layer of problems. 🙂

Torres-ssf (approved):

Dhaiwat10:
@arboleya feel free to merge this one whenever you think it's a good time to do so

.github/workflows/release.yaml
```diff
@@ -1,15 +1,17 @@
...
+    types: completed
+    branches: master
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
 jobs:
   release:
     name: "Release"
+    if: ${{ github.event.workflow_run.conclusion == 'success'}}
```
arboleya, line 14: > [!WARNING]
> **[note to self]**: _Ideally we should not merge this before addressing:_
>  - https://github.com/FuelLabs/fuels-ts/issues/1755

nedsalk (approved):

petertonysmith94 (approved):

petertonysmith94:
Once #3014 is resolved, I believe we should look at merging this.

PR #1747, opened 2024-02-14
feat: migrate over @fuels/assets package into the TS SDK

Dhaiwat10:
This PR migrates the [`@fuels/assets`](https://github.com/fuellabs/fuels-npm-packs/tree/main/packages/assets) npm package to the TS SDK repo.

Todo:
- [x] Someone needs to add the AWS credentials to the TS SDK repo. These are required for the 'Upload assets to S3' workflow added with this PR. cc @luizstacio 

Closes #1541

PR #1590, opened 2023-12-22
fix: flaky test

Torres-ssf:
Fix [this flaky test](https://github.com/FuelLabs/fuels-ts/pull/1495#issuecomment-1867877426)

Screenshot 2023-12-22 at 15 08 29

PR #1773, opened 2024-02-19
fix: flaky test `tweaking-the-blockchain.test.ts`

nedsalk:
This test has been failing in multiple CI runs recently. Dhai updated in in [this commit](https://github.com/FuelLabs/fuels-ts/pull/1747/commits/874a640f62c0eda4799a558c2a77f581866df535) but his PR still needs some reviewing and this one can go through faster.


arboleya:
Is this related to #1755? I'd say we should employ this strategy of spinning up nodes for when we get to #1356 and, before that, ideally fix the flakiness another way like it was done in the past. Can't remember who did it, maybe @Torres-ssf.

danielbate (approved):

apps/docs-snippets/src/guide/testing/tweaking-the-blockchain.test.ts
```diff
@@ -1,10 +1,13 @@
...
+import { Provider, fromTai64ToUnix } from 'fuels';
 
 /**
  * @group node
  */
 test('produceBlocks with custom timestamp docs snippet', async () => {
-  const provider = await Provider.create(FUEL_NETWORK_URL);
+  const { cleanup, ip, port } = await launchNode({});
+  const url = `http://${ip}:${port}/graphql`;
+  const provider = await Provider.create(url);
```
arboleya, line 10: Ok, I think this [was it](https://github.com/FuelLabs/fuels-ts/pull/1773#issuecomment-1952165919) — it turns out we had a very similar problem, solved differently:
 - https://github.com/FuelLabs/fuels-ts/pull/1590

I don't think using `launchNode` here is entirely detrimental, mainly because it fixes the problem. Still, I'd advise adding a `TODO` so we don't forget to replace this with a more definitive solution after #1356.

arboleya (approved):

apps/docs-snippets/src/guide/testing/tweaking-the-blockchain.test.ts
nedsalk: Added todo: https://github.com/FuelLabs/fuels-ts/pull/1773/commits/caa1486880aa1f2335b1427d81e1befc4fbeb44d

arboleya (approved):

Dhaiwat10 (approved):

PR #1777, opened 2024-02-20
feat: accept options as CLI arguments in `create-fuels`

### LINKED ISSUE #1732: Accepts options as CLI arguments in `create-fuels`
Example:

```shell
# usage
npm create fuels <destination|name> [options]

# explicitly
npm create fuels my-dapp --contract --script --pnpm

# shortcuts
npm create fuels my-dapp -cs --pnpm
```

This should be simple by validating and grabbing the `argv`:

https://github.com/FuelLabs/fuels-ts/blob/71c052c9a9ea4414b69cb72144d04cc85d911187/packages/fuels/src/cli.ts#L29-L30

And then skipping the prompts accordingly:

```ts
// pseudo
async function promptForProjectPath() {
  return program.projectName ?? (await prompts({
    type: 'text',
    name: 'projectName',
    message: 'What is the name of your project?',
    initial: 'my-fuel-project',
  })).projectName as string;
}
```

_Originally posted by @arboleya in https://github.com/FuelLabs/fuels-ts/pull/1691#discussion_r1482776559_

Dhaiwat10:
## Summary
With this PR, `create-fuels` will now accept all of its options (and some new shortcuts! ✨) directly via the command line, optionally:

```sh
pnpm create fuels

# or, if you want to pass in your preferences directly from the command line:
pnpm create fuels [project-name] [options]

# eg. pnpm create fuels my-fuel-project --pnpm --contract --predicate
# Note: project-name and all other options are optional
Options:
  -V, --version    output the version number
  -C, --contract   Include contract program
  -P, --predicate  Include predicate program
  -S, --script     Include script program
  --pnpm           Use pnpm as the package manager
  --npm            Use npm as the package manager
  --cs             Include contract and script programs
  --cp             Include contract and predicate programs
  --ps             Include predicate and script programs
  --cps            Include all programs
  -h, --help       display help for command
  ```

Closes #1732

### CODE REVIEW

**Comment by Dhaiwat10** [comment]
@arboleya do you think it would be a good idea to refactor the `runScaffoldCli` function here to accept `process.argv`externally? I am thinking in this direction because currently there is no way for me to test these new CLI options apart from the current hacky approach that I have.

File: packages/create-fuels/src/cli.ts [hunk 1]
```diff
@@ -94,9 +94,23 @@ export const runScaffoldCli = async (
...
+    .arguments('[projectDirectory]')
+    .option('-C, --contract', 'Include contract program')
+    .option('-P, --predicate', 'Include predicate program')
+    .option('-S, --script', 'Include script program')
+    .option('--pnpm', 'Use pnpm as the package manager')
+    .option('--npm', 'Use npm as the package manager')
+    .option('--cs', 'Include contract and script programs')
+    .option('--cp', 'Include contract and predicate programs')
+    .option('--ps', 'Include predicate and script programs')
+    .option('--cps', 'Include all programs')
```
arboleya, line 108: This can be simplified to:

```suggestion
    .option('-c, --contract', 'Include only contract program')
    .option('-p, --predicate', 'Include only predicate program')
    .option('-s, --script', 'Include only script program')
    .option('--pnpm', 'Use pnpm as the package manager')
    .option('--npm', 'Use npm as the package manager')
```

The default is to generate all:

```shell
pnpm create fuels my-project # all
```

Users can opt-in to include only a subset:

```shell
pnpm create fuels my-project -c # contract only
pnpm create fuels my-project -p # predicate only
pnpm create fuels my-project -s # script only
```

Or any combo:

```shell
pnpm create fuels my-project -cp # contract + predicate
pnpm create fuels my-project -cs # contract + script
pnpm create fuels my-project -ps # predicate + script
pnpm create fuels my-project -cps # all, same as not informing anything
```

arboleya (changes requested):

danielbate (approved):
Nice work Dhai, couple thoughts on the structure that may benefit testing.

Rather than a refactor to the signature of `runScaffoldCli`, we could simply introduce a `getProgram` that takes cli arguments and returns a program instance at the top of the `runScaffoldCli` function.

```ts
export function getProgram(args: string[]): Program {
  const program = new Command(packageJson.name)
    .option('--o', 'my option)
    // rest of options
  program.parse(args);
  return program;
}
```

You can then independently test that function to give it various arguments and assert the options it's returned with. Equally we could mock the program in other tests.

packages/create-fuels/src/cli.ts
arboleya: Absolutely! This is valid for boolean flags. 🎉

Torres-ssf (approved):

apps/docs/src/guide/quickstart/index.md
```diff
@@ -6,10 +6,46 @@ You can quickly bootstrap a full-stack Fuel project locally with the following c
...
+  -V, --version    output the version number
+  -c, --contract   Include contract program
+  -p, --predicate  Include predicate program
+  -s, --script     Include script program
+  --pnpm           Use pnpm as the package manager
+  --npm            Use npm as the package manager
+  --cs             Include contract and script programs
+  --cp             Include contract and predicate programs
+  --ps             Include predicate and script programs
+  --cps            Include all programs
```
arboleya, line 25: I suspect you want to use single dashes when combining options. 🤔 

```suggestion
  -cs             Include contract and script programs
  -cp             Include contract and predicate programs
  -ps             Include predicate and script programs
  -cps            Include all programs
```

arboleya (changes requested):

apps/docs/src/guide/quickstart/index.md
nedsalk: I'd favor a one-liner here:
```
-cs, -cp, -sp, -cps    Shorthand to include combination of contract, script and predicate programs
```

Dhaiwat10:
@arboleya re. defaulting to all three programs when no args are specified, I have incorporated that behaviour this way: 

We still prompt the user for their choice of programs, but we pre-select all three options by default. I think strikes a good balance between saving time through a good default and still being flexible enough for first-time or new users who might not be aware of the args.

nedsalk (approved):
Good stuff! :+1:

Torres-ssf (approved):

arboleya (approved):

PR #1775, opened 2024-02-19
chore: internalise `ethers` utilities for `hexlify`, `arrayify`, `concat` and `BytesLike`

danielbate:
Following on from discussions in #1592 to deprecate `ethers` functionality and improve tree-shakability, this PR removes usages of the following `ethers` functions:

 - [x] `hexlify`
 - [x] `arrayify`
 - [x] `concat`
 - [x] `BytesLike`

PR #1779, opened 2024-02-20
chore: add bundled tsup config to `.gitignore`

danielbate:
Initiated by [this comment](https://github.com/FuelLabs/fuels-ts/pull/1775#discussion_r1495613258), we should add bundled tsup configs to the `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,7 +59,7 @@ typings/
 *.tsbuildinfo
 
 # Tsup config bundles
-tsup.config.bundled_*.mjs
+tsup.config.bundled_*
 
 # Optional npm cache directory
 .npm


arboleya (commented):
It is a weird one because this is the [config](https://github.com/FuelLabs/fuels-ts/blob/master/packages/crypto/tsup.config.ts) file that's _fed_ into `tsup`. It never includes itself as an input and, therefore, should never be built by `tsup`. Do you know how this file was built in the first place? 🤔

danielbate:
Hmm it appears after a build, however not all the time. I will try and recreate a scenario where it always produces this file.

arboleya (approved):

Torres-ssf (approved):

Issue #1778, opened 2024-02-20
Add option to build `sway` programs in `release` mode


arboleya:
When one runs `forc build`, `sway` programs are compiled in `debug` mode by default.

We need to add a way to build them in `release` mode.

This must be done in the `fuels` CLI, by introducing a `forcBuildFlags` property in the `fuels.config.ts` file.

Related `sway` issue:
 - https://github.com/FuelLabs/sway/issues/5607

PR #1784, opened 2024-02-21
feat: adding `--release` option to `build` command in `fuels` CLI

arboleya:
- Close #1778

PR #1788, opened 2024-02-22
feat: adding new `forcBuildFlags` property to `fuels` config

arboleya:
- Close https://github.com/FuelLabs/fuels-ts/issues/1778


.github/workflows/pr-release.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
     name: "Release PR to npm"
     runs-on: ubuntu-latest
     # comment out if:false to enable release PR to npm
-    if: false
+    # if: false
```
arboleya, line 11: > [!WARNING]
> Uncomment before merging.

apps/docs/src/guide/cli/config-file.md
```diff
@@ -90,6 +90,20 @@ Port to use when starting a local `fuel-core` node.
 
 <<< @../../../demo-fuels/fuels.config.full.ts#fuelCorePort{ts:line-numbers}
 
+## `forcBuildFlags`
+
+> - _Used by [`fuels build`](./commands.md#fuels-build) and [`fuels deploy`](./commands.md#fuels-deploy)_.
+
+Sway programs are compiled in `debug` mode by default.
+
+Here you can customize all build `flags`, e.g. to build programs in `release` mode.
```
nedsalk, line 99:
```suggestion
Here you can customize all build flags, e.g. to build programs in `release` mode.
```
I think "flags" shouldn't be in backticks here because it's used in its standard meaning.

packages/fuels/src/cli/config/loadConfig.ts
```diff
@@ -44,6 +44,10 @@ export async function loadConfig(cwd: string): Promise<FuelsConfig> {
   const useBuiltinForc = userConfig.useBuiltinForc ?? shouldUseBuiltinForc();
   const useBuiltinFuelCore = userConfig.useBuiltinFuelCore ?? shouldUseBuiltinFuelCore();
 
+  const { forcBuildFlags = [] } = userConfig;
+  const releaseFlag = forcBuildFlags.find((f) => f === '--release');
+  const buildMode = releaseFlag ? 'release' : 'debug';
```
nedsalk, line 49: From what I understood, `buildMode` is only being used in `forcUtils` to get the correct path. If you were to combine the utils that are using the path into one function that returns an object, similar to what `getForcProject` does, you could move this `buildMode` extraction logic out from `loadConfig` and into that function and remove the `buildMode` property from `FuelsConfig`:

```ts
export function getPaths(contractPath: string, { forcBuildFlags: [] }: FuelsConfig){
  const releaseFlag = forcBuildFlags.find((f) => f === '--release');
  const buildMode = releaseFlag ? 'release' : 'debug';
  const outDir = join(contractPath, "out", buildMode);
  
  const projectName = getContractName(contractPath);

  return {
    binaryPath: join(outDir,`${projectName}.bin`),
    storageSlotsPath: join(outDir, `${projectName}-storage_slots.json`),
    abiPath: join(outDir, `${projectName}-abi.json`),
        
  }
}
```

I think that's worth the effort because it removes maintainer questions/ambiguity around these scenarios:

```ts
const fuelsConfig: FuelsConfig = {
  // ...,
  forcBuildFlags: ["--release"],
  buildMode: 'debug',
}

const fuelsConfig: FuelsConfig = {
  // ...,
  forcBuildFlags: ["--debug"],
  buildMode: 'release',
}
```

nedsalk (changes requested):

packages/fuels/src/cli/config/loadConfig.ts
arboleya: Thanks for the analysis.

There is a clear distinction between `UserFuelsConfig` and `FuelsConfig`; the latter is not under the user domain, and we conveniently initialize it only once, immediately after loading the user config. The pattern is borrowed from the `basePath` and `configPath` props that follow the same principle.

Handing all configuration options in a single place provides a clear separation of concerns between configuration and utilities, and having the type clearly defined in the `types.ts` makes it easier to follow, not harder.

The conflicting scenario you described with `debug` x `release` seems made up and purely theoretical since it can never occur in practical terms, only mentally as a hypothesis. We initialize things _only once_ and they are tied together. Full test coverage also guarantees no silly mistakes will slip through.

nedsalk: > The conflicting scenario you described with debug x release seems made up and purely theoretical since it can never occur in practical terms, only mentally as a hypothesis. We initialize things only once and they are tied together. Full test coverage also guarantees no silly mistakes will slip through.

It's not made up, I questioned why it exists and then went to analyze it and produced the suggestion as a way to reduce the ambiguity so that the question doesn't pop up anymore.

nedsalk (approved):

packages/fuels/src/cli/config/loadConfig.ts
arboleya: Made up in the sense this can never occur, only as a mental hypothesis.

```ts
  forcBuildFlags: ["--debug"],
  buildMode: 'release',
```

The same hypothesis would hold regardless of where we'd put his block of code:

```ts
  const releaseFlag = forcBuildFlags.find((f) => f === '--release');
  const buildMode = releaseFlag ? 'release' : 'debug';
```

Torres-ssf (approved):

packages/fuels/src/cli/config/forcUtils.ts
```diff
@@ -70,21 +72,21 @@ export function getContractCamelCase(contractPath: string) {
...
-  return Promise.all(paths.map((path) => getABIPath(path)));
+export function getABIPaths(paths: string[], config: FuelsConfig) {
+  return Promise.all(paths.map((path) => getABIPath(path, config)));
 }
 
-export const getStorageSlotsPath = (contractPath: string) => {
+export const getStorageSlotsPath = (contractPath: string, { buildMode }: FuelsConfig) => {
   const projectName = getContractName(contractPath);
-  return join(contractPath, `/out/debug/${projectName}-storage_slots.json`);
+  return join(contractPath, `/out/${buildMode}/${projectName}-storage_slots.json`);
```
danielbate, line 91: We should probably move `getForcProject` from `test-utils` -> `utils` and add this parameter in so we can utilise it in other places. Ofc it would need the build mode flag.

danielbate (approved):

packages/fuels/src/cli/config/forcUtils.ts
arboleya: Yes, @nedsalk raised the same point [here](https://github.com/FuelLabs/fuels-ts/pull/1784#discussion_r1497519042).

There are a few differences between the two, but yes, we can consider unifying and normalizing them.

Torres-ssf (approved):

Dhaiwat10 (approved):
🔥 🔥

PR #1793, opened 2024-02-23
fix: subscriptions not closing after done

nedsalk:
This fix was part of #1597 which didn't get merged. In essence, the piping done in 
``` ts
const subscriptionStreamReader = response
    .body!.pipeThrough(new FuelSubscriptionStream())
    .getReader();
``` 
was ignoring the value `{value: undefined, done: true}` which indicates a stream is done. This resulted in subscriptions stalling when they aren't cancelled externally with a `break` in a `for-await-of` loop even though their underlying connection got closed properly. I opted for not using piping for transforming streams but rather inlining the byte parsing in the `next` method of the iterator.

You can see more around the investigation of this problem in https://github.com/FuelLabs/fuels-ts/pull/1615#discussion_r1445143916. 

Besides this, the PR also had a test that verified our code correctly handles keep-alive messages for long-running subscriptions which I added here as well.


packages/account/src/providers/fuel-graphql-subscriber.ts
```diff
@@ -7,75 +7,76 @@ type FuelGraphQLSubscriberOptions = {
...
+   */
+  async return(): Promise<IteratorResult<unknown, undefined>> {
+    await this.stream.cancel();
+    this.stream.releaseLock();
+    return { done: true, value: undefined };
+  }
+
+  [Symbol.asyncIterator](): AsyncIterator<unknown, unknown, undefined> {
+    return this;
+  }
```
danielbate, line 81: Not familiar with this syntax. Please can you provide some context? And why do we need this, for the interface?

packages/account/src/providers/fuel-graphql-subscriber.ts
```diff
@@ -7,75 +7,76 @@ type FuelGraphQLSubscriberOptions = {
...
-                FuelError.CODES.INVALID_REQUEST,
-                errors.map((err) => err.message).join('\n\n')
-              )
-            );
-          } else {
-            this.readableStreamController.enqueue(data);
-          }
-        }
+  private async setStream() {
+    const { url, query, variables, fetchFn } = this.options;
```
danielbate, line 19: How does `this.options` get set?

packages/account/src/providers/fuel-graphql-subscriber.ts
```diff
@@ -7,75 +7,76 @@ type FuelGraphQLSubscriberOptions = {
...
+        // eslint-disable-next-line no-continue
+        continue;
+      }
+
+      const { data, errors } = JSON.parse(text.split('data:')[1]);
+
+      if (Array.isArray(errors)) {
+        throw new FuelError(
+          FuelError.CODES.INVALID_REQUEST,
+          errors.map((err) => err.message).join('\n\n')
```
danielbate, line 62: Should we be doing any message formatting here? Or is this just sending through the client response?

packages/account/src/providers/fuel-graphql-subscriber.ts
nedsalk: This is needed for the JS runtime to identify it as an async iterator. Implementing just the interface methods isn't enough. See more on the symbol [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator).

packages/account/src/providers/fuel-graphql-subscriber.ts
```diff
@@ -7,75 +7,76 @@ type FuelGraphQLSubscriberOptions = {
...
-    body: JSON.stringify({
-      query: print(query),
-      variables,
-    }),
-    headers: {
-      'Content-Type': 'application/json',
-      Accept: 'text/event-stream',
-    },
-  });
+  async next(): Promise<IteratorResult<unknown, unknown>> {
```
arboleya, line 37: Nice one! We shouldn't have used that generators* syntax from the beginning. 😃 

Happy we're moving away. 🎉

packages/fuel-gauge/src/transaction-response.test.ts
```diff
@@ -9,6 +9,75 @@ import {
...
+  } while (!hasKeepAliveMessage);
+
+  // The keep-alive message is sent every 15 seconds,
+  // and this assertion verifies that it was indeed sent.
+  // if this fails, check if the duration was changed on the fuel-core side.
+  // As of the time of writing, the latest permalink where this info can be found is:
+  // https://github.com/FuelLabs/fuel-core/blob/bf1b22f47c58a9d078676c5756c942d839f38916/crates/fuel-core/src/graphql_api/service.rs#L247
+  // To get the actual latest info you need to check out the master branch:
+  // https://github.com/FuelLabs/fuel-core/blob/master/crates/fuel-core/src/graphql_api/service.rs#L247
+  // This link can fail because master can change.
```
arboleya, line 36: Does it make sense to have these notes in the source file instead?

packages/fuel-gauge/src/transaction-response.test.ts
```diff
@@ -9,6 +9,75 @@ import {
...
+      const [stream1, stream2] = r.body!.tee();
+      // eslint-disable-next-line no-param-reassign
+      streamHolder.stream = stream1;
+      return new Response(stream2);
+    };
+  }
+
+  const fetchSpy = vi.spyOn(global, 'fetch');
+
+  fetchSpy.mockImplementation(getFetchMock(fetchSpy));
```
arboleya, line 78: Are we using mocks inside `fuel-gauge` now? 🤔

arboleya (commented):
Did a quick pass, will another later on.

Tests could be a bit less abstract, but I like what I see so far! 🚀

packages/account/src/providers/fuel-graphql-subscriber.ts
nedsalk: This is just sending through the client response in the format it was sent.

packages/account/src/providers/fuel-graphql-subscriber.ts
danielbate: LOL my bad

packages/fuel-gauge/src/transaction-response.test.ts
nedsalk: Ah, this is the abstract part :D I agree, it's weird, to say the least. I couldn't figure out a nicer way to _not_ change the behavior while still getting what I need :smile:.

The purpose of this code is to protect the test from false positives if the node's default 15 seconds for sending keep-alive messages was increased over the test's `--poa-interval-period 17s` and thus no keep-alive message was sent. We'd be thinking the test covers keep-alive messages, but it wouldn't be.

As for the implementation and why I "mocked", I couldn't read keep-alive messages any other way because they are ignored by the code and the streams get closed after they are read so I couldn't go with just a simple spy.

In order to get the underlying stream that contains the subscription responses before it gets closed (actual responses and the keep-alive message) and verify that a keep-alive message was sent, I had to "mock" fetch. The mock itself isn't changing anything, it's just duplicating the stream in the case of a subscription so that I can inspect it for the keep-alive message:
- If the call is not to a subscription endpoint, the response will be just as a regular `fetch` response would be. 
- If the call is to a subscription endpoint, the response will _still_ be the same as that of a regular `fetch`, but the stream is duplicated so that I can parse it independently in `verifyKeepAliveMessageWasSent`.

 The `getFetchMock` function looks weird with its `fetchSpy.mockRestore()` and `fetchSpy.mockImplementation(getFetchMock(fetchSpy))` calls precisely because I didn't want to change `fetch` behavior. I had to make sure that the first and _every_ subsequent fetch call are "mocked" with the logic described above so that, for the specific subscription fetch call, I can duplicate the stream before it's closed.

packages/account/src/providers/fuel-graphql-subscriber.ts
nedsalk: Yeah, the `yield` syntax in generators* isn't as commonplace as ...`return`. :smile:

packages/fuel-gauge/src/transaction-response.test.ts
```diff
@@ -9,6 +9,75 @@ import {
...
+      const r = await fetch(...args);
+      fetchSpy.mockImplementation(getFetchMock(fetchSpy));
+
+      const isSubscriptionCall = args[0].toString().endsWith('graphql-sub');
+      if (!isSubscriptionCall) {
+        return r;
+      }
+
+      /**
+       * Tee duplicates a stream and all writes happen to both streams.
```
arboleya, line 63:
```suggestion
       * This duplicates a stream and all writes happen to both streams.
```

nedsalk: This is correct, it's explaining what `r.body!.tee();` below does.

arboleya (approved):
I like the new direction. 🫡

Torres-ssf (approved):

PR #1805, opened 2024-02-26
chore: avoid reassigning inputs and outputs at `BaseInvocationScope`

Torres-ssf:
Changes made in this PR:
- Add `outputVariables` and `missingContractIds` to the return of `estimateTxDependencies`
- Removed `estimatedOutputs` from return of `getTransactionCost`
- Add `outputVariables` and `missingContractIds` to the return of `getTransactionCost`
- Avoid reassigning `inputs` and `outputs` from the estimated TX at `BaseInvocationScope`


packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -642,8 +642,6 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
     this.inputs.forEach((i) => {
       let correspondingInput: TransactionRequestInput | undefined;
       switch (i.type) {
-        case InputType.Contract:
-          return;
```
arboleya, line 646: No more `InputType.Contract`?

arboleya (approved):
🎯

packages/account/src/providers/transaction-request/transaction-request.ts
Torres-ssf: @arboleya Since within the loop we are only concerned with `Coin` and `Message` inputs I am letting the `default` case handle everything else.

danielbate (approved):

Issue #1798, opened 2024-02-23
Add support for adding `CoinOutputs` to the transaction request at `BaseInvocations`


Torres-ssf:
On the Rust SDK, there is a functionality that allows users to easily [add a transfer](https://github.com/FuelLabs/fuels-rs/blob/86e3cc77603642292cb884e7e9f005bc9297cf25/packages/fuels-programs/src/contract.rs#L532-L540) to the transaction request before executing a contract call.

We should enhance the TS-SDK by adding a similar functionality.

PR #1810, opened 2024-02-26
feat: implement `addTransfer` at `BaseInvocationScope`

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/1798


arboleya (approved):

danielbate (approved):
Great improvement @Torres-ssf 👍🏻 

My only issue here is a good amount of this PR seems to relate to the refactoring of Asset IDs. I appreciate the sentiment but potentially scope creep in this PR.

packages/utils/src/test-utils/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const ASSET_A: string = '0x0101010101010101010101010101010101010101010101010101010101010101';
```
danielbate, line 1: FYI @nedsalk who has implemented something similar in #1356.

PR #612, opened 2022-11-20
feat: add better documentation features and convert from Rust repo

camsjams:
Closes #548 

#### 1. Basic Viev
Browse the repo on GitHub or local [using the markdown files](https://github.com/FuelLabs/fuels-ts/tree/cm/issue-548/docs/guide)

#### 2.  Recommended  View
Browse the docs [using Jekyll ](https://jekyllrb.com/docs/) (`bundle exec jekyll serve`) to see how they will appear on GitHub when deployed, which [matches current docs here](https://github.com/FuelLabs/fuels-ts)

## Notable features
The docs will always report what versions of Fuels/Sway/Fuel Core they were built with:
image

New Guide section:
image

Embedded code using real code from library and unit tests:
image

## Notes
- The format and chapters added are sourced from the [Rust SDK Docs](https://fuellabs.github.io/fuels-rs/v0.31.1/) and adjusted as needed
- 99% of code examples are sourced from unit tested code, this took a while :sweat: (see code snippets in action [right here](https://github.com/FuelLabs/fuels-ts/blob/cm/issue-548/docs/_guide/types/address.md?plain=1#L11) for example)
- Some docs were excluded from the docs as the Rust concepts didn't apply
- I added some new Docs and code snippets as needed
- I added the new AbiCoder for B512 as it was needed for an example :sweat_smile: 
- I ended up [making a new plugin](https://xkcd.com/927/) for Typedoc :sweat_smile:  that [copies the `docs/_guide` source folder into `docs/guide`](https://github.com/FuelLabs/fuels-ts/tree/cm/issue-548/scripts/typedoc-plugin-guide-builder) and also pulls in code samples from repo as needed using syntaxt inspired by other Doc tools in other languages (and the [Typedoc code block](https://www.npmjs.com/package/@knodes/typedoc-plugin-code-blocks) plugin which wasn't working well enough for our needs). 
- the [Jekyll data variables defined here](https://github.com/FuelLabs/fuels-ts/blob/cm/issue-548/docs/_data/versions.yml) are pulling in [environment vars passed here ](https://github.com/FuelLabs/fuels-ts/blob/cm/issue-548/scripts/changeset-publish-with-docs.ts#L8)which were [originally parsed out in this code here](https://github.com/FuelLabs/fuels-ts/blob/cm/issue-548/.github/workflows/release.yaml#L28-L30)
[...]

PR #1815, opened 2024-02-27
feat!: improve signing DX

### LINKED ISSUE #1745: Populating transaction witnesses signature naming convention

v0.73.0

### Steps to Reproduce

https://github.com/calldelegation/multisig-predicate/tree/report/signing-predicate-txn

1. Clone the repository
2. Switch to the `report/signing-predicate-txn` branch
3. Run `pnpm install`
4. In another terminal run `fuel-core run --db-type in-memory`
5. Run `pnpm vitest --run ec-recover`

### Expected Behavior

In our first test, the transaction was successfully populated with the witness data using the provided lines of code. However, after making a few minor adjustments for the second test, the witness data no longer seems to populate correctly. I'm unsure if this behavior is intentional.

[TEST 1: `transacts using predicate`](https://github.com/calldelegation/multisig-predicate/blob/report/signing-predicate-txn/frontend/tests/ec-recover.test.ts#L79-#L84)
[TEST 2: `transacts using predicate fails`
](https://github.com/calldelegation/multisig-predicate/blob/report/signing-predicate-txn/frontend/tests/ec-recover.test.ts#L128-#L130)https://github.com/calldelegation/multisig-predicate/blob/report/signing-predicate-txn/frontend/tests/ec-recover.test.ts

The current naming convention adds to the confusion, leading me to believe that the transaction should automatically populate with `.populateTransactionWitnessesSignature()`, eliminating the need for further action on my part. Ideally, I'd prefer a more streamlined approach, perhaps integrating the signature and transaction population into a single function specifically in `.signTransaction()`, utilizing a Last In, First Out (LIFO) stack.

This makes it more intuitive and user-friendly.

Additionally the error message I get is as follows
```
Error: PredicateVerificationFailed(PanicInstruction(PanicInstruction { reason: ContractInstructionNotAllowed, instruction: RVRT { value: 0x0 } (bytes: 36 00 00 00) })): {"response":{"data":null,"errors":[{"message":"PredicateVerificationFailed(PanicInstruction(PanicInstruction { reason: ContractInstructionNotAllowed, instruction: RVRT { value: 0x0 } (bytes: 36 00 00 00) }))","locations":[{"line":2,"column":3}],"path":["submit"]}],"status":200,"headers":{}},"request":{"query":"mutation submit($encodedTransaction: HexString!) {\n  submit(tx: $encodedTransaction) {\n    id\n  }\n}","variables":{"encodedTransaction":
```
This error message is also very misleading. It made me think that there was an error with not the SDK but my predicate when that is not the case. 

### Actual Behavior

TEST 1: `transacts using predicate`
```ts
const signedTransaction = await coreWallet.signTransaction(request);
const transactionRequest =
  await coreWallet.populateTransactionWitnessesSignature({
    ...request,
    witnesses: [signedTransaction],
  });
```

returns

```
stdout | tests/ec-recover.test.ts > ec-recover > transacts using predicate
Witnesses [
  '0x236fa4e110d4273f78244f84e7990cf7c158ccd9745a49ab467062037134d822d115f506ef2405af94d49d426bcb8ae32ed9a1b058e12fa56c60ac876c2a5f37'
]
```

TEST 2: `transacts using predicate fails`
```ts
const signedTransaction = await coreWallet.signTransaction(request);
const transactionRequest = await coreWallet.populateTransactionWitnessesSignature(request);

```

returns 
```
stdout | tests/ec-recover.test.ts > ec-recover > transacts using predicate fails
Witnesses [
  '0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000'
]
```

cc @danielbate

danielbate:
Closes #1745 

- [x] Implementing signing transaction via an `Account` through a wallet connector
- [x] Add multiple external signers to a request through `addAccountWitnesses` functions on `BaseTransactionRequest`
- [x] Rename `createWitness` to `addEmptyWitness` and implement `addWitness` fully
- [x] Add documentation and example for a script and predicate multi sig with external signers (we only ever used `signMessage`, I have added `signTransaction` examples)
- [x] Improvements and fixes to current docs around signing

Please note, the approach discussed in #1576 has not been implemented here however there is merit to it, as shown in the predicate test example.


Dhaiwat10:
@danielbate there are some forc formatting errors that you might want to look at

apps/docs-snippets/src/guide/cookbook/signing-transactions.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    const amountToReceiver = 100;
+
+    // #region multiple-signers-2
+    // #import { Script, BaseAssetId };
+
+    // Create invocation scope
+    const script = new Script(binScript, abiScript, sender);
+    const scope = script.functions
+      .main(signer.address.toB256())
+      .txParams({ gasPrice, gasLimit: 10_000 });
```
Dhaiwat10, line 59: This is an awesome PR. Huge wins here! Still reviewing it, but I noticed this. You don't need to specify these values anymore after #1708

Torres-ssf (dismissed):

apps/docs/src/guide/cookbook/transactions-with-multiple-signers.md
```diff
@@ -0,0 +1,21 @@
...
+
+<<< @/../../docs-snippets/test/fixtures/forc-projects/script-signing/src/main.sw#multiple-signers-1{rust:line-numbers}
+
+Here, through the use of the built-in Sway function `tx_witness_data()` to retrieve the witness signatures and `tx_id()` for the transaction hash. Then we retrieve the signing address to validate the script.
+
+We would interact with this script in the SDK by creating a transaction request from an invocation scope. The same can be done for a contract. Consider the following script:
+
+<<< @/../../docs-snippets/src/guide/cookbook/signing-transactions.test.ts#multiple-signers-2{ts:line-numbers}
+
+The same approach can be used for a predicate. Instead we instantiate a predicate and add it to a transaction request. Consider the following predicate:
```
arboleya, line 15: Couldn't figure out the "instead" part - like instead of what?

See if it makes sense.

```suggestion
The same approach can be used for a predicate by instantiating it and adding it to a transaction request. Consider the following predicate:
```

apps/docs/src/guide/wallets/signing.md
```diff
@@ -1,9 +1,9 @@
 # Signing
 
-Once you've instantiated your wallet in an unlocked state using one of the previously discussed methods, you can sign a message with `wallet.sign_message`. Below is a full example of how to sign and recover a message.
+Once you've instantiated your wallet in an unlocked state using one of the previously discussed methods, you can sign a message with `wallet.signMessage`. Below is a full example of how to sign and recover a message.
```
arboleya, line 3: Out of curiosity, did we have anything like `sign_message` or `sign_transaction`?

Or was it wrong since forever?

apps/docs-snippets/src/guide/cookbook/signing-transactions.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    expect(value.toNumber()).toEqual(1);
+    expect((await receiver.getBalance()).toNumber()).toEqual(amountToReceiver);
+  });
+
+  it('creates a transfer with external signer [predicate]', async () => {
+    const amountToReceiver = 100;
+
+    // #region multiple-signers-4
+    // Create and fund predicate
+    // #import { Predicate, BaseAssetId };
```
arboleya, line 85: Should we remove the "Create and fund predicate" comment?

It's appearing in front of the import, which is weird.

line-break

@Torres-ssf Is this usage of having comments in-between context/import supported?

arboleya (commented):
Awesome PR, Dan! 👏

packages/account/src/connectors/fuel-connector.ts
```diff
@@ -104,6 +104,20 @@ export abstract class FuelConnector extends EventEmitter {
     throw new Error('Method not implemented.');
   }
 
+  /**
+   * Should start the sign transaction process and return
+   * the signed transaction.
+   *
+   * Why do we need address, is it not the address of the account?
```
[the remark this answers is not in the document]
danielbate: Oops, this was a note to self, will disregard

apps/docs/src/guide/cookbook/transactions-with-multiple-signers.md
danielbate: Within context the instead was referring to a script. But I still like yours more.

apps/docs/src/guide/wallets/signing.md
danielbate: This has always been wrong as it was initially taken from the rust repo - #612

apps/docs-snippets/src/guide/cookbook/signing-transactions.test.ts
danielbate: Yes that should not be there, have removed.

apps/docs-snippets/src/guide/cookbook/signing-transactions.test.ts
danielbate: Interestingly, both tests fail if I remove `txParams` 🤔

danielbate:
cc @calldelegation

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -249,6 +265,32 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
...
+    return this;
+  }
+
+  /**
+   * Adds multiple external signers to the transaction.
+   *
+   * @param signers - The signers to add to the transaction.
+   * @returns The transaction with multiple signature witnesses added.
+   */
+  async addSigners(signers: Account[]) {
```
nedsalk, line 288: IMO there is no need to have these plural-style methods because whoever wants to add multiple signers will just create their own ad-hoc implementation just like the one in this method:

```ts
await Promise.all(signers.map((signer) => request.addSigner(signer)));
```

It adds an additional maintenance burden on us because we have to keep the interfaces _and_ descriptions of both methods in sync. I already had to do this one time in another place and it's generally a PITA because you always forget something.

danielbate: Hmm I think I like the pluralised as the use case for a multi sig is pretty common, `addSigners` will probably be more common place than `addSigner`. I can appreciate the additional burden. What about we ditch `addSigner`, only keep `addSigners` and I'll move all the logic there?

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -249,6 +265,32 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
     this.witnesses[index] = witness;
   }
 
+  /**
+   * Helper function to add an external signer to the transaction.
+   *
+   * @param signer - The signer to add to the transaction.
+   * @returns The transaction with the signature witness added.
+   */
+  async addSigner(signer: Account) {
```
nedsalk, line 274: The name of the method `addSigner` doesn't fully carry the meaning of the implementation. It opens up additional questions like "What happens with the signer when I add it?", "Will it be used later for something?", "Does the signer sign the transaction immediately?" - at least for me.

I believe a better approach would be to expand `addWitness` and remove this method:
```ts
  /**
   * Adds a witness to the request.
   * If the value is a `BaseWalletUnlocked`, the wallet signs the transaction and the signature is pushed into the witnesses.
   * If the value is `BytesLike`, it gets directly pushed into the witnesses.
   *
   * @param signerOrSignature - The signature to add to the witness.
   * @returns The index of the created witness.
   */
  addWitness(signerOrSignature: BaseWalletUnlocked | BytesLike) {
    if(signerOrSignature instanceof BaseWalletUnlocked) {
        const hashedTransaction = this.getTransactionId(signer.provider.getChainId());
        const signature = await signer.signTransaction(hashedTransaction);
        this.witnesses.push(signature);
    } else {
        this.witnesses.push(signerOrSignature);
    }

    return this.witnesses.length - 1;
  }
```

packages/account/src/wallet/base-wallet-unlocked.ts
```diff
@@ -77,12 +77,8 @@ export class BaseWalletUnlocked extends Account {
    * @param transactionRequestLike - The transaction request to sign.
    * @returns A promise that resolves to the signature as a ECDSA 64 bytes string.
    */
-  async signTransaction(transactionRequestLike: TransactionRequestLike): Promise<string> {
-    const transactionRequest = transactionRequestify(transactionRequestLike);
-    const chainId = this.provider.getChain().consensusParameters.chainId.toNumber();
-    const hashedTransaction = transactionRequest.getTransactionId(chainId);
+  async signTransaction(hashedTransaction: string): Promise<string> {
```
nedsalk, line 80: This method can be made synchronous because the `await` in `const signature = await this.signer().sign(hashedTransaction);` is unnecessary. It's been bugging me for a while but I never really touched this part of the code :D It'd be great if you would change it.

nedsalk: On the topic of the changed interface of the method, I don't fully understand the motivation in splitting it up as it seemed like a cohesive piece of code. Could you please explain why you split did it?

danielbate: Couple of reasons. Firstly making it compatible with an `Account` and therefore having the same function signature. Wallet extends account, therefore we could support either of those via `addSigners`. The use case came from dev rel who have been working with connectors, so it felt like a safe assumption that we should definitely support the upper level. 
An account can only sign via a connector. I would expect most, if not all of the logic to live within the implementer of the connector. Account is an abstraction and just acts as a middleman, or `throws` if the connector does not support the function (signing, sending etc.). 
So to make these compatible, one had to change. Either `Account.signTransaction()` takes `TransactionRequestLike` or `BaseWalletUnlocked.signTransaction()` takes a string. I opted for the latter as I believe the role of the wallet or account in this instance is just to sign the data, not convert and sign.

packages/account/src/providers/transaction-request/transaction-request.ts
nedsalk: I'm okay with both, but if we make signing synchronous as per the implications of https://github.com/FuelLabs/fuels-ts/pull/1815#discussion_r1511520861, it would be as straightforward as this then:
```ts
signers.forEach(signer => request.addSigner(signer))
```

I like this more than having the plural because I don't like the one-signer use case where an array has to be instantiated just for the signing to work: 
```ts
request.addSigners([signer]);
```

This is just my preference, so if you like the `addSigners` more, I'm okay with you implementing that as well.

danielbate: > I'm okay with both, but if we make signing synchronous as per the implications of https://github.com/FuelLabs/fuels-ts/pull/1815#discussion_r1511520861, it would be as straightforward as this then:

I shall make the signing synchronous, good spot 👍🏻 

> I like this more than having the plural

My preference here is I do like with the explicit `addSigners` as it can be chained onto the tx request for better DX

> This is just my preference, so if you like the addSigners more, I'm okay with you implementing that as well.

True and likewise, I'm appreciating the discussion though 👍🏻  May see if anyone else chimes in with a preference.

nedsalk: Ah, good one with the chaining! I'm on your side now :smile:

apps/docs-snippets/src/guide/cookbook/signing-transactions.test.ts
Dhaiwat10: Oh interesting, what does it say?

packages/account/src/providers/transaction-request/transaction-request.ts
arboleya: Both singular and plural alternatives work fine, but I'd be cautious with type overloading. Although `string | string[]` is obvious and perfectly fine, I'm usually [against](https://github.com/FuelLabs/fuels-ts/issues/748#issuecomment-1400789719) overloading completely different types like `Account | BytesLike`, as abstractions start to creep in. 

@nedsalk The same questions could arise regardless of the terminology.

.github/workflows/test.yaml
```diff
@@ -43,8 +43,9 @@ jobs:
       - name: Verify package.json integrity
         run: pnpm lint:package-jsons
 
-      - name: Forc Format Check
-        run: pnpm forc:check
+      # Temp disable
```
danielbate, line 46: Remove this

packages/account/src/providers/transaction-request/transaction-request.ts
nedsalk: @danielbate and I had a chat around it and opted for a separate method instead of overloading `addWitness`:
```ts
addAccountWitness(account: Account) {...}
```
or something like that.

packages/account/src/wallet/base-wallet-unlocked.ts
nedsalk: @danielbate and I had a chat around this and realized that there are issues in the relationship between `FuelConnector` and `Account`. We did a bit of a dive into it and realized that it's too big of a topic for this PR. The way he did it here is okay in the sense of the current design, but we both agreed that it probably shouldn't be the way forward. I'll try getting more info around this and summarizing the situation in an issue/discussion so that the whole team is on the same page.

apps/docs-snippets/src/guide/cookbook/signing-transactions.test.ts
danielbate: Script throws `OutOfGas` and predicate `Transaction is not inserted. The gas price is too low`.

packages/account/src/providers/transaction-request/transaction-request.ts
danielbate: I have now implemented `addAccountWitnesses` which takes `Account | Account[]`.

packages/account/src/providers/transaction-request/transaction-request.test.ts
```diff
@@ -113,6 +115,40 @@ describe('TransactionRequest', () => {
     expect(transactionRequest.witnesses.length).toEqual(2);
     expect(transactionRequest.witnesses).toStrictEqual([mockSignedTx, mockSignedTx]);
   });
+
+  it.only('adds account based witnesses', async () => {
```
arboleya, line 119: I think this went unnoticed.

```suggestion
  it('adds account based witnesses', async () => {
```

arboleya (changes requested):

packages/account/src/providers/transaction-request/transaction-request.test.ts
arboleya: Oops, I thought this was ready for review.

Sorry for the noise.

danielbate: No problem, you caught it before review 🕵🏻

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -249,6 +266,23 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
     this.witnesses[index] = witness;
   }
 
+  /**
+   * Helper function to add an external signature to the transaction.
+   *
+   * @param account - The account to sign to the transaction.
```
petertonysmith94, line 272:
```suggestion
   * @param account - The account/s to sign to the transaction.
```

packages/account/src/connectors/fuel-connector.ts
```diff
@@ -104,6 +104,19 @@ export abstract class FuelConnector extends EventEmitter {
     throw new Error('Method not implemented.');
   }
 
+  /**
+   * Should start the sign transaction process and return
+   * the signed transaction.
+   *
+   * @param address - The address to sign the transaction
+   * @param message - The transaction to sign
```
petertonysmith94, line 112:
```suggestion
   * @param address - The address to sign the transaction
   * @param transaction - The transaction to sign
```

packages/account/src/account.ts
```diff
@@ -508,6 +508,22 @@ export class Account extends AbstractAccount {
     return this._connector.signMessage(this.address.toString(), message);
   }
 
+  /**
+   * Signs a transaction with the wallet's private key.
+   *
+   * @param transactionRequestLike - The transaction request to sign.
+   * @returns A promise that resolves to the signature of the transaction.
+   */
+  async signTransaction(hashedTransaction: TransactionRequestLike): Promise<string> {
```
petertonysmith94, line 517:
```suggestion
  async signTransaction(transactionRequestLike: TransactionRequestLike): Promise<string> {
```

to match the doc blocks

apps/docs-snippets/src/guide/cookbook/signing-transactions.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    expect(value.toNumber()).toEqual(1);
+    expect((await receiver.getBalance()).toNumber()).toEqual(amountToReceiver);
+  });
+
+  it('creates a transfer with external signer [predicate]', async () => {
+    const amountToReceiver = 100;
+
+    // #region multiple-signers-4
+    // #import { Predicate, BaseAssetId };
```
petertonysmith94, line 84:
```suggestion
    // #import { Predicate, BaseAssetId, ScriptTransactionRequest };
```

apps/docs-snippets/src/guide/cookbook/signing-transactions.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    receiver = WalletUnlocked.generate({
+      provider: sender.provider,
+    });
+  });
+
+  it('creates a transfer with external signer [script]', async () => {
+    const amountToReceiver = 100;
+
+    // #region multiple-signers-2
+    // #import { Script, BaseAssetId };
```
petertonysmith94, line 53:
```suggestion
    // #import { Script, BaseAssetId, FunctionInvocationResult };
```

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -326,7 +360,7 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
 
       // Insert a dummy witness if no witness exists
       if (typeof witnessIndex !== 'number') {
-        witnessIndex = this.createWitness();
+        witnessIndex = this.addEmptyWitness();
```
petertonysmith94, line 363: What condition would we want a dummy witness? Will these be used in production?

danielbate: A dummy witness is used for gas estimation. The correct witness needs to added at the end of the tx preparation flow as it hashes the entire transaction. So when we are just estimating, we just need to make sure we have added 32 bytes to get the correct estimation.

apps/docs-snippets/src/guide/cookbook/signing-transactions.test.ts
arboleya: That's one of the downsides of using `#{context,import}` on code snippets.

I bet we will never fully be able to account for this tho. (cc @Torres-ssf)

packages/account/src/account.ts
danielbate: Good spot

petertonysmith94: 🦅  👀

packages/account/src/wallet/wallet-unlocked.test.ts
```diff
@@ -49,18 +49,18 @@ describe('WalletUnlocked', () => {
   it('Sign a transaction using wallet instance', async () => {
     // #region wallet-transaction-signing
     // #import { WalletUnlocked, Signer };
+
     const provider = await Provider.create(FUEL_NETWORK_URL);
     const wallet = new WalletUnlocked(PRIVATE_KEY, provider);
```
petertonysmith94, line 54:
```suggestion
    const wallet = Wallet.fromPrivateKey(PRIVATE_KEY, provider);
```

Should we favour this instantiation method for users?

danielbate: Hmm as the constructor does not need to be private, I do not mind, however I do prefer the method so will update.

.changeset/tough-parrots-occur.md
```diff
@@ -0,0 +1,7 @@
+---
+"@fuel-ts/account": patch
+---
+
+- Implemented transaction signing through an Account
+- Add signers to a transaction request through `addAccountWitnesses`
+- Renamed `createWitness` -> `addEmptyWitness` and implemented `addWitness`
```
nedsalk, line 7: #1872 got merged and the changeset description should just be the title of the PR, "feat!: improve signing DX".

danielbate: Okay I'll make the change, but this still feels a bit off to me

danielbate: https://github.com/FuelLabs/fuels-ts/pull/1815/commits/4d9baea755c81d101e9ce1b1659188729f51fb46

nedsalk (dismissed):

petertonysmith94 (approved):

arboleya (approved):

Dhaiwat10 (approved):

arboleya (approved):

nedsalk (approved):

Torres-ssf (approved):

apps/docs-snippets/src/guide/cookbook/signing-transactions.test.ts
danielbate: This has been resolved by @Torres-ssf in #1926

PR #1817, opened 2024-02-27
fix: Remove UTXO flag 

calldelegation:
fuel-core binary's `--debug` flag conflicts with `--utxo-validation`


danielbate:
Closing this as disabling `utxo-validation` makes many of predicate tests fail (passing when they should be rejecting), therefore disabling this is dangerous.

arboleya:
Branch deleted. 🧹 🙂

PR #1818, opened 2024-02-28
chore: add update witness tests

danielbate:
Adding some unit tests around updating witnesses to a transaction requests following investigation into #1745 as well as reports of other unexpected behaviour from users.

I have also reorganised this test file slightly and removed some redundant `describe`.


packages/account/src/providers/transaction-request/transaction-request.test.ts
```diff
@@ -15,111 +18,143 @@ import { transactionRequestify } from './utils';
...
-    });
+    expect(inputA?.amount).toEqual(bn(700));
+    expect(inputB?.amount).toEqual(bn(300));
+    expect(inputBase?.amount).toEqual(bn(500));
   });
 
-  describe('fundWithFakeUtxos', () => {
-    it('should fund with the expected quantities', () => {
-      const transactionRequest = new ScriptTransactionRequest();
+  it('updates witnesses', () => {
```
danielbate, line 82: This is the new test.

nedsalk (dismissed):

packages/account/src/providers/transaction-request/transaction-request.test.ts
```diff
@@ -15,111 +18,143 @@ import { transactionRequestify } from './utils';
...
+        assetId: ASSET_A,
+      },
+    ]);
   });
 
-  describe('getCoinOutputsQuantities', () => {
-    it('should correctly map all the coin outputs to CoinQuantity', () => {
-      const transactionRequest = new ScriptTransactionRequest();
+  it('should return an empty array if there are no coin outputs', () => {
+    // Mock the getCoinOutputs method
```
arboleya, line 48: Where's the mock?

```suggestion
```

arboleya (approved):

packages/account/src/providers/transaction-request/transaction-request.test.ts
danielbate: Apologies because I removed a couple of the `describe`s, the history is all over the place. I did not alter that test. The only test I added was `updates witnesses`, but agreed there is no mock here.

packages/account/src/providers/transaction-request/transaction-request.test.ts
danielbate: So the issue I was actually trying to test for seems to be more of a browser performance issue than something wrong with the code, this test matches that of what the user was trying to do, on the level of `ScriptTransactionRequest`. However I am also making changes for #1745 that will impact services above `ScriptTransactionRequest`, but will go into more detail in that PR (when it is ready).

arboleya (approved):

nedsalk (approved):

PR #1822, opened 2024-02-29
feat: make requests extendable by adding `requestMiddleware` to provider options

### LINKED ISSUE #1821: Allow extension of request objects
Some of our apps like [block-explorer-v2](https://github.com/FuelLabs/block-explorer-v2) and [sway-playground](https://github.com/FuelLabs/sway-playground) aren't able to use sticky session cookies of the beta-5 network because there is no easy way to extend the request object to include the `"credentials":"include"` in it, which is necessary in this case to be included for the cookie mechanism to work.

nedsalk:
This PR introduces a `requestMiddleware` callback as a way to modify the request object for _every_ fetch call made towards the network. I didn't ponder about adding the capability for modifying the request object on a per-call basis because that'd be a bigger piece of work and we need at least this functionality for a bugfix on our apps related to sticky session cookies not being included. You can find more on that in the related issue.

closes #1821


packages/account/src/providers/provider.test.ts
```diff
@@ -1045,4 +1038,66 @@ describe('Provider', () => {
...
+  });
+
+  test('requestMiddleware modifies the request before being sent to the node [async]', async () => {
+    const fetchSpy = vi.spyOn(global, 'fetch');
+    await Provider.create(FUEL_NETWORK_URL, {
+      requestMiddleware: (request) => {
+        request.headers ??= {};
+        (request.headers as Record<string, string>)['x-custom-header'] = 'custom-value';
+        return Promise.resolve(request);
+      },
```
Dhaiwat10, line 1066: Was the async variation specifically requested by someone? Trying to understand the need for it / where it could be used

nedsalk: Somebody might want to fetch something from some other API to include in every call:
```ts
      requestMiddleware: (request) => {
        const credentials = await fetchSomeOtherCredentials();
        request.headers ??= {};
        (request.headers as Record<string, string>)['x-my-custom-credential'] = credentials.value;
        return Promise.resolve(request);
      }
```

nedsalk: I took inspiration for the async variation from `GraphQLRequest` which has the same `requestMiddleware` option that can be passed into it. I couldn't just use that library's functionalities, though, because I had to cover subscriptions, for which we have a custom implementation. This will have to be the case for each functionality we extend `fetch` with: we can never benefit from the various `graphql-request` features because we'll always have to roll out custom logic to cover subscriptions.

packages/account/src/providers/provider.ts
```diff
@@ -315,11 +310,18 @@ export default class Provider {
...
+      if (options.requestMiddleware) {
+        fullRequest = await options.requestMiddleware(fullRequest);
+      }
+
+      return fetch(url, fullRequest);
     }, retryOptions);
   }
 
   /**
+
```
Dhaiwat10, line 324: Might be unintentional

Dhaiwat10 (dismissed):
Nice one @nedsalk

Dhaiwat10:
I would change the PR title to something that includes the word '`requestMiddleware` just for more clarity

packages/account/src/providers/provider.ts
```diff
@@ -209,24 +209,19 @@ export type CursorPaginationArgs = {
   before?: string | null;
 };
 
-export type FetchRequestOptions = {
-  method: 'POST';
-  headers: { [key: string]: string };
```
danielbate, line 214: I think extending via `requestMiddleware` / `RequestInit` is definitely a better implementation, but was the use case of custom headers not already achievable here?

danielbate (commented):
Nice work @nedsalk. Should this be documented?

packages/account/src/providers/provider.ts
nedsalk: Yes, but the user would have to implement the whole `fetch`, whereas here they can use the existing fetcher and extend the request object as they please:
```ts
// overrides internal fetch
const provider = await Provider.create(url, {fetch: (fetchUrl, request) => {
  request.headers.something = "whaa"
}});

// modify request object but use internal fetch
const provider = await Provider.create(url, {requestMiddleware: (request) => {
  request.headers.something = "whaa"
}});
```

On the topic of custom `fetch`, IMO we should remove the ability for users to provide custom `fetch`es that override ours altogether because we always have to think about that scenario when extending our internal fetching capabilities.

Torres-ssf (commented):
I agree with @danielbate, we need some docs examples about this feature.

apps/docs-snippets/src/guide/provider/provider.test.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+/**
+ * @group node
+ */
+describe('Provider', () => {
+  it('can be given options', async () => {
+    // #region provider-options
+    await Provider.create(FUEL_NETWORK_URL, {
+      fetch: async (url, requestInit) => {
```
arboleya, line 14: Can we add the argument's type notation here?

apps/docs-snippets/src/guide/provider/provider.test.ts
```diff
@@ -0,0 +1,36 @@
...
+        // do something
+        await sleep(100);
+        return fetch(url, requestInit);
+      },
+      timeout: 2000,
+      cacheUtxo: 1500,
+      requestMiddleware: async (request) => {
```
arboleya, line 21: Same thing about argument's type.

packages/account/src/providers/provider.test.ts
```diff
@@ -209,7 +202,7 @@ describe('Provider', () => {
     const providerUrl2 = 'http://127.0.0.1:8080/graphql';
 
     const provider = await Provider.create(providerUrl1, {
-      fetch: (url: string, options: FetchRequestOptions) =>
+      fetch: (url, options) =>
```
arboleya, line 205: Same thing about argument's type.

apps/docs-snippets/src/guide/provider/provider.test.ts
```diff
@@ -0,0 +1,36 @@
+import { FUEL_NETWORK_URL, Provider, sleep } from 'fuels';
+
+async function fetchSomeExternalCredentials() {
+  return Promise.resolve('credential');
+}
+
+/**
+ * @group node
```
arboleya, line 8: Can we test this on the browser as well?

arboleya (changes requested):

packages/account/src/providers/provider.test.ts
nedsalk: I can see the benefit in adding the type annotations for the docs snippets, but here I prefer leaving it like this as it's an internal test and type inference would warn us if something was wrong.

apps/docs-snippets/src/guide/provider/provider.test.ts
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1822/commits/849a9ca1f42f33fa592f2f7ba51b6481dd850690

packages/account/src/providers/provider.test.ts
arboleya: The downside of needlessly relying on type inference is that it makes it harder for the next person to know the type since one can't simply _read_ the code to figure it out. This requires some extra pirouettes, like hovering the variable, and we don't have a type definition to click and go to.

What's the benefit of purposefully not adding the type notation?

nedsalk: Whoever wants to see the type of `options` can F12 into the `fetch` property and it'll take them directly to the function definition. 

The benefit is less code; we'd be repeating the type of the parameter while it's already defined and inferred in the body of the function. 
Also, doing any type annotation on a function's parameters that can be inferred can lead to a mismatch between the annotated type and the actual underlying type, which'd give compilation errors even though the function's implementation isn't even affected by the change (e.g. adding an additional property to the underlying type that's missing in the annotated type). The only way to _properly_ annotate in these cases and ensure that no compilation mismatch happens is like this:
```ts
{
      fetch: (...[url, options]: Parameters<NonNullable<ProviderOptions['fetch']>>) =>
       getCustomFetch('getVersion', { nodeInfo: { nodeVersion: url } })(url, options),
}
```
It looks silly, but it's the end conclusion of proper type annotation that doesn't give unnecessary compilation errors in the body of the function. One can say that it's easy to change the simple non-silly annotation to make TS happy, but I prefer TS not to interfere when there's no need to interfere (e.g. when the body of the function isn't affected).

In the end, it's a matter of style. I like my TS code inferred based on a single source of truth and not make any assumptions (annotations) in the source's consumers. If the consumer is incompatible with the source type, TS will yell.

arboleya: Less code, in this case, is irrelevant.

The docs exemplify the use of type annotation, and so should the source.

nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1822/commits/c7558749b792d935c5342badb1239fb8ff3035f0

arboleya (approved):

petertonysmith94 (approved):
Super clean 🥇

Torres-ssf (approved):

apps/docs-snippets/src/guide/provider/provider.test.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @group browser
+ */
+describe('Provider', () => {
```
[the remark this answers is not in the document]
nedsalk: Yeah, generally it should be but I experienced troubles using a debugger when tests use the __filename pattern so I always write the string. This was with `jest`, though. I never got the `vitest` debugger to work properly... I'll merge this now and will investigate this a bit later on.

PR #1826, opened 2024-03-04
feat!: accept predicate input data on the `Predicate` constructor

### LINKED ISSUE #1506: Move predicate data setting into constructor
The team agreed in one of the weekly syncs that it's more straightforward and easier to reason about predicate data if we'd be setting it _only_ via the `Predicate` constructor.
```ts
// current
const predicate = new Predicate(...);
predicate.setData(data);

// proposed
const predicate = new Predicate(..., data);
```

This implies removing the `predicate.setData` method.

Dhaiwat10:
## Summary
This PR contains  _BREAKING_ changes since the API for the `Predicate` constructor has changed:

### Breaking Change 1
```ts
// old API
const predicate = new Predicate(bytecode, provider, abi, configurableConstants);

// new API
const predicate = new Predicate({
  bytecode,
  abi, // optional
  provider,
  inputData, // optional
  configurableConstants, // optional
});
```

Notice how the `Predicte` constructor now accepts an _object_ with the following properties:

- `bytecode`: The bytecode of the predicate.
- `abi`: The JSON ABI of the predicate.
- `provider`: The provider for interacting with the predicate.
- `inputData`: The predicate input data (optional).
- `configurableConstants`: The configurable constants for the predicate (optional).

### Rationale
This change was made with readability and ease-of-use in mind, since having too many arguments in a 'flat' constructor can be confusing. Consider a scenario where you want to create a Predicate with configurables but no input data:

```ts
const predicate = new Predicate(
  bytecode,
  provider,
  abi,
  undefined,
  configurableConstants,
);
```

In this case, you would have to pass `undefined` as the `inputData` argument, which is not ideal. By using the object-based constructor, you can now pass an object with the properties you want to set, and the constructor will handle the rest:

```ts
const predicate = new Predicate({
  bytecode,
  abi,
  provider,
  configurableConstants,
});
```

I also chose the name `inputData` over `predicateData` or `data` because it describes the behaviour more accurately.

### Breaking Change 2
The `setData` method has been removed. If you want to pass in the predicate data _after_ instantiating the `Predicate` or if you want to use a different data than the one passed in the constructor, you will have to create a new `Predicate` instance.

Closes #1506


apps/docs/src/guide/predicates/send-and-spend-funds-from-predicates.md
```diff
@@ -18,19 +18,19 @@ Let's use the above predicate to validate our transaction.
 
 Once you've compiled the predicate (`forc build`), you'll obtain two important artifacts: the JSON ABI and the predicate's binary code. These are needed to instantiate a new predicate.
 
-<<< @/../../docs-snippets/src/guide/predicates/send-and-spend-funds-from-predicates.test.ts#send-and-spend-funds-from-predicates-2{ts:line-numbers}
+This is where we also pass in the predicate's data. Note that the `main` function in our predicate example requires a parameter called `input_address` of type `b256`. We will pass this parameter to the `Predicate` constructor along with the binary code and the JSON ABI.
```
danielbate, line 21:
```suggestion
This is where we also pass in the predicate's data. Note that the `main` function in our predicate example requires a parameter called `input_address` of type `b256`. We will pass this parameter to the `Predicate` constructor along with the bytecode and the JSON ABI.
```

packages/account/src/predicate/predicate.ts
```diff
@@ -43,12 +43,14 @@ export class Predicate<ARGS extends InputValue[]> extends Account {
...
    * @param jsonAbi - The JSON ABI of the predicate.
+   * @param predicateData - The predicate data.
    * @param configurableConstants - Optional configurable constants for the predicate.
    */
   constructor(
     bytes: BytesLike,
     provider: Provider,
     jsonAbi?: JsonAbi,
+    predicateData?: ARGS,
     configurableConstants?: { [name: string]: unknown }
```
danielbate, line 54: As we are introducing a breaking change here _anyway_, and we've now got 5 params, is it worth us using an object for the constructor predicate params. This _may_ defend against future breaking changes with the predicate constructor.

```ts
constructor(params: PredicateParams)
const { bytes, provider, jsonAbi, predicateData, configurableConstants } = params;
```

I appreciate this is a bigger breaking change, and your change would not impact anyone instantiating via the required params, but thought I'd suggest as we are now at 5 params.

packages/fuel-gauge/src/predicate/predicate-arguments.test.ts
```diff
@@ -140,20 +144,19 @@ describe('Predicate', () => {
...
 
       // #region predicate-struct-arg
-      // #context const predicate = new Predicate(bytecode, chainId, abi);
-      const tx = await predicate
-        .setData({
-          has_account: true,
-          total_complete: 100,
-        })
-        .transfer(receiver.address, amountToReceiver, BaseAssetId, { gasPrice, gasLimit: 10_000 });
+      // #context const predicate = new Predicate(bytecode, chainId, abi, [{ has_account: true, total_complete: 100 }]);
```
danielbate, line 155: Using context seems unnecessary here, we could just use the actual predicate instantiation and ensure future compatibility.

danielbate (commented):
Looking good, couple comments 💪🏻

apps/docs-snippets/src/guide/predicates/predicate-with-configurable.test.ts
```diff
@@ -28,10 +28,13 @@ describe(__filename, () => {
...
-
-    // set predicate data to be the same as the configurable constant
-    predicate.setData(configurable.WHITELISTED);
+    const predicate = new Predicate(
+      bin,
+      wallet.provider,
+      abi,
+      [configurable.WHITELISTED],
+      configurable
+    );
```
arboleya, line 37: Could we first set the data to a variable to better disambiguate the parameters?

```ts

    // we'll set the predicate data to be the same as the configurable constant
    const data = [configurable.WHITELISTED];

    // instantiate predicate with data and configurable constants
    const predicate = new Predicate(bin, wallet.provider, abi, data, configurable);
```

packages/account/src/predicate/predicate.ts
arboleya: Fair point. The breaking change would be more significant, but the API would be more straightforward.

It's a delicate one; we should think this through.

nedsalk: If we are going to go down that route then I'd suggest also removing the `setData` method. It'd break everybody's `Predicate` usage anyways, so it makes sense to introduce another breaking change that's in the same vein.

packages/account/src/predicate/predicate.ts
```diff
@@ -43,12 +43,14 @@ export class Predicate<ARGS extends InputValue[]> extends Account {
    * @param bytes - The bytes of the predicate.
    * @param provider - The provider used to interact with the blockchain.
    * @param jsonAbi - The JSON ABI of the predicate.
+   * @param predicateData - The predicate data.
    * @param configurableConstants - Optional configurable constants for the predicate.
    */
   constructor(
     bytes: BytesLike,
     provider: Provider,
     jsonAbi?: JsonAbi,
```
Torres-ssf, line 52: @Dhaiwat10 I reckon we must make the `jsonAbi` parameter required on the Predicate constructor.

The parameter is later used on `processPredicateData` which will throw an error if `jsonAbi` is undefined.

By doing so we will also be able to rearrange the Predicate's constructor parameters order more conveniently.

Torres-ssf (changes requested):

packages/account/src/predicate/predicate.ts
Dhaiwat10: @danielbate TBH I'm in favour of using objects for the constructor params too. Judging from the sync, it looks like this is the consensus between the team, so I'll go ahead and incorporate this change.

Dhaiwat10: @nedsalk I'm sorry but I don't follow. You would still need `setData` if you want to pass in some different data after instantiating the predicate. I don't mind the breaking change, but I feel like this method is needed.

If we remove this method, the only way to pass in different data to a predicate would be to re-instantiate it with different data. I don't think that DX would be great. 

I personally think of a `Predicate` instance as a 'pointer' to a predicate. Kinda like how a `Contract` instance is a pointer to a contract. You can call different functions on the contract once you have the instance, and you don't re-instantiate it for every function call.

For predicates, the different function calls are just different inputs:
image

packages/account/src/predicate/predicate.ts
Dhaiwat10: @Torres-ssf I think this could be worth handling in a separate PR, but regardless, I don't think making `jsonAbi` a required property would change the order would it? 🤔

Dhaiwat10:
Please re-visit the PR description whenever you get the chance to review the PR again, guys. I have added some context and rationale there.

packages/account/src/predicate/predicate.ts
nedsalk: @Dhaiwat10 I mentioned it because we kinda had an agreement with the frontend team in one of our syncs that it makes sense to just have it in the constructor and remove `setData`.

The framing you presented is valid. As far as I remember, the logic behind the suggestion was that `predicateData` is mostly seen as an equivalent to a private key for a walletn:
- If you don't have the private key that unlocks the wallet, you can't instantiate the unlocked wallet
- If you don't have the data that unlocks the predicate, there's no point in instantiating the predicate itself

The benefit of not having `setData` from the internal maintenance standpoint is that state management is easier when we know that the data is set _only_ once. Just like we don't allow the changing of a `BaseWalletUnlocked`'s private key but rather you have to instantiate a new `BaseWalletUnlocked`, we wouldn't allow the changing of predicate data.

packages/account/src/predicate/predicate.ts
Torres-ssf: @Dhaiwat10 we can make the `jsonAbi` to be the second parameter if we make it a required parameter. 

TS does not allow a required function parameter to be after an optional parameter. That is why the `predicate` parameter is the second parameter.

Since we are already updating the predicate constructor parameters, I believe this should be done within this PR's scope.

packages/account/src/predicate/predicate.ts
Dhaiwat10: Sounds fair @nedsalk. If this has already been discussed with the frontend team, then I will remove it

packages/account/src/predicate/predicate.ts
Dhaiwat10: @Torres-ssf After some discussions on this PR, the constructor now accepts an object though. Do you think making the `abi` required is still needed?

arboleya: @Dhaiwat10 AFAIK, ABIs should be mandatory for any interaction _[with Contracts]_, as the SDK won't be able to decode anything (errors, logs, outputs) without them. Isn't this also the case for Scripts and Predicates?

Dhaiwat10: @arboleya That _does_ make sense but I insist on handling this change via a separate PR because there are several cases throughout our codebase where the ABI for predicates is not passed / is optional, like this one: https://github.com/FuelLabs/fuels-ts/blob/b1bcfd26a3201c2514479382ed06d6c93d853a3b/packages/fuel-gauge/src/predicate/predicate-evaluations.test.ts#L38

I want to carefully understand why ABIs for predicates were kept optional in the first place.

apps/docs-snippets/src/guide/predicates/index.test.ts
```diff
@@ -18,7 +18,11 @@ describe(__filename, () => {
     // #import { Predicate, Provider, FUEL_NETWORK_URL };
 
     const provider = await Provider.create(FUEL_NETWORK_URL);
-    const predicate = new Predicate(binary, provider, jsonAbi);
+    const predicate = new Predicate({
+      bytecode: binary,
+      provider,
+      abi: jsonAbi,
+    });
```
arboleya, line 25: Would assigning the parameters to a variable and then using the variable in the constructor be more descriptive? I am thinking mainly about the type annotation, but it may not be necessary.

```ts
const predicateParams: Type = { .. }
```

apps/docs-snippets/src/guide/predicates/predicate-with-configurable.test.ts
```diff
@@ -28,10 +28,13 @@ describe(__filename, () => {
...
-
-    // set predicate data to be the same as the configurable constant
-    predicate.setData(configurable.WHITELISTED);
+    const predicate = new Predicate({
+      bytecode: bin,
+      provider: wallet.provider,
+      abi,
+      inputData: [configurable.WHITELISTED],
+      configurableConstants: configurable,
+    });
```
arboleya, line 37: Maybe we could use the generic in `Predicate<T>` for demo purposes if that makes sense.

packages/abi-typegen/src/templates/predicate/factory.hbs
```diff
@@ -57,11 +57,17 @@ export class {{capitalizedName}}__factory {
   static readonly abi = _abi
   static readonly bin = _bin;
 
-  static createInstance(provider: Provider, configurables?: {{capitalizedName}}Configurables) {
+  static createInstance(provider: Provider, predicateData?: InputValue[], configurables?: {{capitalizedName}}Configurables) {
```
arboleya, line 60: I just noticed our typegen template doesn't consider the generic type, but should it? 🤔

packages/account/src/predicate/predicate.ts
```diff
@@ -28,6 +28,14 @@ import type {
 
 import { getPredicateRoot } from './utils';
 
+type PredicateConstructorArgs<T> = {
```
arboleya, line 31: If I'm not mistaken, we use the suffix params in these cases, no?

I am unsure; feel free to dismiss if that's not the case.

Here's a suggestion:

```suggestion
type PredicateParams<T> = {
```

packages/account/src/predicate/predicate.ts
```diff
@@ -40,27 +48,32 @@ export class Predicate<ARGS extends InputValue[]> extends Account {
...
-    jsonAbi?: JsonAbi,
-    configurableConstants?: { [name: string]: unknown }
-  ) {
+  constructor({
+    bytecode,
+    abi,
+    provider,
+    inputData,
+    configurableConstants,
+  }: PredicateConstructorArgs<ARGS>) {
```
arboleya, line 63: Not your doing, but here's another suggestion if you like it:

```suggestion
  }: PredicateConstructorArgs<TInputData>) {
```

arboleya (commented):
I left a couple of minor nits; see what you think.

packages/account/src/predicate/predicate.ts
arboleya: That's alright, but please create a follow-up issue so we don't forget it. 🙂

Dhaiwat10: #1843

apps/docs-snippets/src/guide/predicates/index.test.ts
Dhaiwat10: I think it might not be a good idea in this case because doing this would require us to pass in a shape for the predicate data which might introduce more confusion: 
image

Dhaiwat10: Or maybe we should always encourage people to pass in the input type 🤔

Dhaiwat10: I think it could be kinda confusing because the class automatically infers the type if you do pass in `inputData` like so:

image

packages/abi-typegen/src/templates/predicate/factory.hbs
Dhaiwat10: I think if you pass in `predicateData`, the class will automatically infer the type no? 🤔

nedsalk: > I think if you pass in predicateData, the class will automatically infer the type no? 🤔

I don't think so. The way it is currently, the returned predicate type would be `Predicate<InputValue[]>`.
I believe that it should be something like this for type inference to work properly:

```hbs
static createInstance<Data extends InputValue[]>(provider: Provider, predicateData?: Data, configurables?: {{capitalizedName}}Configurables) {
    const { abi, bin } = {{capitalizedName}}__factory

    return new Predicate<Data>({
      bytecode: bin,
      abi,
      provider,
      inputData: predicateData,
      configurableConstants: configurables,
    })
}
```

Dhaiwat10: Created an issue for this: https://github.com/FuelLabs/fuels-ts/issues/1850

apps/docs-snippets/src/guide/predicates/predicate-with-configurable.test.ts
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/1826/commits/db8be677225d5f6be0a984f75d3f71fa78405a2e

packages/fuel-gauge/src/predicate/predicate-arguments.test.ts
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/1826/commits/2b103dc0a5e7264eb84bb3e945024841ee4d78d7

arboleya (dismissed):

.changeset/many-timers-tease.md
```diff
@@ -0,0 +1,54 @@
...
+  provider,
+  inputData, // optional
+  configurableConstants, // optional
+});
+```
+
+Notice how the `Predicte` constructor now accepts an _object_ with the following properties:
+
+- `bytecode`: The bytecode of the predicate.
+- `abi`: The JSON ABI of the predicate.
```
danielbate, line 26:
```suggestion
- `abi`: The JSON ABI of the predicate (optional).
```

.changeset/many-timers-tease.md
danielbate, line 1: Really enjoying your release notes recently 🔥

packages/account/src/predicate/predicate.ts
```diff
@@ -28,39 +28,52 @@ import type {
...
+
 /**
  * `Predicate` provides methods to populate transaction data with predicate information and sending transactions with them.
  */
-export class Predicate<ARGS extends InputValue[]> extends Account {
+export class Predicate<TInputData extends InputValue[]> extends Account {
   bytes: Uint8Array;
   predicateData: Uint8Array = Uint8Array.from([]);
-  predicateArgs: ARGS = [] as unknown as ARGS;
+  predicateArgs: TInputData = [] as unknown as TInputData;
```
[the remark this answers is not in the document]
danielbate: IMO this is confusing because they aren't really describing their function correctly. We should probably have `predicateData` and `predicateDataBytes`.

danielbate (changes requested):

packages/fuel-gauge/src/predicate/predicate-arguments.test.ts
```diff
@@ -215,14 +230,18 @@ describe('Predicate', () => {
...
+        inputData: [20, 30],
+      });
 
       const initialPredicateBalance = await fundPredicate(wallet, predicate, amountToPredicate);
       const initialReceiverBalance = await receiver.getBalance();
 
       // #region predicate-multi-args
-      // #context const predicate = new Predicate(bytecode, chainId, abi);
-      predicate.setData(20, 30);
+      // #context const predicate = new Predicate(bytecode, chainId, abi, [20, 30]);
```
petertonysmith94, line 244: This appears to be the incorrect signature for `Predicate` :)

packages/fuel-gauge/src/predicate/predicate-arguments.test.ts
```diff
@@ -240,18 +259,48 @@ describe('Predicate', () => {
       );
     });
 
+    it('calls a predicate with valid multiple arguments and returns true - using setData', async () => {
```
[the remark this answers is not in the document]
petertonysmith94: Are we asserting that it returns true?

packages/fuel-gauge/src/predicate/predicate-arguments.test.ts
```diff
@@ -240,18 +259,48 @@ describe('Predicate', () => {
...
+        predicate,
+        receiver,
+        initialPredicateBalance,
+        initialReceiverBalance,
+        amountToPredicate,
+        amountToReceiver
+      );
+    });
+
     it('calls a predicate with invalid multiple arguments and returns false', async () => {
```
petertonysmith94, line 289:
```suggestion
    it('calls a predicate with invalid multiple arguments and throws error', async () => {
```

petertonysmith94 (commented):
Just some minor comments

.changeset/many-timers-tease.md
Dhaiwat10: Appreciate it @danielbate

packages/account/src/predicate/predicate.ts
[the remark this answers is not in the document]
Dhaiwat10: Good point. I've renamed the variables but I noticed that `predicateData`, now named `predicateDataBytes` is never actually _set_ anywhere. 👁️

packages/fuel-gauge/src/predicate/predicate-arguments.test.ts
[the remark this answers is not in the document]
Dhaiwat10: @petertonysmith94 yes, we assert that by asserting the balances. The balances will be updated only if it returns true and the transfer goes through

packages/fuel-gauge/src/predicate/predicate-arguments.test.ts
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/1826/commits/4f921ed92fb41eb7813b02d4dd1d4fd6d4977c36

danielbate (approved):

packages/account/src/predicate/predicate.ts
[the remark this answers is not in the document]
arboleya: @Dhaiwat10 Why do we need it, then?

Will this always use an empty `Uint8Array`?

https://github.com/FuelLabs/fuels-ts/blob/d0c381e53bcd3ea3528ab4accdb44fd87fadca95/packages/account/src/providers/transaction-request/transaction-request.ts#L342

https://github.com/FuelLabs/fuels-ts/blob/d0c381e53bcd3ea3528ab4accdb44fd87fadca95/packages/account/src/providers/transaction-request/transaction-request.ts#L386

I'm confused.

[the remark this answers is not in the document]
nedsalk: The nomenclature for input coins/messages has leaked into the `Predicate` class and IMO a better naming for the `predicate` and `predicateData` properties from the user's POV is `address` (of type `string | Uint8Array`) and `data` (type depends on abi). @Dhaiwat10 sorry for opening yet another topic on this PR, but given that we're breaking stuff this is the ideal opportunity to discuss it.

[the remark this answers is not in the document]
Dhaiwat10: @arboleya I am confused as well. I can remove it and see if all tests pass if you want.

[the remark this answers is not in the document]
Dhaiwat10: @nedsalk Which `predicate` variable are you referring to here?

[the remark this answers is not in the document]
nedsalk: @nedsalk Sorry, the `predicate` property of input coins/messages leaked into my comment as well :sweat_smile: I meant the `bytes` property of `Predicate` that is used to set the `predicate` property of input coins/messages.

petertonysmith94 (approved):

packages/account/src/predicate/predicate.ts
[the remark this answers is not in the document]
Dhaiwat10: In the interest of merging this PR asap since it needs to unblock some follow-up issues

1. I created an issue for the investigation of the seemingly unset class property: https://github.com/FuelLabs/fuels-ts/issues/new
2. @nedsalk I suggest we open a discussion to discuss the nomenclature for the `Predicate` class in general?

PR #1827, opened 2024-03-04
fix: disallow transferring amounts <=0

### LINKED ISSUE #1577: Disallow transferring `<= 0` of an asset
The code below succeeds:
```ts
await wallet.transfer(
      destination.address,
      -1,
      BaseAssetId,
      { gasPrice, gasLimit: 10_000 }
    );
```
We shouldn't allow such inputs. Negative and zero values are generally allowed to be passed as values - they're just numbers after all, but our `transfer` method shouldn't be able to transfer `<= 0` of an asset.

Dhaiwat10:
## Summary
This PR introduces a new error to the `transfer` and `transferToContract` methods on the `Account` class which appears when a user tries to transfer an amount that's less than or equal to zero, as requested in the original issue.

Closes #1577


packages/account/src/account.test.ts
```diff
@@ -623,4 +623,17 @@ describe('Account', () => {
       await result.wait();
     }).rejects.toThrowError(/Gas limit '0' is lower than the required: ./);
   });
+
+  it('should throw when trying to transfer a zero or negative amount', async () => {
+    const sender = await generateTestWallet(provider, [[10_000, BaseAssetId]]);
+    const receiver = Wallet.generate({ provider });
+
+    await expect(async () => {
```
nedsalk, line 631: You should use `expectToThrowFuelError` here.

danielbate (dismissed):

.changeset/perfect-feet-approve.md
```diff
@@ -0,0 +1,6 @@
+---
+"@fuel-ts/account": minor
+"@fuel-ts/errors": minor
```
arboleya, line 3: If the API is the same, and we are just adding new validations, shouldn't this be a patch?

nedsalk (approved):

PR #1764, opened 2024-02-15
chore!: upgrade `fuel-core` to `0.23.0`

Torres-ssf:
- Closes: https://github.com/FuelLabs/fuels-ts/issues/1831
- Closes: https://github.com/FuelLabs/fuels-ts/issues/1700
- Closes: https://github.com/FuelLabs/fuels-ts/issues/1738

Current Tests Passing:
Screenshot 2024-04-02 at 16 12 11

## Breaking Changes
- Requires `Forc` `0.51.1`
- Setting `maxFee` is always required now ( The SDK estimates it by default with `Provider.getTransactionCost` + `Account.fund` )
- `gasPrice` is calculated by the VM so we do not need to inform it anymore. The Policy `gasPrice` [was replaced](https://github.com/FuelLabs/fuels-ts/pull/1764/commits/3ad83744048ce4b4a28a6151048adcdf27aa9566) by `tip`
- `Account.fund` signature has [changed](https://github.com/FuelLabs/fuels-ts/pull/1764/commits/0851940f91620696fe4b481aac63fc3b4ad340ab)
- Graphql URL now includes a versioning path: `http://127.0.0.1:4000/v1/graphql` 
- `calculateTransactionFee` now requires [tip](https://github.com/FuelLabs/fuels-ts/pull/1764/commits/277ead08b124651224b4b8b1934c701b59471f3e), [maxGasPerTx](https://github.com/FuelLabs/fuels-ts/pull/1764/commits/eeb5b1333006253f8fbbb48509b62aca0c798f39), and [gasPrice](https://github.com/FuelLabs/fuels-ts/pull/1764/commits/499d78b887b8779fbd6913ecca710cc5fad3f83c)
- `calculateTransactionFee` does not receive [gasUsed](https://github.com/FuelLabs/fuels-ts/pull/1764/commits/499d78b887b8779fbd6913ecca710cc5fad3f83c) anymore
- [removed](https://github.com/FuelLabs/fuels-ts/pull/1764/commits/87bb4411ef1e1ec767c9fe811ab515240b854182) unneeded properties returned by `Provider.getTransactionCost`

PR #1839, opened 2024-03-06
chore: add try catch when parsing stream response

Torres-ssf:
Adding this because of: https://github.com/FuelLabs/fuels-ts/pull/1764/commits/ccb706ef729928884b69d65598f82115a8443cb5

The stream text data response was slightly modified on `fuel-core` `0.23.0` resulting in an error being thrown without much context. 

With this, we can avoid lost time debugging it in the future if that happens again.

UPDATE:

The stream text data response did not change. The problem was another `data:` present on the text data, resulting in the `String.split` not working as expected.


arboleya (dismissed):
Can we somehow provoke this error and have a test for this?

> **EDIT**: This sort of thing (https://github.com/FuelLabs/fuels-ts/commit/ccb706ef729928884b69d65598f82115a8443cb5) should ideally be done with a `RegExp` instead of `split/[n]` or other string manipulation workarounds.

Dhaiwat10 (dismissed):

arboleya (dismissed):

packages/account/src/providers/provider.test.ts
```diff
@@ -1045,4 +1047,79 @@ describe('Provider', () => {
 
     await Promise.all(promises);
   });
+
+  it('should not throw if the subscrtion stream data string contains more than one "data:"', async () => {
```
petertonysmith94, line 1051: subscrtion -> subscription

packages/account/src/providers/provider.test.ts
```diff
@@ -1045,4 +1047,79 @@ describe('Provider', () => {
...
+      adminWallet.sendTransaction(txRequest, {
+        awaitExecution: true,
+        estimateTxDependencies: false,
+      })
+    ).resolves.toBeTruthy();
+
+    vi.restoreAllMocks();
+  });
+
+  it('should throw an error if stream data string parsing fails for some reason', async () => {
```
petertonysmith94, line 1086: Would suggest `should throw if the subscription stream data...` to be consistent

packages/account/src/providers/provider.test.ts
```diff
@@ -1045,4 +1047,79 @@ describe('Provider', () => {
 
     await Promise.all(promises);
   });
+
+  it('should not throw if the subscription stream data string contains more than one "data:"', async () => {
```
nedsalk, line 1051: Here is a simpler, more focused test that tests the same thing:
```ts
  it('should not throw if the subscription stream data string contains more than one "data:"', async () => {

    vi.spyOn(global, 'fetch').mockImplementationOnce(() => {
      const responseObject = {
        data: {
          submitAndAwait: {
            type: 'SuccessStatus',
            time: 'data: 4611686020137152060',
          },
        },
      };
      const streamedResponse = new TextEncoder().encode(
        `data:${JSON.stringify(responseObject)}\n\n`
      );
      return Promise.resolve(
        new Response(
          new ReadableStream({
            start: (controller) => {
              controller.enqueue(streamedResponse);
              controller.close();
            },
          })
        )
      );
    });

    for await (const { submitAndAwait } of provider.operations.submitAndAwait({
      encodedTransaction: "it's mocked so doesn't matter",
    })) {
      expect(submitAndAwait.type).toEqual('SuccessStatus');
      // @ts-expect-error time isn't inferred
      expect(submitAndAwait.time).toEqual('data: 4611686020137152060');
    }
  });
```

packages/account/src/providers/provider.test.ts
```diff
@@ -1045,4 +1047,79 @@ describe('Provider', () => {
...
+      adminWallet.sendTransaction(txRequest, {
+        awaitExecution: true,
+        estimateTxDependencies: false,
+      })
+    ).resolves.toBeTruthy();
+
+    vi.restoreAllMocks();
+  });
+
+  it('should throw an error if stream data string parsing fails for some reason', async () => {
```
nedsalk, line 1086: Same as above, the test can be simplified:
```ts
  it('should throw an error if stream data string parsing fails for some reason', async () => {

    const badResponse = 'data: whatever';
    vi.spyOn(global, 'fetch').mockImplementationOnce(() => {
      const streamResponse = new TextEncoder().encode(badResponse);
      return Promise.resolve(
        new Response(
          new ReadableStream({
            start: (controller) => {
              controller.enqueue(streamResponse);
              controller.close();
            },
          })
        )
      );
    });

    await expectToThrowFuelError(
      async () => {
        // eslint-disable-next-line @typescript-eslint/no-unused-vars
        for await (const { submitAndAwait } of provider.operations.submitAndAwait({
          encodedTransaction: "it's mocked so doesn't matter",
        })) {
          // shouldn't be reached!
          expect(true).toBeFalsy();
        }
      },
      {
        code: FuelError.CODES.STREAM_PARSING_ERROR,
        message: `Error while parsing stream data response: ${badResponse}`,
      }
    );
  });
```

nedsalk (changes requested):
Good stuff. I requested changes for test simplification.

nedsalk (approved):

packages/account/src/providers/provider.test.ts
```diff
@@ -1039,6 +1040,76 @@ describe('Provider', () => {
...
+            },
+          })
+        )
+      );
+    });
+
+    await expectToThrowFuelError(
+      async () => {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        for await (const { submitAndAwait } of provider.operations.submitAndAwait({
```
arboleya, line 1099: Nit.

```suggestion
        for await (const { submitAndAwait: _ } of provider.operations.submitAndAwait({
```

arboleya (approved):

packages/account/src/providers/provider.test.ts
arboleya: It makes use of [this](https://github.com/FuelLabs/fuels-ts/blob/950f9be3856bbd00ee11c1d41f38729e0ea9c653/.eslintrc.js#L63-L68) rule exception.

Torres-ssf: TS didn't accept that 😢

PR #1832, opened 2024-03-05
feat!: support `v1` encoding in program types

danielbate:
Closes #1739 
Closes #1749 

- [x] Support `v1` decode in scripts
- [x] Support `v1` encode in scripts
- [x] support `v1` encode/decode in predicates
- [x] Support `v1` encode/decode in contracts
- [x] Support str slice (enabling #823)
- [x] Nested heaps (enabling #1880)

This PR is using `forc@0.52.1` for our experimental forc version.

#1187 is currently blocked sway side and will be supported later.

Issue #1556, opened 2023-12-19
Unhelpful and unorganized error messages


arboleya:
We should organize all error messages and properly document them, so there is a consistent, helpful and descriptive set of error messages across the sdk instead of ones I experienced like "too small" with no value or details flagged.

> [!NOTE]
> - Incoming from https://github.com/FuelLabs/fuels-ts/discussions/1545

petertonysmith94:
@danielbate please tag me please

petertonysmith94:
**Documentation**
- [X] Integrate the `@fuels-ts/errors` package with typedoc.
- [ ] Add typedocs for all currently thrown `FuelError`.
- [ ] Add typedocs for `DECODE_ERROR` (skipped for now).
- [ ] Add typedocs for `ENCODE_ERROR` (skipped for now).
- [ ] 

**Further review required**
- [ ] Find all instances of `new Error`, and replace with a FuelError with appropriate error code.
- [ ] 

**Discussion worthy**
- Decoding and encoding error handling has room for improvement, would be good to get peoples thoughts on handling these errors.

PR #1840, opened 2024-03-06
feat: log out custom require messages for error enums when tx reverts

theausicist:
Unfortunately, the SDK does not support debugging revert errors with string messages - at least temporarily. This is because the SDK does not support decoding of Sway `str` slices in the `v0` Sway encoding scheme. This problem will be fixed soon once the `v1` Sway encoding scheme is adopted.

But for now, if your Sway contract has a revert error with a string message like this:

```sw
fn test_function() -> bool {
    require(false, "This is a revert error");
    true
}
```

The SDK will only log out a generic error: `String slices can not be decoded from logs.`

This PR enables the logging of the messages associated with custom error enums in Sway, and recommends users to use these enums to debug contracts that require multiple require statements. After this PR, require reverts using error enums will be logged out like this:

`The script reverted with reason RequireFailed. (Reason: "InvalidInput")`


Dhaiwat10:
@theausicist thanks for this. I had just started investigating the problem. Your PR definitely has some useful insights that will help me. Did you verify that this works though? I tried copying your solution to a local test environment that I have and I did not see any changes. Just curious if you had tested this by any chance :D

theausicist:
> @theausicist thanks for this. I had just started investigating the problem. Your PR definitely has some useful insights that will help me. Did you verify that this works though? I tried copying your solution to a local test environment that I have and I did not see any changes. Just curious if you had tested this by any chance :D

I directly modified/patched the built `index.js` within the `node_modules` of the `@fuel-ts` package, and tried to replicate the changes within the raw code here. It's possible I missed something. But like I said, it is _hacky_! [Here's the full `index.js` file](https://codefile.io/f/Lzt9nB08so) (in `node_modules/@fuel-ts/program/dist/index.js`) and the [original build](https://codefile.io/f/Ckc262Eg2X)

petertonysmith94:
Relates to #1556

Dhaiwat10:
@theausicist that's actually super helpful. Can I ask what version of the SDK that code is from?

theausicist:
> @theausicist that's actually super helpful. Can I ask what version of the SDK that code is from?

`v0.76.0`

danielbate:
@Dhaiwat10 IMO regardless of where this issue was coming from (i.e. adding str slice support), adding logs to an error message, probably with some formatting or interpolation, is a good idea. Are you implementing this yourself or should we review this PR?

Dhaiwat10:
@theausicist I tried replacing my `node_modules/@fuel-ts/program/dist/index.js` with the file that you provided and I'm still only seeing this error:

```
FuelError: String slices can not be decoded from logs. Convert the slice to `str[N]` with `__to_str_array`
```

:( I'm gonna continue investigating to see if there is some other way to log out the reason string, but the approach you suggested doesn't seem to be working. v1 encoding solves this anyways, but that's still several weeks away.

theausicist:
> @theausicist I tried replacing my `node_modules/@fuel-ts/program/dist/index.js` with the file that you provided and I'm still only seeing this error:
> 
> ```
> FuelError: String slices can not be decoded from logs. Convert the slice to `str[N]` with `__to_str_array`
> ```
> 
> :( I'm gonna continue investigating to see if there is some other way to log out the reason string, but the approach you suggested doesn't seem to be working. v1 encoding solves this anyways, but that's still several weeks away.

you'll get this error because you're using raw error strings in the `require` method which fails with the error you mentioned above. If you use an Error enum, fuel-ts can decode this nicely, following which this PR can help.

instead of:
```sway
require(false, "Raw Error");
```
do:
```sway
enum Error {
    DecodeableError: ()
}

...
require(false, Error::DecodeableError);
```

Dhaiwat10:
@theausicist I see what you mean. I just tried this out and it works. :) 

@FuelLabs/sdk-ts Adding support for decoding str slices in the v0 scheme doesn't seem to be all that feasible, especially since we are gonna move to v1 very soon. I think a fair compromise would be supporting at least error enums as mentioned in @theausicist's comment above, and recommending the usage of error enums until v1 support drops with a notice in our docs.

Please let me know if this sounds fair, or if anyone has any other ideas.

apps/docs/src/guide/errors/debugging-revert-errors.md
```diff
@@ -0,0 +1,27 @@
+# Debugging Revert Errors
+
+Unfortunately, the SDK does not support debugging revert errors with string messages - at least temporarily. This is because the SDK does not support decoding of Sway `str` slices in the `v0` Sway encoding scheme. This problem will be fixed soon once the `v1` Sway encoding scheme is adopted.
```
Torres-ssf, line 3: @Dhaiwat10 If we tell users to use `__to_str_array()` it will be possible to decode the string error from the logs, like:

```rs
    fn test_function() -> bool {
        require(false, __to_str_array("This is a revert error"));
        true
    }
```

theausicist: this is also a "hack" because it won't work for slightly longer arbitrary-length strings. Per my observations, AbiEncoder/AbiDecoder within the Sway lib needs to implement the trait for str[32] onwards. 

may be slightly inaccurate with the numbers here...

Torres-ssf: @theausicist In my tests, I was able to log a string with a length of 280:

Screenshot 2024-03-13 at 16 05 12

Everything works when decoding the log receipts:

Screenshot 2024-03-13 at 16 11 20

I don't see it as a hack since a `str[280]` is enough for an error message.

Dhaiwat10: Great suggestion @Torres-ssf

apps/docs-snippets/src/guide/errors/debugging-revert-errors.test.ts
nedsalk, line 1: 1. These tests could be put into `fuel-gauge/src/revert-error.test.ts`
2. Both of these tests are working with contracts. Can you write a test for scripts? I'm unsure if it'll succeed, though. If it turns out to require more work to cover this for scripts, you can create an issue for that and rename this PR into `feat: log out custom require messages for error enums when tx reverts (in contract calls)`.

apps/docs/src/guide/errors/debugging-revert-errors.md
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/1840/commits/2f5e137c64486b34a6f847b17fb0dd9afca5fdfa

apps/docs-snippets/src/guide/errors/debugging-revert-errors.test.ts
Dhaiwat10: @nedsalk 
1. I put these tests into the `docs-snippets` folder because I think the convention in our docs is to reference snippets from `docs-snippets` over `fuel-gauge` wherever possible. 

2. Let me check that

Dhaiwat10: It works for scripts: https://github.com/FuelLabs/fuels-ts/pull/1840/commits/c3fe9ec162c44ed6801fb475fda34a8318501f1c

packages/program/src/revert/revert-error.ts
```diff
@@ -62,8 +62,18 @@ export class RevertError extends Error {
...
    */
-  constructor(receipt: TransactionResultRevertReceipt, reason: RevertReason) {
-    super(`The script reverted with reason ${reason}`);
+  constructor(receipt: TransactionResultRevertReceipt, reason: RevertReason, logs: Array<unknown>) {
+    // @TODO: this is a "hack"
+    // motivation: a custom error revert string is the only "string" type item in the `logs` array
+    // the `logs.filter for string` will make sense only for `RequireFailed` errors, so could potentially
+    // just restrict the following to `RequireFailed` class by modifying `this.message`
+    // produces error messages like so:
+    // ```RequireRevertError: The script reverted with reason RequireFailed (reason: "FailImplicitly")```
```
petertonysmith94, line 71: Not a fan of todo's in the codebase, would advise removing and creating an issue to follow up
```suggestion

```

packages/program/src/revert/revert-error.ts
```diff
@@ -62,8 +62,18 @@ export class RevertError extends Error {
...
+    super(
+      `The script reverted with reason ${reason}. (Reason: "${logs.filter(
+        (l) => typeof l === 'string'
+      )}")`
```
petertonysmith94, line 75:
```suggestion
      `The script reverted with reason ${reason}. (Reason: "${this.extractErrorFromLogs(logs)}")`
```

Might be easier to read for the end consumer

petertonysmith94 (commented):
Noticed the type for `logs` higher in the dependency chain  is `Array<any>`. Pretty sure this isn't going to cause issues, just thought I'd raise it. 

Examples:
- https://github.com/FuelLabs/fuels-ts/blob/f5466b6c3972d98c3cf76dd192c1f61c99498687/packages/program/src/script-request.ts#L113
- https://github.com/FuelLabs/fuels-ts/blob/ef168fc95c830955f09f95b35e09f9cafae42662/packages/program/src/contract-call-script.ts#L178

packages/program/src/revert/revert-error.ts
Dhaiwat10: I think we can just remove it then. We don't need a follow-up because we already have proper support for the decoding of str slices in the v1 encoding scheme.

apps/docs/src/guide/errors/index.md
```diff
@@ -0,0 +1,3 @@
+# Errors
+
+<!-- Will replace this file's contents with the content from PR #1866 once it's merged -->
```
Dhaiwat10, line 3: Don't merge before #1866 since both this PR and #1866 are adding this new file, but #1866 contains the contents that we actually want for this specific file

packages/program/src/revert/revert-error.ts
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/1840/commits/3ae162dba97a55a08c7f7e3fb521f0b639b2be55

petertonysmith94 (dismissed):
Looks good to me 🏅

apps/docs/src/guide/errors/debugging-revert-errors.md
```diff
@@ -0,0 +1,39 @@
...
```
arboleya, line 3: Here, we state that the problem will be fixed in v1.

Can we have an issue for this so it doesn't go unnoticed?

We must be on top of this to guarantee this is indeed only temporary.

The new encoding is taking good shape already; this could be something to go top of it. cc @danielbate

arboleya (approved):

apps/docs/src/guide/errors/debugging-revert-errors.md
danielbate: This has now been proved as working in #1832 and I've added a test for the revert failing gracefully, rather than unsupported str s;oce. I am only testing for `RequireFailed`, I will leave the specifics extracted from the logs to this PR.

danielbate (approved):

nedsalk (approved):

danielbate (approved):

petertonysmith94 (approved):

PR #1847, opened 2024-03-08
fix: release ci failing when there are only empty changesets

### LINKED ISSUE #1701: Problems in CI release workflow
It currently fails when a PR is merged with an empty changeset.

Examples:
 - PR: https://github.com/FuelLabs/fuels-ts/pull/1694
 - Failed workflow: https://github.com/FuelLabs/fuels-ts/actions/runs/7725165033/job/21058804783

We can see the same doesn't happen on the `Wallet` repo:
 - PR: https://github.com/FuelLabs/fuels-wallet/pull/1069
 - Successful workflow: https://github.com/FuelLabs/fuels-wallet/actions/runs/7662160626/job/20882925915

nedsalk:
Do the following to see the bug in action:
1. `git checkout 7b6f3b60b956e58cd58450787c29f67d76a87035` (a commit with a [failing CI](https://github.com/FuelLabs/fuels-ts/actions/runs/8147986790/job/22269864641#step:7:48))
2. run `pnpm changeset:version-with-docs` - nothing happens
3. run `pnpm changeset version` - the `quick-swans-lie.md` file gets deleted and the commit, thus, isn't empty

Our custom script `changeset-version-with-docs.ts` was running the `execSync("changeset version")` call only if one of the versions got changed, which isn't the case when there are only empty changesets. Given that running `changeset version` deletes all changesets (empty ones as well), skipping its execution would make the `FuelLabs/changesets-action@main` action try to create a PR that doesn't differ from `master` because the action didn't commit anything ([see relevant code of action](https://github.com/FuelLabs/changesets-action/blob/5866ff9f4cd625a76d86c0735aded055aeacf814/src/run.ts#L378)). This is the cause of the reported error 
```
RequestError [HttpError]: Validation Failed: {"resource":"PullRequest","code":"custom","message":"No commits between master and changeset-release/master"}
```

closes #1701


scripts/changeset-version-with-docs.ts
```diff
@@ -8,17 +8,14 @@ import { error } from 'console';
...
-    if (!isGitClean) {
+    if (versionsChanged) {
       execSync(`git add packages/versions/src/lib/getBuiltinVersions.ts`);
       execSync(`git commit -m"ci(scripts): update versions"`);
-
-      // Run changeset version
-      execSync(`changeset version`);
     }
+
+    execSync(`changeset version`);
```
arboleya, line 18: Wouldn't it be enough to move this line out of the if statement and preserve everything as it was?

arboleya (commented):
Thanks for getting to the bottom of this.

To prevent the resulting knowledge of this fix from being shadowed, we could distill the PR description/explanation into a concise version and add it as a comment on this script.

scripts/changeset-version-with-docs.ts
nedsalk: Yeah, I just refactored it as I was going through and understanding the code and I feel like it's more descriptive now.

arboleya (approved):

Torres-ssf (approved):

PR #1848, opened 2024-03-09
fix: changeset PR update not running workflows

### LINKED ISSUE #1734: Fix `changeset` automation PR
The `changesets` PR is created correctly, using the `fuels-service-user` via the `REPO_TOKEN `:

https://github.com/FuelLabs/fuels-ts/blob/53babd507ac4ed9c775237cb6500df8e20409cd7/.github/workflows/release.yaml#L60

Then, all subsequent updates to it seem to come from the GH bot, and the workflows stop running automatically.
 - These updates happen when other PRs are merged into `master`
 - Only the `fuels-service-user` can trigger CI automatically
 - GH bot can't trigger CI

At this point, someone must go and close/open the PR and rename its title to something else and back.

This is a follow-up issue on top of:
 - https://github.com/FuelLabs/fuels-ts/issues/465

nedsalk:
The `FuelLabs/changesets-action@main` action is supposed to trigger workflows in the changesets branch after it force-pushes commits because it's [given](https://github.com/FuelLabs/fuels-ts/blob/b83d6d259c443f145b2741046cf53ea9990d44dc/.github/workflows/release.yaml#L60) the `REPO_TOKEN` - which is a personal access token (PAT) of our [fuel-service-user](https://github.com/fuel-service-user) - but the workflows aren't being run because apparently the `GITHUB_TOKEN` is still being used when the PR is updated.

Upon analyzing the discussions and linked PRs in https://github.com/changesets/action/issues/70, it seems that stopping `actions/checkout` from persisting credentials (via `persist-credentials: false`) should do the trick.

closes #1734

#### Where's the bug?

The bug most probably lies somewhere in the interplay between [`actions/checkout`](https://github.com/actions/checkout) which [by default](https://github.com/actions/checkout?tab=readme-ov-file#checkout-v4) persists the `GITHUB_TOKEN` into git's local config and  [actions/toolkit](https://github.com/actions/toolkit) which `FuelLabs/changesets-action@main` [uses](https://github.com/FuelLabs/changesets-action/blob/5866ff9f4cd625a76d86c0735aded055aeacf814/src/run.ts#L328) to interact with GitHub.

> [!IMPORTANT]
> It seems that the default behavior of `persist-credentials:true` is generally a security issue (https://github.com/actions/checkout/issues/485). We should consider disabling it in all of our workflows.


Dhaiwat10 (approved):
Nice one @nedsalk, might be a good idea to file a follow-up issue to disable persisting credentials for all workflows

danielbate (approved):
Nice one, hopefully this does the trick, more CI testing fun.

arboleya (approved):
Nice find! 🎯

PR #1849, opened 2024-03-09
feat: strategy for patching old releases

### LINKED ISSUE #1746: Implement strategy for patching old releases
Let's imagine we have these three _past_ releases:

 1. `0.70.0` — any release
 1. `0.71.2` — introduces a bug 👈
 1. `0.72.0` — network / fuel-core upgrade

The second release introduces a bug that goes unnoticed until after `latest`.

Now, consider we need to publish `0.71.3` with the fix without affecting the `latest` / `0.72.0`.

Currently, we can't do that. We can only release a new release on top of the latest.

This is the scenario we need to address.

nedsalk:
closes #1746


Dhaiwat10:
This is excellent @nedsalk and something that we desperately need 👏🏻 

Quick question: instead of creating these new release branches ourselves, do you think we could re-use the branches created by Github automatically for each release like this one: https://github.com/FuelLabs/fuels-ts/tree/v0.76.0?

nedsalk:
@Dhaiwat10 Thanks! I didn't see that GH creates it automatically; I don't see why it couldn't and there'd be even less ceremony in the workflows. Thanks for the suggestion, I'll look into it.

nedsalk:
@Dhaiwat10 The linked branch isn't actually a branch but just a tag added by `FuelLabs/changesets-action@main` (see [here](https://github.com/FuelLabs/changesets-action/blob/5866ff9f4cd625a76d86c0735aded055aeacf814/src/run.ts#L64)), and it's not possible to create PRs against tags. I tried the following:
- `git checkout -b ns/testing-tags v0.76.0` - this checks out the branch on that specific v0.76.0 tag
- I added a file for a PR to be possible
- Tried to create a PR from [`ns/testing-tags`](https://github.com/FuelLabs/fuels-ts/tree/ns/testing-tags) against base that I chose via the tag in the dropdown
- Only the diff is shown and it's not possible to create the PR

It seems like we have to create a branch for a PR to be possible.

.github/workflows/test.yaml
```diff
@@ -66,6 +66,7 @@ jobs:
           path: coverage/environments/${{ matrix.env.name }}
 
   test:
+    if: github.base_ref == 'master'
```
danielbate, line 69: Why would we not want to run tests when based against a release?

.github/workflows/test.yaml
```diff
@@ -123,7 +124,7 @@ jobs:
 
   test-e2e:
     name: e2e
-    if: github.head_ref == 'changeset-release/master'
+    if: startsWith(github.head_ref, 'changeset-release/') == 'true'
```
danielbate, line 127: We probably can't run live e2e for patch releases of old versions, as they'll get run against a beta-5 url and will fail.

danielbate (changes requested):
Really nice work @nedsalk 🚀 

Couple thoughts around testing release branches. And IMO I prefer basing against branches than tags so this flow works better for me.

nedsalk:
@arboleya We don't necessarily have to create the `release/*` branches, we can create them manually as the need arises from tags:
1. `git checkout release/v0.76.0 v0.76.0` and push it to origin
2. Do stuff as explained in proposal, release `v0.76.1`
3. Remove branch

The `FuelLabs/changesets-action@main` action will create a tag for `v0.76.1` as well and it'll exist even if its branch was removed.

.github/workflows/test.yaml
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1849/commits/9a4987ee81db0410ef8916cb598ff856d9c9b261

.github/workflows/test.yaml
nedsalk: We are running tests; this step is just collecting coverage. I started implementing CI coverage handling for release branches as well but it started touching upon multiple files and I felt it diluted the original idea of the PR too much. I'd rather implement it in another PR - if it needs to be implemented at all.

danielbate: Ah, of course!

danielbate: Really want to change the name of that stage.

danielbate:
If we are moving to manual branch creation, is it worth having a workflow to create a `release/***` branch so we can still have the same branch creations as `master`? I don't like the idea of people being able to commit directly to release branches.

nedsalk:
@danielbate The same branch protection rules for `master` should be turned on for these branches so nobody would be able to push to them.

arboleya:
@danielbate Yep. I presume this is one of the reasons why the `Wallet` repo uses a single pre-existent `release` branch.

danielbate:
Okay, well IMO `release/*` should be protected. We could have a manually triggered workflow to create `release/*` branches so that it can still be locked down.

arboleya:
My hesitation is that this `release/*` approach can send us back to the `rc/*` permissions from which we escaped.

danielbate:
Hmm yes I am getting some deja vu with my comment, maybe @FuelLabs/frontend can weigh in with their single `release` branch approach. At some point in the flow, we need to merge a branch into another. If we have a latest `release` and a legacy patch `release` operating in tandem, I don't understand how that would work, as you can't merge into a tag.

nedsalk:
I see three ways forward:

1. Use `fuel-service-user` to generate a `release/*` branch when a release happens (as was the initial proposal)
2. Have a workflow that creates the appropriate `release/*` branch and which gets triggered when something is done in relation to a release commit. _(This came to mind as I was writing this comment; I don't know how it'd look or if it would even make sense, but I wanted to mention it just so that we have it on the map.)_
3. Do what the frontend team does and have only one `release` branch onto which we'd be force-pushing commits:
    1. Checkout a release tag and force-push it onto the `release` branch
    2. Create a PR against the `release` branch and merge it
    3. The same changeset workflow is started on the `release` branch; we approve it and it gets merged and the current `release` branch is released and we get the corresponding new tag in our git history.

To be honest, I don't see a problem with the third, force-push approach. The first one is "more proper", but we'll be getting tags from both anyways so we won't be losing any info.

arboleya:
@nedsalk Actually, I think we _can_ get away with the approach from the PR description.

These `release/*` branches could be automatically deleted after the release as well.

They'll be short-lived, detached, and never be merged to `master`.

Never going to `master` should keep us away from `rc/*` territory. 🤔

nedsalk:
@arboleya So we'd have have the same branch protection rules on them as `master` except that we'd set the "Restrict creations" rule to allow all code owners to create them?

danielbate:
> So we'd have the same branch protection rules on them as master except that we'd set the "Restrict creations" rule to allow all code owners to create them?

Or we restrict it via a workflow, which I think I am more in favour of

danielbate:
It's usage, if manually triggered, should be pretty sporadic though.

We could then also use the workflow to validate the patch version on branch creation rather than on release creation.

petertonysmith94 (commented):
@nedsalk just wanted to run past my understanding of this change...

---
The following example is for releasing a patch for `v0.69.0` -> `v0.69.1`

### Patch release

- Create branch for new patch release (from `v0.69.0`), from the tag (`v0.69.0` -> `release/v0.69.0`)
  - Changeset PR will be created against the release branch

### Developer workflow
- PR's created against the release branch
  - (`fix/bug1` -> `release/v0.69.0`)
  - (`fix/bug2` -> `release/v0.69.0`)
- PR's merged into the release branch

### Release
- Changeset PR merged into the release branch (`ci(changeset)` -> `release/v0.69.0`)
- Publish release as `v0.69.1` and tagged as `v0.69.1`.
- Delete the release branch (`release/v0.69.0`)

nedsalk:
@petertonysmith94 You're correct, except that the changeset PR won't be created for the branch until something is merged into it (e.g. `fix/bug1`).

arboleya (commented):
I think we should document how to patch and old release in the [CONTRIBUTING.md](https://github.com/FuelLabs/fuels-ts/blob/ns/feat/patch-old-releases/CONTRIBUTING.md) document.

arboleya (approved):

Torres-ssf (approved):

PR #1855, opened 2024-03-11
feat: warn when running against unofficial `fuel-core` nodes

### LINKED ISSUE #1841: Improve error handling for `fuel-core` version mismatch
It seems that nightly builds can sometimes contain **_breaking changes_**.

Consider this version:
 - `0.22.0+nightly.20240212.5cbe7e2984`

If we try the aforementioned `fuel-core` version with `fuels@0.76`, we get this:

```
Error: Field "dryRun" argument "txs" of type "Mutation" is required but not provided
```

For non-official/nightly builds, we should consider either:
 1. Throwing an `error`
 2. Adding a descriptive `warning`

This should make it even easier for users to spot version incompatibilities.

nedsalk:
I opted for a `console.warn` instead of throwing an error because somebody might actually want to run against nightly versions.

closes #1841


packages/versions/src/lib/checkFuelCoreVersionCompatibility.test.ts
```diff
@@ -67,4 +67,35 @@ describe('getDifferenceToUserFuelCoreVersion', () => {
...
+      FUELS: '1', // not under test
+      FORC: '1', // not under test
+      FUEL_CORE: '0.1.2',
+    }));
+
+    checkFuelCoreVersionCompatibility('0.1.2+nightly.20240212.5cbe7e2984');
+
+    expect(consoleWarnSpy).toHaveBeenCalledOnce();
+    expect(consoleWarnSpy).toHaveBeenCalledWith(
+      "You're running against a non-official fuel-core version: 0.1.2+nightly.20240212.5cbe7e2984. Things may work as expected, but it's not guaranteed. Please use an official release."
```
petertonysmith94, line 84:
```suggestion
      "You're running against an un-released fuel-core version: 0.1.2+nightly.20240212.5cbe7e2984. Things may work as expected, but it's not guaranteed. Please use a released version."
```

petertonysmith94 (commented):
Small comment on the choice of non-official

packages/versions/src/lib/checkFuelCoreVersionCompatibility.test.ts
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1855/commits/95971325b7e3886b8b3cd52ca15d0d52c3b70e9e

packages/versions/src/lib/checkFuelCoreVersionCompatibility.ts
```diff
@@ -4,6 +4,12 @@ import { majorEq, minorEq, patchEq } from './semver';
 export function checkFuelCoreVersionCompatibility(networkVersion: string) {
   const { FUEL_CORE: supportedVersion } = getBuiltinVersions();
 
+  if (networkVersion.match(/^[0-9]+\.[0-9]+\.[0-9]+$/) === null) {
```
arboleya, line 7:
```suggestion
  if (/^[0-9]+\.[0-9]+\.[0-9]+$/m.test(networkVersion)) {
```

packages/versions/src/lib/checkFuelCoreVersionCompatibility.ts
```diff
@@ -4,6 +4,12 @@ import { majorEq, minorEq, patchEq } from './semver';
...
+    console.warn(
```
arboleya, line 8: The `console` lint errors can be circumvented by the usage below.

It seems more elegant than inline-disabling the rule.

```ts
import { warn } from 'console';

warn(...)
```

packages/versions/src/lib/checkFuelCoreVersionCompatibility.ts
```diff
@@ -4,6 +4,12 @@ import { majorEq, minorEq, patchEq } from './semver';
...
+      `You're running against an unreleased fuel-core version: ${networkVersion}. Things may work as expected, but it's not guaranteed. Please use a released version.`
+    );
```
Torres-ssf, line 10: I believe the warning message should include something like (not the same sentence but something with the same meaning):

_This non-official `fuel-core` build may include features and updates not yet supported by this version of the TS-SDK._

packages/versions/src/lib/checkFuelCoreVersionCompatibility.ts
nedsalk: I've got a rollup error that doesn't allow me to do this:
```
fuels:build: RollupError: "warn" is not exported by "__vite-browser-external", imported by "../versions/dist/index.mjs".
```

I'll have to go the inline disabling route unfortunately.

packages/versions/src/lib/checkFuelCoreVersionCompatibility.ts
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1855/commits/a519b9baeee95db92d9219cd165d8c0ad9562408

packages/versions/src/lib/checkFuelCoreVersionCompatibility.ts
danielbate: Not 100%, but you may need to polyfill `console` in browser which you can add [here](https://github.com/FuelLabs/fuels-ts/blob/master/vite.browser.config.mts#L15).

danielbate (approved):

PR #1861, opened 2024-03-12
chore: upgrade fuels on demo wallet sdk react

Torres-ssf:
- Fix: https://github.com/FuelLabs/fuels-ts/actions/runs/8248086191/job/22557643625?pr=1860


apps/demo-wallet-sdk-react/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "@fuel-wallet/react": "0.15.2",
-    "fuels": "0.73.0",
+    "fuels": "0.76.0",
```
arboleya, line 13: Should it be using it from the `workspace:*`?

Torres-ssf: NIce catch 👍

danielbate (approved):

Dhaiwat10:
@Torres-ssf the reason why I had set it to `0.73.0` is because that's the version that the wallet SDK is compatible with. I'm afraid bumping it to `workspace` would mean that the app doesn't work anymore. It might build fine still, but we need to verify that it actually works. Can you please make sure that it works after this change?

arboleya:
> the reason why I had set it to `0.73.0` is because that's the version that the wallet SDK is compatible with.

@Dhaiwat10 Is it helpful to have it like this so we'll know when and if something breaks?

I mean, CI should ideally break if something is off.

Dhaiwat10:
@arboleya you mean always having it set to `workspace` and relying on the `build` of the app?

petertonysmith94 (approved):

arboleya:
@Dhaiwat10 Yes. IIRC, you recently documented something about using React hooks, right? I am assuming this is related to that, and if that's the case, my concern is to have something in place that guarantees that the latest SDK version is working for this. If I'm mixing things up, please ignore.

arboleya:
But it's a `peerDependency` and it's not pinned but `>=0.73.0`. 🤔

Dhaiwat10:
@arboleya It's not pinned there but I have often heard from the frontend team when something doesn't work with the latest TS SDK version to just go back and use the version specified in their package.json. Maybe @luizstacio could give more clarity here.

PR #1867, opened 2024-03-12
feat: predicate factory generics

### LINKED ISSUE #1850: The `Predicate` typegen template does not consider the generic type
> Note: This issue was created after some findings from https://github.com/FuelLabs/fuels-ts/pull/1826 and needs for it to be merged before this can be worked on

Currently, when you create a `Predicate` instance using the typegen factory, this is what it looks like:

https://github.com/FuelLabs/fuels-ts/blob/b83d6d259c443f145b2741046cf53ea9990d44dc/apps/demo-typegen/src/demo.test.ts#L144

This code snippet creates a `Predicate` instance of type `Predicate<InputValue[]>`. It doesn't actually infer the type for the input data from the typegen. For example, if the `Predicate` in question accepts one string argument, `createInstance` should return an instance of type `Predicate<[string]>`. 

This is important because it helps users pass in the input data in the correct shape.

> Incoming from https://github.com/FuelLabs/fuels-ts/pull/1826#discussion_r1515139427

petertonysmith94:
Closes #1850 

- Added generic data input for the predicate factory.


arboleya (commented):
@Dhaiwat10 Could you elaborate on your idea of inferring this type automatically from the ABI, if that still makes sense?

Dhaiwat10:
@arboleya sure.

@petertonysmith94 I think the ideal DX for this `createInstance` method would be to not have the user pass in a generic, but _infer_ it from the ABI/typegen. We already know the 'shape' of the input that the predicate will accept when we generate types for it using the ABI. We should ideally use this information about the input's shape to pass that into the `Predicate` constructor. Does that make sense? Happy to elaborate over a huddle if unclear.

Dhaiwat10 (changes requested):
Awesome work on the PR @petertonysmith94 👏🏻 

Could you please also export the type for the inputs from the typegen file in case someone wants to import it and use it like so:

```ts
import { PredicateInputType, PredicateFactory } from '@/types';

const predicateInputData: PredicateInputType = [20, 30];
const predicate = PredicateFactory.createInstance(provider, predicateInputData);
```

arboleya (approved):

Dhaiwat10 (approved):

Issue #1862, opened 2024-03-12
New flaky test failling with test timeout


Torres-ssf:
We have a new flaky test: https://github.com/FuelLabs/fuels-ts/actions/runs/8248315155/job/22558365399?pr=1861#step:9:547

Maybe it is just a matter of increasing the test [timeout](https://github.com/FuelLabs/fuels-ts/blob/41aa6a1c198567f54c54be3cf33b93548cb40d63/packages/fuel-gauge/src/transaction-response.test.ts#L280)

UPDATE
This same test failure was also mentioned [here](https://github.com/FuelLabs/fuels-ts/issues/1755), but most likely it is not related

arboleya:
Should this mention go as a comment on https://github.com/FuelLabs/fuels-ts/issues/1755?

PR #1888, opened 2024-03-18
fix: increase timeout of `TransactionResponse` test

nedsalk:
- Close: https://github.com/FuelLabs/fuels-ts/issues/1862

This test has been timing out _a lot_ in our workflows recently and I increased its timeout to 10s in the hopes of stopping the timeouts.


nedsalk:
@arboleya haha, this is comedy. I change the node settings a bit and reran the workflows a couple of times and they passed each time. I'd merge it like this and we can see if it'll pop up anymore.

petertonysmith94 (approved):
Seems legit - nice one @nedsalk!

arboleya (approved):

PR #1892, opened 2024-03-19
ci(changesets): versioning packages for branch `master`

fuel-service-user:
# RELEASE - v0.78.0

# Breaking

- Features
    - [#1815](https://github.com/FuelLabs/fuels-ts/pull/1815), Improve signing DX, by [@danielbate](https://github.com/danielbate)

---

# Fixes
- [#1883](https://github.com/FuelLabs/fuels-ts/pull/1883), Command `fuels version` not working, by [@arboleya](https://github.com/arboleya)

PR #1901, opened 2024-03-20
chore: testing patch of older releases

nedsalk:
Testing out logic added in #1849.


arboleya:
Also went to the wrong [changeset](https://github.com/FuelLabs/fuels-ts/pull/1892).

Issue #1117, opened 2023-07-24
Remove externalLoggedTypes from `Interface`


nedsalk:
Currently `Interface.externalLoggedTypes` is being used as a way to decode logged types during multicall. This shouldn't be the case, as an `Interface` shouldn't be concerned with some other `Interface`. The `externalLoggedTypes` logic should be moved elsewhere; probably somewhere in `BaseInvocationScope` or similar.

PR #1936, opened 2024-03-25
chore!: remove `externalLoggedTypes` from `Interface` class

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/1117


packages/fuel-gauge/src/advanced-logging.test.ts
```diff
@@ -115,4 +124,77 @@ describe('Advanced Logging', () => {
...
+      'fuelfuel',
+    ];
+
+    logs.forEach((log, i) => {
+      if (typeof log === 'object') {
+        expect(JSON.stringify(log)).toBe(JSON.stringify(expectedLogs[i]));
+      } else {
+        expect(log).toBe(expectedLogs[i]);
+      }
+    });
```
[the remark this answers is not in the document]
Torres-ssf: `toBe` cannot be used to compare objects directly

packages/account/src/providers/transaction-response/getDecodedLogs.ts
```diff
@@ -1,23 +1,44 @@
...
+      const data =
+        receipt.type === ReceiptType.Log
+          ? new BigNumberCoder('u64').encode(receipt.val0)
+          : receipt.data;
 
-    if (r.type === ReceiptType.Log) {
-      logs.push(
-        abiInterface.decodeLog(new BigNumberCoder('u64').encode(r.val0), r.val1.toNumber(), r.id)[0]
-      );
+      const [decodedLog] = interfaceToUse.decodeLog(data, receipt.val1.toNumber());
```
danielbate, line 40: What is `val0` and `val1` in this context?

packages/program/src/functions/invocation-results.ts
```diff
@@ -73,6 +72,27 @@ export class InvocationResult<T = any> {
     return this.functionScopes[0].getCallConfig();
   }
 
+  /**
+   * Gets the ABI from all calls.
+   *
+   * @returns The ABIs from all calls.
+   */
+  getAbiFromAllCalls = () =>
```
danielbate, line 80: Please add a return type for this function.

packages/program/src/functions/invocation-results.ts
```diff
@@ -73,6 +72,27 @@ export class InvocationResult<T = any> {
...
+    this.functionScopes.reduce((acc, funcScope, i) => {
+      const { program, externalAbis } = funcScope.getCallConfig();
+
+      if (i === 0) {
```
danielbate, line 84: Can we always assume that the first ABI is the non external?

danielbate (commented):
Nice work and good tests, couple questions.

packages/program/src/functions/invocation-results.ts
Torres-ssf: Yes because the first function scope entry is always the instance that initiates the call scope

```ts
// main ABI belongs to contractA
contractA.functions.x_y_z(10).addContracts([contractB])

or

// main ABI belongs to contractB
contractB.multicall(...)

or

// main ABI belongs to script
scriptA.functions.main()
```

Since the main ABI can belong to either a script or contract, we cannot index it by an `id`. I am treating it differently from the other ABIs

The following ABIs will always belong to contracts since we can't execute a multi-call using a script invocation or even make a script to call another script.

packages/account/src/providers/transaction-response/getDecodedLogs.ts
Torres-ssf: @danielbate They are properties from the 2 receipts related to logs: `ReceiptLog` and `ReceiptLogData`. 

These values are used to decode the receipts. The only change here is that now this helper receives the dictionary with the indexed contracts IDs with ABIs so the logged values can be properly decoded by accessing the types present on the ABI.

Before this dictionary was being cached within the `Interface` class within the property `externalLoggedTypes`

packages/account/src/providers/transaction-response/getDecodedLogs.ts
```diff
@@ -1,23 +1,44 @@
...
+   * @param receipts - The array of transaction result receipts.
+   * @param mainAbi - The ABI of the main contract.
+   * @param externalAbis - The record of external contract ABIs.
+   * @returns An array of decoded logs from Sway projects.
+   */
+  return receipts.reduce((logs: T[], receipt) => {
+    if (receipt.type === ReceiptType.LogData || receipt.type === ReceiptType.Log) {
+      const interfaceToUse = externalAbis[receipt.id]
+        ? new Interface(externalAbis[receipt.id])
+        : new Interface(mainAbi);
```
nedsalk, line 33:
```suggestion
      const interfaceToUse = new Interface(externalAbis[receipt.id] ?? mainAbi);
```

packages/fuel-gauge/src/advanced-logging.test.ts
```diff
@@ -115,4 +124,69 @@ describe('Advanced Logging', () => {
       INPUT,
     ]);
   });
+
+  it('should properly decode all logs in a multicall with inter-contract calls', async () => {
```
nedsalk, line 128: :clap: :+1:

nedsalk (approved):

danielbate (approved):

packages/account/src/providers/transaction-response/getDecodedLogs.ts
Torres-ssf: @nedsalk Since the PR is already approved I will make this suggestion at: https://github.com/FuelLabs/fuels-ts/pull/1940

Issue #1644, opened 2024-01-15
Use receipts from `TransactionStatus`


nedsalk:
### Discussed in https://github.com/FuelLabs/fuels-ts/discussions/1588

<div type='discussions-op-text'>

<sup>Originally posted by **nedsalk** December 15, 2023</sup>
Post `fuel-core v0.22.0`, the node now returns receipts with a `TransactionStatus`, so we can use them right away instead of running another fetch. From my understanding, this will only become relevant after #1374 is merged, because we're currently using the `getTransactionWithReceipts` endpoint.</div>

The way to go about implementing this is to use the `TransactionRequest` sent to the node in `provider.sendTransaction` because it contains the same transaction as the one that's fetched from the node inside of `TransactionResponse` via `getTransactionWithReceipts`, only that after fetching it from the node the receipts are there as well. 

Because the receipts can now be received from the subscription directly, the same transaction summary can be generated based off of the `TransactionRequest` and the receipts received from the subscription. A working implementation can be found on [this branch](https://github.com/FuelLabs/fuels-ts/tree/ns/feat/subscription-receipts).

nedsalk:
Blocked by merge of #1415 and #1615.

arboleya:
The only utility for a discussion is to discuss something _before_ it becomes an issue.

arboleya:
So the discussion is not needed anymore.

nedsalk:
Blocked by:
- https://github.com/FuelLabs/fuel-core/issues/1724

arboleya:
Unblocked by:
 - https://github.com/FuelLabs/fuel-core/pull/2061

PR #1961, opened 2024-03-27
docs: add essentials section for Transactions

### LINKED ISSUE #1923: Docs: Transactions
Things we need more clarity:

 - Transaction response?
 - Transaction Policies
 - Custom Transactions
   - Custom Contract call
     - Crosslink to `Scripts` > `Custom Script Call`?

danielbate:
Closes #1923 

This PR introduces a section to the essentials docs around Transactions. It includes pages on the following:
- Transaction Request
- Transaction Response
- Transaction Parameters
- Transaction Policies

> It also crosslinks parameters with contract call parameters hence having removed that page in contracts. I have not included custom transactions as I think transaction request adds enough information on this, and we have the cookbooks.


apps/docs/src/guide/transactions/transaction-response.md
```diff
@@ -0,0 +1,23 @@
+# Transaction Response
+
+Once a transaction has been submitted, you may want to extract information regarding the result of the transaction. The SDK offers a `TransactionResponse` class with helper methods to expose the following information:
+
+- The transaction ID
+- The status (submitted, success, squeezed out, or failure)
+- Receipts (return data, logs, mints/burns, transfers and panic/reverts)
+- Gas fees and usages
+- The raw payload of the transaction including inputs and outputs
```
nedsalk, line 9: This one would be removed in #1644 so I'm not sure if we should be mentioning it here at all. I'd really like people to _not_ know about it so that they don't break their code when it's removed.

apps/docs-snippets/src/guide/transactions/transaction-response.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    transactionRequest.setData(scriptAbi, scriptMainFunctionArguments);
+    transactionRequest.addResources(resources);
+    const response: TransactionResponse = await wallet.sendTransaction(transactionRequest);
+
+    // #region transaction-response-3
+    // #import { TransactionResponse };
+
+    // Get a transaction ID from the transaction request, this could have been stored from a
+    // previous transaction request. You will need the chain ID to obtain this.
+    const transactionId = transactionRequest.getTransactionId(await provider.getChainId());
```
nedsalk, line 87:
```suggestion
    const transactionId = transactionRequest.getTransactionId(provider.getChainId());
```

`await` is redundant here.

apps/docs/src/guide/essentials/index.md
```diff
@@ -1,3 +1,5 @@
 # Essentials
 
-This guide will cover the essentials required for interacting with programs built using Sway.
\ No newline at end of file
+This guide will cover the essentials required for interacting with the Fuel Network.
+
+This includes building and modifying transactions and interacting with Sway program types; contracts, scripts and predicates.
```
nedsalk, line 5:
```suggestion
This includes building and modifying transactions and interacting with Sway program types: contracts, scripts and predicates.
```

: instead of ;.

apps/docs/src/guide/transactions/index.md
```diff
@@ -0,0 +1,13 @@
+# Transactions
+
+A transaction is any interaction with the blockchain. This is the primary way of interacting with the Fuel Network and can include actions like transferring assets, deploying contracts and minting tokens.
```
nedsalk, line 3: Not all interactions with the blockchain are transactions, e.g. `provider.operations.getChain()`. Nothing comes to mind for rephrasing it, though.

apps/docs/src/guide/transactions/index.md
```diff
@@ -0,0 +1,13 @@
...
+
+Within the Fuel Network, we have the following transaction types:
+
+- Script
+- Create
+- Mint
+
+The SDK provides class helpers for handling script and create transactions, `ScriptTransactionRequest` and `CreateTransactionRequest` respectively. Mint transactions are recognized by the SDK but can only be created by the block producer so do not have use any use outside of block creation. Therefore there is no need to manually create them.
```
nedsalk, line 11:
```suggestion
The SDK provides class helpers for handling script and create transactions: `ScriptTransactionRequest` and `CreateTransactionRequest`, respectively. 
Mint transactions can only be created by the block producer and do not have any use outside of block creation. Therefore, the SDK only provides the ability to decode them.
```

If this rewording doesn't go through, there is an additional "use" in the original wording that should be removed.

apps/docs/src/guide/transactions/index.md
```diff
@@ -0,0 +1,13 @@
...
+
+This guide will discuss how to modify transactions, set their transaction parameters and policies, and submit them to the network to then produce an extractable transaction response.
```
nedsalk, line 13:
```suggestion
This guide will discuss how to create, modify and submit transactions to the network, as well as how to get info on transactions that have previously executed.
```

apps/docs/src/guide/transactions/transaction-parameters.md
```diff
@@ -0,0 +1,25 @@
...
+
+<<< @/../../docs-snippets/src/guide/transactions/transaction-parameters.test.ts#transaction-parameters-1{ts:line-numbers}
+
+## Setting Transaction Parameters
+
+To set the transaction parameters, you have access to the `txParams` method on a transaction request.
+
+<<< @/../../docs-snippets/src/guide/transactions/transaction-parameters.test.ts#transaction-parameters-2{ts:line-numbers}
+
+The same method is also accessible within a function invocation scope, so it can also be used when calling contract functions.
```
petertonysmith94, line 21:
```suggestion
The same method is also accessible when calling contract functions.
```

Does the user need to know this?

apps/docs/src/guide/transactions/transaction-parameters.md
```diff
@@ -0,0 +1,25 @@
...
+
+1. Gas Price - The price you're willing to pay for each unit of gas consumed during the transaction execution.
+1. Gas Limit - The maximum amount of gas you're willing to allow the transaction to consume. If the transaction execution requires more gas than the specified gas limit, the transaction will fail and any state changes made during the execution will be reverted.
+1. Variable Outputs - The number of variable outputs allowed in the transaction. Variable outputs are used in transactions that have a dynamic number of outputs, such as those involving multiple recipients or complex contract interactions. By setting this value, you can control the number of variable outputs permitted in the transaction, which can be useful for managing transaction size and complexity.
+
+> **Note**: Setting transaction parameters is optional. If you don't specify them, the SDK will fetch some sensible defaults from the chain.
+
+All available parameters are shown below:
+
+<<< @/../../docs-snippets/src/guide/transactions/transaction-parameters.test.ts#transaction-parameters-1{ts:line-numbers}
```
petertonysmith94, line 13: I think explaining each parameter would be beneficial like [here](https://docs.fuel.network/docs/fuels-ts/cli/config-file/).

You've outlined 3 parameters in the list above, but there are 6 in total? What do the other ones do?

apps/docs-snippets/package.json
```diff
@@ -12,7 +12,8 @@
     "@fuel-ts/errors": "workspace:*",
     "@fuel-ts/utils": "workspace:*",
     "ethers": "^6.7.1",
-    "fuels": "workspace:*"
+    "fuels": "workspace:*",
+    "vitest": "^1.1.3"
```
nedsalk, line 16: Why did you add this? Shouldn't it be globally available?

apps/docs-snippets/src/guide/transactions/transaction-request.test.ts
```diff
@@ -0,0 +1,180 @@
...
+  it('creates a transaction request from ScriptTransactionRequest', () => {
+    const scriptMainFunctionArguments = [1];
+
+    // #region transaction-request-1
+    // #import { ScriptTransactionRequest };
+
+    // Instantiate the transaction request using a ScriptTransactionRequest
+    const transactionRequest = new ScriptTransactionRequest({ script: scriptBytecode });
+    // Set the script main function arguments (can also be passed in the class constructor)
+    transactionRequest.setData(scriptAbi, scriptMainFunctionArguments);
```
petertonysmith94, line 68: Given the structure of our documentation.

The typegen section is directly before this example. Is it worth syncing up the example from the [typegen section](https://docs.fuel.network/docs/fuels-ts/abi-typegen/using-generated-types/#script), with this example, to use the same `Script` and then building on the knowledge from there.

Then we can back link to that section and it creates a better flow between sections?

petertonysmith94 (commented):
image

- For me, whitespace between the comments on the docs-snippets, would be easier on the eyes.
- Also, it appears that types aren't being imported correctly in the final render.

---

I would favour omitting types from docs-snippets, so then they can be leveraged by both JavaScript and TypeScript developers, what are your thoughts?

apps/docs-snippets/src/guide/transactions/transaction-parameters.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    const txParams: TxParams = {
+      gasPrice: bn(1), // BigNumberish
+      gasLimit: bn(1), // BigNumberish
+      maturity: 1, // number or undefined
+      maxFee: bn(1), // BigNumberish or undefined
+      witnessLimit: bn(1), // BigNumberish or undefined
+      variableOutputs: 1, // number
+    };
+    // #endregion transaction-parameters-1
```
petertonysmith94, line 41: I think explaining each parameter would be beneficial like [here](https://docs.fuel.network/docs/fuels-ts/cli/config-file/).

You've outlined 3 parameters in the [list](https://github.com/FuelLabs/fuels-ts/pull/1961/files#diff-58920915a7f022eb88089c997f230b4268730b90c9fa2f848e9d130bfe0b249fR5), but there are 6 in total? What do the other ones do?

apps/docs/src/guide/transactions/transaction-response.md
danielbate: By one, you mean just `reciepts`?

apps/docs/src/guide/essentials/index.md
danielbate: One day I'll be able to use a semi colon correctly

apps/docs/src/guide/transactions/index.md
danielbate: Hmmm interesting, I'd originally taken that from [here](https://docs.fuel.network/docs/intro/glossary/#transaction).

"A transaction is a way of interacting with the blockchain"?

apps/docs-snippets/package.json
danielbate: Previously it was [failing](https://github.com/FuelLabs/fuels-ts/actions/runs/8455979446/job/23164701406#step:17:18) without it. Vitest only exposes `describe`, `expect`, `it` globally and I used `epectTypeOf`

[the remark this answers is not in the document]
danielbate: Moving to above [comment](https://github.com/FuelLabs/fuels-ts/pull/1961#discussion_r1542729264)

apps/docs/src/guide/transactions/transaction-parameters.md
danielbate: Definitely, this used to have a page under contracts but I've deleted that in place of this doc, so good to keep a reference.

apps/docs-snippets/src/guide/transactions/transaction-parameters.test.ts
danielbate: https://github.com/FuelLabs/fuels-ts/pull/1961/commits/14b3ca3b74bae11ec1850ae55c3fc29b5d38b641

apps/docs-snippets/src/guide/transactions/transaction-request.test.ts
danielbate: Hmm I agree we can probably improve flow between sections, but I don't think moving transactions into basics is the answer. There's quite a few elements in this section that I don't think everyone needs to know about.

IMO the better approach is to link to transactions from typegen?

petertonysmith94: I wasn't thinking moving the transactions into the basics section.

More building upon the example from the previous section, however, potentially something worth re-visiting later, once the revamp has been completed.

```ts
import { ScriptAbi__factory } from './types'; 

const request = new ScriptTransactionRequest({
  script: ScriptAbi__factory.bin,
})

request.setData(ScriptAbi__factory.abi, [
  // Add the inputs here
])
```

Happy for you to resolve this as you see fit 😄

danielbate:
@petertonysmith94 

> For me, whitespace between the comments on the docs-snippets, would be easier on the eyes.

Agreed - https://github.com/FuelLabs/fuels-ts/pull/1961/commits/79ece0c2c1915c00fa1c5b756faca0b6496247ed

> I would favour omitting types from docs-snippets, so then they can be leveraged by both JavaScript and TypeScript developers, what are your thoughts?

I wouldn't wish doing this in pure JS on my enemies 😆 I think if we do this we should offer both JS and TS snippets with an option to switch between. Additionally, we are quite explicit throughout the docs that this is a TS SDK. It should be an SDK wide change.

apps/docs-snippets/src/guide/transactions/transaction-request.test.ts
danielbate: Will see if any others have opinions on this 👍🏻 I am just conscious of greying the lines between sections. But I agree we want flow.

apps/docs-snippets/src/guide/transactions/transaction-request.test.ts
```diff
@@ -0,0 +1,191 @@
...
+    transactionRequest.addCoinOutput(recipientAddress, 1000, BaseAssetId);
+
+    // Adding message inputs
+    transactionRequest.addMessageInput(message);
+    // #endregion transaction-request-3
+
+    expect(transactionRequest.script).toEqual(arrayify(scriptBytecode));
+    expect(transactionRequest.inputs.length).toEqual(4);
+    expect(transactionRequest.outputs.length).toEqual(2);
+    expect(transactionRequest.witnesses.length).toEqual(1);
```
Torres-ssf, line 114: The flow should be:
```ts
    const transactionRequest = new ScriptTransactionRequest({ script: scriptBytecode });

    transactionRequest.addCoinOutput(recipientAddress, 1000, BaseAssetId);

    // Adding message inputs
    // Is there a reason why a message is being added? If so it needs to be added before estimating the TX
    transactionRequest.addMessageInput(message);

    const { gasUsed, requiredQuantities, maxFee } = provider.getTransactionCost(transactionRequest);

     // Adding resources (coins or messages)
    await senderWallet.fund(transactionRequest, requiredQuantities, maxFee);
    // #endregion transaction-request-3
```

The reason is that we should avoid adding resources without knowing the real amount the TX will require ( in this specific case it's the transfer and the fee ) and avoid running `OutOfGas`.

apps/docs-snippets/src/guide/transactions/transaction-request.test.ts
```diff
@@ -0,0 +1,191 @@
...
+    // Instantiate the transaction request
+    const transactionRequest = new ScriptTransactionRequest({ script: scriptBytecode });
+
+    // Add a witness directly
+    transactionRequest.addWitness(witness);
+
+    // Add a witness using an account
+    const account: Account = WalletUnlocked.generate({ provider });
+    await transactionRequest.addAccountWitnesses(account);
+    // #endregion transaction-request-6
```
Torres-ssf, line 167: The docs-related text should explicitly say that after the TX is signed at:

```ts
await transactionRequest.addAccountWitnesses(account);
```

Any changes to it will result in the signing not working anymore, because modifying the TX will result in modifying its ID ( only the witnesses property can be changed )

So before signing it, the user needs to add everything that the TX needs, including funds

```ts
    // Instantiate the transaction request
    const transactionRequest = new ScriptTransactionRequest({ script: scriptBytecode });

    transactionRequest.addCoinOutput(recipientAddress, 1000, BaseAssetId);

    const { gasUsed, requiredQuantities, maxFee } = provider.getTransactionCost(
      transactionRequest, [], { signatureCallback: (tx) => tx.addAccountWitnesses(account)}
     );

     // Adding resources (coins or messages)
    await senderWallet.fund(transactionRequest, requiredQuantities, maxFee);

    // Add a witness directly
    transactionRequest.addWitness(witness);

    // Add a witness using an account
    const account: Account = WalletUnlocked.generate({ provider });
    await transactionRequest.addAccountWitnesses(account);
    // #endregion transaction-request-6
```

Since this case is more tricky, it is worth adding a complete test: using a script that requires a signature to make the TX succeed. By doing so, if any changes related to this flow happens, we will remember to update this doc section.

apps/docs/src/guide/transactions/transaction-response.md
nedsalk: - The raw payload of the transaction including inputs and outputs

nedsalk: The whole raw payload

apps/docs/src/guide/transactions/transaction-request.md
```diff
@@ -0,0 +1,51 @@
...
+
+The SDK provides a way of either modifying the witnesses for a transaction request directly, or by passing accounts. This will then signing the transaction request with the account's private key. Below will detail how to add a witness to a transaction request:
+
+<<< @/../../docs-snippets/src/guide/transactions/transaction-request.test.ts#transaction-request-6{ts:line-numbers}
+
+### Getting the Transaction ID for a Transaction Request
+
+The transaction ID is a SHA-256 hash of the entire transaction request. This can be useful for tracking the transaction on chain. To get the transaction ID, you can use the following method:
+
+<<< @/../../docs-snippets/src/guide/transactions/transaction-request.test.ts#transaction-request-7{ts:line-numbers}
```
Torres-ssf, line 51: We should say here that if we change the TX request it will result in changing the TX ID.

apps/docs-snippets/package.json
nedsalk: A bit weird that they'd expose only a part of their functionality in their "make global" opt-in feature.

apps/docs-snippets/src/guide/transactions/transaction-response.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    // previous transaction request. You will need the chain ID to obtain this.
+    const transactionId = transactionRequest.getTransactionId(provider.getChainId());
+
+    // Retrieve the transaction response from the transaction ID
+    const transactionResponse = await TransactionResponse.create(transactionId, provider);
+    
+    // Generate the transaction summary
+    const transactionSummary = await transactionResponse.getTransactionSummary();
+    // #endregion transaction-response-3
+
```
Torres-ssf, line 100: This flow can be misleading because if the TX request was never submitted to the chain, getting its `TransactionResponse` will result in an error.

Here we want to use an ID from a TX that was already submitted to the node, like:

```ts
  ...

  const { id: transactionId } = await wallet.transfer(recipient, 100, assetId)

  // Retrieve the transaction response from the transaction ID
  const transactionResponse = await TransactionResponse.create(transactionId, provider);

  // Generate the transaction summary
  const transactionSummary = await transactionResponse.getTransactionSummary();
```

or even

```ts
  // id from an already processed TX
  const transactionId = '0x...'

  // Retrieve the transaction response from the transaction ID
  const transactionResponse = await TransactionResponse.create(transactionId, provider);
```

Torres-ssf (changes requested):

apps/docs/src/guide/transactions/transaction-response.md
danielbate: Based on #1644 you're right this does make the response redundant for a tx just submitted. But we will still want this to fetch info about a tx that was submitted historically?

apps/docs/src/guide/transactions/index.md
danielbate: https://github.com/FuelLabs/fuels-ts/pull/1961/commits/5fcb542f03019c177c4234ea07a05c343dddd034

apps/docs/src/guide/transactions/transaction-request.md
danielbate: https://github.com/FuelLabs/fuels-ts/pull/1961/commits/5fcb542f03019c177c4234ea07a05c343dddd034

apps/docs-snippets/src/guide/transactions/transaction-response.test.ts
danielbate: So for this doc I give 3 examples. One for a contract call, one for a transaction request and this example was for an already processed TX. But I agree with your sentiment, so the latter example is more correct.

danielbate: https://github.com/FuelLabs/fuels-ts/pull/1961/commits/f8fd5dea744b62d2e384a7c4d0b0895ca574dbcf

apps/docs-snippets/src/guide/transactions/transaction-request.test.ts
danielbate: So I have added a note [here](https://github.com/FuelLabs/fuels-ts/pull/1961/commits/f622966217675595eb83db74f79141bc5e8b6e97).

The more complete example around tx signing remains at `apps/docs/src/guide/cookbook/transactions-with-multiple-signers.md`. I think that suffices but I will link that here. Do you think that is enough? The example above is mainly to demonstrate the methods accessible to a tx request.

danielbate: https://github.com/FuelLabs/fuels-ts/pull/1961/commits/ddb2f9a287608a5c5b2d970a912dc462c9088a27

apps/docs-snippets/src/guide/transactions/transaction-request.test.ts
danielbate: Hmm this snippet is attempting to demo methods available to a tx request to modify it, rather than flow. Hence the more verbose approach and transaction submission is not included. And we are only testing the shape of the tx. Do you still think we need to ensure the flow is correct?

apps/docs/src/guide/transactions/index.md
```diff
@@ -0,0 +1,14 @@
+# Transactions
+
+A transaction is a way of interacting with Fuel Network's blockchain and can include actions like transferring assets, deploying contracts and minting tokens.
```
Dhaiwat10, line 3: Feel like we should prefer using this wording over '_the_ xyz blockchain' because we are a VM-focused company, and not an L2 necessarily. 

```suggestion
A transaction is a way of interacting with a Fuel blockchain and can include actions like transferring assets, deploying contracts and minting tokens.
```

Dhaiwat10: Our SDK is supposed to work for _any_ blockchain using the FuelVM and not just some one chain that Fuel would deploy.

apps/docs/src/guide/transactions/index.md
```diff
@@ -0,0 +1,14 @@
...
+
+Within the Fuel Network, we have the following transaction types:
```
Dhaiwat10, line 5:
```suggestion
Within Fuel, we have the following transaction types:
```

Same as above. Reads better, and also more aligned with our narrative imo.

apps/docs/src/guide/transactions/index.md
```diff
@@ -0,0 +1,14 @@
...
+
+- Script
+- Create
+- Mint
+
+The SDK provides class helpers for handling script and create transactions: `ScriptTransactionRequest` and `CreateTransactionRequest`, respectively.
+Mint transactions can only be created by the block producer and do not have any use outside of block creation. Therefore, the SDK only provides the ability to decode them.
```
Dhaiwat10, line 12: Feel like this paragraph should be:

1. A _note_ rather than a main paragraph
2. Below the paragraph which is currently the last on this page

Dhaiwat10: I feel this way because this paragraph contains informationa about more advanced concepts like customt script transactions etc. We should tell users about the 'straightforward' transactions first, which is what the last paragraph does.

Dhaiwat10 (changes requested):
Great job @danielbate, a lot of work with this PR 👏🏻 

I must also say that the flow of these docs feels a bit off to me.

These docs are a part of the esssentials section and the first thing that they teach me shouldn't be how to create a custom transaction which requires custom script bytecode etc. It is not easy to replicate this and follow along.

We should rather start with 'normal' transactions like wallet transfers and contract calls, and then slowly go into the more advanced stuff like custom script transactions with custom inputs etc.

Thoughts?

danielbate:
Good points @Dhaiwat10. I'm going to edit the `transactions/index.md` so it makes no reference to those classes and move that to `transaction-request.md`. I think I'm going to add another page called `Simple Transactions`(?) that has a wallet transfer and basic contract call. Then I'll change `Transaction Request` to `Modifying the Transaction Request`?

danielbate:
@Dhaiwat10 I've actually [opted](https://github.com/FuelLabs/fuels-ts/pull/1961/commits/7b20becfd6431fc1f8ff726d812551ec6b66c2ff) to flesh out the index page, adding information about the more common tx types (transfers and contract calls) and move the custom tx specifics to the tx request page. i decided to just do links for contracts as I think we reference them enough in contracts. Let me know what you think.

apps/docs-snippets/src/guide/transactions/transaction-request.test.ts
danielbate: I've made some improvements to the flow based on @Dhaiwat10 comment, let me know what you think - https://github.com/FuelLabs/fuels-ts/pull/1961/commits/7b20becfd6431fc1f8ff726d812551ec6b66c2ff

apps/docs/src/guide/transactions/index.md
danielbate: Made some changes [here](https://github.com/FuelLabs/fuels-ts/pull/1961/commits/7b20becfd6431fc1f8ff726d812551ec6b66c2ff) based on these comments.

Dhaiwat10:
Thanks @danielbate, that sounds very fair

apps/docs-snippets/src/guide/transactions/transaction-request.test.ts
petertonysmith94: Nice one - the links are a good touch.

petertonysmith94 (approved):

apps/docs/src/guide/transactions/transaction-parameters.md
```diff
@@ -0,0 +1,28 @@
+# Transaction Parameters
+
+Transaction parameters allow you to configure various aspects of your blockchain transactions. The parameters are:
+
+1. Gas Price - The price you're willing to pay for each unit of [gas](https://docs.fuel.network/docs/intro/glossary/#gas) consumed during the transaction execution.
+1. Gas Limit - The maximum amount of gas you're willing to allow the transaction to consume. If the transaction execution requires more gas than the specified gas limit, the transaction will fail and any state changes made during the execution will be reverted.
+1. Maturity - The number of blocks that must pass before the transaction can be included in a block. This is useful for time-sensitive transactions, such as those involving time-locked assets.
+1. Max Fee - The maximum amount you're willing to pay for the transaction.
```
Dhaiwat10, line 8: The difference between the max fee and the gas limit is a bit unclear to me. Can you have another go at this?

apps/docs/src/guide/transactions/transaction-policies.md
```diff
@@ -0,0 +1,20 @@
+# Transaction Policies
+
+Transaction policies are rules that can govern how a transaction is processed, based off the transaction parameters that are passed to a transaction request. The available policies are as follows:
```
Dhaiwat10, line 3: Also, what's the difference between setting transaction parameters and setting transaction policies? Both seem to have the same set of fields.

Torres-ssf (approved):

apps/docs/src/guide/transactions/transaction-request.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Adding a Predicate to a Transaction Request
+
+Predicates are used to define the conditions under which a transaction can be executed. Therefore you may want to add a predicate to a transaction request to unlock funds that are utilized by a script. This can be added like so:
+
+<<< @/../../docs-snippets/src/guide/transactions/transaction-request.test.ts#transaction-request-5{ts:line-numbers}
+
+### Adding a Witness and Signing a Transaction Request
+
+The SDK provides a way of either modifying the witnesses for a transaction request directly, or by passing accounts. This will then signing the transaction request with the account's private key. Below will detail how to add a witness to a transaction request:
```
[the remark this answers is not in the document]
danielbate: https://github.com/FuelLabs/fuels-ts/pull/1961/commits/fd94e8fd0a848a082e001d61172d6aa72b01123b

apps/docs/src/guide/transactions/transaction-request.md
```diff
@@ -0,0 +1,67 @@
...
+
+Scripts can perform multiple actions on chain, therefore you may want to chain contract calls. For this you will need to add a contract to the transaction request. This can be done like so:
+
+<<< @/../../docs-snippets/src/guide/transactions/transaction-request.test.ts#transaction-request-4{ts:line-numbers}
+
+### Adding a Predicate to a Transaction Request
+
+Predicates are used to define the conditions under which a transaction can be executed. Therefore you may want to add a predicate to a transaction request to unlock funds that are utilized by a script. This can be added like so:
+
+<<< @/../../docs-snippets/src/guide/transactions/transaction-request.test.ts#transaction-request-5{ts:line-numbers}
```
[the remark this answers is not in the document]
danielbate: @Dhaiwat10 makes sense, some more info added [here](https://github.com/FuelLabs/fuels-ts/pull/1961/commits/1e64d97c955099bf196ab9e76b123d70123c1b39). I decided to not cover too much about the predicate to not add noise around the tx request, but I have now also linked to the predicate page.

@Torres-ssf yes conceptually unlock IMO means successfully validating the predicate. I had seen it used in that context before and feels like quite an encompassing word. However if it's not clear, I could use a more clear term?

apps/docs/src/guide/transactions/transaction-parameters.md
danielbate: I've had a go but it is tricky. Gas limit is the total gas usage limit. And max fee is like gas used * gas price limit. Not sure if there's a better way of describing that?

https://github.com/FuelLabs/fuels-ts/pull/1961/commits/6daff848b53832ef7d88350fe850155fbf5fd5f2

Dhaiwat10: Can you just maybe add this -> `gas used * gas price limit` to the description too? That would help

danielbate: https://github.com/FuelLabs/fuels-ts/pull/1961/commits/20871111460a759c3f845bb17eca973851defe6a

Dhaiwat10 (approved):

apps/docs/src/guide/transactions/transaction-response.md
danielbate: Going to resolve this conversation and leave this to be unimplemented by #1644, looks like there are blockers up the chain.

PR #1963, opened 2024-03-28
chore: reset base forc version to `0.49.3` except for experimental builds

nedsalk:
The official `beta-5` network version of `forc` is `v0.49.3`. This PR resets it to that version except for experimental builds that are dependent on newer versions (e.g. the new encoding schema).


packages/fuel-gauge/FORC_VERSION_EXPERIMENTAL
danielbate, line 1: I think this is better in `pacakges/forc/VERSION_EXPERIMENTAL`?

packages/fuel-gauge/package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "pretest": "run-s build:forc build:forc-experimental build:process-predicates",
     "build:forc": "pnpm fuels-forc build -p test/fixtures/forc-projects --release",
-    "build:forc-experimental": "pnpm fuels-forc build -p test/fixtures/forc-projects-experimental --experimental-new-encoding --release",
+    "build:forc-experimental": "tsx ./scripts/build-forc-experimental.ts",
```
danielbate, line 10: IMO it would be better to split these up into `install:forc-experimental` and `build:forc-experimental` commands. Then you can go straight to `package.json` to alter build flags as normal.

packages/fuel-gauge/scripts/build-forc-experimental.ts
```diff
@@ -0,0 +1,25 @@
...
+if (installedForcVersion !== EXPECTED_FORC_VERSION) {
+  // eslint-disable-next-line no-console
+  console.log(
+    `Please install the correct version of forc. Installed version: ${installedForcVersion}, expected version: ${EXPECTED_FORC_VERSION}`
+  );
+  process.exit(1);
+}
+
+execSync(
+  `${FORC} build -p test/fixtures/forc-projects-experimental --experimental-new-encoding --release`,
```
danielbate, line 23: See above for comment to move this to `package.json`

packages/fuel-gauge/package.json
nedsalk: There is no installing going on in `build:forc-experimental`, just the usage of `forc` that was installed via the CI. Do you want me to move the logic from the [CI](https://github.com/FuelLabs/fuels-ts/pull/1963/files#diff-ad01a716da494dcbc95adfbc4041860456aabe99d5a1c39b49a61b821e303ed9R42) to an `install:forc-experimental` script?

packages/fuel-gauge/scripts/build-forc-experimental.ts
nedsalk: Continued in https://github.com/FuelLabs/fuels-ts/pull/1963#discussion_r1542537886

packages/fuel-gauge/FORC_VERSION_EXPERIMENTAL
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1963/commits/e801acf60e7ccea097adb438f717749d191efffd

packages/fuel-gauge/package.json
danielbate: Ah okay I had misinterpreted the code. I think the CI code is fine. I think having the `forc` command actually inside `package.json` is the only thing I'd prefer. But the validate script would have to run before anyway. Potentially this isn't worth the benefit of just having to go inside one additional file. It's only different at the moment because we have 2 different approaches to forc installation, and this is indeed a proof of concept.

danielbate (approved):

danielbate:
Probably best to get 3 reviews here due to the new approach.

petertonysmith94 (approved):

arboleya (approved):

PR #1966, opened 2024-03-28
chore: upgrading `forc` to `0.51.1`

arboleya:
This PR adds back support to `forc@0.51.1`, removed in:
 - #1963


.changeset/tidy-olives-dance.md
```diff
@@ -0,0 +1,7 @@
+---
+"@fuel-ts/abi-typegen": patch
+"@fuel-ts/forc": patch
+"@fuel-ts/versions": patch
+---
+
+chore: upgrading forc to `0.51.1`
```
arboleya, line 7: > [!WARNING]
> Do not release before public network compatibility.

nedsalk:
This should be a breaking change `chore!` PR. It'd create an @rc-1966 release for every merge of `master` into it, though, if we designate it as a breaking change PR now. Not sure if we want that for the time being.

packages/fuel-gauge/src/script-with-vectors.test.ts
```diff
@@ -41,13 +41,13 @@ describe('Script With Vectors', () => {
     expect(formattedLog).toEqual([
       7,
       'vector.buf.ptr',
-      11256,
+      11232,
```
maschad, line 44: Out of curiosity where are these _magic_ numbers retrieved from? It would be nice if we could replace it with some kind of named generator function.

arboleya: They are pertinent to the data returned from the Sway script.

https://github.com/FuelLabs/fuels-ts/blob/08241a90eeac2d4c8d2dd00ba8e7a53e24c2ca60/packages/fuel-gauge/test/fixtures/forc-projects/script-with-vector/src/main.sw#L14-L20

maschad: Noted -  perhaps I'm being a bit pedantic in this case but I am generally against use of magic numbers in code,  I think we should have them as named constants. Should these values change again it would be easier to make inferences about what is changing i.e. the pointer or the address of the vector.

arboleya: Hmm, let me see if I'm following.

Do you mean to have something like this:

```ts
const XYZ = 123;

expect(something).toEqual(XYZ);
```

Instead of this:

```ts
expect(something).toEqual(123);
```

Whereas `XYZ` would be a more descriptive name?

maschad: Exactly, so that going forward, if the value changes, at least we can more easily determine if that is an expected change or a bug.

arboleya: I agree, and that's usually how I write my stuff.

One thing to keep in mind is upgrading `forc` and `fuel-core` usually requires updating [several](https://github.com/FuelLabs/fuels-ts/pull/2017/files) files, and sometimes refactoring all legacy code it touches, although tempting, is not the best move.

This is an easy one tho, and because this PR is about to be closed, I did it on its [successor](https://github.com/FuelLabs/fuels-ts/pull/2017/commits/5be60b4fafc3717f46d039c1a7a17a8e74ab20fe).

.changeset/tidy-olives-dance.md
arboleya: I moved the warning note to the PR description and marked it as a draft.

PR will be deleted soon.

packages/fuel-gauge/src/script-with-vectors.test.ts
maschad: Noted, thanks for the clarification.

arboleya:
Superseded by:
 - https://github.com/FuelLabs/fuels-ts/pull/2017

PR #1976, opened 2024-03-29
feat!: fetch base asset ID from chain

### LINKED ISSUE #1948: Dynamic base asset ID
Soon, we'll move to a more dynamic way of loading/interacting with the base `AssetID`. This change will affect the Fuel VM, the client, and SDKs.

These are some of the _known_ the changes that need to be done in the SDK: 

## Remove `BaseAssetId`

```tsx
export const ZeroBytes32 = '0x0000000000000000000000000000000000000000000000000000000000000000';
- export const BaseAssetId = ZeroBytes32;
export const EmptyRoot = '0xe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855';
```

## Fetch base ID in `Provider` & add `getBaseAssetId()`

The `fetchChain()` method of the `Provider` class already fetches chain configuration. This function simply needs to add the base asset ID to its query and save this value in the `consensusParameters` property.

Additionally, a `getBaseAssetId()` method would be added to return this value.

## Replace  `BaseAssetId`  with  `getBaseAssetId()` in codebase

Any portion of the codebase that uses `BaseAssetId` must be updated to call the new `getBaseAssetId()` function.

Most usages are simply in unit tests, however there are some instances of it used in the main packages, such as for funding a transaction with gas funds.

```tsx
export const fundPredicate = async <T extends InputValue[]>(
  wallet: WalletUnlocked,
  predicate: Predicate<T>,
  amountToPredicate: BigNumberish
): Promise<BN> => {
  const { minGasPrice } = wallet.provider.getGasConfig();

  const request = new ScriptTransactionRequest({
    gasPrice: minGasPrice,
  });

  - request.addCoinOutput(predicate.address, amountToPredicate, BaseAssetId);
  + const baseAssetId = predicate.provider().getBaseAssetId();
  + request.addCoinOutput(predicate.address, amountToPredicate, BaseAssetId);
  const { minFee, requiredQuantities, gasUsed } = await wallet.provider.getTransactionCost(request);
  request.gasLimit = gasUsed;
  await wallet.fund(request, requiredQuantities, minFee);

  await wallet.sendTransaction(request, { awaitExecution: true });

  return predicate.getBalance();
};
```

Note that there might be other places where we're pulling the base asset ID. Part of the work here is scanning the codebase for it, and making sure we're dynamically pulling the base asset ID.

danielbate:
Closes #1948 

### Breaking Changes:

- `BaseAssetId` is no longer exported by `fuels`. It must be fetched from the chain as it is configurable by the chain.

```ts
// Before
import { BaseAssetId } from `fuels`;

// After
const baseAssetId = provider.getBaseAssetId();
```
- `TransactionRequest.addCoinOutput` now requires an `assetId`, it no longer defaults to the `baseAssetId`.
- `TransactionRequest.addChangeOutput` now requires an `assetId`, it no longer defaults to the `baseAssetId`.
- `TransactionRequest.fundWithFakeUtxos` now requires passing the `baseAssetId` as a function parameter. This is the only function that is base asset aware, so that it can be used specifically to estimate the transaction cost. 
- `CoinQuantityLike` now requires an `AssetId`. Previously most of it's usages would default to the `BaseAssetId`, as this must now be fetched, then it must be passed to the type.


packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -105,6 +107,8 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
   outputs: TransactionRequestOutput[] = [];
   /** List of witnesses */
   witnesses: TransactionRequestWitness[] = [];
+  /** Base asset ID */
+  baseAssetId: string = ZeroBytes32;
```
danielbate, line 111: Left the default in as it minimises breaking changes and provides a better DX for users on local test net. All docs and docs snippets will reference fetching and setting this value.

If we want to fetch a sensible default value, we'll need to pass in a provider to the transaction request.

fuel-service-user:
✨ A PR has been created under the `rc-1976` tag on `fuels-wallet` repo.
https://github.com/FuelLabs/fuels-wallet/pull/1212

packages/account/src/providers/transaction-request/transaction-request.ts
danielbate: If it's not set and the user does not own the `baseAssetId` then the tx will fail validation as the user does not have enough of the required coins to fund the transaction. Or they don't set it, and the base asset ID is `ZeroBytes32`, and the user does have the asset then tx will succeed (equal to current behaviour).

packages/script/src/script.test.ts
```diff
@@ -46,6 +46,7 @@ const callScript = async <TData, TResult>(
   response: TransactionResponse;
 }> => {
   const { minGasPrice } = account.provider.getGasConfig();
+  const BaseAssetId = account.provider.getBaseAssetId();
```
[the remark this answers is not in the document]
danielbate: https://github.com/FuelLabs/fuels-ts/pull/1976/commits/1032e67212f5f055ff200683083004aa9824bcbe

packages/account/src/providers/coin-quantity.ts
```diff
@@ -1,12 +1,11 @@
...
 import type { BytesLike } from '@fuel-ts/interfaces';
 import type { BigNumberish, BN } from '@fuel-ts/math';
 import { bn } from '@fuel-ts/math';
 import { hexlify } from '@fuel-ts/utils';
 
 export type CoinQuantityLike =
-  | [amount: BigNumberish, assetId?: BytesLike, max?: BigNumberish]
-  | { amount: BigNumberish; assetId?: BytesLike; max?: BigNumberish };
+  | [amount: BigNumberish, assetId: BytesLike, max?: BigNumberish]
+  | { amount: BigNumberish; assetId: BytesLike; max?: BigNumberish };
```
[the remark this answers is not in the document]
danielbate: Added.

packages/account/src/providers/transaction-request/transaction-request.ts
arboleya: I share the frustration, and I also understand this would rely on TX validation (runtime), but it might be better to fail earlier (compile time). Of course, it's a DX trade-off, but something tells me we should not favor local test net for this decision.

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -105,6 +107,8 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
   outputs: TransactionRequestOutput[] = [];
   /** List of witnesses */
   witnesses: TransactionRequestWitness[] = [];
+  /** Base asset ID - should be fetched from the chain */
+  baseAssetId: string = ZeroBytes32;
```
Torres-ssf, line 111: If we are going to have `baseAssetId` on `TransactionRequest` we need to make it an optional property, and this needs to be set later on.

Setting this early as `ZeroBytes32` can be misleading

packages/account/src/providers/transaction-request/transaction-request.ts
Torres-ssf: If we are going to have `baseAssetId` on `TransactionRequest` we need to make it an optional property, and this needs to be set later on.

Setting this early as `ZeroBytes32` can be misleading

packages/program/src/contract-call-script.ts
```diff
@@ -283,7 +283,7 @@ export const getScriptDataV1: ContractCallScriptFn = (
   // 1. Amount
   scriptData.push(new BigNumberCoder('u64').encode(call.amount || 0));
   // 2. Asset ID
-  scriptData.push(new B256Coder().encode(call.assetId?.toString() || BaseAssetId));
+  scriptData.push(new B256Coder().encode(call.assetId?.toString() || ZeroBytes32));
```
Torres-ssf, line 286: We should not use `ZeroBytes32` here as a default value. 

If the used chain config sets the BaseAsset ID to a different value, this will result in an error. 

We need to make the related functions have the base asset ID as a parameter.

internal/check-imports/src/references.ts
```diff
@@ -63,7 +63,7 @@ log(runCliAction);
  * address
  */
 log(Address);
-log(BaseAssetId);
+log(ZeroBytes32);
```
Torres-ssf, line 66: We should remove this log as it does not have a purpose anymore

Torres-ssf (changes requested):

internal/check-imports/src/references.ts
danielbate: https://github.com/FuelLabs/fuels-ts/pull/1976/commits/cef28a387a71fc2615fb6112bb6466be98e81acd

packages/program/src/contract-call-script.ts
arboleya: Could we assume `BaseAssetId` will be the most used? If yes, are there any downsides to defaulting to it contextually, and could this reduce necessary breaking change adjustments?

```ts
call.assetId ?? (call.amount ? BaseAssetId : ZeroBytes32)
```

arboleya (approved):

Dhaiwat10:
Sorry @danielbate I'm very late to the party here but is there a way for us to fetch the base asset ID just like the chainInfo and nodeInfo when we initialize the `Provider`, and then cache it? This way we can avoid the breaking change around having to fetch and pass it in via all the transactions. This breaking change really degrades the DX imo.

arboleya:
@Dhaiwat10 I guess that's the same thing I [said here](https://github.com/FuelLabs/fuels-ts/pull/1976#discussion_r1565515951).

arboleya (approved):

apps/docs-snippets/src/guide/transactions/transaction-request.test.ts
```diff
@@ -91,26 +98,32 @@ describe('Transaction Request', () => {
...
+    transactionRequest.addCoinOutput(recipientAddress, 1000, baseAssetId);
 
     // Adding message inputs
     transactionRequest.addMessageInput(message);
     // #endregion transaction-request-3
 
     expect(transactionRequest.script).toEqual(arrayify(scriptBytecode));
     expect(transactionRequest.inputs.length).toEqual(4);
-    expect(transactionRequest.outputs.length).toEqual(2);
+    expect(transactionRequest.outputs.length).toEqual(4);
```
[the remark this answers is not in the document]
danielbate: This was a mistake - https://github.com/FuelLabs/fuels-ts/pull/1976/commits/50b9defa6ba814beaec4303af2d8138c885986a9

arboleya (dismissed):

packages/account/src/providers/transaction-request/transaction-request.ts
danielbate: Went through the design and DX implications with @nedsalk earlier and attempting to mitigate breaking changes. There isn't much value in moving the breaking change from the constructor to the function signatures. And making it optional would require checks where it is used.

The real value would be in making a transaction request agnostic from the idea of a base asset ID. Asset IDs should just be passed, if it should be a base asset ID, then that is up to the user. This is a design decision right now to improve DX. However the only function that we can see that requires knowing the base asset ID specifically is `addMessageInput` which uses the base asset to create a change output.

Given the above is out of scope here, and this approach is low risk as requires the user to pass the base asset at tx instantiation, I am going to resolve this conversation.

packages/program/src/contract-call-script.ts
danielbate: @Torres-ssf let me know what you think of the above or resolve accordingly, as I'll merge after this one is closed 👍🏻 

But as discussed, `ZeroBytes32` is a place holder here for the required property space rather than the value of the asset to forward.

The base asset ID is set on the request level. For an invocation scope, this is actually pretty straightforward and negates breaking changes as we have access to the provider from the program.

packages/account/src/providers/transaction-request/transaction-request.ts
Torres-ssf: @danielbate I don't follow. A `MessageCoin` has the property `assetId` set on it. 

Why TransactionRequest needs to know the BaseAsset when adding an `OutputChange` for it?

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -424,7 +427,7 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
     this.pushInput(input);
 
     // Insert a ChangeOutput if it does not exist
-    this.addChangeOutput(recipient, assetId);
+    this.addChangeOutput(recipient, this.baseAssetId);
```
Torres-ssf, line 430:
```suggestion
    this.addChangeOutput(recipient, message.assetId);
```

All resources are fetched from `Provider.getResourcesToSpend`, which is the graphQL query `coinsToSpend`. 

When the user owns messages that have not been spent yet, they will be returned with the `assetId` property (base asset) already [set to it](https://github.com/FuelLabs/fuels-ts/blob/cb2da0408e38b4af9a1198fa4f3b504929274b7f/packages/account/src/providers/provider.ts#L1087-L1095).

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -121,14 +125,16 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
...
+    baseAssetId,
+  }: BaseTransactionRequestLike) {
     this.gasPrice = bn(gasPrice);
     this.maturity = maturity ?? 0;
     this.witnessLimit = witnessLimit ? bn(witnessLimit) : undefined;
     this.maxFee = maxFee ? bn(maxFee) : undefined;
     this.inputs = inputs ?? [];
     this.outputs = outputs ?? [];
     this.witnesses = witnesses ?? [];
+    this.baseAssetId = baseAssetId;
```
Torres-ssf, line 137: We should remove the `baseAssetId` from the `TransactionRequest`

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -73,6 +73,8 @@ export interface BaseTransactionRequestLike {
   outputs?: TransactionRequestOutput[];
   /** List of witnesses */
   witnesses?: TransactionRequestWitness[];
+  /** Base asset ID - should be fetched from the chain */
+  baseAssetId: string;
```
Torres-ssf, line 77: We should remove the `baseAssetId` from the `TransactionRequest`

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -494,12 +497,12 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
    * @param amount - Amount of coin.
    * @param assetId - Asset ID of coin.
    */
-  addCoinOutput(to: AddressLike, amount: BigNumberish, assetId: BytesLike = BaseAssetId) {
+  addCoinOutput(to: AddressLike, amount: BigNumberish, assetId?: BytesLike) {
```
Torres-ssf, line 500: `assetId` must be always required here

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -494,12 +497,12 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
...
     this.pushOutput({
       type: OutputType.Coin,
       to: addressify(to).toB256(),
       amount,
-      assetId,
+      assetId: assetId ?? this.baseAssetId,
```
Torres-ssf, line 505:
```suggestion
      assetId,
```

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -530,7 +533,7 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
    * @param to - Address of the owner.
    * @param assetId - Asset ID of coin.
    */
-  addChangeOutput(to: AddressLike, assetId: BytesLike = BaseAssetId) {
+  addChangeOutput(to: AddressLike, assetId?: BytesLike) {
```
Torres-ssf, line 536: `assetId` must be always required here

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -541,7 +544,7 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
       this.pushOutput({
         type: OutputType.Change,
         to: addressify(to).toB256(),
-        assetId,
+        assetId: assetId ?? this.baseAssetId,
```
Torres-ssf, line 547:
```suggestion
        assetId,
```

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -627,7 +630,7 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
       }
     };
 
-    updateAssetInput(BaseAssetId, bn(100_000_000_000));
+    updateAssetInput(this.baseAssetId, bn(100_000_000_000));
```
Torres-ssf, line 633: This method should receive the `baseAssetId` as a parameter

Torres-ssf (changes requested):
Not sure if I am missing something, but I believe we do not need the `baseAssetId` specified in the `TransactionRequest`.

If I understood correctly, this change is happening because we need the `BaseAssetId` when setting the `assetId` for a `MessageInput`. 

However, a `MessageCoin` already carries the `assetId` property and can be used when adding an `OutputCoin` or `OutputChange` to the Transaction

Aside this, is there another reason to have the `baseAssetId` being required to instantiate a `TransactionRequest`?

danielbate:
> Not sure if I am missing something, but I believe we do not need the baseAssetId specified in the TransactionRequest

Agreed, but from a DX POV, we are now handing off the responsibility of ensuring the base asset is passed around correctly completely to the consumer. Previously we were using optional checks and default values for asset IDs that would then evaluate to the base asset.

I am actually more for this change, however when I initially went for making the property optional, and to hand off responsibility, it was deemed [too risky](https://github.com/FuelLabs/fuels-ts/pull/1976#discussion_r1556141118).

> If I understood correctly, this change is happening because we need the BaseAssetId when setting the assetId for a MessageInput. However, a MessageCoin already carries the assetId property and can be used when adding an OutputCoin or OutputChange to the Transaction

When we create the corresponding change output, we are creating it specifically for the [base asset](https://github.com/FuelLabs/fuels-ts/blob/6cf0051a9958430db8a2e59f33f63df6b204b8c9/packages/account/src/providers/transaction-request/transaction-request.ts#L398). Is there any risk in returning funds for the same asset as the message? This area is not very well tested.

But agreed, if we wanted to completely decouple that class from the concept of a base asset, `addMessageInput` right now seems to be the only area that makes a specific reference to it.

> cc @Torres-ssf

packages/account/src/providers/transaction-request/transaction-request.ts
danielbate: Responded [here](https://github.com/FuelLabs/fuels-ts/pull/1976#issuecomment-2065849482).

Torres-ssf:
> When we create the corresponding change output, we are creating it specifically for the [base asset](https://github.com/FuelLabs/fuels-ts/blob/6cf0051a9958430db8a2e59f33f63df6b204b8c9/packages/account/src/providers/transaction-request/transaction-request.ts#L398). Is there any risk in returning funds for the same asset as the message? This area is not very well tested.

This code was added because all `MessageCoin` has as the `assetId` the Base asset.

MessageCoins are bridged ETH, brought to Fuel from the main layer.

For instance, take a look at the messages that we define at `chainConfig`:

```ts
    "messages": [
      {
        "sender": "0xc43454aa38dd91f88109a4b7aef5efb96ce34e3f24992fe0f81d233ca686f80f",
        "recipient": "0x69a2b736b60159b43bb8a4f98c0589f6da5fa3a3d101e8e269c499eb942753ba",
        "nonce": "0101010101010101010101010101010101010101010101010101010101010101",
        "amount": "0x000000000000FFFF",
        "data": "",
        "da_height": "0x00"
      },
      ...
``` 

As you can see there is no `assetId`. If you start the node using this chain config and try the following query on the playground:
```
query {
  coinsToSpend(owner: "0x69a2b736b60159b43bb8a4f98c0589f6da5fa3a3d101e8e269c499eb942753ba", queryPerAsset: [{ assetId: "0x0000000000000000000000000000000000000000000000000000000000000000", amount: "1"}]) {
  	... on MessageCoin {
      amount
      assetId
      nonce
    }
  }
}
```

 The query will return the specific message, as a MessageCoin, carrying the `assetId` property.

danielbate:
Thanks @Torres-ssf, that makes sense.

If we're happy then, we'll hand off ownership of the base asset ID completely to the consumer. I'll make the `TransactionRequest` have no knowledge of the base asset, and instead it just passes asset IDs.

packages/account/src/providers/transaction-request/transaction-request.ts
danielbate: https://github.com/FuelLabs/fuels-ts/pull/1976/commits/1239a4bbaac8a320ced1c66eeab51224269e0d8d

packages/account/src/providers/transaction-request/transaction-request.ts
danielbate: https://github.com/FuelLabs/fuels-ts/pull/1976/commits/1239a4bbaac8a320ced1c66eeab51224269e0d8d

packages/account/src/providers/transaction-request/transaction-request.ts
danielbate: https://github.com/FuelLabs/fuels-ts/pull/1976/commits/1239a4bbaac8a320ced1c66eeab51224269e0d8d

packages/account/src/providers/transaction-request/transaction-request.ts
danielbate: https://github.com/FuelLabs/fuels-ts/pull/1976/commits/1239a4bbaac8a320ced1c66eeab51224269e0d8d

packages/account/src/providers/transaction-request/transaction-request.ts
danielbate: Tx request no longer requires a base asset ID. https://github.com/FuelLabs/fuels-ts/pull/1976/commits/1239a4bbaac8a320ced1c66eeab51224269e0d8d

packages/program/src/contract-call-script.ts
danielbate: Base asset ID has been removed from tx request. So if a call is forwarding assets, the asset ID will be passed through to the contract call script. If no assets are being forwards, `ZeroBytes32` is used as the property space placeholder. https://github.com/FuelLabs/fuels-ts/pull/1976/commits/1239a4bbaac8a320ced1c66eeab51224269e0d8d

packages/account/src/providers/transaction-request/transaction-request.ts
danielbate: I agreed here, this is the only reference to `baseAssetId` as this is the only place where we'd want it specifically. https://github.com/FuelLabs/fuels-ts/pull/1976/commits/1239a4bbaac8a320ced1c66eeab51224269e0d8d

.github/workflows/pr-release.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
     name: "Release PR to npm"
     runs-on: ubuntu-latest
     # comment out if:false to enable release PR to npm
-    if: false
+    # if: false
```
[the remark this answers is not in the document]
danielbate: Do we want to be creating a PR release for everything? 🤔

packages/account/test/fuel-wallet-connector.test.ts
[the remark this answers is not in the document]
danielbate: Most likely yes but it depends how these are being used.

> cc @FuelLabs/frontend

packages/account/src/providers/transaction-request/transaction-request.ts
Dhaiwat10: @danielbate can you update the PR description?

.github/workflows/pr-release.yaml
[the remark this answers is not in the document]
danielbate: Yep definitely for that one. Will resolve this then and we'll leave the RC for @Torres-ssf when conflicts are resolved in #2021.

Torres-ssf (approved):

arboleya (approved):

Issue #1515, opened 2023-12-13
Review UTXO caching logic


Torres-ssf:
We recently experienced [flaky tests](https://github.com/FuelLabs/fuels-ts/pull/1514) related to the [optional UTXO caching](https://github.com/FuelLabs/fuels-ts/pull/930) that happens on the Provider class.

This logic was added to solve an [issue](https://github.com/FuelLabs/fuels-ts/issues/865) that was happening on the test net (beta-3?).

Now that we are moving to a new test net, and have identified that something might be wrong with how the UTXOs are being cached, we need to reevaluate this logic and validate if it needs to be updated or if we could remove it enterely.

arboleya:
What have we identified that might be wrong?

Torres-ssf:
@arboleya I am unsure about what was causing this one.

I would need to investigate this further. I can take a look at it this week.

PR #1978, opened 2024-03-30
docs: update provider docs

### LINKED ISSUE #1918: Docs: Provider
We need to improve current docs over [here](https://fuellabs.github.io/fuels-ts/guide/providers/), here's an initial draft:

- Provider
   - What is it? 
   - Instantiating a Provider
   - Provider Options
     - `fetch`
     - `timeout`
     - `cacheUtxo`
     - `retryOptions` 
     - `requestMiddleware`

Related:
  - https://github.com/FuelLabs/fuels-ts/issues/975

> _Consider documenting options like [this](https://fuellabs.github.io/fuels-ts/guide/cli/config-file.html), as it is superior to the auto-generated Typedocs._

nedsalk:
closes #1918


apps/docs/src/guide/provider/index.md
```diff
@@ -0,0 +1,48 @@
+# Provider
+
+The [`Provider`](../../api/Account/Provider.md) class lets you connect to a Fuel node ([_*local*_](../introduction/getting-started.md#connecting-to-a-local-node) or [_*external*_](../introduction/getting-started.md#connecting-to-the-testnet)) and interact with it, encapsulating common client operations in the SDK.
+
+<<< @/../../docs-snippets/src/guide/provider/provider.test.ts#provider-definition{ts:line-numbers}
+
+## Provider Options
```
nedsalk, line 7: I did not describe `cacheUtxo` here because I couldn't figure out how to convey why it'd be important to the end user.  #1515 questions the need for this functionality as well. Also, because it's an opt-in feature, I doubt that our users were even using it to begin with, and I don't remember any reports in `beta-4` and `beta-5` being made for the issue this caching solved. Perhaps we don't need it anymore.

If somebody can provide a user-relevant explanation for it, I'd appreciate it very much. If not, we can handle this part of the documentation alongside #1515 as well.

apps/docs/src/guide/provider/index.md
```diff
@@ -0,0 +1,47 @@
...
+_NOTE: retrying is only done when a connection cannot be established. If the connection is established and the node throws an error, no retry will happen._
+
+You can provide the following settings:
+
+- `maxRetries` - Amount of attempts to retry after initial attempt before failing the call.
+- `backoff` - Strategy used to define the intervals between attempts.
+  - `exponential` _(default)_: Doubles the waiting time with each attempt.
+  - `linear` - Increases the waiting time linearly with each attempt.
+  - `fixed`: Uses a constant waiting time between attempts.
+- `baseDelay` - Base time in milliseconds for the backoff strategy.
```
petertonysmith94, line 25: - Worth noting it defaults to 150ms if not set.
- Is it worth changing this to just `delay`?
- Could we reference `delay` instead of `waiting time` in backoff options?

petertonysmith94 (commented):
Small nits

apps/docs/src/guide/provider/index.md
```diff
@@ -0,0 +1,47 @@
...
+<<< @/../../docs-snippets/src/guide/provider/provider.test.ts#provider-definition{ts:line-numbers}
+
+## Provider Options
+
+You can provide various [options](../../api/Account/index.md#provideroptions) on `Provider` instantiation to modify its behavior.
+
+### `retryOptions`
+
+Calls done via the `Provider` towards a fuel node will fail if the connection cannot be established.
+Specifying retry options allows you to customize the way you want to handle that failure scenario before ultimately throwing an error.
```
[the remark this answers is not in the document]
nedsalk: The way you phrased it implies we retry by default, but we don't. It's an opt-in feature.

apps/docs/src/guide/provider/index.md
nedsalk: - https://github.com/FuelLabs/fuels-ts/pull/1978/commits/f883024802c324b381c53faf69b3ea6835559454
- I chose `baseDelay` because it implies that it's not _the delay_ but something used to get the real delay value between retries
- https://github.com/FuelLabs/fuels-ts/pull/1978/commits/4a2c0d09f0ea2121fab156ceae7e79f8ddb826f6

apps/docs/src/guide/provider/index.md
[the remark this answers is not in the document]
petertonysmith94: That's fair.
Will resolve 👍🏼

apps/docs/src/guide/provider/index.md
```diff
@@ -0,0 +1,47 @@
+# Provider
+
+The [`Provider`](../../api/Account/Provider.md) class lets you connect to a Fuel node ([_*local*_](../introduction/getting-started.md#connecting-to-a-local-node) or [_*external*_](../introduction/getting-started.md#connecting-to-the-testnet)) and interact with it, encapsulating common client operations in the SDK.
+
+<<< @/../../docs-snippets/src/guide/provider/provider.test.ts#provider-definition{ts:line-numbers}
```
Dhaiwat10, line 5: After this section, I think it would be beneficial to have a section that gives people context about why the `Provider` is important, i.e. telling them where it is used in 'real-world' code snippets. 

For example, we shuold tell them that the provider is important because it is used every time you want to initialize a wallet, deploy a contract, read data from the chain, etc. All those use cases. We should give people a brief idea. Thoughts?

Dhaiwat10: You should probably also change the code snippet at the top and show something like:

- Initiating a provider
- Creating a wallet using the provider
- And then sending a transaction using a wallet

Dhaiwat10:
Thoughts on moving this page to the 'Basics' section?

apps/docs/src/guide/provider/index.md
```diff
@@ -0,0 +1,47 @@
+# Provider
+
+The [`Provider`](../../api/Account/Provider.md) class lets you connect to a Fuel node ([_*local*_](../introduction/getting-started.md#connecting-to-a-local-node) or [_*external*_](../introduction/getting-started.md#connecting-to-the-testnet)) and interact with it, encapsulating common client operations in the SDK.
```
danielbate, line 3: I think this could be made a bit more general about the kind of functions you can perform with a provider. Then info on connecting using it.

danielbate (commented):
@Dhaiwat10 agreed provider is a basic concept, but we should split out the options into a separate page.And add some more snippets to the provider index page on operations it performs.

apps/docs/src/guide/provider/index.md
arboleya: If the flag is there, we should document it.

If later it is removed, docs can also removed with it.

apps/docs/src/guide/provider/index.md
nedsalk: @Dhaiwat10 I have updated the description. I also changed the code snippet, but opted for fetching the balance of the wallet instead of sending a transaction because the transaction first has to be prepared before being sent, which would take too much space and take focus away from the topic at hand which is the `Provider`.

I'll wait until #1961 is merged so that I can link to relevant pages so that the transaction sending part is referenced, but the text on the page seems okay to me right now.

nedsalk:
@Dhaiwat10 @danielbate I moved it to the `Basics` section and put `Provider Options` into a separate page. Because there already was a `Querying the Chain` page, I just referenced it at the end of `provider/index.md` instead of writing multiple examples. I did update the single example to showcase how a `Provider` is used in conjunction with a wallet, though.

apps/docs/src/guide/provider/index.md
```diff
@@ -0,0 +1,9 @@
+# Provider
+
+The [`Provider`](../../api/Account/Provider.md) lets you connect to a Fuel node ([_*local*_](../introduction/getting-started.md#connecting-to-a-local-node) or [_*external*_](../introduction/getting-started.md#connecting-to-the-testnet)) and interact with it, encapsulating common client operations in the SDK. Those operations include querying the blockchain for network-, block-, and transaction-related info (and [more](../../api/Account/Provider.md)), as well as sending transactions to the blockchain.
+
+All higher-level abstractions (e.g. [`Wallet`](../wallets/index.md), [`Contract`](../contracts/index.md)) that interact with the blockchain go through the `Provider`, so it's used for various actions like getting a wallet's balance, deploying contracts, querying their state, etc.
```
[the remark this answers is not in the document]
nedsalk: I wrote it like this because https://github.com/FuelLabs/fuels-ts/pull/1978#discussion_r1550989141 requested a mention of some examples.

apps/docs/src/guide/provider/index.md
arboleya: @Torres-ssf Objections?

Torres-ssf: @arboleya I do agree that we should document this, but for now, I am not certain why this was added.

I will take a look into this and add some relevant explanations to our docs.

Torres-ssf: https://github.com/FuelLabs/fuels-ts/issues/2056

apps/docs/src/guide/provider/querying-the-chain.md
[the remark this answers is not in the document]
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/1978/commits/a81be024bcfbcc0530fcfbb8648d77e854687a80

Note: I commented out message-related examples because they were using hard-coded values that aren't copy-pastable and put a todo related to #1356 which adds tooling that'd make these examples self-sufficient.

arboleya (approved):

PR #1986, opened 2024-04-01
docs: parsing and formatting numbers

### LINKED ISSUE #1554: No unit (decimals) handling documentation
Nowhere in the documentation does it mention anything to do with units or handling large / small units with different levels of precision. This is very basic, and should be mentioned / included as to a proper way to do so. You can see the format/parse unit facilities of Ethers.js to see what that looks like.

> [!NOTE]

petertonysmith94:
Closes #1554 

---

Documentation around parsing and formatting number (of varied units).

Drawing reference from:
- https://docs.ethers.org/v5/api/utils/display-logic/#utils-parseUnits
- https://docs.ethers.org/v5/api/utils/display-logic/#utils-formatUnits

### DIFF
--- a/apps/docs-snippets/src/guide/utilities/unit-conversion.test.ts
+++ b/apps/docs-snippets/src/guide/utilities/unit-conversion.test.ts
@@ -1,0 +1,67 @@
+import { bn } from 'fuels';
+
+/**
+ * @group node
+ * @group browser
+ */
+describe('unit-conversion', () => {
+  it('should convert one gwei to wei', () => {
+    // #region parse-units-1
+    const oneGwei = bn.parseUnits('0.000000001')
+
+    // 1
+    const result = oneGwei.toNumber();
+    // #endregion parse-units-1
+
+    expect(result).toEqual(1);
+  })
+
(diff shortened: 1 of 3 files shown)


petertonysmith94:
@FuelLabs/sdk-ts Does anyone have thoughts about exporting [common units](https://docs.ethers.org/v5/api/utils/display-logic/#display-logic--named-units)?

```ts
// Units of Ether
export const UNITS_WEI = 0;
export const UNITS_KWEI = 3;
export const UNITS_MWEI = 6;
export const UNITS_GWEI = 9;
export const UNITS_ETHER = 18;
```

danielbate:
@petertonysmith94 I think it should be decimals rather than units, I think that is used more universally.

danielbate:
@petertonysmith94 yes I think the decimal prefix is better. And I do like the ethers API with second param being the unit of measure.

maschad (changes requested):
Looks good overall, I just have one requested change

maschad (approved):

Torres-ssf (approved):

Issue #1782, opened 2024-02-20
Error for decode erros in contract


luisburigo:
### What version of fuels-ts are you using?

0.73.0

### Steps to Reproduce

1.  Create two contracts: 
```sway
abi ContractB {
    #[storage(read)]
    fn method();
}

storage {
    initialized: bool = true, // the value 'true' is to reproduce the error
}

impl ContractB for Contract {
    #[storage(read)]
    fn method() {
       let initalized = storage.initialized.try_read().unwrap_or(false);
       require(!initalized, ContractBError::AlreadyInitialized);
    }
}
```

```sway
abi ContractA {
    fn method(contract_b_id: ContractId);
}

impl ContractA for Contract {
    fn method(contract_b_id: ContractId) {
       let contract_b = abi(ContractB, contract_b_id.into());
       contract_b.method();
    }
}
```

2. Execute the **ContractA** method in **TS**:
```ts
const contractAAbi = ContractAAbi__factory.connect(contractA, wallet);
const result = await contractAAbi.functions.method({
    value: contractB
}).txParams({
    gasPrice: 1,
    gasLimit: 1_000_000,
}).call();
```

### Expected Behavior

The error from `require` in contract:
```shell
return new RequireRevertError(receipt, reason);
             ^
RequireRevertError: The script reverted with reason RequireFailed
```

### Actual Behavior

It triggers the error from `@fuel-ts/abi-coder` instead of triggering the error from the contract.
```bash
throw new FuelError(ErrorCode.ELEMENT_NOT_FOUND, 'Element not found in the array.');
          ^
_FuelError: Element not found in the array.
```

On my case the application that is calling the contract only has knowledge of the contract A and not contract B.

luisburigo:
Repository to reproduce the error: [https://github.com/luisburigo/fuel-ts-issue.git](https://github.com/luisburigo/fuel-ts-issue.git)

luisburigo:
I debugged the code and found the error in this function:
https://github.com/FuelLabs/fuels-ts/blob/116d642e553784159f29c72989a24cd50aa5c73b/packages/account/src/providers/transaction-response/getDecodedLogs.ts#L8-L22

petertonysmith94:
Hey @luisburigo, hope you are well.

I have been playing around with your example (thank you btw), not 100% on the root cause just yet. I did find a potential workaround (hopefully it can unblock you). 

```sway
abi ContractA {
  fn method(contract_b_id: ContractId);
}

[...]

PR #2000, opened 2024-04-02
chore: helper functions for common JsonAbi features

### LINKED ISSUE #1556: Unhelpful and unorganized error messages

petertonysmith94:
Closes #1556 

---

- Added helper functions for finding common JsonAbi features.
  - `findFunctionByName`
  - `findNonEmptyInputs`
  - `findTypeById`
- Removed `findOrThrow` utility function.

### DIFF
--- a/.changeset/sour-books-flow.md
+++ b/.changeset/sour-books-flow.md
@@ -1,0 +1,5 @@
+---
+"create-fuels": minor
+---
+
+fix: user resolvable errors no longer cause the cli to exit
--- a/packages/create-fuels/src/cli.ts
+++ b/packages/create-fuels/src/cli.ts
@@ -33,27 +33,33 @@ const processWorkspaceToml = (fileContents: string, programsToInclude: ProgramsT
 };
 
 async function promptForProjectPath() {
-  const res = await prompts({
-    type: 'text',
-    name: 'projectName',
-    message: 'What is the name of your project?',
-    initial: 'my-fuel-project',
-  });
+  const res = await prompts(
+    {
+      type: 'text',
+      name: 'projectName',
+      message: 'What is the name of your project?',
+      initial: 'my-fuel-project',
+    },
+    { onCancel: () => process.exit(0) }
+  );
 
   return res.projectName as string;
 }
 
 async function promptForPackageManager() {
-  const packageManagerInput = await prompts({
-    type: 'select',
-    name: 'packageManager',
-    message: 'Select a package manager',
-    choices: [
-      { title: 'pnpm', value: 'pnpm' },
-      { title: 'npm', value: 'npm' },
-    ],
-    initial: 0,
-  });
+  const packageManagerInput = await prompts(
+    {
+      type: 'select',
+      name: 'packageManager',
+      message: 'Select a package manager',
+      choices: [
+        { title: 'pnpm', value: 'pnpm' },
+        { title: 'npm', value: 'npm' },
+      ],
+      initial: 0,
+    },
+    { onCancel: () => process.exit(0) }
+  );
   return packageManagerInput.packageManager as string;
 }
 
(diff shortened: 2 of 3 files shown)


packages/abi-coder/src/Interface.ts
```diff
@@ -83,40 +83,34 @@ export class Interface<TAbi extends JsonAbi = JsonAbi> {
...
 
   decodeLog(data: BytesLike, logId: number): any {
-    const { loggedType } = findOrThrow(this.jsonAbi.loggedTypes, (type) => type.logId === logId);
+    const loggedType = this.jsonAbi.loggedTypes.find((type) => type.logId === logId);
+    if (!loggedType) {
+      throw new FuelError(
+        ErrorCode.LOG_TYPE_NOT_FOUND,
+        `Log type with logId '${logId}' doesn't exist in the ABI.`
+      );
+    }
```
arboleya, line 92: This is something I mentioned in the last sync.

When dealing with multi-contracts ([see this](https://github.com/FuelLabs/fuels-ts/issues/1782)), users might call `ConractA`, which calls `ContractB`. Although `ContracA` will require an ABI, `ContractB` will not. So, if `ContractB` has some `log()` calls, those may bubble up to `ContractA`, which won't be able to decode them due to the lack of an ABI definition, and then we throw. The issue is that the transaction may have been successful, in which case throwing can be confusing.

Another way to handle this could be to print a warning instead of throwing an error when trying to decode `ContractB` logs as they come via `ContractA`. This approach could acknowledge the possible successful transaction while alerting the user to the decoding issue. It's worth considering but requires more in-depth thought to understand if it can cause other problems.

Feel free to leave this to another issue/PR.

---

@Torres-ssf Not sure how exactly this connects to the things you did in your PR. See if there's something else you'd want to add here.

petertonysmith94: Do we know that the contract is external as the `id` of the [receipt](https://github.com/FuelLabs/fuels-ts/blob/master/packages/transactions/src/coders/receipt.ts) is non-zero.

We could disable throwing in these scenarios as this level, or a try/catch in the `getDecodedLogs` function.

Will leave this for a follow up issue.

petertonysmith94: > It's worth considering but requires more in-depth thought to understand if it can cause other problems.

Added a follow up [issue](https://github.com/FuelLabs/fuels-ts/issues/2004) for this 😄

packages/abi-coder/src/utils/json-abi.ts
```diff
@@ -0,0 +1,30 @@
+import { ErrorCode, FuelError } from '@fuel-ts/errors';
+
+import type { JsonAbi, JsonAbiArgument, JsonAbiType } from '../types/JsonAbi';
+
+export const findFunctionByName = (abi: JsonAbi, name: string) => {
```
danielbate, line 5: I think we should add a return type here.

packages/abi-coder/src/utils/json-abi.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (!fn) {
+    throw new FuelError(
+      ErrorCode.FUNCTION_NOT_FOUND,
+      `Function with name '${name}' doesn't exist in the ABI`
+    );
+  }
+  return fn;
+};
+
+export const findTypeById = (abi: JsonAbi, typeId: number): JsonAbiType => {
```
danielbate, line 16: Doc blocks would be good for each of these.

danielbate (approved):

packages/abi-coder/src/encoding/strategies/getCoderV1.ts
```diff
@@ -114,7 +113,15 @@ export const getCoder: GetCoderFn = (
...
-    const arg = findOrThrow(components, (c) => c.name === 'buf').originalTypeArguments?.[0];
+    const bufferComponent = components.find((c) => c.name === 'buf');
+    if (!bufferComponent) {
+      throw new FuelError(
+        ErrorCode.INVALID_COMPONENT,
+        `The provided Vec type is missing the 'buf' component.`
+      );
+    }
+
+    const arg = bufferComponent.originalTypeArguments?.[0];
```
maschad, line 124: May I'm missing something but aren't these lines just a duplicate of [CoderV0](https://github.com/FuelLabs/fuels-ts/pull/2000/files#diff-0ee699e8c8af0ac5e0e4ed4059dc3a58071ec13a4a31934c92f29918a8a24ebcR113-R121) perhaps this should be shared function.

Torres-ssf (dismissed):

maschad (changes requested):
Looks good, just some final comments

danielbate (approved):

maschad (approved):
Great work!

PR #2002, opened 2024-04-02
chore: remove redundant error codes

### LINKED ISSUE #1556: Unhelpful and unorganized error messages

petertonysmith94:
Related to #1556 

---

- Removed redundant error codes.

### DIFF
--- a/packages/contract/src/contract-factory.ts
+++ b/packages/contract/src/contract-factory.ts
@@ -7,9 +7,9 @@ import { ErrorCode, FuelError } from '@fuel-ts/errors';
 import type { BytesLike } from '@fuel-ts/interfaces';
 import { Contract } from '@fuel-ts/program';
 import type { StorageSlot } from '@fuel-ts/transactions';
-import { arrayify } from '@fuel-ts/utils';
+import { arrayify, hexlify } from '@fuel-ts/utils';
 
-import { getContractId, getContractStorageRoot, hexlifyWithPrefix } from './util';
+import { getContractId, getContractStorageRoot } from './util';
 
 /**
  * Options for deploying a contract.
@@ -92,8 +92,8 @@ export default class ContractFactory {
   createTransactionRequest(deployContractOptions?: DeployContractOptions) {
     const storageSlots = deployContractOptions?.storageSlots
       ?.map(({ key, value }) => ({
-        key: hexlifyWithPrefix(key),
-        value: hexlifyWithPrefix(value),
+        key: hexlify(key),
+        value: hexlify(value),
       }))
       .sort(({ key: keyA }, { key: keyB }) => keyA.localeCompare(keyB));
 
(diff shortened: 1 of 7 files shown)


Torres-ssf (approved):

packages/contract/src/contract-factory.ts
```diff
@@ -92,8 +92,8 @@ export default class ContractFactory {
   createTransactionRequest(deployContractOptions?: DeployContractOptions) {
     const storageSlots = deployContractOptions?.storageSlots
       ?.map(({ key, value }) => ({
-        key: hexlifyWithPrefix(key, true),
-        value: hexlifyWithPrefix(value, true),
+        key: hexlifyWithPrefix(key),
+        value: hexlifyWithPrefix(value),
```
danielbate, line 96: Added by mistake? Although that whole function is redundant, we can just use `hexlify`.

danielbate, line 96: We could just call `hexlify` here?

packages/contract/src/util.ts
```diff
@@ -64,17 +63,8 @@ export const getContractId = (
...
-    return hexlify(value);
-  }
-
-  if (isKnownHex) {
-    return hexlify(`0x${value}`);
-  }
-
-  throw new FuelError(FuelError.CODES.UNEXPECTED_HEX_VALUE, `Value should be hex string ${value}.`);
-};
+export const hexlifyWithPrefix = (value: string) =>
```
danielbate, line 69: IMO this function is pretty redundant. Hexlifying a hex value will produce the same value. Defo an edge case and unlike but we risk this passing with `0xNotAHexVal`. May as well just call `hexlify`.

danielbate (changes requested):
Just one thought about `hexlify` usage, rest looks good

packages/contract/src/contract-factory.ts
petertonysmith94: Took a lil time to think over this one and I do agree.

https://github.com/FuelLabs/fuels-ts/pull/2002/commits/e73dc54934d0e14b4172a4cfa60129d23258c407

danielbate (approved):

packages/contract/src/util.ts
petertonysmith94: Reverted in 23995888b8e82a863f03e60f4fd8896b7a2ab711.

As storage slots have key, values without the `0x` prefix, it [fails](https://github.com/FuelLabs/fuels-ts/actions/runs/8542109255/job/23403007707?pr=2002#step:3:1011).

danielbate (approved):

Dhaiwat10 (approved):

PR #2008, opened 2024-04-03
feat: add `isReadOnly` helper for functions

### LINKED ISSUE #1931: Add a helper that helps users identify if a given contract function is read-only or not
Currently, there's no way for users to programatically identify if a given function is read-only or not directly from the SDK.

This feature was requested after a user gave feedback about them having to pay gas even for a read-only function call on the Sway Playground. The Sway Playground dynamically generates the UI for the contract written by the user and always `.call()`s the functions because there is no way to tell which function is read-only and which one isn't. 

If we had a helper that could identify if a given function is read-only, then in this case the Sway Playground can `.get()` those functions instead of `.call()`ing them.

cc @sdankel

Dhaiwat10:
## Summary

This PR adds a new helper called `isReadOnly` to the `InvokeFunction` type that helps users identify if a given function is read-only or not.

## Usage
```ts
  const isReadOnly = contract.functions.get_count.isReadOnly();

  if (isReadOnly) {
    await contract.functions.get_count().get();
  } else {
    await contract.functions.get_count().call();
  }
 ```

Closes #1931

### CODE REVIEW

File: apps/docs-snippets/src/guide/contracts/is-function-readonly.test.ts [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * @group node
+ */
+test('isReadOnly returns true for read-only functions', async () => {
+  const contract = await createAndDeployContractFromProject(DocSnippetProjectsEnum.COUNTER);
+
+  // #region is-function-readonly-1
+  const isReadOnly = contract.functions.get_count().isReadOnly();
+
+  expect(isReadOnly).toBe(true);
```
danielbate, line 15: We should keep assertions out of snippets to not add any noise to the snippet.

apps/docs-snippets/src/guide/contracts/is-function-readonly.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+  // #endregion is-function-readonly-1
+});
+
+test('isReadOnly returns false for functions containing write operations', async () => {
+  const contract = await createAndDeployContractFromProject(DocSnippetProjectsEnum.COUNTER);
+
+  const isReadOnly = contract.functions.increment_count(bn(1)).isReadOnly();
+
+  expect(isReadOnly).toBe(false);
```
danielbate, line 30: Was this test added for the coverage? We should have both a true and false test case in `fuel-gauge` and keep `docs-snippets` just for docs.

apps/docs/src/guide/contracts/methods.md
```diff
@@ -29,3 +29,13 @@ The `call` method should be used to submit a real contract call transaction to t
...
 
 <<< @/../../docs-snippets/src/guide/contracts/interacting-with-contracts.test.ts#interacting-with-contracts-4{ts:line-numbers}
+
+## `isReadOnly` (utility)
+
+If you want to figure out whether a function is read-only, you can use the `isReadOnly` method:
+
+<<< @/../../docs-snippets/src/guide/contracts/is-function-readonly.test.ts#is-function-readonly-1{ts:line-numbers}
+
+Usually, if the function is read-only, you can use the `get` method to read the data from the blockchain without using resources.
```
danielbate, line 39: Why only usually, is this not always applicable?

Dhaiwat10: I am not *sure* about this. There could be a case where you might want to use your _real_ resources as inputs even in a read-only function.

packages/program/src/contract.ts
```diff
@@ -89,7 +89,17 @@ export default class Contract implements AbstractContract {
...
+      const funcInvocationScopeCreator = (...args: Array<unknown>) =>
+        new FunctionInvocationScope(this, func, args);
+
+      Object.defineProperty(funcInvocationScopeCreator, 'isReadOnly', {
+        value: () => func.isReadOnly(),
+        writable: false,
+      });
+
+      return funcInvocationScopeCreator;
+    })() as InvokeFunction;
```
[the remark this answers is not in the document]
Dhaiwat10: Interesting point. I'll explore this possibility

packages/fuel-gauge/src/is-function-readonly.test.ts
```diff
@@ -0,0 +1,22 @@
...
+  expect(isReadOnly).toBe(true);
+});
+
+test('isReadOnly returns false for a function containing write operations', async () => {
+  const contract = await getSetupContract(FuelGaugeProjectsEnum.STORAGE_TEST_CONTRACT)();
+
+  const isReadOnly = contract.functions.increment_counter.isReadOnly();
+
+  expect(isReadOnly).toBe(false);
+});
```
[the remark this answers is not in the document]
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/2008/commits/44408be732cd32fa82cf2515a80ebe92cbb69330

packages/program/src/contract.ts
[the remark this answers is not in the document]
Dhaiwat10: @petertonysmith94 so we _could_ do that but again, the user would have to pass in arguments to the function like so in order do access that field:

```ts
contract.functions.increment_counter(bn(1)).isReadOnly
```

We don't want that.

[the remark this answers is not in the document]
arboleya: Yes, it's weird to call the fn with args in advance to discover if it's `readOnly.`

Would it make sense to use another `functionsAttrs` namespace or something similar?

```ts
contract.functionsAttrs.increment_counter; // write, read, etc.
```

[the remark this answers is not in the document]
Dhaiwat10: @arboleya this seems more intuitive than what you proposed: (which is the current solution in this PR)

```ts
contract.functions.increment_counter.isReadOnly()
```

Not a fan of adding another namespace as that's one more thing people would have to remember/learn.

Torres-ssf (approved):

apps/docs/src/guide/contracts/methods.md
danielbate:
```suggestion
If the function is read-only, you can use the `get` method to read the data from the blockchain without using resources.
```

Oh yeah agreed, they could be added in the tx builder but just not spent. I think we can add certainty here and remove unusally.

packages/abi-coder/src/FunctionFragment.ts
```diff
@@ -210,4 +210,14 @@ export class FunctionFragment<
 
     return coder.decode(bytes, 0) as [DecodedValue | undefined, number];
   }
+
+  /**
+   * Checks if the function is read-only i.e. it only reads from storage, does not write to it.
+   *
+   * @returns True if the function is read-only, false otherwise.
+   */
+  isReadOnly() {
```
danielbate, line 219:
```suggestion
  isReadOnly(): boolean {
```

Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/2008/commits/4d5646b5a5cbdd60f06074a7166e25c15b77ab05

packages/abi-coder/src/FunctionFragment.ts
```diff
@@ -221,4 +221,14 @@ export class FunctionFragment<
...
   }
+
+  /**
+   * Checks if the function is read-only i.e. it only reads from storage, does not write to it.
+   *
+   * @returns True if the function is read-only, false otherwise.
+   */
+  isReadOnly(): boolean {
+    const storageAttribute = this.attributes.find((attr) => attr.name === 'storage');
+    return !storageAttribute?.arguments.includes('write');
```
maschad, line 232: If `arguments` is _undefined_, this will cause a runtime error, is that an unrealistic edge case?

packages/program/src/contract.ts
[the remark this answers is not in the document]
arboleya: Are you injecting function within a function scope?

```ts
// like this
function outer () {}
outer.inner = function () {}
```

packages/abi-coder/src/FunctionFragment.ts
maschad: That's correct, but if the `storageAttribute` is `undefined` and there is no storage, then should it return `true` or should it throw an error that there is no `storageAttribute` ?

packages/program/src/contract.ts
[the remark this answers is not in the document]
Dhaiwat10: @arboleya I am actually not sure whether to say yes or no 😅 I just added a 'property' called isReadOnly to the function itself. So when the function itself is called, it runs whatever code it is supposed to, but function.isReadOnly is another separate function. Does that make sense?

packages/abi-coder/src/FunctionFragment.ts
Dhaiwat10: @maschad that's a good point. Maybe in that case I should return `undefined`?

maschad: It depends, I am not as familiar with Sway yet, but from my understanding in pure and view functions in solidity would be read-only, so they wouldn't have a storage attribute, if the case is different in Sway, then I think throwing an error would actually be more appropriate otherwise you can return `true`

Dhaiwat10: @maschad https://github.com/FuelLabs/fuels-ts/pull/2008/commits/ec413f537cc0dad3de2be532d6466f3e25f873df

I've added a testcase for a pure function. `isReadOnly` will return `true` for a function that does not use storage at all, since the user does not need to `call()` that function. `get()` is enough.

I've also updated the docstring.

apps/docs/src/guide/contracts/methods.md
```diff
@@ -29,3 +29,13 @@ The `call` method should be used to submit a real contract call transaction to t
...
 
 <<< @/../../docs-snippets/src/guide/contracts/interacting-with-contracts.test.ts#interacting-with-contracts-4{ts:line-numbers}
+
+## `isReadOnly` (utility)
+
+If you want to figure out whether a function is read-only, you can use the `isReadOnly` method:
+
+<<< @/../../docs-snippets/src/guide/contracts/is-function-readonly.test.ts#is-function-readonly-1{ts:line-numbers}
+
+If the function is read-only or pure, you can use the `get` method to read the data from the blockchain without using resources.
```
maschad, line 39:
```suggestion
If the function is read-only, you can use the `get` method to retrieve on-chain data without spending gas.
````

File: apps/docs/src/guide/contracts/methods.md [hunk 9]
```diff
@@ -29,3 +29,13 @@ The `call` method should be used to submit a real contract call transaction to t
...
+
+If the function is not read-only, you can use the `call` method to submit a 'real' transaction to the blockchain which _will_ consume resources.
```
maschad, line 41:
```suggestion
If the function is not read-only you will have to use the `call` method to submit a transaction onchain which incurs  a gas fee.
```

packages/abi-coder/src/FunctionFragment.ts
maschad: Understood, I suppose a user would only want to know if it's read only or not.

maschad (changes requested):
Looks good, I've left some suggestions around wording in the docs.

danielbate (approved):

packages/program/src/contract.ts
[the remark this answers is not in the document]
Dhaiwat10: Resolving this conversation for now to merge the PR but feel free to re-open or re-discuss this if there is anything in the future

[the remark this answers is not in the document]
arboleya: No, this is fine. Thanks. 🙂

PR #2011, opened 2024-04-04
feat: hide unnecessary logs and add loading spinners for `create-fuels`

### LINKED ISSUE #2006: `npm create fuels` logs out too many things during installation
When you run `npm create fuels`, it logs out every single log message from the child processes it executes under the hood like `pnpm install`. This is unnecessary and makes the whole experience unpleasant. This should be abstracted away under some nice loader animations.

Dhaiwat10:
## Summary
This PR hides unnecessary logs from the 'install dependencies' step for the `create-fuels` CLI, and adds some nice loading spinners for the two steps that we have.

Users can opt into verbose logging via the `--verbose` flag.

https://github.com/FuelLabs/fuels-ts/assets/39617427/549e7b50-9bd8-406a-bc7e-3570952453d1

Closes #2006


arboleya:
@Dhaiwat10 I couldn't help but notice how you ran the tool in the video:

```
node packages/create-fuels/dist/bin.global.js
```

Is this still a thing, or why did you demo it like this?

Dhaiwat10:
@arboleya that is because I am using the locally built binaries to run the CLI for the demo, without releasing it anywhere

PR #2019, opened 2024-04-05
fix: remove unused dependency `fuels` from the `create-fuels` package

### LINKED ISSUE #2013: `npm create fuels`  'cold-boot' time is very slow
When someone runs `npm create fuels`, especially when they run it for the very first time, it takes a lot of time to install the package. It takes around 1-2 minutes on a very fast internet connection, and only then do the users see a prompt that they can interact with.

This is very slow, and not ideal. We should investigate if we can improve this experience.

Dhaiwat10:
## Summary
We had `fuels` listed in the dependencies for `create-fuels` even though we were not using it, and this was causing the initial install/execution of the `pnpm create fuels` command to be extremely slow. Now it's almost instant :D 

Closes #2013


maschad:
Good find, but we probably should integrate a tool like [depcheck](https://github.com/depcheck/depcheck) to see what other unused dependencies are polluting the repo.

arboleya (approved):

Issue #1457, opened 2023-11-27
Improve DX for `Predicate` class


arboleya:
It should follow the same DX we have on `Contract`, supporting things like:

 - `txParams`
 - `callParams`
 - `getTransactionRequest`
 - Etc.

This increases flexibility and ease of use, giving users complete control of everything.

Related:
 - https://github.com/FuelLabs/fuels-ts/issues/1451
 - https://github.com/FuelLabs/fuels-ts/issues/1458

arboleya:
Pasting more insights from the 8th item from here:
 - https://github.com/FuelLabs/fuels-ts/discussions/1545

> _There is no prestaging of txs, i.e. if I use predicate.transfer, I can't see the tx_id unless I construct things manually, consider a prestage like predicate.prepare.transfer so that we can see the final tx before we send. I know this is already being considered._

arboleya:
Postponed until we sort out the breaking change it will cause.

PR #2022, opened 2024-04-07
docs: add docs for pre-staging predicate transactions

Dhaiwat10:
Related to #1457

We received some feedback that there is currently no way to pre-stage a predicate transcation. Well we _do_ have a way (via the `createTransfer` method), but we hadn't documented it. This PR adds that piece of documentation.


maschad (approved):
Looks good!

Torres-ssf (approved):

PR #2033, opened 2024-04-09
chore: integrate syncpack for synchronizing dependencies

### LINKED ISSUE #2028: chore: ensure consistent dependency versions across packages
We have some [multiple packages requiring the same dependency with different the versions](https://github.com/FuelLabs/fuels-ts/blob/master/pnpm-lock.yaml#L7689-L7698) it would be useful to sync them using something like [syncpack](https://github.com/JamieMason/syncpack)

richardgreg:
closes #2028


maschad:
Hey @richardgreg thanks for taking this on! I'm going to put the PR in draft now and you can open it up for review once you are ready.

richardgreg:
Hi @maschad, could you please review my draft and tell me if I'm on the right track?

If it's okay I'll close the open PR and and rerun the syncpack command on a new up-to-date branch cause rebasing with the current branch will be a pain 😁

maschad:
> Hi @maschad, could you please review my draft and tell me if I'm on the right track?
> 
> If it's okay I'll close the open PR and and rerun the syncpack command on a new up-to-date branch cause rebasing with the current branch will be a pain 😁

So a few things: 

- Given the conflicts, it may be best to _cherry-pick_ your commits onto a new branch checked out from master.
- It may be best to get started by actually setting up the mechanism to resolve mismatches, than trying to resolve them locally -  so essentially we would have a command in the root `package.json` that would [list the mismatches](https://jamiemason.github.io/syncpack/command/list-mismatches/) and then run that command in CI, similar to the [depcheck](https://github.com/FuelLabs/fuels-ts/blob/master/.github/workflows/pr-lint.yaml#L28) command. Then we can move from there
- when I briefly went through the [getting started guide](https://jamiemason.github.io/syncpack/guide/getting-started/) the recommendation was to _start small_ with the production dependencies, so you may want to update your config.

maschad:
> All sounds reasonable. I have the following question:
> 
> * After the CI lists mismatches and fixes them, it should also commit the changes, right?

That would be ideal but I would say just to have CI failing first with a intelligible error thrown would be a great first step.

Issue #1999, opened 2024-04-02
Fix wrong fee estimation for TX on multicall


Torres-ssf:
Incoming from this forum post:

https://forum.fuel.network/t/insufficientfeeamount-error/4985

PR #2021, opened 2024-04-06
chore!: upgrade `fuel-core` to `0.24.3`

Torres-ssf:
- Closes: https://github.com/FuelLabs/fuels-ts/issues/1831
- Closes: https://github.com/FuelLabs/fuels-ts/issues/1700
- Closes: https://github.com/FuelLabs/fuels-ts/issues/1738

## Breaking Changes
- Requires `Forc` `0.51.1`
- Setting `maxFee` is always required now ( The SDK estimates it by default with `Provider.getTransactionCost` + `Account.fund` )
- `gasPrice` is calculated by the VM so we do not need to inform it anymore. The Policy `gasPrice` [was replaced](https://github.com/FuelLabs/fuels-ts/pull/1764/commits/3ad83744048ce4b4a28a6151048adcdf27aa9566) by `tip`
- `Account.fund` signature has [changed](https://github.com/FuelLabs/fuels-ts/pull/1764/commits/0851940f91620696fe4b481aac63fc3b4ad340ab)
- Graphql URL now includes a versioning path: `http://127.0.0.1:4000/v1/graphql` 
- `calculateTransactionFee` now requires [tip](https://github.com/FuelLabs/fuels-ts/pull/1764/commits/277ead08b124651224b4b8b1934c701b59471f3e), [maxGasPerTx](https://github.com/FuelLabs/fuels-ts/pull/1764/commits/eeb5b1333006253f8fbbb48509b62aca0c798f39), and [gasPrice](https://github.com/FuelLabs/fuels-ts/pull/1764/commits/499d78b887b8779fbd6913ecca710cc5fad3f83c)
- `calculateTransactionFee` does not receive [gasUsed](https://github.com/FuelLabs/fuels-ts/pull/1764/commits/499d78b887b8779fbd6913ecca710cc5fad3f83c) anymore
- [removed](https://github.com/FuelLabs/fuels-ts/pull/1764/commits/87bb4411ef1e1ec767c9fe811ab515240b854182) unneeded properties returned by `Provider.getTransactionCost`

Current Tests Passing:
Screenshot 2024-04-08 at 10 38 54

PR #2034, opened 2024-04-10
fix!: fee estimation for multicall

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/1999


fuel-service-user:
✨ A PR has been created under the `rc-2034` tag on `fuels-wallet` repo.
https://github.com/FuelLabs/fuels-wallet/pull/1221

packages/account/src/providers/provider.ts
```diff
@@ -775,6 +774,46 @@ export default class Provider {
...
   }
 
+  /**
+   * Estimates the transaction gas and fee based on the provided transaction request.
+   * @param transactionRequest - The transaction request object.
+   * @param optimizeGas - Optional. Specifies whether to optimize the gas. Default is false.
+   * @returns An object containing the estimated minimum gas, minimum fee, maximum gas, and maximum fee.
+   */
+  estimateTxGasAndFee(params: { transactionRequest: TransactionRequest; optimizeGas?: boolean }) {
+    const { transactionRequest } = params;
```
arboleya, line 784: The parameter `optimizeGas` it not in use.

```suggestion
  /**
   * Estimates the transaction gas and fee based on the provided transaction request.
   * @param transactionRequest - The transaction request object.
   * @returns An object containing the estimated minimum gas, minimum fee, maximum gas, and maximum fee.
   */
  estimateTxGasAndFee(transactionRequest: TransactionRequest) {
```

arboleya (changes requested):

packages/account/src/providers/provider.test.ts
```diff
@@ -1007,12 +1007,11 @@ describe('Provider', () => {
       .spyOn(gasMod, 'calculatePriceWithFactor')
       .mockReturnValue(bn(0));
 
-    const { minFee, maxFee, usedFee } = await provider.getTransactionCost(request);
+    const { minFee, maxFee } = await provider.getTransactionCost(request);
 
-    expect(calculatePriceWithFactorMock).toHaveBeenCalledTimes(3);
+    expect(calculatePriceWithFactorMock).toHaveBeenCalledTimes(4);
```
maschad, line 1012: I'm trying to understand what is the purpose of this assertion, why is the calculatePriceWithFactor is this being called 4 times? As opposed to 5, 6 or 7 times?

packages/account/src/providers/provider.ts
```diff
@@ -775,6 +774,46 @@ export default class Provider {
     };
   }
 
+  /**
+   * Estimates the transaction gas and fee based on the provided transaction request.
+   * @param transactionRequest - The transaction request object.
+   * @param optimizeGas - Optional. Specifies whether to optimize the gas. Default is false.
+   * @returns An object containing the estimated minimum gas, minimum fee, maximum gas, and maximum fee.
+   */
+  estimateTxGasAndFee(params: { transactionRequest: TransactionRequest; optimizeGas?: boolean }) {
```
maschad, line 783: Should we have a test for this function?

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -615,12 +609,12 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
       const assetInput = findAssetInput(assetId);
 
       if (assetInput && 'assetId' in assetInput) {
-        assetInput.id = generateId();
+        assetInput.id = hexlify(randomBytes(33));
```
maschad, line 612: What's the significance of `33` here? Seems like a magic number

danielbate: Type of `32`? 🤔

maschad: If it is, I think that further consolidates the notion that we should move away from [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming) in our code as much as possible.

Torres-ssf: UTXOs IDs have 33 bytes:
- The first 32 bytes (`b256`) represent the ID of the transaction that created the UTXO.
- The last part, 1 byte (`u8`), indicates the output index where the corresponding [OutputCoin](https://specs.fuel.network/master/tx-format/output.html#outputcoin) was located within the Outputs of the originating Transaction. ( on VM version 0.48.0, this [changed](https://github.com/FuelLabs/fuels-ts/pull/2021/commits/a94624ddae4e8c74527fb4d723e3c554c5b7627f) to 2 bytes (`u16`) )

I can create a constant with this value to be used instead of the number.

packages/account/src/providers/transaction-request/input.ts
```diff
@@ -97,7 +97,7 @@ export const inputify = (value: TransactionRequestInput): Input => {
       return {
         type: InputType.Coin,
         txID: hexlify(arrayify(value.id).slice(0, 32)),
-        outputIndex: arrayify(value.id)[32],
+        outputIndex: toNumber(arrayify(value.id).slice(32, 33)),
```
maschad, line 100: Could you also convert these to named constants as well?

packages/account/src/providers/transaction-request/transaction-request.ts
maschad: Thanks for clarifying. Yes I agree would be good to name those constants.

maschad (changes requested):

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -13,6 +13,7 @@ import {
   TransactionType,
 } from '@fuel-ts/transactions';
 import { concat, hexlify } from '@fuel-ts/utils';
+import { randomBytes } from 'ethers';
```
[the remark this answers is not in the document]
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2034/commits/27f69783a9b1e8ebc718a2e4e9fb614457ebce18

packages/account/src/providers/provider.test.ts
Torres-ssf: I've just updated the existing test case to match the code after it was updated.

Previously, it was being called [3 times](https://github.com/FuelLabs/fuels-ts/blob/155b6f2fe28e988b277dac231af6d6a0cff1df0c/packages/account/src/providers/provider.ts#L907-L914). One of which was to calculate the prop `usedFee`, which it is not needed and I've removed it.

Now it is being called 4 times:

- 2 times when we are [estimating](https://github.com/FuelLabs/fuels-ts/blob/cde750c0402e2dfae9440625edf6fa94750d0cf2/packages/account/src/providers/provider.ts#L792-L818) the value for the estimation dry-run.
- 2 times [again](https://github.com/FuelLabs/fuels-ts/blob/cde750c0402e2dfae9440625edf6fa94750d0cf2/packages/account/src/providers/provider.ts#L956-L958) after the dry-run and having the `gasUsed` by the TX and using it as `gasLimit`, to have more accurate values.

packages/account/src/providers/transaction-request/input.ts
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2034/commits/7fc0e27d3f5bc103001248f7f2dbc8211c16f00c

packages/account/src/providers/transaction-request/transaction-request.ts
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2034/commits/7fc0e27d3f5bc103001248f7f2dbc8211c16f00c

packages/account/src/providers/provider.ts
arboleya: @Torres-ssf I see you kept the `params` object instead, and that's fine, so I'm resolving this.

packages/account/src/providers/provider.ts
arboleya: Hasn't this already been addressed by testing the `getTransactionCost` method? 🤔

arboleya (dismissed):

packages/account/src/providers/provider.ts
Torres-ssf: @arboleya indeed. This method is being executed by almost all `e2e` tests. 

We also already have unit tests for the methods that it is using: https://github.com/FuelLabs/fuels-ts/blob/master/packages/account/src/providers/utils/gas.ts

packages/account/src/providers/provider.test.ts
maschad: Understood, thanks for explaining. I still think that intuition is not conveyed through the test, could you document it in the test through some comments [similar to this](https://github.com/FuelLabs/fuels-ts/blob/cde750c0402e2dfae9440625edf6fa94750d0cf2/packages/account/src/providers/provider.ts#L801-L810)

The precedent is that we should be able to determine what the code is doing based on the tests, otherwise the comments.

maschad (changes requested):

packages/account/src/providers/provider.test.ts
Torres-ssf: I've updated the test assert as the amount of times that the method is called does not have importance on the testing case: https://github.com/FuelLabs/fuels-ts/pull/2034/commits/bb26747bbf06f54c488dfcf998d7411a5c2dce82

arboleya (approved):

maschad (approved):
Great job!!

Issue #2029, opened 2024-04-09
Transaction hash differs from the fuel-tx/vm


luizstacio:
### What version of fuels-ts are you using?

0.79.0

### Steps to Reproduce

Create a transaction that includes a predicate that uses a tx_id() to validate the TX, to deploy a contract.

### Expected Behavior

Hash transaction should generate the same hash of fuel-tx/vm. By zeroing witnessIndex on all predicted inputs and also removing the witness data that is removed from the link.

### Actual Behavior

When adding input coins in a transaction the SDK populates witnesses and witnessIndex. And once we call `populateTransactionPredicateData`, the input coin turns into a predicate by receiving the predicate bytecode. And the `witnessIndex` is not zeroed and `witnesses` is more removed.

- `witnessIndex` is not zeroed. The `hashTransaction` includes the value of `witnessIndex` into hash. This is different the expected behavior that is zeroed  `witnessIndex` for predicate inputs.
- `witnesses` is never removed creating junk bytes on the transaction by keeping a [zeroed witness on the tx](https://app.fuel.network/tx/0x6547366c297fcb384c40d507d736c66ccba4863bd26f1f9b006ee24083dfef47/advanced) increasing the price of the TX.``

Current workaround for fixing the hashing of the tx: https://github.com/FuelLabs/fuel-connectors/pull/46/files

PR #2037, opened 2024-04-11
fix!: avoid add witnesses for predicates

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/2029

## Breaking Changes:

- `TransactionRequest.addPredicateResources` was [removed](https://github.com/FuelLabs/fuels-ts/commit/3566c2ab9f9f79e32571716dc1164207b40c260f)


fuel-service-user:
✨ A PR has been created under the `rc-2037` tag on `fuels-wallet` repo.
https://github.com/FuelLabs/fuels-wallet/pull/1224

packages/account/src/predicate/predicate.ts
```diff
@@ -93,6 +96,16 @@ export class Predicate<TInputData extends InputValue[]> extends Account {
...
+
+        /**
+         * Not sure if this is needed. The only scenario that this could help is if
+         * the user has fetched the predicate resources using the provider `getResourcesToSpend` method,
+         * as the method will not populate the predicate property and the resources are going to be added
+         * as common resources. We should enforce users to always fetch resources using the predicate instance,
+         * as they will need it to populate the predicate data anyway using this method.
+         */
+        // eslint-disable-next-line no-param-reassign
+        input.witnessIndex = 0;
```
maschad, line 108: Should we throw an Error then, suggesting something along the lines of the `witnessIndex` is missing - please ensure you are fetching resources using the predicate instance?

packages/account/src/predicate/predicate.ts
```diff
@@ -255,4 +291,37 @@ export class Predicate<TInputData extends InputValue[]> extends Account {
 
     return mutatedBytes;
   }
+
+  private getIndexFromPlaceholderWitness(request: TransactionRequest): number {
```
danielbate, line 295: Not really sure how the index would be negative? Maybe a short doc on this function would be beneficial.

packages/account/src/providers/transaction-request/helpers.ts
```diff
@@ -0,0 +1,54 @@
...
+  isRequestInputCoin(input) || isRequestInputMessage(input);
+
+export const getRequestInputResourceOwner = (
+  input: CoinTransactionRequestInput | MessageTransactionRequestInput
+) => {
+  if (isRequestInputCoin(input)) {
+    return input.owner;
+  }
+  return input.recipient;
+};
```
danielbate, line 33:
```suggestion
export const getRequestInputResourceOwner = (
  input: CoinTransactionRequestInput | MessageTransactionRequestInput
) => (isRequestInputCoin(input) ? input.owner : input.recipient);
```

danielbate (dismissed):

danielbate (commented):
Sorry accidental approve, I have left comments.

packages/account/src/predicate/predicate.ts
Torres-ssf: @danielbate The index is negative when removing an entry from the `TransactionRequest.witnesses` is unnecessary. I will add a description

Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2037/commits/427dbd655c5d066955d5142e3965e6b96b135036

maschad (dismissed):
Great work!

packages/account/src/providers/transaction-request/helpers.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const isRequestInputMessage = (
+  input: TransactionRequestInput
+): input is MessageTransactionRequestInput => input.type === InputType.Message;
+
+export const isRequestInputResource = (
+  input: TransactionRequestInput
+): input is CoinTransactionRequestInput | MessageTransactionRequestInput =>
+  isRequestInputCoin(input) || isRequestInputMessage(input);
+
```
nedsalk, line 25:
```suggestion
export const isRequestInputResource = (
  input: TransactionRequestInput
): input is CoinTransactionRequestInput | MessageTransactionRequestInput =>
  input.type === InputType.Coin || input.type === InputType.Message;
```

There is no need for the two functions because they aren't used except internally in this file and they're too simple to necessitate an abstraction. This suggestion, if merged as-is, would break `helpers.test.ts`.

nedsalk (approved):

packages/fuel-gauge/src/predicate/predicate-populate-witness.test.ts
```diff
@@ -0,0 +1,325 @@
...
+
+      const predicate1WrongResources = await provider.getResourcesToSpend(
+        predicate1.address,
+        quantity
+      );
+
+      transactionRequest.addResources(predicate1WrongResources); // will add a placeholder witness
+
+      // The request carries 1 placeholder witnesses that was added for the predicate resources
+      expect(transactionRequest.witnesses.length).toBe(1);
```
danielbate, line 80:
```suggestion
      expect(transactionRequest.witnesses.length).toBe(1);
      expect(transactionRequest.witnesses[-1]).toBeTruthy();
```

Just to check it's specifically the place holder withness.

packages/fuel-gauge/src/predicate/predicate-populate-witness.test.ts
```diff
@@ -0,0 +1,325 @@
...
+        quantity
+      );
+
+      transactionRequest.addResources([
+        ...predicate1WrongResources, // witnessIndex 0 but will add placeholder witness
+        ...resources1, // witnessIndex will be 1 and will need to be ajusted to 0
+      ]);
+
+      // The request carries 2 placeholder witnesses, 1 added for the predicate resources
+      expect(transactionRequest.witnesses.length).toBe(2);
```
danielbate, line 111: Again should we be checking positions rather than just length?

danielbate (commented):
Really good work. Just wondering if some of those assertions could be improved to actually check the indexes rather than just witnesses length? As that seems to be the real value in the placeholder witness.

packages/fuel-gauge/src/predicate/predicate-populate-witness.test.ts
Torres-ssf: @danielbate All witnesses entries will be zeroed placeholders as the wallets resources were not signed at this point.

apps/docs-snippets/src/guide/transactions/transaction-request.test.ts
```diff
@@ -173,7 +173,7 @@ describe('Transaction Request', () => {
     });
 
     const predicateCoins = await predicate.getResourcesToSpend([
-      { amount: 1000, assetId: baseAssetId },
+      { amount: 2000, assetId: baseAssetId },
```
maschad, line 176: This is just to reduce probability of a failure?

packages/account/src/test-utils/seedTestWallet.ts
```diff
@@ -5,20 +5,29 @@ import { coinQuantityfy, ScriptTransactionRequest } from '../providers';
...
+  quantities: CoinQuantityLike[],
+  utxosAmount = 1
+) => {
+  const toFundAccounts = Array.isArray(wallet) ? wallet : [wallet];
+
   const genesisWallet = new WalletUnlocked(
     process.env.GENESIS_SECRET || randomBytes(32),
-    wallet.provider
+    // Connect to the same Provider as wallet
+    toFundAccounts[0].provider
```
maschad, line 18: This is a bit less intuitive than `wallet.provider`
```suggestion
   // There may be multiple wallets, so want to use the same provider for them all
   const walletProvider = toFundAccounts[0].provider
 ```

File: packages/fuel-gauge/src/policies.test.ts [hunk 9]
```diff
@@ -63,7 +63,7 @@ describe('Policies', () => {
 
     const txParams: CustomTxParams = {
       tip: 10,
-      maturity: randomNumber(1, 2),
+      maturity: 1,
```
maschad, line 66: Is this in relation to https://github.com/FuelLabs/fuels-ts/issues/2201 ?

maschad (commented):
Looking good, just left some comments inline

packages/account/src/test-utils/seedTestWallet.ts
Torres-ssf: @maschad do you think it is better now? https://github.com/FuelLabs/fuels-ts/pull/2037/commits/a657ee5ea356626a0312504239beeb419b5a8c01

packages/fuel-gauge/src/policies.test.ts
Torres-ssf: I made this change because `maturity` value should not be higher than the block height https://github.com/FuelLabs/fuels-ts/pull/2037/commits/596c6770ac00fefd34c981216d30d703d3461829

apps/docs-snippets/src/guide/transactions/transaction-request.test.ts
Torres-ssf: @maschad Yes

maschad (approved):
Great work!

packages/account/src/predicate/predicate.ts
```diff
@@ -250,4 +265,43 @@ export class Predicate<TInputData extends InputValue[]> extends Account {
...
+    const hasEmptyPredicateInputs = predicateInputs.find((input) => !input.predicate);
+
+    if (hasEmptyPredicateInputs) {
+      index = hasEmptyPredicateInputs.witnessIndex;
+
+      const allInputsAreEmpty = predicateInputs.every((input) => !input.predicate);
+
+      if (!allInputsAreEmpty) {
+        /**
+         * If at least one resource was added as predicate resource, we need to check if it was the
```
arboleya, line 291: Nit.

```suggestion
         * If at least one resource was added as a predicate resource, we need to check if it was the
```

packages/account/src/predicate/predicate.ts
```diff
@@ -250,4 +265,43 @@ export class Predicate<TInputData extends InputValue[]> extends Account {
...
+         * first one. If that is the case, we don't need to remove the witness placeholder
+         * as this was added with the "witnessIndex" as 0 and without a placeholder witness. Later if
+         * any resource without predicate is added, it will have the same witnessIndex because it has the
```
arboleya, line 294: Nit.

```suggestion
         * any resource without a predicate is added, it will have the same witnessIndex because it has the
```

arboleya (approved):
Good stuff, extensive test coverage! 💯 

Just left two minor optional nits.

arboleya (approved):

packages/fuel-gauge/src/e2e-script.test.ts
```diff
@@ -86,6 +86,7 @@ describe('Live Script Test', () => {
 
     const provider = await Provider.create(process.env.FUEL_TESTNET_NETWORK_URL);
     const wallet = new WalletUnlocked(process.env.TEST_WALLET_PVT_KEY, provider);
+    console.log(wallet.address);
```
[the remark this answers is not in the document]
Torres-ssf: 🤦‍♂️ https://github.com/FuelLabs/fuels-ts/pull/2037/commits/d5f64fd954a2c29f6d15664638611c9e57e69ca5

Torres-ssf:
I am holding this until the fix https://github.com/FuelLabs/fuels-ts/pull/2267 is released. 

I am unsure if the changes here will be reflected in more required updates on the PR that will upgrade https://github.com/FuelLabs/fuel-connectors to use the latest `fuels` version

danielbate (approved):
Nice one 🧙🏻 

One thing that I like that they do in the RS SDK, is they will add funding / set gas limits to the exact amount required by the tx. So rather than an increase from `1000` > `2000`, it would be increased to (ex. `1432`). Then if we have a fail due to an increase consumption, we can actually validate that it should have increased. Thoughts?

Torres-ssf:
> Nice one 🧙🏻
> 
> One thing that I like that they do in the RS SDK, is they will add funding / set gas limits to the exact amount required by the tx. So rather than an increase from `1000` > `2000`, it would be increased to (ex. `1432`). Then if we have a fail due to an increase in consumption, we can actually validate that it should have increased. Thoughts?

@danielbate I've failed to understand what you mean mate. Are you talking about adding tests to validate that a TX has consumed the amount of gas it was supposed to consume? And throw an error if they do consume a different value?

Issue #1890, opened 2024-03-19
Remove or reimplement `getNetwork`


danielbate:
We have the following function on our `Provider`:

https://github.com/FuelLabs/fuels-ts/blob/bc72cb67e78b7b92146c38ced9e99af00af5883b/packages/account/src/providers/provider.ts#L504-L511

It's purpose is to return a network configuration object, as defined and shaped by `ethers.js`. In our efforts to deprecate `ethers` (#1592, #1640, #1775 etc), we should look to either remove or reimplement this function.

IMO we already provide enough network info through `getNode` and `getChain`, and therefore I would like to deprecate `getNetwork`, removing the `ethers` dependency here. We could also keep `getNetwork` and have it perform the same function as the other 2 functions.

PR #2047, opened 2024-04-11
fix!: deprecate `getNetwork`

maschad:
Closes https://github.com/FuelLabs/fuels-ts/issues/1890

This PR seeks to deprecate the `getNetwork` method and also remove `ethers`(as apart of the larger effort discussed in https://github.com/FuelLabs/fuels-ts/discussions/1592#discussioncomment-8003515) 

`ethers` is now removed from:
- `account` 
- `crypto`

As a result  there are more `ethers` functions being internalized:
- `ripemd160` 
-  `pbkdf2`
- `hmac`
- `dataSlice`
- `encodeBase58`
- `decodeBase58`
- `getBytes`


packages/account/src/hdwallet/hdwallet.ts
```diff
@@ -221,7 +221,8 @@ class HDWallet {
   }
 
   static fromExtendedKey(extendedKey: string) {
-    const decoded = toBeHex(decodeBase58(extendedKey));
+    let decoded = toHex(decodeBase58(extendedKey));
+    decoded = `${decoded.substring(0, 2)}0${decoded.substring(2)}`;
```
maschad, line 225: This was a strange addition on my part needed to make the tests pass, essentially `toHex` omits a padded `0` at the third index, thus `0x0488b21e03587302...` becomes `0x488b21e03587302....` but I would think that  if `toHex` had a bug it would've been discovered much earlier... any thoughts?

danielbate: This is weird but I agree. `toHex` also just wraps `BN` functionality. Should we not be updating the test case rather than having this specific logic in place?

packages/crypto/src/node/pbkdf2.ts
```diff
@@ -0,0 +1,46 @@
...
+    iterations: number,
+    keylen: number,
+    algo: 'sha256' | 'sha512'
+  ) => BytesLike
+) => {
+  if (locked) {
+    throw new Error('pbkdf2 is locked');
+  }
+  pBkdf2 = func;
+};
```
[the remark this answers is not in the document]
maschad: They are more for dealing with async operations, which isn't really applicable to our use case atm, but they are used to handle concurrent asynchronous calls to pbkdf2Async with the same salt. They ensure that only one computation is performed at a time for a given salt, preventing duplicate computations and potential race conditions.

packages/crypto/test/hmac.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+// #TODO: This computes the wrong value on node yet the other tests pass, I'm not sure why
+/**
+ * @group browser
+ */
+describe('computeHmac node & browser', () => {
+  it('should compute HMAC correctly', () => {
+    const key = '0x0102030405060708090a0b0c0d0e0f10';
+    const data = '0x11121314151617181920212223242526';
+    const expectedSha256 = '0x5ce589e9b1295be92db7f5e08a5672e8e5f2e80163cc706a9584269f294fe3bd';
```
maschad, line 13: To my understanding these are the correct values

based on https://cryptotools.net/hmac

packages/crypto/test/pbkdf2.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    const algo = 'sha256';
+    const expectedResult = '0x90eceedd899d5cdcdfd9b315ad6e2c3391bf95cc131b6f0f016339db5ee60494';
+
+    const result = pbkdf2(passwordBuffer, saltBuffer, iterations, keylen, algo);
+
+    expect(result).toBe(expectedResult);
+  });
+});
+
+// #TODO: Even when I run the filter:browser, these tests still run
```
maschad, line 22: I'm skipping these are they are not critical tests but locally even when I run the browser filter these tests still run, could it be because they are in the same file?

packages/account/src/hdwallet/hdwallet.ts
maschad: @danielbate The [test cases look right](https://github.com/FuelLabs/fuels-ts/blob/master/packages/account/src/hdwallet/hdwallet.test.ts#L88-L129) as they are just key derivation

maschad:
@arboleya was dbd1d0d4cdf5b04e3caf80df66b0c24fedf9d536 supposed to resolve [this issue](https://github.com/FuelLabs/fuels-ts/actions/runs/8757354421/job/24035806488?pr=2047) ?

arboleya:
@maschad Hmm I don't think so; they are different workflows.
 - [.github/workflows/pr-release.yaml](https://github.com/FuelLabs/fuels-ts/blob/master/.github/workflows/pr-release.yaml) — the one I fixed
 - [.github/workflows/breaking-change-pr-release.yaml](https://github.com/FuelLabs/fuels-ts/blob/master/.github/workflows/breaking-change-pr-release.yaml) — the one you reported

maschad:
> @maschad Hmm I don't think so; they are different workflows.
> 
> * [.github/workflows/pr-release.yaml](https://github.com/FuelLabs/fuels-ts/blob/master/.github/workflows/pr-release.yaml?rgh-link-date=2024-04-19T18%3A57%3A27Z) — the one I fixed
> * [.github/workflows/breaking-change-pr-release.yaml](https://github.com/FuelLabs/fuels-ts/blob/master/.github/workflows/breaking-change-pr-release.yaml?rgh-link-date=2024-04-19T18%3A57%3A27Z) — the one you reported

My bad, thanks for pointing that out. @Torres-ssf could it be that the [auth credentials here](https://github.com/FuelLabs/fuels-ts/blob/master/.github/workflows/breaking-change-pr-release.yaml#L60-L62) have changed resulting in [this failure](https://github.com/FuelLabs/fuels-ts/actions/runs/8757354421/job/24035806488?pr=2047)? The [action config](https://github.com/FuelLabs/github-actions/blob/master/setups/npm/action.yaml) looks fine to me.

maschad:
Closing in favour of https://github.com/FuelLabs/fuels-ts/pull/2130

Issue #2035, opened 2024-04-10
ci: Introduce `dependabot` to manage dependency upgrades


maschad:
The primary benefit is to be up to date on potential vulnerabilities, secondly I think it can potentially help reduce the friction involved doing `fuel-core` and `forc` upgrades, granted I don't know much of what's involved in that process, but at the very least having all our dependencies up to date can rule out those possibilities.

More info: https://docs.github.com/en/code-security/dependabot/working-with-dependabot

PR #2062, opened 2024-04-15
chore: configuring dependabot

arboleya:
Closes https://github.com/FuelLabs/fuels-ts/issues/2035


maschad (approved):

maschad (changes requested):
We will probably need some configuration adjustments for the changesets

See https://github.com/FuelLabs/fuels-ts/pull/2063

arboleya:
@maschad Should we merge it like this and address the changeset in another PR?

Meanwhile, could we manually add changesets to the `dependabot` PRs?

danielbate (approved):
Approving as I agree with [this comment](https://github.com/FuelLabs/fuels-ts/pull/2062#issuecomment-2061323384).

Depenbabot PR will fail validation without a changeset also? So we shouldn't miss it.

maschad:
> Approving as I agree with [this comment](https://github.com/FuelLabs/fuels-ts/pull/2062#issuecomment-2061323384).
> 
> Depenbabot PR will fail validation without a changeset also? So we shouldn't miss it.

That's a good point, I was planning to address the issue here though to avoid the manual work in the future. But I will approve to unblock the work.

maschad (approved):

PR #2071, opened 2024-04-16
feat: generalize subscription event parsing

### LINKED ISSUE #2053: Error while parsing stream data response
Seems to have happened [here](https://github.com/FuelLabs/fuels-ts/issues/1755#issuecomment-2046080940), [here](https://github.com/FuelLabs/fuels-ts/issues/1755#issuecomment-2047932603) and [here](https://github.com/FuelLabs/fuels-ts/actions/runs/8687248195/job/23820387487#step:9:79).

nedsalk:
This PR updates the event stream parsing logic to properly parse events streamed in multiple chunks and multiple events streamed in one chunk (which was the issue reported in #2053).

closes #2053 and adds infrastructure necessary for #1644


packages/account/src/providers/fuel-graphql-subscriber.ts
```diff
@@ -34,53 +34,66 @@ export class FuelGraphqlSubscriber implements AsyncIterator<unknown> {
     this.stream = response.body!.getReader();
   }
 
+  private events: Array<{ data: unknown; errors: unknown }> = [];
```
maschad, line 37: Should `errors` be optional?

```suggestion
  private events: Array<{ data: unknown; errors?: unknown }> = [];
```

packages/account/src/providers/fuel-graphql-subscriber.ts
```diff
@@ -34,53 +34,66 @@ export class FuelGraphqlSubscriber implements AsyncIterator<unknown> {
...
   async next(): Promise<IteratorResult<unknown, unknown>> {
     if (!this.stream) {
       await this.setStream();
     }
 
     // eslint-disable-next-line no-constant-condition
     while (true) {
+      if (this.events.length > 0) {
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        const { data, errors } = this.events.shift()!;
```
maschad, line 49: if `errors` is optional then I think we can remove the no-null assertion
```suggestion
        const { data, errors } = this.events.shift();
```

nedsalk: This no-null assertion is related to [array.shift](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/shift) returning `T | undefined` because running `[].shift()` returns `undefined` so they had to cover that case in the type as well.

maschad (dismissed):
Great work!

arboleya (approved):

packages/account/src/providers/fuel-graphql-subscriber.ts
```diff
@@ -34,53 +34,66 @@ export class FuelGraphqlSubscriber implements AsyncIterator<unknown> {
     this.stream = response.body!.getReader();
   }
 
+  private events: Array<{ data: unknown; errors?: unknown }> = [];
```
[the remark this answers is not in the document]
nedsalk: Good catch! https://github.com/FuelLabs/fuels-ts/pull/2071/commits/3c8213f6347a2dc5de5911f1a937a838effcbce5

maschad (approved):

PR #2079, opened 2024-04-16
fix: `MemoryCache` flaky tests 

Torres-ssf:
This PR is intended to fix these annoying flaky tests:

https://github.com/FuelLabs/fuels-ts/actions/runs/8711150584/job/23894625257?pr=2037#step:10:512

### DIFF
--- a/packages/account/src/providers/memory-cache.test.ts
+++ b/packages/account/src/providers/memory-cache.test.ts
@@ -142,6 +142,12 @@ describe('Memory Cache', () => {
       setTimeout(resolve, 10);
     });
 
+    /*
+      MemoryCache uses a global cache with values from
+      several instances, all returned by `getActiveData()`.
+      However, we only want to check the ones from this
+      test, so we use `containSubset`.
+    */
     expect(memCache.getAllData()).containSubset(EXPECTED);
   });
 
--- a/packages/account/src/providers/memory-cache.test.ts
+++ b/packages/account/src/providers/memory-cache.test.ts
@@ -148,11 +148,11 @@ describe('Memory Cache', () => {
   it('should validate that MemoryCache uses a global cache', async () => {
     const oldValue = randomBytes(8);
 
-    const instance1 = new MemoryCache(3_000);
+    const instance1 = new MemoryCache(1000);
     instance1.set(oldValue);
 
     await new Promise((resolve) => {
-      setTimeout(resolve, 1_000);
+      setTimeout(resolve, 200);
     });
 
     const newValue = randomBytes(8);
--- a/packages/account/src/providers/memory-cache.test.ts
+++ b/packages/account/src/providers/memory-cache.test.ts
@@ -1,5 +1,6 @@
 import { randomBytes } from '@fuel-ts/crypto';
 import type { BytesLike } from '@fuel-ts/interfaces';
+import { hexlify } from '@fuel-ts/utils';
 
 import { MemoryCache } from './memory-cache';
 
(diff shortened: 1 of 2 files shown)


packages/account/src/providers/memory-cache.test.ts
```diff
@@ -53,81 +50,97 @@ describe('Memory Cache', () => {
...
+    const EXPECTED: BytesLike[] = [value1, value2, value3];
+
     const memCache = new MemoryCache(100);
 
-    expect(memCache.getActiveData()).toStrictEqual(EXPECTED);
+    memCache.set(value1);
+    memCache.set(value2);
+    memCache.set(value3);
+
+    expect(memCache.getActiveData()).containSubset(EXPECTED);
```
arboleya, line 124: Why `containSubset` instead of `toStrictEqual`?

Torres-ssf: @arboleya because it won't be strictly equal. The class `MemoryCache` uses a global cache and the test suite interacts with it multiple times.

arboleya: Considering that values 1-3 are `identical`, and let's say those three `.set()` calls occur between a 100ms interval, does it mean that `getActiveData()` will always return a single item?

Torres-ssf: I believe `getActiveData` returns all cached data that has not expired.

packages/account/src/providers/memory-cache.test.ts
```diff
@@ -1,11 +1,8 @@
 import { randomBytes } from '@fuel-ts/crypto';
 import type { BytesLike } from '@fuel-ts/interfaces';
-import { hexlify } from '@fuel-ts/utils';
 
 import { MemoryCache } from './memory-cache';
 
-const CACHE_ITEMS = [hexlify(randomBytes(8)), randomBytes(8), randomBytes(8)];
```
nedsalk, line 7: `CACHE_ITEMS[0]` is hexlified and was testing that `MemoryCache` works with string keys as well. Can you please add one more test that uses `hexlify`?

```ts
  it('can set (string)', () => {
    const ttl = 1000;
    const expiresAt = Date.now() + ttl;
    const memCache = new MemoryCache(ttl);
    const value = hexlify(randomBytes(8));

    expect(memCache.set(value)).toBeGreaterThanOrEqual(expiresAt);
  });
```

nedsalk (changes requested):

packages/account/src/providers/memory-cache.test.ts
arboleya: In this case, it should always be `[value1, value2, value3]`. For the `toStrictEqual()` assertion to fail, the interval between the instantiation and the test assertion needs to be greater than the cache TTL—in this case, it is 100ms.

If the above understanding is correct, the flakiness issue could be attributed to this; some tests executed in `<100ms` and others in `>=100`, and the latter would fail.

Having a greater `ttl` and keeping with `strictEqual()` assertions could work best because of the improved predictability. The `containSubset` creates some uncertainty; it is not obvious what the expected result is and why.

Let me know what you think.

```ts
/* 100% pseudo-code */

const value1 = randomBytes(8);
const value2 = randomBytes(8);
const value3 = randomBytes(8);

const allValues: BytesLike[] = [value1, value2, value];

const ttl = 500;
const timeStart = new Date().getTime();
const memCache = new MemoryCache(ttl);

memCache.set(value1);
memCache.set(value2);
memCache.set(value3);

const activeData = memCache.getActiveData();
const timeDiff = new Date().getTime() - timeStart;

expect(timeDiff).toBeLessThan(ttl); // test must finish before ttl
expect(activeData).toBeStrictEqual(allValues); // everything must be active
```

Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2079/commits/189bdb8d97f0a3301bfe182de5611abd25b68ae0

https://github.com/FuelLabs/fuels-ts/actions/runs/8720023700/job/23920681093?pr=2079#step:10:509

Like I said before, the `MemoryCache` class uses a [global cache](https://github.com/FuelLabs/fuels-ts/blob/124099b1a134607b8d12953d61ac7065c34fc7ed/packages/account/src/providers/memory-cache.ts#L11), the instance only controls how long the values set by it are going to live within the cache.

```ts
const old = randomValues(8)

const instance1 = new MemoryCache(2000)
instance1.set(old)

await new Promise((resolve) => {
  setTimeout(resolve, 100);
});

const new = randomValues(8)

const instance2 = new MemoryCache(100)
instance2.set(new)

instance2. getActiveData() // contains both [old, new]
```

`toBeStrictEqual` was the reason this test case was flaky in the first place.

There are other test cases setting values within the _global_ cache and by the time this specific test case uses `toStrictEqual` to validate only some specific values, the validation fails because there is still other active data within the cache.

arboleya: Got it; my bad.

Maybe a comment could make it more straightforward.

```ts
/*
  MemoryCache uses a global cache with values from
  several instances, all returned by `getActiveData()`.
  However, we only want to check the ones from this
  test, so we use `containSubset`.
*/
expect(memCache.getActiveData()).containSubset(EXPECTED);
```

Torres-ssf: Done https://github.com/FuelLabs/fuels-ts/pull/2079/commits/5dc5dcf4d24642365203fdb35bfa4b4c0be4102d

arboleya (approved):

packages/account/src/providers/memory-cache.test.ts
arboleya: 🚢 ?

PR #2106, opened 2024-04-18
chore: use `.env` variables in tests

danielbate:
If you want to run the live node tests locally, you'll need to override the wallet and network url. This PR provides a way of doing exactly that using a local `.env` file and updates the contributing guidelines for using them.


.env.example
```diff
@@ -0,0 +1,2 @@
+FUEL_NETWORK_URL=
+TEST_WALLET_PVT_KEY=
```
arboleya, line 2: Why has this become a necessity?

Should we document something about this in the [contributing](https://github.com/FuelLabs/fuels-ts/blob/db/chore/vitest-env-vars/CONTRIBUTING.md#testing) guide?

danielbate: It is not a necessity, but you'll see in the [live tests](https://github.com/FuelLabs/fuels-ts/blob/384c5b348bc89392f8f2c7fa750d4862750ce368/packages/fuel-gauge/src/e2e-script.test.ts#L82-L84), if these variables are not set, then the test is skipped. So unless someone has these set, we are only running them in CI. I would like the option to run them locally. Useful for when I wanted to test experimental work on a live node without pushing work up.

But adding some info in contibruting would be helpful.

Torres-ssf: I am wondering the same thing. Why do we need this change now?

danielbate: @Torres-ssf pls see my response [above](https://github.com/FuelLabs/fuels-ts/pull/2106#discussion_r1570857714), we commented at basically the same time so just checking you've seen it 😅

Torres-ssf: @danielbate Sorry but when I submitted mine, yours was not visible to me yet lol

danielbate: https://github.com/FuelLabs/fuels-ts/pull/2106/commits/9bd994ec76d0057727bf3576e1e1ba5bb80a0672

maschad (approved):

PR #2108, opened 2024-04-18
docs: pinned version of fuels for `beta-5` docs

### LINKED ISSUE #2107: Patch `beta-5` docs to suggest installation of specific version
Soon, we will lose compatibility with `beta-5`, so we should consider orchestrating a quick patch/update the docs on the hub for beta-5, instructing users to [install](https://docs.fuel.network/docs/fuels-ts/introduction/getting-started/#installation) a specific version instead of the latest - see [example](https://docs.fuel.network/docs/fuels-rs/getting-started/#importing-the-fuel-rust-sdk).

It'd look something like this:

```console
npm install fuels@0.81.0
```

And this:
```console
npm create fuels@0.81.0
```

petertonysmith94:
Closes #2107

---

- Fixed `fuels` documentation version to the current version (`{{fuels}}`), for beta-5 compatibility.

### DIFF
--- a/apps/docs/src/guide/fuels-cli/commands.md
+++ b/apps/docs/src/guide/fuels-cli/commands.md
@@ -55,7 +55,7 @@ In a nutshell:
 
 ## `fuels build`
 
-```console
+```console-vue
 npx fuels@{{fuels}} help build
 ```
 


arboleya (commented):
Isn't there any way we could use variables for this? 👀 

It could make it so much more reliable. 🤔

maschad (dismissed):

arboleya (changes requested):

apps/docs/src/getting-started.md
```diff
@@ -17,16 +17,18 @@ To begin, you need to add the `fuels` dependency to your project. You can do thi
...
 
-```sh [pnpm]
-pnpm add fuels
+```sh-vue [pnpm]
+pnpm add fuels@{{fuelsInstallVersion}}
 ```
 
 :::
 
+**Note**: We use `{{fuelsInstallVersion}}` to ensure compatibility with our latest testnet.
```
arboleya, line 30: Should we be more specific here?

```suggestion
**Note**: Use version `{{fuelsInstallVersion}}` to ensure compatibility with `beta-5` network—check the [docs](https://docs.fuel.network/guides/installation/#using-the-latest-toolchain).
```

arboleya (approved):
Good stuff! ✨

Thanks for getting this out of the way! 🫡

maschad (approved):
Nice work!

Torres-ssf (approved):

apps/docs/src/getting-started.md
```diff
@@ -17,16 +17,18 @@ To begin, you need to add the `fuels` dependency to your project. You can do thi
...
 
-```sh [npm]
-npm install fuels --save
+```sh-vue [npm]
+npm install fuels@{{fuelsInstallVersion}} --save
 ```
 
-```sh [pnpm]
-pnpm add fuels
+```sh-vue [pnpm]
```
Dhaiwat10, line 24: What is ```sh-vue ?

Dhaiwat10 (approved):

apps/docs/src/getting-started.md
arboleya: Is there any chance this can be misinterpreted on the docs hub? 🤔

danielbate (approved):

Torres-ssf (dismissed):

apps/docs/src/getting-started.md
arboleya: I know I said it first, but I don't know if there's any advantage in using a new variable then.

Perhaps we could use the version from `@fuel-ts/versions` and call it a day.

What do you think?

arboleya: Let's not discard the option of using the same variable, as it could solve the issue without relying on external things.

In this approach of using different variables, do we need to wait for https://github.com/FuelLabs/docs-hub/pull/247?

If that doesn't get merged in time, what would we do?

arboleya: @petertonysmith94 As we spoke, let's go without the second variable. 🙁

arboleya (approved):

nedsalk (approved):

danielbate (approved):

Issue #2070, opened 2024-04-16
Create infrastructure for using `typegen` in integration tests


nedsalk:
We're currently using `getForcProject` and aren't dogfooding `typegen` in our integration tests. This work can be split into two issues:
1. Create the necessary infrastructure
2. Apply it on the whole test base

This issue deals with the first point. 
The second point should be done after #1356 is merged because they might conflict with each other.

#284 won't fully benefit from this because contract binaries aren't put in the typegen-generated code, only script and predicate binaries are.

PR #2122, opened 2024-04-19
feat: support `()` type in typegen

nedsalk:
While implementing #2070 I came across an issue with `call-test-contract`'s [`foobar` function](https://github.com/FuelLabs/fuels-ts/blob/555d60a8b025cc79e5d91d660db5639126a5c3d2/packages/fuel-gauge/test/fixtures/forc-projects/call-test-contract/src/main.sw#L44) that takes in an empty `()` type. Typegen couldn't handle it and was throwing so I implemented the logic for supporting `()` types.
Now, empty inputs are ignored in `typegen`. This corresponds to how we're encoding empty arguments during runtime - we ignore them. I added [tests](https://github.com/FuelLabs/fuels-ts/pull/2122/files#diff-cbc1b15016d49087e1c4ec7b22076e14d9f7e6500665dbe59e8f835804e7723cR48) to verify this runtime behavior as well.


arboleya:
> Now, functions that take in an empty type _mandate_ passing `undefined` into them:
> 
> ```ts
> const fn = contract.functions.foobar(undefined); // InvokeFunction<[value: undefined], ...>
> ```

Oof. It's a tough one - I'm not sure I can handle it. 😅

nedsalk:
@arboleya I think it's not worth treating this as a special optional case because we'd then have to cover all of the below:

```rs
abi TestContract {
    fn foobar_0(value: ()) -> u8;
    fn foobar_1(value: (), v1: u8) -> u8;
    fn foobar_2(v1: u8, value: ()) -> u8;
    fn foobar_3(v1: u8, value: (), v2: u8) -> u8;
}
```

The generated contract functions would have to be

```ts
foobar_0: InvokeFunction<[value?: undefined], BN>;
foobar_1: InvokeFunction<[value: undefined, v1: BigNumberish], BN>;
foobar_2: InvokeFunction<[v1: BigNumberish, value?: undefined], BN>;
foobar_3: InvokeFunction<[v1: BigNumberish, value: undefined, v2: BigNumberish], BN>;
```

With this PR's solution the generated code looks like this:

```ts
foobar_0: InvokeFunction<[value: undefined], BN>;
foobar_1: InvokeFunction<[value: undefined, v1: BigNumberish], BN>;
foobar_2: InvokeFunction<[v1: BigNumberish, value: undefined], BN>;
foobar_3: InvokeFunction<[v1: BigNumberish, value: undefined, v2: BigNumberish], BN>;
```

We'd have to introduce some additional infrastructure for these cases that is specific _only_ for the `()` type in the context of it being used directly as a parameter, and IMO it's not worth it because this is more an edge case the language allows and we have to account for rather than something practical developers would be writing. 

Also, the same way nobody writes in typescript `function my_function(input: undefined): number`, nobody will write in sway `fn my_function(input: ()) -> u8` so that we have to spend time catering to their DX when writing such oddities.

nedsalk:
Converted to draft because I found some issues.

arboleya (commented):
These feel ok:

```ts
contract.functions.empty().call();
contract.functions.value_then_empty(35).call();
```

These do not:

```ts
contract.functions.empty_then_value(35).call();
contract.functions.value_then_empty_then_value(35, 35).call();
```

I believe the expected usage would be somewhat similar to the following:

```ts
contract.functions.empty_then_value(null, 35).call();
contract.functions.value_then_empty_then_value(35, null, 35).call();
```

**EDIT**: If empty types are always ignored, should they exist in Sway and compile successfully? Do you know their purpose?

nedsalk:
@arboleya I don't know the purpose of the empty types.

I agree with you that the expected behavior should be the one you mentioned for all cases. I didn't implement it like that because it required changes to the runtime behavior of the ABI coder, because currently the abi-coder also removes the empty types and fails if you provide it a value like in your proposal. 
I didn't go down that route because of the critical work going on with the new ABI encoding. These edge case runtime tests I added will have to be green for the new encoding as well and maybe we'll have to change the runtime behavior for empty types due to them. Thus, IMO we should handle these oddities in a separate PR once the dust settles. What do you think?

packages/abi-typegen/src/abi/types/EmptyType.ts
```diff
@@ -0,0 +1,32 @@
...
+  public name = 'empty';
+
+  static MATCH_REGEX: RegExp = /^\(\)$/m;
+
+  constructor(params: { rawAbiType: IRawAbiTypeRoot }) {
+    super(params);
+    this.attributes = {
+      /**
+       * The empty type is always ignored in function inputs
+       * and it's an bug in the typegen if it makes its way into a function's inputs list
```
arboleya, line 18: Nit.

```suggestion
       * The empty type is always ignored in function inputs. If it makes
       * its way into a function's inputs list, it's a bug in the typegen.
```

packages/abi-typegen/src/utils/parseTypeArguments.test.ts
```diff
@@ -108,7 +108,7 @@ describe('parseTypeArguments.ts', () => {
   test('should fallback to void for null outputs', () => {
     const project = getTypegenForcProject(AbiTypegenProjectsEnum.FN_VOID);
 
-    const types = bundleTypes([]);
+    const types = bundleTypes(project.abiContents.types);
```
arboleya, line 111: Why was this necessary?

arboleya (dismissed):
Please create a detailed follow-up issue with the adjustments you mentioned on the ABI Coder.

Using our tests to dogfood Typegen was a nice touch! 👏 🚀 

I hope this is also beneficial for browser testing.

packages/abi-typegen/src/utils/parseTypeArguments.test.ts
nedsalk: Because this test is testing that the output is `void`, which is now [handled](https://github.com/FuelLabs/fuels-ts/blob/0a7cf759cac2d2e934be714654852e7037b8c700/packages/abi-typegen/src/abi/types/EmptyType.ts#L21) by the `EmptyType` and so we need the empty type in the types array:
```ts
[
    {
      "typeId": 0,
      "type": "()",
      "components": [],
      "typeParameters": null
    }
]
```
This type comes included in `project.abiContents.types`.

Issue #2109, opened 2024-04-18
Remove `v0` encoding logic


danielbate:
Once #2050 has been merged and [sway encoding improvements](https://github.com/FuelLabs/sway/issues/5727) to `v1` have been completed, we should remove all `v0` specific logic:

- Encoding should now fallback to `v1`, strategy can remain in place for future custom encoding
- Remove redundant `v0` coders, some redistribution may be required for anything still being used by transactions
- Remove `v0` function selector encoding and contract call script
- Remove heap pointer logic
- Remove heap type validation for multicalls
- Remove experimental tests
- Ensure documentation type support is up to date

arboleya:
@danielbate I think we can start this right after https://github.com/FuelLabs/fuels-ts/pull/2050, regardless of https://github.com/FuelLabs/sway/issues/5727.

Only v1 will be accepted anyway, and the stabilization should be transparent.

danielbate:
Agreed 👍🏻

PR #2126, opened 2024-04-22
fix: encoding version type

### LINKED ISSUE #2124: JsonAbi encoding version type too strict
The [`encoding`](https://github.com/FuelLabs/fuels-ts/blob/master/packages/abi-coder/src/types/JsonAbi.ts) field type for the `JsonAbi` is too strict and preventing users from using a raw JSON with the `ContractFactory`. We should relax this type to allow users to pass a JSON object to the ContractFactory.

```console
@fuel-bridge/test-utils:build:     Types of property 'encoding' are incompatible.
@fuel-bridge/test-utils:build:       Type 'string' is not assignable to type 'EncodingVersion'.
@fuel-bridge/test-utils:build: src/utils/fuels/getOrDeployFuelTokenContract.ts(56,7): error TS2345: Argument of type '{ encoding: string; types: { typeId: number; type: string; components: { name: string; type: number; typeArguments: any; }[]; typeParameters: number[]; }[]; functions: { inputs: { name: string; type: number; typeArguments: any; }[]; name: string; output: { ...; }; attributes: { ...; }[]; }[]; loggedTypes: { ...; }[]...' is not assignable to parameter of type 'JsonAbi | Interface<JsonAbi>'.
```
[Source](https://github.com/FuelLabs/fuel-bridge/blob/main/packages/test-utils/src/utils/fuels/getOrDeployFuelTokenContract.ts#L56)

petertonysmith94:
Closes #2124 

---

- Changed the `version` field on the `JsonAbi` to a string.
- Added validation on the encoding field.


packages/abi-coder/src/utils/json-abi.ts
```diff
@@ -3,6 +3,39 @@ import { ErrorCode, FuelError } from '@fuel-ts/errors';
...
+    case ENCODING_V0:
+      return ENCODING_V0;
+
+    default:
+      throw new FuelError(
+        ErrorCode.UNSUPPORTED_ENCODING_VERSION,
+        `Encoding version '${encoding}' is unsupported.`
+      );
+  }
+};
```
nedsalk, line 37:
```suggestion
export const getEncodingVersion = (encoding?: string): EncodingVersion => {
  switch (encoding) {
    case undefined:
    case ENCODING_V0:
        return ENCODING_V0;

    case ENCODING_V1:
      return ENCODING_V1;

    default:
      throw new FuelError(
        ErrorCode.UNSUPPORTED_ENCODING_VERSION,
        `Encoding version '${encoding}' is unsupported.`
      );
  }
};
```
And you can remove the DEFAULT_ENCODING_VERSION constant with this.

packages/abi-coder/src/types/JsonAbi.ts
```diff
@@ -2,15 +2,12 @@
...
-
-import type { EncodingVersion } from '../utils/constants';
-
 export interface JsonAbi {
   readonly types: readonly JsonAbiType[];
   readonly loggedTypes: readonly JsonAbiLoggedType[];
   readonly functions: readonly JsonAbiFunction[];
   readonly configurables: readonly JsonAbiConfigurable[];
-  readonly encoding?: EncodingVersion;
+  readonly encoding?: string;
```
nedsalk, line 10: Can you explain why we'd go from `EncodingVersion` to `string` when this field is optional to begin with so the user doesn't have to type it at all?

packages/abi-coder/src/utils/constants.ts
```diff
@@ -25,9 +25,11 @@ export const genericRegEx = /^generic (?<name>\w+)$/;
 /**
  * Encoding versions
  */
-export type EncodingVersion = '0' | '1';
 export const ENCODING_V0 = '0';
 export const ENCODING_V1 = '1';
+export const DEFAULT_ENCODING_VERSION = ENCODING_V0;
```
maschad, line 30: Is the default always going to be `v0` ? If so then I don't think we need this constant and then @nedsalk 's [comment applies](https://github.com/FuelLabs/fuels-ts/pull/2126/files#r1574608969)

maschad (changes requested):
Nice, but also need to address the comments that you made as well.

maschad (approved):
Great work!

nedsalk (approved):

danielbate (approved):
Good find 🦊 

This PR introduces some duplication for encoding version retrieval, however approving as this resolves the bug and I think that can be cleaned up #2109

PR #2130, opened 2024-04-22
fix: internalize `ethers` functionality and remove dependency

maschad:
Closes https://github.com/FuelLabs/fuels-ts/issues/2153

`ethers` is now removed from:
- `account` 
- `crypto`

As a result  there are more `ethers` functions being internalized:
- `ripemd160` 
-  `pbkdf2`
- `hmac`
- `dataSlice`
- `encodeBase58`
- `decodeBase58`
- `getBytes`

This can be merged as a follow up to https://github.com/FuelLabs/fuels-ts/pull/2152


packages/crypto/test/pbkdf2.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    const algo = 'sha256';
+    const expectedResult = '0x90eceedd899d5cdcdfd9b315ad6e2c3391bf95cc131b6f0f016339db5ee60494';
+
+    const result = pbkdf2(passwordBuffer, saltBuffer, iterations, keylen, algo);
+
+    expect(result).toBe(expectedResult);
+  });
+});
+
+// #TODO: Even when I run the filter:browser, these tests still run
```
maschad, line 22: I'm skipping these are they are not critical tests but locally even when I run the browser filter these tests still run, could it be because they are in the same file?

packages/account/src/hdwallet/hdwallet.ts
```diff
@@ -221,7 +221,8 @@ class HDWallet {
   }
 
   static fromExtendedKey(extendedKey: string) {
-    const decoded = toBeHex(decodeBase58(extendedKey));
+    let decoded = toHex(decodeBase58(extendedKey));
+    decoded = `${decoded.substring(0, 2)}0${decoded.substring(2)}`;
```
maschad, line 225: This was a strange addition on my part needed to make the tests pass, essentially toHex omits a padded 0 at the third index, thus 0x0488b21e03587302... becomes 0x488b21e03587302.... but I would think that if toHex had a bug it would've been discovered much earlier... any thoughts?

fuel-service-user:
✨ A PR has been created under the `rc-2130` tag on `fuels-wallet` repo.
https://github.com/FuelLabs/fuels-wallet/pull/1242

packages/crypto/test/pbkdf2.test.ts
Torres-ssf: That's odd 🤔 . 

But I reckon you are right, maybe this unintended behavior is happening because they are within the same file

arboleya (commented):
The `getNetwork()` work here appears to be small compared to the rest, making the PR title slightly misleading.

Should we split the PR into two, dealing with the `getNetwork()` in a subsequent PR while re-shaping this one to be concerned only with replacing ethers utilities?

maschad:
> The `getNetwork()` work here appears to be small compared to the rest, making the PR title slightly misleading.
> 
> Should we split the PR into two, dealing with the `getNetwork()` in a subsequent PR while re-shaping this one to be concerned only with replacing ethers utilities?

That's a good point @arboleya I'll do that.

.changeset/grumpy-years-stare.md
```diff
@@ -0,0 +1,8 @@
+---
+"@fuel-ts/account": minor
+"@fuel-ts/crypto": minor
+"@fuel-ts/utils": minor
```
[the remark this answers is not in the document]
maschad: Good catch, my bad.

packages/account/src/hdwallet/hdwallet.ts
maschad: I'm not sure I agree with that, but regardless the [`toHex` function](https://github.com/FuelLabs/fuels-ts/blob/refactor/internalize-ethers-deprecate-getnetwork/packages/math/src/functional.ts#L15) was implemented some time ago is used pretty widely across the repo so adjusting it would be a substantial undertaking beyond the scope of this PR.

packages/errors/src/error-codes.ts
```diff
@@ -103,5 +103,5 @@ export enum ErrorCode {
   STREAM_PARSING_ERROR = 'stream-parsing-error',
 
   // coder
-  // ...
+  BUFFER_OVERRUN = 'buffer-overrun',
```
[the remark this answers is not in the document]
maschad: I decided against adding a new error as `INVALID_DATA` suits this use case.

packages/crypto/test/pbkdf2.test.ts
maschad: I decided to move them to seperate files for now and then we can re-adjust the script to be more flexible in a follow up PR if needs be.

packages/crypto/test/hmac.node.test.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import { computeHmac } from '..';
+/**
+ * @group node
+ */
+describe('computeHmac node', () => {
+  it('should use the registered HMAC function', () => {
+    const key = '0x0102030405060708090a0b0c0d0e0f10';
+    const data = '0x11121314151617181920212223242526';
+    const expectedHmac = '0x1234567890abcdef';
```
danielbate, line 11: Were these test cases taken from `ethers`? If not how did you get the expected value?

packages/crypto/test/hmac.node.test.ts
danielbate, line 1: How come the `node` file name?

packages/crypto/test/pbkdf2.node.test.ts
```diff
@@ -0,0 +1,38 @@
+import { bufferFromString, pbkdf2 } from '..';
+
+/**
+ * @group node
```
danielbate, line 4:
```suggestion
 * @group node
 * @group browser
```

danielbate (commented):
Great work @maschad, another good step towards decoupling from `ethers`.

What was the reasoning behind implementing `getBytes`? As it looks exactly the same as our own `arrayify`?

packages/account/src/hdwallet/hdwallet.ts
arboleya: cc @nedsalk — is this relevant to the problems you had with `toHex` vs. `hexlify`?

packages/crypto/test/pbkdf2.node.test.ts
maschad: These tests are intended to run specifically on Node as they relate to a method that is useful for multi-threaded node environments which is not applicable in the browser.

packages/crypto/test/hmac.node.test.ts
maschad: These are node specific tests as opposed to the other `hmac` test file which is for both browser and node,  it's a limitation of the current test setup. (See https://github.com/FuelLabs/fuels-ts/pull/2130#discussion_r1586479472 for wider discussion)

maschad:
> Great work @maschad, another good step towards decoupling from `ethers`.
> 
> What was the reasoning behind implementing `getBytes`? As it looks exactly the same as our own `arrayify`?

Thanks! There are some [additional parameters](https://github.com/FuelLabs/fuels-ts/blob/refactor/internalize-ethers-deprecate-getnetwork/packages/utils/src/utils/getBytes.ts#L12) included in the function signature for `getBytes` that I would need for [some use cases](https://github.com/FuelLabs/fuels-ts/blob/refactor/internalize-ethers-deprecate-getnetwork/packages/crypto/src/node/hmac.ts#L20-L21) I opted not to modify the `arrayify` to reduce the surface of these changes, but if that's acceptable I can include them in this PR.

packages/crypto/src/browser/pbkdf2.ts
```diff
@@ -0,0 +1,18 @@
...
+import { sha512 } from '@noble/hashes/sha512';
+
+import type { CryptoApi } from '../types';
+
+export const pbkdf2: CryptoApi['pbkdf2'] = (
+  password: BytesLike,
+  salt: BytesLike,
+  iterations: number,
+  keylen: number,
+  _algo: 'sha256' | 'sha512'
```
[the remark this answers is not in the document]
maschad: Good catch!

packages/crypto/test/hmac.node.test.ts
maschad: You have a variety of tools to derive the HMAC value such as [this](https://cryptotools.net/hmac) but this particular test is just evaluating that if we register an amended algorithm, that it will return the right value.

danielbate:
IMO that is acceptable here, I'd rather have one point of entry for byte array conversion. I probably would have done the same when we initially internalised `ethers` if it didn't work with all the use cases.

danielbate (changes requested):
As [discussed](https://github.com/FuelLabs/fuels-ts/pull/2130#issuecomment-2092028930), amalgamate the functionality of `getBytes` and `arrayify`.

maschad:
> As [discussed](https://github.com/FuelLabs/fuels-ts/pull/2130#issuecomment-2092028930), amalgamate the functionality of `getBytes` and `arrayify`.

Thanks @danielbate I've updated in 69bf67a

packages/account/src/hdwallet/hdwallet.ts
nedsalk: I had an issue where I was using `toHex(randomBytes(32))` instead of `hexlify(randomBytes(32))` for generating asset ids in a custom chain config, which the `fuel-core` node sometimes accepted and sometimes couldn't interpret as a valid asset id. After switching to `hexlify`, the problem went away. From the names of the functions, one would expect the same behavior; I didn't investigate what their differences are.

@maschad Did you try `hexlify(decodeBase58(extendedKey))` here?

maschad: @nedsalk thanks for the suggestion, I actually ended up converting it to bytes then hexilify-ing it.

danielbate (approved):
Nice work 💪🏻 

Jw, what `ethers` deps are left after this?

packages/account/src/hdwallet/hdwallet.ts
Torres-ssf: @maschad Is `toHex` still omiting the leading zero?

maschad:
> Nice work 💪🏻
> 
> Jw, what `ethers` deps are left after this?

Thanks 😄 

Looks like just `doc-snippets` and `abi-coder` now

packages/account/src/hdwallet/hdwallet.ts
maschad: That's my understanding, but I am using `hexlify` here

PR #2135, opened 2024-04-22
docs: reorganize docs navigation

sarahschwartz:


apps/docs/src/guide/contracts/deploying-contracts.md
```diff
@@ -37,4 +37,4 @@ Now that the contract is deployed, you can interact with it. In the following st
 
 <<< @/../../docs-snippets/src/guide/contracts/deploying-contracts.test.ts#contract-setup-4{ts:line-numbers}
 
-For a more comprehensive TypeScript-backed Fuel usage, learn how to [generate types from ABI](../typegen/generating-types.md)
+For a more comprehensive TypeScript-backed Fuel usage, learn how to [generate types from ABI](../fuels-cli/generating-types.md)
```
maschad, line 40: This may not be directly related to this PR but what does _TypeScript-backed_ mean? Is it that the generated types are Typescript types?

maschad (approved):
Great work! Looks good, just left a question.

danielbate (dismissed):

apps/docs/src/guide/contracts/deploying-contracts.md
sarahschwartz: Yes exactly, but if you think that isn't clear it's probably worth rewording this at some point.

Torres-ssf (approved):

PR #2125, opened 2024-04-19
build: upgrade `pnpm` version

maschad:
The [lint step fails on CI](https://github.com/FuelLabs/fuels-ts/actions/runs/8757757991/job/24037100858?pr=2117) but this isn't happening for me locally. I am running `pnpm` `9.0.2` so figured I would upgrade it across the board.

PR #2136, opened 2024-04-22
build: update `pnpm-lock.yaml`

petertonysmith94:
Updating the `pnpm-lock.yaml` after the version [update](https://github.com/FuelLabs/fuels-ts/pull/2125/files).


maschad (dismissed):
Good catch, we probably should introduce a CI step that ensures the `lockfile` remains synchronized.

PR #2165, opened 2024-04-25
feat: add `getMessageByNonce` to `Provider`

### LINKED ISSUE #1538: Implement `getMessageByNonce` or similar
We probably need to request this feature from `fuel-core.`

> [!NOTE]
> - Incoming from https://github.com/FuelLabs/fuels-ts/discussions/1537

Dhaiwat10:
## Summary

This PR adds a new method called `getMessageByNonce` to the `Provider` class which lets users retrieve a message using its nonce.

Closes #1538


Torres-ssf (dismissed):

packages/account/src/providers/provider.ts
```diff
@@ -1635,4 +1636,15 @@ export default class Provider {
...
     return new TransactionResponse(transactionId, this);
   }
+
+  /**
+   * Returns Message for given nonce.
+   *
+   * @param nonce - The nonce of the message to retrieve.
+   * @returns A promise that resolves to the Message object.
+   */
+  async getMessageByNonce(nonce: string): Promise<GqlMessage | undefined | null> {
```
maschad, line 1646: Should we just return a `GqlMessage` with the values set to a default as opposed to a `undefined` | `null` ? It's quite difficult to work with APIs that could return `null` | `undefined` as well as a type

Dhaiwat10: @maschad I don't think we should or we can fill in defaults since that would be unexpected behaviour. I think we should simplify the API though. I will return `undefined` in case there's no message found, and the actual message if we find something.

Dhaiwat10:
Git history messed up here after changing base. Closing in favour of #2174

PR #2167, opened 2024-04-25
build: create changeset PR only if version will be updated

### LINKED ISSUE #2068: Create changeset PR only if merged commit meets criteria
Merging a commit to `master` currently always creates or updates the changeset PR, even if the merged commit doesn't have a changeset that affects any package.

This is problematic when a changeset PR is _newly created_ because its name [depends](https://github.com/FuelLabs/fuels-ts/blob/29d530367505aebe6652920d415414dee8cbc7fa/.github/workflows/release.yaml#L54) on the new version it'd release upon merging into its base branch. If a new version isn't released because no changeset file has a patch/minor/major update in it, it'll take the latest-released version for its name, which is confusing and wrong.

> [!IMPORTANT]
> Apply this logic only on creation of a changeset PR, not on its updating. The updating should happen _always_, even if the merged commit has an empty changeset, because the release/publish behavior depends on _all_ changeset files being removed, as explained in #2067. The changeset PR can remove all changesets from its base branch only if it's updated every time the base branch is updated.

nedsalk:
closes #2068


maschad (approved):
Nice work

arboleya (approved):

Issue #2160, opened 2024-04-24
chore: remove `sway-libs` dependency


maschad:
`sway-libs` doesn't keep compatibility with `forc`, resulting in [broken code snippets](https://github.com/FuelLabs/fuels-ts/pull/2045/files#diff-1a11e2c177f1f12253aa42005f3290260aa32da5c9362433ee63dadd2aa5c5fa) during `forc` upgrades 

_Originally posted by @arboleya in https://github.com/FuelLabs/fuels-ts/pull/2045#discussion_r1578445319_

arboleya:
I am already doing this as part of #2042. 👋

arboleya:
Relevant places to change should involve the ones included in these commits:
 - https://github.com/FuelLabs/fuels-ts/pull/2045/commits/47c511eb3fe19e7cd1b39f0948649cd47f4c3e0a
 - https://github.com/FuelLabs/fuels-ts/pull/2045/commits/be3accd9594a4b2be5d061fbfdf0b8451d802cb2

PR #2018, opened 2024-04-05
feat: accept `Uint8Array`s as inputs for `Vec`s, add docs for bytecode inputs

Dhaiwat10:
## Summary

Before this PR, the `VecCoder` wasn't accepting input arrays of type `Uint8Array`. In some cases, like the `compute_bytecode_root` function from the [bytecode Sway Lib](https://github.com/FuelLabs/sway-libs/tree/master/libs/src/bytecode), you need to pass in a `Uint8Array` to the Sway function.

I have tweaked the validation inside the `VecCoder` for the inputs, and also added some documentation around how to pass in bytecode to Sway functions.

You can pass in bytecode to any Sway function like so now:

```ts
import { arrayify } from 'fuels';

const { logs } = await contract.functions
  .compute_bytecode_root(arrayify(bytecodeFromFile))
  .call();
```

Closes #1950 

This PR will unblock the Fuelnaut project, too.

Issue #1950, opened 2024-03-26
Add a helper for the conversion: bytecode file contents -> `Vec<u8>`


Dhaiwat10:
We need a helper function to read bytecode from a file and convert that to a bytearray that users can pass into our `sway-libs` functions related to bytecode: https://github.com/FuelLabs/sway-libs/tree/master/libs/bytecode

This conversion is required because the bytecode from a file is a `string`, but the type that these Sway functions accept for the bytecode is a `Vec<u8>` - which in JavaScript-land is a `Vec<BigNumberish>`.

This helper would convert the string to the proper type and ensure all the data is correctly packed.

## Possible API
```ts
const bytecodeFromFile = readFile(pathToBytecode); // string

const bytecodeVec = generateBytecodeVecFromString(bytecodeFromFile); // our helper -> returns `Vec<BigNumberish>`

await contract.functions.verify_bytecode(bytecodeVec);
```

> _Incoming from feedback from the development work being done on the Fuelnaut project: https://github.com/FuelLabs/docs-hub/pull/203_

The Fuelnaut project is blocked for now until we figure this out.

Dhaiwat10:
cc @sarahschwartz

arboleya:
@Dhaiwat10 You mean the contract bytecode needs to be sent as an argument to another contract/function call?

Dhaiwat10:
@arboleya precisely

nedsalk:
If we were to treat `Vec<u8>` as `number[] | Uint8Array` then we can do simply do this:
```ts
import { arrayify } from 'fuels';

const bytecodeFromFile = readFile(pathToBytecode); // string

await contract.functions.verify_bytecode(arrayify(bytecodeFromFile));
```

nedsalk:
@sarahschwartz Could you try my proposal from above? You'd have to do something like this:
```ts
await contract.functions.verify_bytecode(arrayify(bytecodeFromFile) as unknown as number[]);
```

sarahschwartz:
> @sarahschwartz Could you try my proposal from above? You'd have to do something like this:
> 
> ```ts
> await contract.functions.verify_bytecode(arrayify(bytecodeFromFile) as unknown as number[]);
> ```

I tried this, but i get this error: 

```
FuelnautLevel.tsx:136 Error deploying new instance: FuelError: Expected array value.
    at VecCoder.encode (index.mjs:723:13)
    at eval (index.mjs:687:31)
[...]

PR #2189, opened 2024-04-26
chore: remove `sway-libs` dep from testing fixtures

maschad:
Closes https://github.com/FuelLabs/fuels-ts/issues/2160

- Mocks the `compute_bytecode` functions in the sway contracts


packages/fuel-gauge/test/fixtures/forc-projects/bytecode-sway-lib/src/main.sw
```diff
@@ -1,33 +1,25 @@
...
 }
 
 impl MyContract for Contract {
     fn compute_bytecode_root(bytecode_input: Vec<u8>) {
-        // let mut bytecode = bytecode_input;
-        // let root = compute_bytecode_root(bytecode);
-        // log(root);
+        // simply logs the hexidecimal b256 string of the bytecode input for testing purposes
+        let ret: b256 = 0x0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20;
+        log(ret);
```
arboleya, line 15: Would the documentation lose its meaning if we changed the snippet?


doc-snippet

maschad: That's a good point, but how do you intend to have this mocked if we still rely on the `bytecode` lib from sway libs?

arboleya: Yeah, good question. I didn't think this through enough.

One suggestion @Dhaiwat10 gave was to use a hardcoded Sway snippet, but that would be a temporary solution.

We should reevaluate whether this page is really needed or rethink/rewrite it differently.

maschad: I don't think this page is particularly relevant to the TS SDK - if we begin to delve into documenting how to work with Sway libraries it would be determine where we draw the line. I think this is under the purview of the Sway library.

arboleya: Well, @Dhaiwat10 can confirm this, but I'd assume it was added because someone had difficulties achieving the described approach. In this case, it is beneficial and valuable to have it there.

However, we need to reframe the message.

 1. New title
 2. Condensed paragraphs
 3. Improved TS snippet readability
 4. Fixed TS imports
 5. Removal of Sway snippet
 6. Pointing to Sway docs instead of source code

In this scenario, your new Sway snippet can stay as is as its only purpose will be to ensure that the Typescript snippet is valid and working. I think that seals the deal.

---

# Converting Bytecode to `Vec<u8>`

Some Sway functions require you to pass in bytecode to the function. The type of the bytecode parameter is usually `Vec<u8>`. For example, take the `compute_bytecode_root` function from the bytecode Sway library ([docs](https://fuellabs.github.io/sway-libs/book/bytecode/index.html?highlight=bytecode#using-the-bytecode-library-in-sway)).

To pass bytecode to this function, you can make use of the `arrayify` function to convert the bytecode file contents into a `UInt8Array`, the Typescript compatible type for Sway's `Vec<u8>` type, and pass it the function like so:

```ts
import { arrayify } from 'fuels';
 
const bytecode = readFileSync(bytecodePath);
const bytecodeAsVecU8 = arrayify(bytecode);
 
const { value: bytecodeRoot } = await bytecodeContract.functions
  .compute_bytecode_root(bytecodeAsVecU8)
  .call();

console.log(`Bytecode root is: ${bytecodeRoot}`);
```

maschad:
@Dhaiwat10 @arboleya any ideas why [this test](https://github.com/FuelLabs/fuels-ts/actions/runs/8992479675/job/24702732336?pr=2189#step:10:642) would hang on CI?  It works fine locally
Screenshot 2024-05-07 at 4 33 09 PM

apps/docs-snippets/src/guide/types/vector.test.ts
```diff
@@ -69,9 +68,10 @@ describe(__filename, () => {
     // #import { arrayify, readFile };
 
     const bytecode = await readFile(bytecodePath);
+    const bytecodeAsVecU8 = arrayify(bytecode);
 
     const { value: bytecodeRoot } = await bytecodeContract.functions
-      .compute_bytecode_root(arrayify(bytecode))
+      .compute_bytecode_root(Array.from(bytecodeAsVecU8))
```
arboleya, line 74: How come `Array.from()` was necessary now if it wasn't before when using the non-mocked Sway implementation?

arboleya (commented):
> any ideas why [this test](https://github.com/FuelLabs/fuels-ts/actions/runs/8992479675/job/24702732336?pr=2189#step:10:642) would hang on CI? It works fine locally

Nothing I can see only by reading it.

apps/docs-snippets/src/guide/types/vector.test.ts
maschad: I forgot to leave a comment regarding this, it seems the encoding has changed with the upgrade, @nedsalk you had some thoughts on this?

danielbate (commented):
One concern here is the docs still references the sway std lib function, so we are miss communicating the dependency. If the sway std lib signature changed, we are not in sync as we have a mocked version. I think we should be more generally referencing this as working with bytes with both a `Vec<u8>` and `Bytes` example. Thoughts?

apps/docs-snippets/src/guide/types/vector.test.ts
nedsalk: No special thoughts, I was just mentioning #1950 which enabled the `arrayify(bytecode))` which is being changed here.

maschad:
> One concern here is the docs still references the sway std lib function, so we are miss communicating the dependency. If the sway std lib signature changed, we are not in sync as we have a mocked version. I think we should be more generally referencing this as working with bytes with both a `Vec<u8>` and `Bytes` example. Thoughts?

Whether it's mocked or not wouldn't have an effect as we would still be using a particular version as our [dependency](https://github.com/FuelLabs/fuels-ts/pull/2189/files#diff-ae3432834f7a896fa4e4716d9bf083f99533aff2aea0457722399e387b8fd043L10)

the benefit with this at least would be should we try to upgrade, if there is a breaking change Rust would let us know due to it's [semver](https://docs.rs/semver/latest/semver/) toolchain - but this would still be dependent on our manual attempt to upgrade. 

In summary, I'm not sure if there is an uncomplicated way to discuss working with bytecode or `Vec<u8> ` without using a concrete example from a sway library.

apps/docs-snippets/src/guide/types/vector.test.ts
maschad: It seems the current encoding post-upgrade of Forc expects an array now?

Dhaiwat10: cc @danielbate any thoughts/insights that you might have?

danielbate:
> Whether it's mocked or not wouldn't have an effect as it relates to breaking changes as we would still be using a particular version as our [dependency](https://github.com/FuelLabs/fuels-ts/pull/2189/files#diff-ae3432834f7a896fa4e4716d9bf083f99533aff2aea0457722399e387b8fd043L10)

But agreed we should remove the dependency, like you have done.

> In summary, I'm not sure if there is an uncomplicated way to discuss working with bytecode or Vec<u8>  without using a concrete example from a sway library.

I would personally prefer removing the reference to the sway std lib in the docs and just more generally speak about bytecode.

```md
Some functions require you to pass in bytecode to the function. The type of the bytecode parameter is usually `Vec<u8>`, this would be handled like so:

<-- code snippet -->
```

apps/docs-snippets/src/guide/types/vector.test.ts
danielbate: Looks like @Dhaiwat10 in #2018 change to add `UInt8Array` as a viable `Vec` input to the `VecCoder` did not make it's way to the `V1` coder. My b I must've missed it, will push a change now.

danielbate: #2289 should resolve this @maschad.

maschad:
> I would personally prefer removing the reference to the sway std lib in the docs and just more generally speak about bytecode.

2f6af7d

arboleya (approved):

danielbate (approved):

Torres-ssf (approved):

PR #2191, opened 2024-04-29
build: create changeset PR only if version will be updated

nedsalk:


nedsalk:
Closing because CI is behaving strangely, will reopen.

arboleya:
Do we also have an issue for this?

nedsalk:
@arboleya I thought I created one but I couldn't find it so I created #2511. I'll be closing this because I can't get to the root of the problem of why I can't get the correct commit message in the CI and it's a really low priority issue.

PR #2199, opened 2024-04-29
ci: add workflow to close stale issues

maschad:
Closes https://github.com/FuelLabs/fuels-ts/issues/2203

Introduces a workflow to clean up potentially stale issues / bugs raised where we need author input to progress.

This workflow requires two new labels `need-author-input` and `stale`

So the sequence of actions would be this:

1. A user (whether code owner or not) creates an issue, no label is assigned to it.
2. A code owner responds to that issue, but determines that more information is needed to proceed, they assign the `need-author-input` label
3. Only one that `need-author-input` label has been assigned to the issue, the countdown (of 14 days in this case) begins, after 14 days the label stale is assigned to the issue and a generated warning is commented by the github bot that the issue will close in 7 days
4. After 7 days of the issue being stale it is closed by the github bot.


.github/workflows/stale.yml
```diff
@@ -0,0 +1,24 @@
...
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v3
+        with:
+          repo-token: ${{ github.token }}
+          stale-issue-message: "Oops, seems like we needed more information for this issue, please comment with more details or this issue will be closed in 7 days."
+          close-issue-message: "This issue was closed because it is missing author input."
+          stale-issue-label: "stale"
+          any-of-labels: "need-author-input"
+          exempt-issue-labels: "need/triage,need/community-input,need/maintainer-input,need/maintainers-input,need/analysis,status/blocked,status/in-progress,status/ready,status/deferred,status/inactive"
```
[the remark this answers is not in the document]
maschad: definitely not 😅

.github/workflows/stale.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v3
+        with:
+          repo-token: ${{ github.token }}
```
arboleya, line 16: `github.token`?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,23 @@
...
+          stale-issue-message: "Oops, seems like we needed more information for this issue, please comment with more details or this issue will be closed in 7 days."
+          close-issue-message: "This issue was closed because it is missing author input."
+          stale-issue-label: "stale"
+          any-of-labels: "need-author-input"
+          days-before-issue-stale: 6
+          days-before-issue-close: 7
+          enable-statistics: true
```
arboleya, line 23: IIUC, statistics depend on [enabling logs](https://github.com/actions/stale?tab=readme-ov-file#debugging) - is it worth it?

maschad: probably not at this time, we can re-visit in the future

.github/workflows/stale.yml
maschad: copy pasta 🤪

.github/workflows/stale.yml
```diff
@@ -0,0 +1,22 @@
...
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          stale-issue-message: "Oops, seems like we needed more information for this issue, please comment with more details or this issue will be closed in 7 days."
+          close-issue-message: "This issue was closed because it is missing author input."
+          stale-issue-label: "stale"
+          any-of-labels: "need-author-input"
+          days-before-issue-stale: 6
```
arboleya, line 21: Is 6x days enough to mark things as stale?

Could you please clarify the criteria by which it happens?

For example, will it apply only to issues created by non-code owners?

danielbate: Could you add this to the PR description?

danielbate (approved):

.github/workflows/stale.yml
```diff
@@ -0,0 +1,22 @@
...
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          stale-issue-message: "Oops, seems like we needed more information for this issue, please comment with more details or this issue will be closed in 7 days."
+          close-issue-message: "This issue was closed because it is missing author input."
+          stale-issue-label: "stale"
+          any-of-labels: "need-author-input"
```
arboleya, line 20: We should use a shorter version more aligned with the other labels.

```suggestion
          any-of-labels: "awaiting"
```

arboleya (changes requested):

.github/workflows/stale.yml
arboleya: @maschad Perfect, thanks for clarifying.

I think 6x days may not be enough, can we start with 14 days and see how it goes?

maschad: Sure

.github/workflows/stale.yml
maschad: I can appreciate the brevity but I think _awaiting_ is a bit ambiguous, I would suggest _awaiting-input_ but that's just two characters shorter than _need-author-input_

arboleya: Apologies, I expressed myself poorly. I was referring to the label's simplicity (single word, no hyphens). I believe it's a no-brainer for us to understand what it means, and we're the only ones who actually need to know.

A question like `Awaiting for what?` has arguably the same weight as `Blocked by what?`. It doesn't seem like a problem to me, but if it becomes one, we could disambiguate in the label's description if necessary.

Either way, it's worth considering its use will be marginal right now, another reason I'd favor the simplicity of words and broad terms. If external participation increases and it becomes strictly necessary (as in, a _problem_), we can always revisit and rename it. 🙂

maschad: I take your point

arboleya (approved):
Thanks for the PR and adjustments! 🙌

Torres-ssf (approved):
🚀

Issue #2147, opened 2024-04-23
Upgrade `fuel-core` to `0.26.0`


arboleya:
Changelog:
 - https://github.com/FuelLabs/fuel-core/releases/tag/v0.25.2
 - https://github.com/FuelLabs/fuel-core/releases/tag/v0.25.3
 - https://github.com/FuelLabs/fuel-core/releases/tag/v0.26.0

Issue #1578, opened 2023-12-21
Possible issue with `getTransactionSummary`


mpoplavkov:
> `getTransactionSummary` uses current chain configuration for fee calculation. Even for old transactions.

### What version of fuels-ts are you using?

master

### Steps to Reproduce

Just call the [getTransactionSummary](https://github.com/FuelLabs/fuels-ts/blob/53dafb1c6e5e02e47f805c111a8f348278101a5f/packages/providers/src/transaction-summary/get-transaction-summary.ts#L25) method. 

It takes the parameters such as `gasPriceFactor` and `gasCosts` at the time of Provider creation, not the time of transaction execution: [proof](https://github.com/FuelLabs/fuels-ts/blob/53dafb1c6e5e02e47f805c111a8f348278101a5f/packages/providers/src/transaction-summary/get-transaction-summary.ts#L48-L50). Those parameters are later [used](https://github.com/FuelLabs/fuels-ts/blob/53dafb1c6e5e02e47f805c111a8f348278101a5f/packages/providers/src/transaction-summary/assemble-transaction-summary.ts#L68-L78) for fees calculation.

### Expected Behavior

The SDK will use consensus parameters at the time of transaction execution, not the current time.

### Actual Behavior

Current consensus parameters are used for fee calculation of all transactions. If those parameters change, the fees for previous transaction change as well. But those fees should remain the same

maschad:
I'm not sure there's much that we can do right now because the gas price and other related parameters cannot be fetched for a given period of time, we only have access to the current values. Maybe this can be solved with the indexer?

PR #2217, opened 2024-04-30
chore!: upgrade `fuel-core` to `0.26.0`

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/2147
- Close: https://github.com/FuelLabs/fuels-ts/issues/1578

## Breaking Changes
- The type [ConsensusParameters](https://github.com/FuelLabs/fuels-ts/blob/19530e049f5f001146dbe790c30a1d546cca26c7/packages/account/src/providers/provider.ts#L107-L117) was [modified](https://github.com/FuelLabs/fuels-ts/pull/2217/commits/0e883b2186fd3dcdcd9cea69d569390da85dd963#diff-69a6670d0426c74935ed2830ce552e7df2379c2d112a582306e57569fcd17029L88-L106) to match the [GraphQL schema](https://github.com/FuelLabs/fuels-ts/blob/19530e049f5f001146dbe790c30a1d546cca26c7/packages/account/src/providers/fuel-core-schema.graphql#L190-L202).


packages/account/src/test-utils/launchNode.ts
```diff
@@ -198,10 +198,12 @@ export const launchNode = async ({
       const chainConfigWritePath = path.join(tempDirPath, 'chainConfig.json');
       const stateConfigWritePath = path.join(tempDirPath, 'stateConfig.json');
       const metadataWritePath = path.join(tempDirPath, 'metadata.json');
+      const stateTransitionWritePath = path.join(tempDirPath, 'state_transition_bytecode.wasm');
 
       writeFileSync(chainConfigWritePath, JSON.stringify(chainConfigJson), 'utf8');
       writeFileSync(stateConfigWritePath, fixedStateConfigJSON, 'utf8');
       writeFileSync(metadataWritePath, JSON.stringify(metadataJson), 'utf8');
+      writeFileSync(stateTransitionWritePath, JSON.stringify(''));
```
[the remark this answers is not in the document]
Torres-ssf: @petertonysmith94 I am just creating this file without adding anything to it.

Starting on [fuel-core 0.25](https://github.com/FuelLabs/fuel-core/pull/1826) we need the file `state_transition_bytecode.wasm` within the snapshot directory. 

Without this file, `fuel-core` does not start.

Since we are going to use it right now, is does not need any data within it, but it needs to be there whenever we are going to run `fuel-core`

packages/account/src/providers/provider.ts
```diff
@@ -86,24 +93,27 @@ export type ContractResult = {
...
-  maxPredicateDataLength: BN;
-  maxGasPerPredicate: BN;
-  gasPriceFactor: BN;
-  gasPerByte: BN;
-  maxMessageDataLength: BN;
+type ModifyStringToBN<T> = {
+  [P in keyof T]: P extends 'version' ? T[P] : T[P] extends string ? BN : T[P];
+};
+
+export type FeeParameters = Omit<GqlFeeParameters, '__typename'>;
```
nedsalk, line 100: If you modify the `codegen.json` file to include the below, you wouldn't need these intermediary types that get rid of `__typename`:
```ts
{
      ...
      "plugins": [
        {
          "typescript": {
            "skipTypename": true
          }
        },
        ...
      ],
    ...
}
```

packages/account/src/providers/provider.ts
```diff
@@ -1647,4 +1688,18 @@ export default class Provider {
 
     return message;
   }
+
+  async getRelayedTransactionStatus(
```
nedsalk, line 1692: I couldn't find any usages of this method anywhere. Should it be deleted?

packages/account/src/providers/provider.ts
Torres-ssf: @nedsalk Nice suggestion. Let me try that.

Torres-ssf: @nedsalk There is a place where `__typename` is used:

- https://github.com/FuelLabs/fuels-ts/blob/master/packages/account/src/providers/utils/gas.ts#L27-L31

packages/account/src/providers/provider.ts
Torres-ssf: @nedsalk Well, I reckon no one is going to use this query for now but I thought it was good to have it in place already.

arboleya (approved):

maschad (approved):
LGTM

packages/account/src/providers/provider.ts
nedsalk: Hmm, I believe that should be `type` and not `__typename`.

arboleya: This can be done on a follow-up PR if necessary. 👍 

cc @Torres-ssf

Issue #2195, opened 2024-04-29
Replace `eslint` and `prettier` with `biome`


nedsalk:

maschad:
I've been experiencing some strange errors with eslint locally and also [on ci](https://github.com/FuelLabs/fuels-ts/actions/runs/8914039554/job/24480786358) despite not upgrading it, running `biome` I am already impressed with it's performance and it's output, so I'll push forward with this.
<img width="835" alt="Screenshot 2024-05-01 at 2 41 33 PM" src="https://github.com/FuelLabs/fuels-ts/assets/9755286/79f56dbd-a1c1-4707-abe7-fde4e93fdaa4">

maschad:
I think we can revisit this in the future but I wouldn't say this adds value at the moment based on the description https://github.com/FuelLabs/fuels-ts/pull/2219

danielbate:
@FuelLabs/sdk-ts I'd like to propose we close this for now.

arboleya:
Agreed.

PR #2219, opened 2024-05-01
chore: remove `prettier` and `eslint` in favour of `biome`

maschad:
Closes https://github.com/FuelLabs/fuels-ts/issues/2195

**Pros**

- `Biome` would centralize all our `linting` and `formatting` into one file i.e. the [`biome.json`](https://github.com/FuelLabs/fuels-ts/pull/2219/files#diff-2bc8a1f5e9380d5a187a4e90f11b4dd36c3abad6aea44c84be354a4f44cdec55) 
- `Biome`  is [~15x faster than ESLint and ~20x than prettier](https://github.com/biomejs/biome/blob/main/benchmark/README.md)
- Addresses a lot of the [inconsistencies with prettier](https://biomejs.dev/formatter/differences-with-prettier/)

*Cons*

- Extremely disruptive change with no value add
- Not nearly as widely adopted
- It's not stable yet nor is there support for plugins.


packages/abi-coder/src/utils/utilities.test.ts
```diff
@@ -197,14 +197,55 @@ describe('Abi Coder Utilities', () => {
...
     const DATA_OFFSET = 0;
-    // prettier-ignore
     const EXPECTED = new Uint8Array([
-      0,       0,       0,       0,       0,       0,       0,       0,
-      0,       0,       0,       0,       0,       0,       0,       2,
-      0,       0,       0,       0,       0,       0,       0,       2,
-      0,       0,       0,       0,       0,       0,       0,       48,
-      0,       0,       0,       0,       0,       0,       0,       1,
-      0,       0,       0,       0,       0,       0,       0,       1,
+      0,
```
nedsalk, line 201: The previous (manual?) formatting for this was better and it'd be good to leave it as it was.

packages/abi-coder/test/Interface.test.ts
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
+/*
```
nedsalk, line 1: These `/*` look weird

packages/abi-coder/test/Interface.test.ts
```diff
@@ -505,7 +505,7 @@ describe('Abi interface', () => {
             ],
           ],
           encodedValue: (input?: any, _offset: number = 0) => {
-            // eslint-disable-next-line no-param-reassign
+
```
nedsalk, line 508: No need for the whitespace now

packages/account/src/wallet/wallets.ts
```diff
@@ -16,8 +16,8 @@ import { decryptKeystoreWallet } from './keystore-wallet';
 
 export interface GenerateOptions {
   /** Additional entropy for the random bytes */
-  entropy?: BytesLike;
-  provider?: Provider;
+entropy ?: BytesLike;
```
nedsalk, line 19: Didn't know this was even possible, but I like the `entropy?:` without whitespace version

packages/fuel-gauge/src/transaction-response.test.ts
```diff
@@ -1,4 +1,4 @@
-import type {} from '@fuel-ts/account/dist/providers/__generated__/operations';
+import type { } from '@fuel-ts/account/dist/providers/__generated__/operations';
```
nedsalk, line 1: 1. IMO we shouldn't allow empty imports,
2. This import is from the `/dist/` folder, which shouldn't happen (just an observation, it was there previously, you don't need to tackle it here)

packages/program/src/script-request.ts
```diff
@@ -38,9 +38,9 @@ export type ScriptResult = {
   receipts: TransactionResultReceipt[];
   scriptResultReceipt: TransactionResultScriptResultReceipt;
   returnReceipt:
-    | TransactionResultReturnReceipt
-    | TransactionResultReturnDataReceipt
-    | TransactionResultRevertReceipt;
+  | TransactionResultReturnReceipt
```
nedsalk, line 41: I like this way of formatting discriminated unions, the whitespace is unnecessary because the `|` gives a visual distinction.

packages/transactions/src/coders/receipt.ts
```diff
@@ -9,18 +9,18 @@ import { ByteArrayCoder } from './byte-array';
...
-  Revert = 4,
-  Log = 5,
-  LogData = 6,
-  Transfer = 7,
-  TransferOut = 8,
-  ScriptResult = 9,
-  MessageOut = 10,
-  Mint = 11,
-  Burn = 12,
+    Return = 1,
```
nedsalk, line 12: This indentation is weird

packages/account/src/wallet/wallets.ts
danielbate: Agreed.

arboleya (commented):
I second what @petertonysmith94 said. If we are swapping libraries, it must be a drop-in replacement.

Also, do we really want to push biome on others via the `create-fuels` template? I haven't thought much about it, but my instinct says we should not.

maschad:
Agreed @arboleya and @petertonysmith94 the intention is to migrate the rules over, this initial step was just removing eslint. Once that migration is complete I think the PR will be ready for review. I appreciate the initial feedback on PR especially since it touches so many files.

maschad:
Thanks for the first pass @nedsalk , a lot of these were actually bugs hence the compilation issues. I've addressed them, I'll open the PR for review when ready.

maschad:
I think this project would be good to revisit in the future but I don't think it's worth the change yet based on reasons listen in the description.

PR #2235, opened 2024-05-05
chore: dependencies version consistency

### LINKED ISSUE #2028: chore: ensure consistent dependency versions across packages

richardgreg:
Fixes #2028


.github/workflows/pr-lint.yaml
```diff
@@ -126,3 +126,32 @@ jobs:
...
+          fetch-depth: 0
+
+      - name: CI Setup
+        uses: ./.github/actions/ci-setup
+
+      - name: Install Dependencies
+        run: pnpm install
+
+      - name: Install syncpack
+        run: pnpm add --save-dev syncpack
```
danielbate, line 146: Better to have this as a development dependency on the monorepo, so we can version it properly. And we have the config file committed anyways.

danielbate (changes requested):

maschad (changes requested):
Hi @richardgreg I think this is a great start. To finish up:

- you may want to define a syncpack config to avoid mismatches such as `workspace:*`
- Apparently there is a [Github action](https://github.com/marketplace/actions/syncpack-synchronise-monorepo-dependency-versions) which might save you some trouble

After that you can[ fix the mismatches ](https://jamiemason.github.io/syncpack/command/fix-mismatches/) then this should be ready for review

maschad:
Thanks for this work @richardgreg

I must apologize for an oversight on my part as it seems you would need [elevated permissions](https://github.com/JamieMason/syncpack-github-action/tree/main?tab=readme-ov-file#-installation) to be able to run this workflow accurately, which is why you are getting this [error](https://github.com/FuelLabs/fuels-ts/actions/runs/9308025797/job/25620535193?pr=2235#step:4:329)

I will take over the PR from here, merging your changes into one of my branches. Thanks for all of your work!

maschad:
Hi @richardgreg feel free to look through our [good first issues](https://github.com/FuelLabs/fuels-ts/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) and see if there's anything that piques your interest.

Issue #2240, opened 2024-05-06
Exclude Duplicated `Fragment` Suffix


Torres-ssf:
Currently, `@graphql-codegen` generates fragments with an additional `Fragment` suffix. 

This happens because we follow the standard to manually append `Fragment` to the fragment names during creation, to differentiate them from type names.

However, `@graphql-codegen` automatically appends the `Fragment` suffix to all generated fragments. Consequently, this results in the [duplication](https://github.com/FuelLabs/fuels-ts/blob/31f60b7ed7560a0ba97109e5192abc22725b9feb/packages/account/src/providers/provider.ts#L27) of the suffix.

Issue #2239, opened 2024-05-06
Exclude `__typename` from GraphQL Types


Torres-ssf:
Currently, all GraphQL types generated via @graphql-codegen include the `__typename` property by default. 

However, in most cases, this property serves no practical purpose for us. Consequently, we often find ourselves [rearranging code](https://github.com/FuelLabs/fuels-ts/blob/31f60b7ed7560a0ba97109e5192abc22725b9feb/packages/account/src/providers/provider.ts#L100-L105) to eliminate `__typename` from certain types.

PR #2241, opened 2024-05-06
chore!: remove `__typename` from GraphQL types

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/2239
- Close: https://github.com/FuelLabs/fuels-ts/issues/2240

## Breaking
- The `__typename` property was [removed](https://github.com/FuelLabs/fuels-ts/commit/edd80ae39630e9914e1b0f40058e30edd97c2e31) from all GraphQL types
- The property `__typename` was [renamed](https://github.com/FuelLabs/fuels-ts/commit/107f51ad31b4e2c2f6bb65b59c56d76d5a7c4f91) to `type` on `CoinFragment` and `MessageCoinFragment` types 
- All GraphQL fragment-related types were [renamed](https://github.com/FuelLabs/fuels-ts/commit/af1a14d687cddf60409927eda7024e7cf236f0f8) to remove the duplicated `Fragment` suffix.


fuel-service-user:
✨ A PR has been created under the `rc-2241` tag on `fuels-wallet` repo.
https://github.com/FuelLabs/fuels-wallet/pull/1272

packages/account/codegen.json
```diff
@@ -30,7 +30,8 @@
           "UtxoId": "string"
         },
         "useTypeImports": true,
-        "nonOptionalTypename": true,
+        "dedupeOperationSuffix": true,
```
maschad, line 33: are there scenarios in the future where we would want these to be set to `false`?

packages/account/codegen.json
```diff
@@ -30,7 +30,8 @@
...
+        "skipTypename": true,
         "typesPrefix": "Gql"
```
maschad, line 35: Should we remove this as well? seems like it just ends up [becoming unnecessary](https://github.com/FuelLabs/fuels-ts/pull/2241/files#diff-69a6670d0426c74935ed2830ce552e7df2379c2d112a582306e57569fcd17029R98-R103)

packages/account/src/providers/provider.ts
```diff
@@ -21,11 +21,11 @@ import type { Predicate } from '../predicate';
 
 import { getSdk as getOperationsSdk } from './__generated__/operations';
 import type {
-  GqlChainInfoFragmentFragment,
+  GqlChainInfoFragment,
```
maschad, line 24: I made a previous comment which I think should be the better solution but otherwise we should prefer
```typescript
import type {
	GqlFeeParameters as FeeParameters,
	GqlPredicateParameters as PredicateParameters,
	...
```

maschad (changes requested):

packages/account/codegen.json
Torres-ssf: @maschad What flag specifically? `dedupeOperationSuffix` ? 

I don't see a case where we want to use this flag as false since [its purpose](https://the-guild.dev/graphql/codegen/plugins/typescript/typescript-apollo-next#dedupeoperationsuffix) is to remove the duplicated suffix from operation types

packages/account/codegen.json
Torres-ssf: @maschad I would prefer not to since this prefix indicates that this type is related to GraphQL.

Also, we do not use [this approach](https://github.com/FuelLabs/fuels-ts/pull/2241/files#diff-69a6670d0426c74935ed2830ce552e7df2379c2d112a582306e57569fcd17029R98-R103) for every GraphQL type

packages/account/src/providers/provider.ts
Torres-ssf: @maschad https://github.com/FuelLabs/fuels-ts/pull/2241/commits/372fb6fb673209e15aa2230bc334ccfa2f7e034f

packages/account/codegen.json
maschad: I'm referring to both `dedupeOperationSuffix` and `skipTypename` but I just realize that [this flag](https://the-guild.dev/graphql/codegen/plugins/typescript/typescript-operations) is for gql-codegen, otherwise I would have said that these shouldn't be flags but the codegen itself should be modified. Given it's an external lib my comment is no longer relevant though.

maschad (approved):
Great work!

arboleya (approved):

nedsalk (approved):

danielbate (approved):

Issue #2266, opened 2024-05-08
Add Tests to Confirm Changes in `Account.fund` Method


Torres-ssf:
During https://github.com/FuelLabs/fuels-ts/pull/2128 the `Account.fund` method was [modified](https://github.com/FuelLabs/fuels-ts/pull/2128/files#diff-2b61add1a6638c810f012c577ce094c87ca7685c03bb7f9c71b3260cf11fa875L272-L273) to avoid unnecessary validation of whether the resources already present in the `TransactionRequest` belong to the wallet owner. 

Prior to this change, the `fund` method would attempt to add funds even if the `TransactionRequest` already contained sufficient resources, as long as these resources didn't belong to the wallet owner. Now, the method will only add funds if any are missing.

However, no tests have been implemented to verify this behavior. We need to add tests to confirm that the `fund` method functions correctly after this modification.

PR #2267, opened 2024-05-08
chore: add tests for updated `Account.fund` logic

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/2266


packages/fuel-gauge/src/funding-transaction.test.ts
```diff
@@ -232,4 +236,109 @@ describe(__filename, () => {
...
+  it('should ensure a partially funded Transaction will require only missing funds', async () => {
+    const receiver = Wallet.generate({ provider });
+    const wallet1 = Wallet.generate({ provider });
+    const wallet2 = Wallet.generate({ provider });
+
+    const totalInBaseAsset = 15_000;
+    const totalInAssetA = 20_000;
+    const partiallyInAssetA = totalInAssetA / 2;
+
+    // wallet1 does not have enough funds to cover total required in assetA and
```
maschad, line 249: Not understanding this comment

packages/fuel-gauge/src/funding-transaction.test.ts
```diff
@@ -232,4 +236,109 @@ describe(__filename, () => {
...
+    ]);
+
+    const baseAssetResource = partiallyResources.find((r) => r.assetId === baseAssetId);
+    const assetAResource = partiallyResources.find((r) => r.assetId === assetA);
+
+    // expect to have the correct amount of resources, not enough to cover the TX needs
+    expect(baseAssetResource?.amount.toString()).toBe(totalInBaseAsset.toString());
+    expect(assetAResource?.amount.toString()).toBe(partiallyInAssetA.toString());
+
+    transactionRequest.addResources(partiallyResources);
```
maschad, line 282: Maybe I am misinterpreting the issue but shouldn't this test essentially ensure that the `fund` method is only called  when resources if they are missing? Or perhaps the method should reject an attempt to add resources if they are present?

maschad (commented):
It's difficult to evaluate what is being tested here

packages/fuel-gauge/src/funding-transaction.test.ts
Torres-ssf: @maschad is it more clear now? https://github.com/FuelLabs/fuels-ts/pull/2267/commits/bb25beffc6021b04462596c4d88ccad695b6c998

packages/fuel-gauge/src/funding-transaction.test.ts
Torres-ssf: @maschad No, the issue is to validate that the `Account.fund` method will not try to fetch and add more funds to a TX that is already funded.

maschad (approved):
Thanks for the clarification, nice work!

danielbate (approved):

PR #2268, opened 2024-05-08
fix: flaky `TransactionResponse` test

### LINKED ISSUE #2200: CI: Inconsistent `TransactionResponse` tests
[Post-upgrade](9c3c094d810b8400e8c0b9ac2830f75ca2ca70b5) I've noticed a sharp increase in [CI failures](https://github.com/FuelLabs/fuels-ts/actions/runs/8880413928/job/24380375347?pr=2199) related to the `transaction-response-test`

```rust
FAIL  packages/fuel-gauge/src/transaction-response.test.ts > TransactionResponse > should throw error for a SqueezedOut status update [submitAndAwait]
Error: Passed-in lambda didn't throw.
 ❯ Module.expectToThrowFuelError packages/errors/src/test-utils/safeExec.ts:101:11
     12| 
     13|   return { error, result };
     14| };
       |             ^
     15| 
 ❯ packages/fuel-gauge/src/transaction-response.test.ts:280:5
 ```

### PR DESCRIPTION
**Pull request by nedsalk**
I retried the test 100s of times locally and could not get it to fail. My hunch is that it's because of the CI environment being slow. 

I increased the `--tx-pool-ttl` to 1 second to allow for some more room before the transaction is squeezed out and `--poa-interval-period` to 4s to ensure that the transaction doesn't get processed before it's squeezed out. Still, these are just informed shots in the dark; I don't know the underlying root cause.

Due to the flakiness of the test and me not knowing the root cause, I also added `{ retry: 10 }` to it so that it's retried if it fails. This should at least reduce the flakiness _a lot_ if the changes to the flag values aren't enough.

Also note that this _is_ the way to test the squeezing out of a transaction and I was given guidance by the `fuel-core` team on how to write it.

closes #2200

### CODE REVIEW

File: packages/fuel-gauge/src/transaction-response.test.ts [hunk 1]
```diff
@@ -247,42 +246,46 @@ describe('TransactionResponse', () => {
...
+
+      await expectToThrowFuelError(
+        async () => {
+          await nodeProvider.sendTransaction(request, { awaitExecution: true });
+        },
+        { code: ErrorCode.TRANSACTION_SQUEEZED_OUT }
+      );
+      cleanup();
+    },
+    { retry: 10 }
```
Torres-ssf, line 289: So the only way to avoid this test being flaky is to retry it?

Torres-ssf (approved):

packages/fuel-gauge/src/transaction-response.test.ts
nedsalk: I don't know the underlying reason for its flakiness so, afaik, yes. The good thing is that if the test is failing due to valid reasons then it'll fail after the retries anyways.

maschad (approved):

packages/fuel-gauge/src/transaction-response.test.ts
arboleya: It looks like a temporary solution that could hide the real [bigger?] issue.

It could have been valid to have an explanation and a TODO for future review.

PR #2262, opened 2024-05-08
fix: avoid overriding user `gasLimit` and `maxFee` inputs

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/2257
- Close: https://github.com/FuelLabs/fuels-ts/issues/2243

PR #2271, opened 2024-05-08
build: ensure whitespace is trimmed for vscode formatting

maschad:
See https://github.com/FuelLabs/fuels-ts/pull/2262#discussion_r1594349239


.vscode/settings.json
```diff
@@ -12,6 +12,7 @@
     "**/.turbo": true,
     "**/dist": true
   },
+  "files.trimTrailingWhitespace": true,
```
arboleya, line 15: This shouldn't be necessary, considering we already have this:

https://github.com/FuelLabs/fuels-ts/blob/7b570d965fe3bdabf39d80624b37a32709f7dae8/.vscode/settings.json#L4-L7

maschad: But we don't have the [trailing spaces](https://eslint.org/docs/latest/rules/no-trailing-spaces) rule enabled. I've added it in 84188e9fc

danielbate (approved):

.vscode/settings.json
arboleya: I may have presumed it wrong, but we never have whitespace problems. Or rarely.

Could Prettier be the one trimming them off?

https://github.com/FuelLabs/fuels-ts/blob/7b570d965fe3bdabf39d80624b37a32709f7dae8/.vscode/settings.json#L16-L18

My only point against a VSCode-exclusive solution is that it will only work within VSCode.

The goal should be something like ESLint/Prettier, as it encompasses CI, Neovim (and others), etc. 

I'm curious to hear back from @nedsalk, as we had other Linux lint issues in the past with other team members.

maschad: I agree with you @arboleya but I'm not sure of your objection, I've added the [eslint rule](https://github.com/FuelLabs/fuels-ts/pull/2271/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R30) to cover non-vscode users as well.

nedsalk: Ideally the `.vscode` directory should be in `.gitignore` and we'd merge only the eslint rule @maschad added in this PR. Then, whoever has troubles with it conflicting with VSCode - it's their problem.

arboleya: @maschad I wrongly presumed ESlint was removing whitespaces so far, but since you pointed out that we didn't even have the rule enabled, that hypothesis is discarded. Thanks for adding the rule. Next in line is Prettier. Was it doing this job then and not ESLint? If yes, the same question remains—how did whitespaces still pass through?

As I said, I suspect we may have a misconfiguration affecting Linux users. My objective is to understand the root cause, and enabling multiple strategies simultaneously can make it challenging to pinpoint the reason.

@nedsalk Did you understand how whitespace slipped through your commits?

nedsalk: @arboleya I might have just broken off the sentences at the white spaces and forgot to remove them and it might be the case that @Torres-ssf has a local configuration that trims whitespaces whereas I don't and both versions are accepted by our linting rules. It couldn't have slipped through if our linting rules didn't allow it and that's why I'd just merge the eslint rule and remove/ignore the `.vscode` directory in our repo.

arboleya: I don't think it is @Torres-ssf specific configs since everyone but you seems to be in the same boat.

The only times I've seen these slips happening were with Linux users.

While I don't like being VSCode-specific for certain things, I'd not dismiss the fact that all of you use it.

arboleya: Validation would be simple: Open file, add trailing whitespace, save, git diff.

danielbate: IMO eslint/biome rules are source of truth. It'll remain that way as it runs in CI.

Then surely any IDE specific config we add doesn't hurt _that much_. We also recommend vscode plugins.

Torres-ssf (approved):

PR #2275, opened 2024-05-09
chore: treat minor changes as breaking

### LINKED ISSUE #2244: Add new workflow to validate version bumps
Before we hit stable v1, our semver requires a slight shift:

 - `major` — never (for now)
 - `minor` — when we introduce a breaking change
 - `patch` — everything else (non-breaking)

This has been a source of misunderstanding on several PRs, something a simple workflow could automate.

One way to do this is to _only_ allow `minor` bumps when the PR has breaking changes (look for `!` in the title).

nedsalk:
closes #2244


.github/workflows/pr-lint.yaml
```diff
@@ -93,6 +93,14 @@ jobs:
             exit 0
           fi
 
+          # Minor changes are treated as breaking until v1.0.0 is released.
+          if echo "$AFFECTED_PACKAGES" | grep -q 'minor'; then
+            if ! echo "$PR_TITLE" | grep -q '!'; then
+              echo "Changeset has minor changes. Please mark the PR as breaking by adding an exclamation mark to its title (e.g. fix!: xxx)."
```
arboleya, line 99: Nit.

```suggestion
              echo "Changeset has `minor` changes. Please mark the PR as breaking by adding an exclamation mark to its title (e.g. fix!: xxx) or use a `patch` instead."
```

arboleya (dismissed):

arboleya (dismissed):

Torres-ssf (dismissed):

.github/workflows/pr-lint.yaml
```diff
@@ -93,6 +93,14 @@ jobs:
             exit 0
           fi
```
maschad, line 95:
```suggestion
#TODO: Remove once v1.0.0 is released.
```

maschad (changes requested):
Looks good, just want to include a TODO so it doesn't get lost in the ether... 😉

danielbate (dismissed):

arboleya (approved):

PR #2276, opened 2024-05-09
chore: cleanup typedoc for `Account` and `Wallet`

### LINKED ISSUE #2249: Review outdated Typedocs
Some methods on the `Provider` and `Wallet` classes seem to have outdated Typedocs attached.

Of course, automating validations like this would be awesome, but given that this sort of thing can become a deep rabbit hole, let's consider the possibility of making manual revisions that could more simply target and mitigate the perceived confusion.

> Reported by @LuizAsFight

danielbate:
Closes #2249


packages/account/src/account.ts
```diff
@@ -117,12 +121,12 @@ export class Account extends AbstractAccount {
   /**
    * Retrieves resources satisfying the spend query for the account.
    *
-   * @param quantities - IDs of coins to exclude.
+   * @param quantities - IDs of coins to obtains.
```
maschad, line 124:
```suggestion
   * @param quantities - IDs of coins to be obtained.
```

maschad (changes requested):
Spelling correction but looks good otherwise.

packages/account/src/account.ts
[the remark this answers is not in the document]
danielbate: Which line?

maschad (approved):

Torres-ssf (approved):

arboleya (dismissed):

arboleya:
@danielbate :shipit: ?

packages/account/src/providers/provider.ts
```diff
@@ -1058,7 +1104,8 @@ Supported fuel-core version: ${supportedVersion}.`
    * transaction. The default value is 0.2 or 20%
    *
    * @param transactionRequestLike - The transaction request object.
-   * @param tolerance - The tolerance to add on top of the gasUsed.
+   * @param transactionCostParams - The transaction cost parameters (optional).
+   *
```
Torres-ssf, line 1108: Since tolerance was removed we should also remove:

```
   * @privateRemarks
   * The tolerance is add on top of the gasUsed calculated
   * from the node, this create a safe margin costs like
   * change states on transfer that don't occur on the dryRun
   * transaction. The default value is 0.2 or 20%
```

Torres-ssf (changes requested):
Great work 👍

packages/account/src/providers/provider.ts
danielbate: https://github.com/FuelLabs/fuels-ts/pull/2276/commits/5955ab1ebd70f04d9fd1661ee4199d2236c5970c

Torres-ssf (approved):

Issue #2170, opened 2024-04-25
Create integration test for `create fuels`


arboleya:
# Summary

The idea is that every PR would publish a PR tag and have an install/assertion routine.

The automation would probably need to drive the UI and assert the results.

Below are the steps of how one could do it _manually_:

# Manual Steps

1) Ensure you have the correct versions of `forc` and `fuel-core`:

```shell
 fuelup component add forc@0.55.0
 fuelup component add fuel-core@0.24.3
```

2) Scaffold a new project using the `next` tag:

```shell
pnpm create fuels@next
```

3) Make the scaffolded app to use the `next` tag:

```shell
cd ./my-app-name # <—— add app name here
pnpm add fuels@next
```

4) Run `fuels dev` in terminal 1
```shell
pnpm fuels:dev
```

5) Run `next dev` in terminal 2
```shell
pnpm dev
```

# Final Notes

This flow will test all CLIs (below) at once and also the SDK's basic functionality:
 1. `create-fuels`
 1. `fuels`
 1. `fuels-typegen`

arboleya:
This would also satisfy all requirements of this older issue:
 - https://github.com/FuelLabs/fuels-ts/issues/568

Issue #2048, opened 2024-04-12
Top up button for `create-fuels` doesn't refresh balance


petertonysmith94:
## Summary

When clicking the `Top-up Wallet` button, the balance doesn't refresh.

### Expected

The current balance to be updated with the new balance.

### Actual

The balance does not change.

Dhaiwat10:
This used to work fine but now this seems to be a theme in almost all interactions where the balance needs to be re-fetched. It is indeed re-fetched but for some reason the update isn't coming in time. The query is almost too 'fast'. This should absolutely be investigated

PR #2278, opened 2024-05-09
chore: add integration test for the `create fuels` template

### LINKED ISSUE #2264: Add e2e tests for the create fuels template app
> As a part of #2170 

Currently, we have unit tests for the create fuels CLI tool itself but we don't have any automated testing suites for the UI of the template app. We should have some tests that

1. Run a new local node
2. Deploy contracts to this node
3. Load the template UI
4. Click the 'increment counter' button
5. Assert that the counter was incremented

Dhaiwat10:
## Summary
This PR adds an e2e test for the `create fuels` template. This test simulates the user opening the page, topping up their burner wallet, clicking 'increment', and verifies that the counter is incremented as expected.

While this PR only adds a test for the counter contract, we should add tests for the predicate and script examples in the template as well - maybe in separate PRs.

Closes #2264


e2e-tests/create-fuels.test.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  const topUpWalletButton = page.getByText('Top-up Wallet');
+  await topUpWalletButton.click();
+
+  const welcomeToFuelText = page.getByText('Welcome to Fuel');
+  await expect(welcomeToFuelText).toBeVisible();
+
+  await page.waitForTimeout(2000);
+
+  await page.reload();
```
Dhaiwat10, line 16: Unfortunately we need to reload the page after topping up the wallet because for some reason the balance isn't being updated properly as mentioned in #2048

e2e-tests/create-fuels.test.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  await page.waitForTimeout(2000);
```
Dhaiwat10, line 18: All of the`waitForTimeout`s are waiting for read or write calls to the node.

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          node-version: ${{ matrix.env.version || 20 }}
+      - name: Run Fuel Node
+        run: (cd templates/nextjs && pnpm run fuels:dev) &
+      - name: Copy .env.example and create .env.local off it
+        run: cp templates/nextjs/.env.example templates/nextjs/.env.local
+      - name: Run Next.js server
+        run: (cd templates/nextjs && pnpm run dev) &
+      - name: Install Playwright Browsers
+        run: pnpm exec playwright install --with-deps
```
danielbate, line 24: Whats this doing? Can it not just be a dependency?

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build
+        uses: ./.github/actions/test-setup
+        with:
+          node-version: ${{ matrix.env.version || 20 }}
+      - name: Run Fuel Node
+        run: (cd templates/nextjs && pnpm run fuels:dev) &
+      - name: Copy .env.example and create .env.local off it
+        run: cp templates/nextjs/.env.example templates/nextjs/.env.local
```
danielbate, line 20: Add a `.env.test` instead?

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Run Next.js server
+        run: (cd templates/nextjs && pnpm run dev) &
+      - name: Install Playwright Browsers
+        run: pnpm exec playwright install --with-deps
+      - name: Run Playwright tests
+        run: pnpm exec playwright test
+      - uses: actions/upload-artifact@v4
```
danielbate, line 27: What are we using the uploaded artifact for here?

danielbate (commented):
Great work Dhai, couple questions

e2e-tests/create-fuels.test.ts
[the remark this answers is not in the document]
Dhaiwat10: @petertonysmith94 that's a good point but I'm not sure if we want to add more things into our template which I feel already has quite a few things for a 'getting started' template

.github/workflows/playwright.yml
Dhaiwat10: @danielbate unfortunately not. When you install playwright, it does not install the browsers along with it

.github/workflows/playwright.yml
Dhaiwat10: I could but there is nothing quite different about the test env and the env people would usually use, so I don't think this change is needed

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,32 @@
...
+jobs:
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build
+        uses: ./.github/actions/test-setup
+        with:
+          node-version: ${{ matrix.env.version || 20 }}
```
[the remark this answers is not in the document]
Dhaiwat10: Oh I see, should it just be:

```yml
- name: Build
        uses: ./.github/actions/test-setup
        with:
          node-version: 20
```

instead?

.github/workflows/playwright.yml
Dhaiwat10: That's my bad. It's not needed. I forgot to remove this from the snippet I copied from the playwright docs

.changeset/great-teachers-yell.md
```diff
@@ -0,0 +1,4 @@
+---
+---
+
+feat: add e2e test for the `create fuels` template
```
maschad, line 4: I'm not sure if this is really a `feat` considering it's a test.

```suggestion
chore: add e2e test for the `create fuels` template
```

maschad (changes requested):
Great work on this 🚀  Left a suggestion

e2e-tests/create-fuels.test.ts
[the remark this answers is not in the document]
Dhaiwat10: @petertonysmith94 Not sure if it's worth the hassle because if we do that then we'll have to remove this folder when we ship the template.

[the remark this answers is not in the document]
Dhaiwat10: and tbh the e2e test does kind of encompass the whole SDK in terms of the functionality that it covers and tests.

.github/workflows/test.yaml
```diff
@@ -72,6 +72,21 @@ jobs:
           TEST_WALLET_ADDRESS: ${{ secrets.TEST_WALLET_ADDRESS }}
           FUEL_TESTNET_NETWORK_URL: ${{ secrets.FUEL_TESTNET_NETWORK_URL }}
 
+  create-fuels-e2e:
```
maschad, line 75: Also I suppose @arboleya may have to do this, but we would want this to be a required workflow on the repo as well correct?

maschad (approved):

.github/workflows/test.yaml
```diff
@@ -72,6 +72,21 @@ jobs:
...
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
```
nedsalk, line 77:
```suggestion
    needs: [environments]
    if: ${{ !github.event.pull_request.draft }}
    runs-on: ubuntu-latest
```

nedsalk: After analyzing the code I realized that we're running these tests against a local node so it's not needed.

maschad (approved):
Awesome job on this!

danielbate (approved):

Torres-ssf (approved):

danielbate (commented):
Was meant to request changes.

danielbate (changes requested):
Why is this e2e, as it is running on a local node? 🤔

Dhaiwat10:
@danielbate I have named it e2e in reference of the UI. Instead of testing isolated parts of the UI, we are testing the whole 'flow' in the UI from start to finish.

nedsalk (dismissed):

.github/workflows/test.yaml
arboleya: @Dhaiwat10 I can do it right after this goes to `master`.

arboleya (commented):
Good stuff!

I think I agree with @danielbate regarding it not being an e2e but an integration.

At least terminologically, we have been thinking about e2e using an external node connection.

maschad (approved):
Great job on this!

playwright-tests/create-fuels.test.ts
```diff
@@ -0,0 +1,27 @@
...
+  await page.reload();
+
+  await page.waitForTimeout(2000);
+
+  const incrementButton = page.getByText('Increment Counter');
+  await incrementButton.click();
+
+  await page.waitForTimeout(2000);
+
+  const counter = page.getByText('1', { exact: true });
```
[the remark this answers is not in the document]
Dhaiwat10: This seems to have done the trick

maschad (approved):

scripts/create-fuels-template-integration.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+(cd templates/nextjs && pnpm run fuels:dev) &
+cp templates/nextjs/.env.example templates/nextjs/.env.local
+(cd templates/nextjs && pnpm run dev) &
+pnpm exec playwright install --with-deps
+pnpm exec playwright test
```
arboleya, line 7: I may be a bit out of context, so here's a couple questions:
 1. Have you successfully run this on CI?
 1. Should we be concerned about shutting down the `dev` processes to avoid blocking?
 1. Should we run the validations using the published version from NPM? (cc @petertonysmith94)
    - Example: [packages/create-fuels/test/e2e.test.ts](https://github.com/FuelLabs/fuels-ts/pull/2364/files#diff-adad36ad07b62532f988d67772fa9eac2fa210d765683f38b6a78c6d6965ab9a)

Am I missing something?

Dhaiwat10: Merging this for now. If anything is required, we'll take care of it in #2170

PR #2279, opened 2024-05-09
fix: added CDN path for icon assets

### LINKED ISSUE #1893: `iconPath`s for assets not being resolved on build
Unlike the original `assets` package:

https://github.com/FuelLabs/fuels-npm-packs/blob/09d21738540674531df678d17f40a00ba6db36c3/packages/assets/src/index.ts#L10

We are not resolving the icon paths for assets when we export them. We chose not to do this because we did not want default exports (if I remember correctly), but this has resulted in a missing feature/behaviour as well. 

___

Incoming from a Slack message by @helciofranco:

> assets is not returning imgUrl with the complete URL as it was with @fuels/assets.
previously it had https://cdn.fuel.network/assets/ prefix for each imgUrl
currently it has only the filename (eth.svg for example).

petertonysmith94:
Closes #1893

### DIFF
--- a/.github/workflows/pr-release.yaml
+++ b/.github/workflows/pr-release.yaml
@@ -8,7 +8,7 @@ jobs:
     name: "Release PR to npm"
     runs-on: ubuntu-latest
     # comment out if:false to enable release PR to npm
-    # if: false
+    if: false
     permissions: write-all
     steps:
       - name: Checkout


maschad (approved):
Nice!

danielbate (approved):

packages/account/src/providers/assets/assets.ts
```diff
@@ -0,0 +1,38 @@
...
+        type: 'fuel',
+        chainId: CHAIN_IDS.fuel.beta5,
+        decimals: 9,
+        assetId: '0x0000000000000000000000000000000000000000000000000000000000000000',
+      },
+      {
+        type: 'fuel',
+        chainId: CHAIN_IDS.fuel.devnet,
+        decimals: 9,
+        assetId: '0x0000000000000000000000000000000000000000000000000000000000000000',
```
arboleya, line 32: Our `BaseAssetId` is dynamic now.

PR #2282, opened 2024-05-10
chore: simplify E2E test

danielbate:
The purpose of the E2E test is to ensure compatibility with the current live testnet. IMO the real value here is testing that the transaction format is valid, not the encoding format of the call data. If that was the case, we should be testing for all types, rather than the struct vec setup we currently have. 

Therefore, as we are currently facing issues with running out of funds on our test address, we should simplify this test to minimise it's fund consumption.

This PR reduces gas consumption of the E2E test from `51326` to `1226`.


nedsalk:
> [...] IMO the real value here is testing that the transaction format is valid...

I agree. Should we then have (simple) e2e tests involving all transaction types with all types of programs? It can be a follow-up PR but can be done in this PR as well, afaic.

danielbate:
@nedsalk agreed however more txs will mean more gas, and part of the reason here is to reduce gas consumption for the e2e wallet 😬  Will do as follow up once we've resolved the wallet draining with the bridge or something.

nedsalk:
@danielbate yeah, but it'll still be significantly less than what we spend now given that the tests will be simple. We'll get more tests while spending 5-10x less gas in the end.

nedsalk (dismissed):

Torres-ssf (approved):

maschad (approved):

PR #2302, opened 2024-05-14
fix: ignore libraries in `fuels` CLI

### LINKED ISSUE #2300: `fuels` will throw for libraries
Currently, `fuels` parses program types, checks them and pushes them to program type member arrays.

https://github.com/FuelLabs/fuels-ts/blob/81f62a0bb394331767ee2e3298772d53bcc4e06b/packages/fuels/src/cli/config/loadConfig.ts#L96-L101

As it is checking for a key in `config`, if the key doesn't exist it shall throw:

```sh
TypeError: Cannot read properties of undefined (reading 'push')
```

We should validate the program type before pushing.

danielbate:
Closes #2300


arboleya (approved):
Nice catch! 🙌

maschad (approved):

Torres-ssf (approved):

PR #2308, opened 2024-05-15
fix!: remove offset from `decodeLog` and `decodeFunctionResult` methods

DZakh:
It feels like public `decode` functions should return the decoded value itself. Not a tuple with an internal offset.

### DIFF
--- a/.changeset/perfect-kings-camp.md
+++ b/.changeset/perfect-kings-camp.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/abi-coder": minor
+---
+
+Changed decodeLog and decodeFunctionResult return types to have only decoded value without an offset
--- a/.changeset/perfect-kings-camp.md
+++ b/.changeset/perfect-kings-camp.md
@@ -2,4 +2,4 @@
 "@fuel-ts/abi-coder": minor
 ---
 
-Changed decodeLog and decodeFunctionResult return types to have only decoded value without an offset
+fix: Changed decodeLog and decodeFunctionResult return types to have only decoded value without an offset
--- a/packages/abi-coder/src/Interface.ts
+++ b/packages/abi-coder/src/Interface.ts
@@ -75,7 +75,7 @@ export class Interface<TAbi extends JsonAbi = JsonAbi> {
   }
 
   decodeLog(data: BytesLike, logId: string): any {
-    const loggedType = this.jsonAbi.loggedTypes.find((type) => type.logId === logId);
+    const loggedType = this.jsonAbi.loggedTypes.find((type) => type.logId.toString() === logId);
     if (!loggedType) {
       throw new FuelError(
         ErrorCode.LOG_TYPE_NOT_FOUND,
--- a/packages/abi-coder/src/Interface.ts
+++ b/packages/abi-coder/src/Interface.ts
@@ -71,7 +71,7 @@ export class Interface<TAbi extends JsonAbi = JsonAbi> {
     const fragment =
       typeof functionFragment === 'string' ? this.getFunction(functionFragment) : functionFragment;
 
-    return fragment.decodeOutput(data);
+    return fragment.decodeOutput(data)[0];
   }
 
   decodeLog(data: BytesLike, logId: string): any {
@@ -85,7 +85,7 @@ export class Interface<TAbi extends JsonAbi = JsonAbi> {
 
     return AbiCoder.decode(this.jsonAbi, loggedType.loggedType, arrayify(data), 0, {
       encoding: this.encoding,
-    });
+    })[0];
   }
 
   encodeConfigurable(name: string, value: InputValue) {
--- a/packages/abi-coder/src/Interface.ts
+++ b/packages/abi-coder/src/Interface.ts
@@ -45,7 +45,7 @@ export class Interface<TAbi extends JsonAbi = JsonAbi> {
 
     throw new FuelError(
       ErrorCode.FUNCTION_NOT_FOUND,
-      `function ${nameOrSignatureOrSelector} not found: ${JSON.stringify(fn)}.`
+      `Function ${nameOrSignatureOrSelector} not found.`
     );
   }
 
--- a/packages/abi-coder/src/Interface.ts
+++ b/packages/abi-coder/src/Interface.ts
@@ -75,7 +75,7 @@ export class Interface<TAbi extends JsonAbi = JsonAbi> {
   }
 
   decodeLog(data: BytesLike, logId: string): any {
-    const loggedType = this.jsonAbi.loggedTypes.find((type) => type.logId.toString() === logId);
+    const loggedType = this.jsonAbi.loggedTypes.find((type) => type.logId === logId);
     if (!loggedType) {
       throw new FuelError(
         ErrorCode.LOG_TYPE_NOT_FOUND,
--- a/packages/abi-coder/test/Interface.test.ts
+++ b/packages/abi-coder/test/Interface.test.ts
@@ -1,6 +1,8 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
+import { ErrorCode, FuelError } from '@fuel-ts/errors';
+import { expectToThrowFuelError } from '@fuel-ts/errors/test-utils';
 import type { BN } from '@fuel-ts/math';
-import { concat } from '@fuel-ts/utils';
+import { concat, hexlify } from '@fuel-ts/utils';
 
 import { Interface } from '../src';
 /** @knipignore */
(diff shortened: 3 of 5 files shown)


DZakh:
I don't have a strong feeling about the change, and it's not required for my project.
It was mostly done after the following reasoning:
- The offset is not needed as a result of a high-level function such as `decodeLog`
- The `decodeFunctionData` doesn't return an offset

If you say it's intentional, it would be nice to change the return type from `any` to a tuple with an offset as a second argument. Because it took me some time to understand what was wrong with the decoded data.

danielbate:
Hey @DZakh, after investigation I do agree with you and I think it was oversight that these entry points for function and log value decoding would include the offset. Most SDK consumers will utilise the `TransactionResponse` or `FunctionInvocationResult` classes that we return on `sendTransaction()` or `call()` that internally will pick off the decoded values rather than returning the tuple.

To get this over the line, it'd be great if you can add a changeset (run `pnpm changeset` on the monorepo and it would also be great to get some tests so this is not missed in the future, I would reccomend adding them to `packages/abi-coder/test/Interface.test.ts`. Let me know if I can help with either of these.

danielbate (commented):
Hey @DZakh, after investigation I do agree with you and I think it was oversight that these entry points for function and log value decoding would include the offset. Most SDK consumers will utilise the TransactionResponse or FunctionInvocationResult classes that we return on `sendTransaction()` or `call()` that internally will pick off the decoded values rather than returning the tuple.

To get this over the line, it'd be great if you can add a changeset (run pnpm changeset on the monorepo and it would also be great to get some tests so this is not missed in the future, I would reccomend adding them to packages/abi-coder/test/Interface.test.ts. Let me know if I can help with either of these.

DZakh:
> Hey @DZakh, after investigation I do agree with you and I think it was oversight that these entry points for function and log value decoding would include the offset. Most SDK consumers will utilise the TransactionResponse or FunctionInvocationResult classes that we return on sendTransaction() or call() that internally will pick off the decoded values rather than returning the tuple.
> 
> To get this over the line, it'd be great if you can add a changeset (run pnpm changeset on the monorepo and it would also be great to get some tests so this is not missed in the future, I would reccomend adding them to packages/abi-coder/test/Interface.test.ts. Let me know if I can help with either of these.

I'll try to find some time during the week 👌

packages/abi-coder/src/Interface.ts
```diff
@@ -71,11 +71,11 @@ export class Interface<TAbi extends JsonAbi = JsonAbi> {
     const fragment =
       typeof functionFragment === 'string' ? this.getFunction(functionFragment) : functionFragment;
 
-    return fragment.decodeOutput(data);
+    return fragment.decodeOutput(data)[0];
   }
 
   decodeLog(data: BytesLike, logId: string): any {
-    const loggedType = this.jsonAbi.loggedTypes.find((type) => type.logId === logId);
+    const loggedType = this.jsonAbi.loggedTypes.find((type) => type.logId.toString() === logId);
```
DZakh, line 78: There's a bug in latest version after `logId` was changed to `string`.
Will `rb` become a `string` type? Why was the change needed?

packages/abi-coder/test/Interface.test.ts
```diff
@@ -766,4 +766,20 @@ describe('Abi interface', () => {
       );
     });
   });
+
+  describe('decodeLog', () => {
+    it('should return decoded log by id', () => {
+      const data = exhaustiveExamplesInterface.decodeLog('0x01000000000000000000000000000020', '0');
+      expect(data).toEqual({
+        a: true,
+        b: 32,
```
DZakh, line 775: The test is failing. The decoded field value is `0`. Probably something is broken.

DZakh (commented):
Could you take a look, please

packages/abi-coder/src/Interface.ts
```diff
@@ -45,7 +45,7 @@ export class Interface<TAbi extends JsonAbi = JsonAbi> {
 
     throw new FuelError(
       ErrorCode.FUNCTION_NOT_FOUND,
-      `function ${nameOrSignatureOrSelector} not found: ${JSON.stringify(fn)}.`
+      `Function ${nameOrSignatureOrSelector} not found.`
```
DZakh, line 48: `fn` is undefined here

packages/abi-coder/src/Interface.ts
danielbate: cc @nedsalk

nedsalk: @DZakh the change is part of work on [#2313](https://github.com/FuelLabs/fuels-ts/issues/2313), specifically tied to https://github.com/FuelLabs/sway/issues/5953.

What is the bug you're referring to? Could you give me an example?

packages/abi-coder/test/Interface.test.ts
```diff
@@ -766,4 +766,42 @@ describe('Abi interface', () => {
       );
     });
   });
+
+  describe('decodeLog', () => {
+    it('should return decoded log by id', () => {
+      const data = exhaustiveExamplesInterface.decodeLog('0x01000000000000000000000000000020', '0');
```
nedsalk, line 772: The `logId` passed here should be `"8500535089865083573"` when the sway program is built with `forc 0.59.0`.

packages/abi-coder/src/Interface.ts
nedsalk: I think that you might have tested this solution out with a `forc` version that's not `0.59.0`. After you build with it, you won't need this `type.logId.toString()` change because the `logId`s are strings.

packages/abi-coder/test/Interface.test.ts
```diff
@@ -766,4 +766,42 @@ describe('Abi interface', () => {
...
+      expect(data).toEqual({
+        a: true,
+        b: 32,
+      });
+    });
+
+    it('should throw an error when log does not exist', () => {
+      expect(() =>
+        exhaustiveExamplesInterface.decodeLog('0x01000000000000000000000000000020', '1')
+      ).toThrowError(`Log type with logId '1' doesn't exist in the ABI.`);
```
nedsalk, line 782: You should use the `expectToThrowFuelError` utility here ([example](https://github.com/FuelLabs/fuels-ts/blob/ba0f3d0f7001cd1f261d658bec0a356a1a40a4ee/packages/abi-coder/src/encoding/coders/ArrayCoder.test.ts#L93)).

packages/abi-coder/test/Interface.test.ts
```diff
@@ -766,4 +766,42 @@ describe('Abi interface', () => {
...
+      });
+    });
+
+    it('should throw an error when function does not exist', () => {
+      expect(() => {
+        exhaustiveExamplesInterface.decodeFunctionResult(
+          'doesnt_exist',
+          '0x01000000000000000000000000000020'
+        );
+      }).toThrowError(/^Function doesnt_exist not found\.$/);
```
nedsalk, line 804: You should use the `expectToThrowFuelError` utility here ([example](https://github.com/FuelLabs/fuels-ts/blob/ba0f3d0f7001cd1f261d658bec0a356a1a40a4ee/packages/abi-coder/src/encoding/coders/ArrayCoder.test.ts#L93)).

.changeset/perfect-kings-camp.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/abi-coder": minor
+---
+
+Changed decodeLog and decodeFunctionResult return types to have only decoded value without an offset
```
nedsalk, line 5:
```suggestion
fix!: don't return offset from public decode functions
```

We have some CI logic that mandates that this text is equal to the pr title.

nedsalk:
@DZakh Thanks for the work! I agree that the offset shouldn't be returned.

packages/abi-coder/src/Interface.ts
DZakh: That's probably the case

DZakh:
I've just returned from a one-week vacation. I didn't touch my laptop at all, and now I'm a little bit lost on what's left for the PR 😅 Could somebody summarize it for me, please?

danielbate:
@DZakh I am doing some cleanup of `abi-coder` and altering leftover logic from the `v1` migration, if it's easier I can include that there, and we can close this PR?

nedsalk:
@DZakh Thanks for the work here, I've also assigned myself to the PR and will take this over the line.

DZakh:
> @DZakh Thanks for the work here, I've also assigned myself to the PR and will take this over the line.

Thanks

nedsalk:
@DZakh I don't have the rights to push to your fork. You can find my changes on [this branch](https://github.com/FuelLabs/fuels-ts/tree/ns/fix/remove-offset-from-decode-functions). Please merge them into your fork so that we can continue with the review.

DZakh:
@nedsalk, I don't mind if we continue with your PR. Let's just close this one and reference yours.

nedsalk:
Superseded by #2514

PR #2309, opened 2024-05-15
feature: Add decodeTypeApplication

DZakh:


DZakh:
I was pointed to the `ABICoder.decode()`, which is exactly what I wanted. 

https://forum.fuel.network/t/abi-coders-request-for-the-decodetypeapplication-method/5298

PR #2143, opened 2024-04-23
chore!: remove built-in binaries for `forc` and `fuel-core`

petertonysmith94:
Closes #2084
Closes #2277

---

### Summary

- Internalises the `forc` and `fuel-core` binaries (removed usage for end-consumer).
- Adds the ability to pass in custom binary paths via `forcPath` and `fuelCorePath` to the `fuels.config.ts`.
  - Defaults to `forc` and `fuel-core`
- Binary not found and version mismatches errors for `fuels` CLI.
- Suppressed the following related logs (experienced in the CI):
```console
/bin/sh: 1: fuel-core: not found
/bin/sh: 1: forc: not found
```

### Breaking change

1. Built-in binaries have been removed, so the following means of running `forc` or `fuel-core` are not available. 
```console
# Before
npx fuels forc
npx fuels-forc
# AND
npx fuels core
npx fuels-core
```

Suggested to [install](https://docs.fuel.network/guides/installation/) `fuelup` and associated binaries (`forc` and `fuel-core`).
```console
# After
forc
fuel-core
```

2. `fuels.config` - removed options `useBuiltinForc` and `useBuiltinFuelCore`

3. `launchNode` - removed option for `useSystemFuelCore`

---
<details>
  <summary>Scenarios</summary>
Both instances will stop the process until the issue/s are resolved.

1. Non-existent binaries 

```console
> template-nextjs@0.1.3 fuels:dev /fuel/0.0.0-pr-2143-20240513112950
> fuels dev

Unable to find the following binaries on the filesystem:
 -> 'forc' at path 'forc'
 -> 'fuel-core' at path 'fuel-core'

Visit https://docs.fuel.network/guides/installation/ for an installation guide.
```

2. Version mismatch

```console
> template-nextjs@0.1.3 fuels:dev /fuel/0.0.0-pr-2143-20240513112950
> fuels dev

The following binaries on the filesystem are outdated:
 -> 'forc' is currently '0.55.0', but requires '0.56.1'.
 -> 'fuel-core' is currently '0.23.0', but requires '0.26.0'.
```
</details>

PR #2123, opened 2024-04-19
build: infra to use typegen in tests

nedsalk:
This PR introduces infrastructure for running `typegen` in our internal test suite.

closes #2070

PR #2312, opened 2024-05-15
chore: use typegen in tests

danielbate:
Building on the fantastic work from @nedsalk in #2123

- Uses `fuels` to build projects, means we don't have to be blocked by #2143 as the underlying binaries can be swapped without impacting the command entrypoint, so no need for additional scripts
- Still having issues with browser polyfills


danielbate:
This work was intergrated in #2312 so will supercede this with a new branch

PR #2322, opened 2024-05-17
feat!: remove nested option size validation

### LINKED ISSUE #2321: `StructCoder.decode` throwing Invalid struct data size
This issue happens when the following contract function `test_function` is called:

```rs
contract;

pub enum FoodType {
    Tomatoes: (),
}

pub struct Food {
    pub name: FoodType,
    pub time_planted: Option<u64>,
}

impl Food {
    pub fn new(name: FoodType, time_planted: Option<u64>) -> Self {
        Self {
            name,
            time_planted,
        }
    }
}

pub struct GardenVector {
    pub inner: [Option<Food>; 10],
}

impl GardenVector {
    pub fn new() -> Self {
        let initial_val: Option<Food> = None;
        Self {
            inner: [initial_val; 10],
        }
    }
}

abi MyContract {
    fn test_function() -> GardenVector;
}

    fn test_function() -> GardenVector {
       GardenVector::new()
    }
} 
```

I could validate after some investigation that the issue arises when using `None` for an `Option` value that contains yet another `Option` value, such as the `Food` struct defined in the contract code.

danielbate:
Closes #2321 

For `v0` encoding, an option would contain a `WORD_SIZE` for the enum case key (`Some` or `None`) and the property space for the value (i.e `u64` would be another `WORD_SIZE`. Even if the value was `None`, we would still receive the expected value property space, meaning we could validate correct for both values. However in `v1`, `None` will now only return the case bytes, and no value bytes. Meaning we can no longer validate the expected byte length before checking the case value.

We are still validating at the option value level (i.e. `Option<u8>` will be validated in the `NumberCoder` so we should not be too concerned about underflow/overflow here, but there are definitely improvements that could be made, that I'll investigate [here](https://github.com/FuelLabs/fuels-ts/issues/2323). But for now, this PR will remove the invalid check for options.

### Breaking Changes: 

Any parent type that contains a nested `option` will no longer throw due to size validation.

### DIFF
--- a/packages/abi-coder/src/encoding/coders/VecCoder.ts
+++ b/packages/abi-coder/src/encoding/coders/VecCoder.ts
@@ -46,7 +46,7 @@ export class VecCoder<TCoder extends Coder> extends Coder<
   }
 
   decode(data: Uint8Array, offset: number): [DecodedValueOf<TCoder>, number] {
-    if (!this.#hasNestedOption && (data.length < this.encodedLength || data.length > MAX_BYTES)) {
+    if ((!this.#hasNestedOption && data.length < this.encodedLength) || data.length > MAX_BYTES) {
       throw new FuelError(ErrorCode.DECODE_ERROR, `Invalid vec data size.`);
     }
 
--- a/packages/abi-coder/src/encoding/coders/VecCoder.ts
+++ b/packages/abi-coder/src/encoding/coders/VecCoder.ts
@@ -2,13 +2,12 @@ import { ErrorCode, FuelError } from '@fuel-ts/errors';
 import { bn } from '@fuel-ts/math';
 import { concatBytes } from '@fuel-ts/utils';
 
-import { MAX_BYTES, WORD_SIZE } from '../../utils/constants';
+import { MAX_BYTES, OPTION_CODER_TYPE, WORD_SIZE } from '../../utils/constants';
 import { isUint8Array } from '../../utils/utilities';
 
 import { Coder } from './AbstractCoder';
 import type { TypesOfCoder } from './AbstractCoder';
 import { BigNumberCoder } from './BigNumberCoder';
-import { OptionCoder } from './OptionCoder';
 
 type InputValueOf<TCoder extends Coder> = Array<TypesOfCoder<TCoder>['Input']> | Uint8Array;
 type DecodedValueOf<TCoder extends Coder> = Array<TypesOfCoder<TCoder>['Decoded']>;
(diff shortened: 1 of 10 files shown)


fuel-service-user:
✨ A PR has been created under the `rc-2322` tag on `fuels-wallet` repo.
https://github.com/FuelLabs/fuels-wallet/pull/1307

packages/abi-coder/src/encoding/coders/VecCoder.ts
```diff
@@ -47,7 +46,7 @@ export class VecCoder<TCoder extends Coder> extends Coder<
   }
 
   decode(data: Uint8Array, offset: number): [DecodedValueOf<TCoder>, number] {
-    if (!this.#isOptionVec && (data.length < this.encodedLength || data.length > MAX_BYTES)) {
+    if (!this.#hasNestedOption && (data.length < this.encodedLength || data.length > MAX_BYTES)) {
```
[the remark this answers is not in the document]
danielbate: https://github.com/FuelLabs/fuels-ts/pull/2322/commits/92d64c52d78a77b7be9659fc376ea7b77de92df0

Voxelot (approved):

packages/fuel-gauge/test/fixtures/forc-projects/options/src/main.sw
```diff
@@ -0,0 +1,46 @@
...
+    fn echo_option(arg: Option<u8>) -> Option<u8>;
+    fn echo_struct_enum_option(arg: OptionStruct) -> OptionStruct;
+    fn echo_vec_option(arg: Vec<Option<u32>>) -> Vec<Option<u32>>;
+    fn echo_tuple_option(arg: (Option<u8>, Option<u16>)) -> (Option<u8>, Option<u16>);
+    fn echo_enum_option(arg: OptionEnum) -> OptionEnum;
+    fn echo_array_option(arg: [Option<u16>; 3]) -> [Option<u16>; 3];
+}
+
+impl OptionContract for Contract {
+    fn echo_option(arg: Option<u8>) -> Option<u8> {
```
maschad, line 23: I think we should have tested for the internal array of `None` case. 

```rust
fn echo_array_option() -> [OptionEnum; 3] {
       [None; 3]
}

Comment (danielbate) [reply packages/fuel-gauge/test/fixtures/forc-projects/options/src/main.sw:23]: It's a good point but `undefined` when encoded as `Option<u8>` == `None`.

Take a look at the [OptionCoder.toSwayOption()](https://github.com/FuelLabs/fuels-ts/blob/63a9d55a18e8b16bd10edda5535610c118770765/packages/abi-coder/src/encoding/coders/OptionCoder.ts#L14-L19) where `undefined` is converted to `None` before being encoded by an `EnumCoder`.

I usually start with a print test to just test byte decoding and to see how we expect the bytes to be arranged:

```rs
fn print_option() -> Option<u8> {
    Option::None
}
```

and then will move to an echo test for encode and decode:

```rs
fn echo_option(arg: Option<u8>) -> Option<u8> {
    arg
}
```

Which is why most of our type tests are echoes rather than prints. So then we can use TS to represent Sway values correctly. So IMO this [test](https://github.com/FuelLabs/fuels-ts/pull/2322/files#diff-5118910301c29105538f3849ae3483c491fd11733ea5d7d78f4ad178a5ff905aR135-R141) does do the job.

maschad: Good point, noted

PR #2324, opened 2024-05-17
ci: removed e2e dependance on other tests

petertonysmith94:
We made the `e2e` test dependent on the `environments` tests (https://github.com/FuelLabs/fuels-ts/pull/2254), to reduce our coin consumption. Due to the [simplification](https://github.com/FuelLabs/fuels-ts/pull/2282) of our `e2e` test, we no longer consume as many coins.

This change removes that dependency to increase our pipeline throughput.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -50,7 +50,6 @@ jobs:
   e2e:
     runs-on: ubuntu-latest
     timeout-minutes: 10
-    needs: [environments]
     if: ${{ !github.event.pull_request.draft }}
     steps:
       - name: Checkout


nedsalk (commented):
I'm fine with it remaining dependent because it reduces the empty-e2e-wallet-breaks-CI occurrences. Our CI is not super-duper fast to the point that the 3-4 minutes for the e2e workflow slow us down too much. I won't give an approval, but this is not a hard opinion by any means; I understand the counter points and am okay if the PR gets approved and merged.

arboleya (approved):

arboleya:
I agree with @nedsalk; no big deal to leave it as dependent.

One thing we could do at some point is try to optimize the steps themselves; it's not urgent, though.

Torres-ssf:
I am in favor of leaving this the way it is. After updating our chain config to use the gas values set [here](https://github.com/FuelLabs/fuel-core/pull/1894), the gas fee will greatly increase.

Issue #2323, opened 2024-05-17
Improve `Option::None` size validation


danielbate:
In #2322 we removed the expected size validation for nested options. That should be reintroduced.

arboleya:
Is this still a thing?

danielbate:
It is yes, as we are not [correctly validating option length](https://github.com/FuelLabs/fuels-ts/blob/35057ffd6e0c4bdb1771ba6c24b7b0d11c88d70c/packages/abi-coder/src/encoding/coders/EnumCoder.ts#L47). Not a massive deal as `Option::Some` is correct and `None` is empty bytes. But with the correct design we can solve for this. However improved length validation in #1795 would also solve this.

arboleya:
Two more questions:
1. Can this be a blocker for `mainnet`?
2. Is there any relationship between this issue and this PR?
   - https://github.com/FuelLabs/fuels-ts/pull/2515

danielbate:
1. No it should not. At some level, an option will need to have a native type, at which point it will be validated correctly
2. No, there is a single line missing coverage in the `EnumCoder` that this will resolve.

danielbate:
Going to un-assign my self as this could be picked up by anyone and is not on my immediate radar.

danielbate:
> At some level, an option will need to have a native type, at which point it will be validated correctly

For the above reason, I think we can deprioritise this issue. Additionally this code has now been audited in a few ways and it remains sound. There is enough validation in other coders and around the empty option that we can have confidence.

IMO the real value will come from improved coder wide length validation in #1795.

PR #2325, opened 2024-05-17
feat: find deeply nested `option` coders

danielbate:
Extension of #2322 to include deeply nested search for `option`.


arboleya (approved):

arboleya:
Would this also close #2323, or that should require another PR?

If the latter, should we release without it or only when it's completed?

danielbate:
@arboleya it does improve it but until we have actual validation checks for `Option::None` I don't want to close it. But I don't think we need to ship it with it. As we still have validation on the option child level (`i.e` `Option<u8>` would be validated at the `u8` level by `NumberCoder`.

Issue #2303, opened 2024-05-14
Use Testnet configuration for local testnet


Torres-ssf:
We should update our default local chain configuration to match the Testnet's configuration. 

This change is inspired by the adjustments made in [this PR](https://github.com/FuelLabs/fuel-core/pull/1894).

PR #2334, opened 2024-05-17
chore: use dev-net chain config for local node

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/2303

## Summary:
- Updated the 2 chains config to use the same values from the dev-net.
- Updated the base asset ID to be the same value as the dev-net.
- Updated funding amounts within our tests, as the calculated fee value increased.


.fuel-core/configs/chainConfig.json
```diff
@@ -98,168 +98,168 @@
...
-              "base": 1645,
-              "units_per_gas": 14
+            "HeavyOperation": {
+              "base": 5254820,
+              "gas_per_unit": 0
             }
           }
         }
       },
       "base_asset_id": "0000000000000000000000000000000000000000000000000000000000000000",
```
Torres-ssf, line 261: @FuelLabs/sdk-ts Should we use the `base_asset_id` to the same ID used by the Testnet as well?

Dhaiwat10 (approved):
🚀

danielbate (approved):

Torres-ssf:
@nedsalk I wonder if merging this PR now will result in breaking https://github.com/FuelLabs/fuels-ts/pull/1356.

nedsalk:
@Torres-ssf I don't think it will break because only the values of the configs changed, not the interface.

arboleya (approved):

Issue #2340, opened 2024-05-18
Use `V1` Encode Logic For `Upgrade` and `Upload` TXs


Torres-ssf:
The de/encode for `Upgrade` and `Upload` transactions is not adjusted with the new V1 encoding logic. As a result, these transaction types are not properly encoded or decoded. 

This needs to be addressed to ensure full compatibility with the latest transaction formats.

PR #2341, opened 2024-05-18
fix: use `V1` encode for `Upgrade` and `Upload` transaction coders

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/2340


arboleya (approved):

Torres-ssf:
> Does this change mean we can perform the `Upgrade` and `Upload` transactions? Or is it just ensuring the coders for the operations align correctly?

With this change, the SDK can properly query and decode them but it does not support submitting them.

PR #2345, opened 2024-05-20
docs: add more thorough examples for predicates

maschad:
This PR seeks to improve the documentation surrounding predicates in the SDK, this includes

- documentation of the [Predicate methods](https://github.com/FuelLabs/fuels-ts/pull/2345/files#diff-8f528b24de2456d5421a8cd534e5d9d63d73c009665a62461f5c9565aa6b72ea)
- A walkthrough [example](https://github.com/FuelLabs/fuels-ts/pull/2345/files#diff-69098d51c0733a29214f157d838eb613e8ba64ea594b1e2e526b0beac1197435) guide of how one would modify the existing [create fuels predicate example](https://github.com/FuelLabs/fuels-ts/blob/master/templates/nextjs/src/pages/predicate.tsx) to use a configurable predicate.

Closes https://github.com/FuelLabs/fuels-ts/issues/2311

PR #2351, opened 2024-05-20
feat: implement `generateFakeResources` on `Account` class

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/2363

After the changes made in this PR we can work on: https://github.com/FuelLabs/fuels-ts/issues/2362


packages/account/src/account.ts
```diff
@@ -642,6 +646,22 @@ export class Account extends AbstractAccount {
     return this.provider.simulate(transactionRequest, { estimateTxDependencies: false });
   }
 
+  /**
+   * Generates an array of fake resources based on the provided coins.
+   *
```
maschad, line 651: We should probably indicate the use case of this method
```suggestion
* This method is useful for unfunded wallets to perform dry runs
```

packages/account/src/predicate/predicate.ts
```diff
@@ -195,6 +196,20 @@ export class Predicate<TInputData extends InputValue[]> extends Account {
     }));
   }
 
+  /**
+   * Generates an array of fake resources based on the provided coins.
+   *
```
maschad, line 201: Likewise, then it should be added to [methods](https://github.com/FuelLabs/fuels-ts/pull/2345/files#diff-8f528b24de2456d5421a8cd534e5d9d63d73c009665a62461f5c9565aa6b72ea)

maschad (changes requested):
Nice job on this, we should definitely document as this is useful for a few scenarios.

packages/account/src/account.ts
Torres-ssf: @maschad Thanks for the suggestion but I reckon my description suits the method better.

Torres-ssf: @maschad Thanks for the suggestion but I reckon the current description suits the method better.

packages/account/src/predicate/predicate.ts
Torres-ssf: @maschad good point. I will add some docs for this.

packages/account/src/account.ts
```diff
@@ -640,6 +644,22 @@ export class Account extends AbstractAccount {
...
+   * @param coins - An array of `FakeResources` objects representing the coins.
+   * @returns An array of `Resource` objects with generated properties.
+   */
+  generateFakeResources(coins: FakeResources[]): Array<Resource> {
```
nedsalk, line 653: Moving this utility from `Account`/`Predicate`/(...`Wallet` in the future?) to a separate function would have the same effect but wouldn't introduce the OOP inheritance complexity of which `Account`+inheritors unfortunately already suffer. 

```ts
export function generateFakeResources(account: Account, coins: FakeResources[]): Array<Resource> {
  return coins.map((coin) => ({
      id: hexlify(randomBytes(UTXO_ID_LEN)),
      owner: this.address,
      blockCreated: bn(1),
      txCreatedIdx: bn(1),
      predicate: account instanceof Predicate ? hexlify(account.bytes) : undefined,
      predicateData: account instanceof Predicate ? hexlify(account.predicateData) : undefined
      ...coin,
    }));
}
```

nedsalk (changes requested):

packages/account/src/account.ts
Torres-ssf: @nedsalk that is a valid concern. However, using the OOP inheritance was the most elegant way that I found to solve this.

To give you more context, this work is a pre-stage for this [issue](https://github.com/FuelLabs/fuels-ts/issues/2362). By removing the fake funding part from `provider.getTransactionCost` we can slim down a little the amount of excessive work that this method does while we get rid of the flags: [resourcesOwner, and quantitiesToContract](https://github.com/FuelLabs/fuels-ts/blob/7c08593037e65ce6256c6cc63a8d0a2fe2f6c0b8/packages/account/src/providers/provider.ts#L1066C7-L1066C21).

We will also be able to get rid of [this](https://github.com/FuelLabs/fuels-ts/blob/7c08593037e65ce6256c6cc63a8d0a2fe2f6c0b8/packages/account/src/providers/provider.ts#L1094-L1096) weird hacky cast that we had to add.

The only reason I am using inheritance is to avoid the casting part ( `account instanceof Predicate` ) since predicate fake resources will be properly generated with `predicate` and `predicateData` and we will not need to inject them later.

There is no need to implement this method in a specific wallet class since only predicates have different resources/inputs.

Also, `instanceof` has proven to be [error-prone](https://github.com/FuelLabs/fuels-ts/pull/1702). We would end up using something similar to [this](https://github.com/FuelLabs/fuels-ts/blob/7c08593037e65ce6256c6cc63a8d0a2fe2f6c0b8/packages/account/src/providers/provider.ts#L1094-L1096).

packages/account/src/predicate/predicate.ts
Torres-ssf: @maschad I've decided to not add a documentation page for this.

I remembered that some users had to learn how to create fake resources on their own to execute dry-run on contract calls to get value from the node. Since these contract calls were merely to read data, there was no need to waste resources submitting a real transaction. 

In the past, we did not provide a straightforward way to do so but this was addressed on https://github.com/FuelLabs/fuels-ts/pull/1860

Since I cannot think of another use case where users would want to generate fake resources on their own, I don't see much value in a documentation page for this since there won't be a real problem to solve.

However, this method is relevant for the changes that are going to be made [here](https://github.com/FuelLabs/fuels-ts/issues/2362), as predicate resources will be properly generated, without the need to be injected later on.

packages/account/src/account.ts
nedsalk: > The only reason I am using inheritance is to avoid the casting part ( `account instanceof Predicate` )

It's easier to grok this `account instanceof Predicate` (or equivalent) check inside of one function than two methods where one is overriding another to get the intended behavior. The `account instanceof Predicate` check is explicit whereas the overriding is not.

This `generateFakeResources` functionality is not really integral to `Account`/`Predicate`/`Wallet` to the point that it _has_ to be on the class like e.g. `transfer` and similar methods, so we have more liberty to move it out into a helper.

I have a relatively strong opinion on the design of this _minor_ thing. It's not some grand thing worthy of extensive discussion and argumentation so feel free to resolve the conversation if you're still not convinced.

Torres-ssf: @nedsalk I don't see this as a problem since resources are intrinsically related to the Account abstraction, as it is an account that owns and spend resources. 

Again, there won't be a case to implement this for wallets because the wallets classes extend from the Account class.

packages/account/src/predicate/predicate.ts
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2351/commits/33c46397835fa686a17407de41f2a28f885213ff

maschad (dismissed):
Great work!

arboleya (commented):
I second what @danielbate said [here](https://github.com/FuelLabs/fuels-ts/pull/2351#discussion_r1622227590).

apps/docs/src/guide/cookbook/generate-fake-resources.md
```diff
@@ -0,0 +1,13 @@
+# Generate Fake Resources
+
+When working with an unfunded account, you can generate fake resources to perform a dry-run on your transactions. This is useful for testing purposes without the need for real funds.
+
+Below is an example script that returns the value `1337`. You can use fake resources to execute a dry-run of this script and obtain the returned value.
+
+<<< @/../../docs-snippets/test/fixtures/forc-projects/return-script/src/main.sw#generate-fake-resources-1{rust:line-numbers}
+
+To execute a dry-run, use the `Provider.call` method. Ensure you set the `utxo_validation` flag to true, as this script uses fake UTXOs:
```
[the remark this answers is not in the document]
Torres-ssf: @danielbate @arboleya The Provider class doesn't have a method called `dryRun`. Instead, it is called [call](https://github.com/FuelLabs/fuels-ts/blob/7dfac30d3d12eb71bd59cf233f4959916b76298f/packages/account/src/providers/provider.ts#L745).  

I've just created an [issue](https://github.com/FuelLabs/fuels-ts/issues/2451) to rename it

Dhaiwat10 (approved):

nedsalk (approved):

arboleya (approved):

PR #2354, opened 2024-05-20
chore: add e2e test for decode txs

Torres-ssf:
This PR adds e2e tests for querying and decoding `Mint`, `Upgrade`, and `Upload` TXs from the devnet


danielbate (approved):

packages/fuel-gauge/src/e2e-script.test.ts
```diff
@@ -8,14 +8,30 @@ import { getScript } from './utils';
  * @group e2e
  */
 describe('Live Script Test', () => {
-  it('can use script against live Fuel Node', async () => {
+  const MINT_TX_ID = '0x03299946676ddc0044a52a675dd201d3173886c998a7301262141334b6d5a29e';
+  const UPGRADE_TX_ID = '0xe2c03044fe708e9b112027881baf9f892e6b64a630a629998922c1cab918c094';
+  const UPLOAD_TX_ID = '0x94bc2a189b8211796c8fe5b9c6b67624fe97d2007e104bf1b30739944f43bd73';
```
arboleya, line 13: Why use hardcoded IDs?

How will this behave if we reset the network or upgrade to a new one?

Torres-ssf: @arboleya If the state of the chain is reset we would need a new set of IDs. If the chain is upgraded, we would not need to update the IDs. 

Also, we would need to update these IDs after changing from the dev net to the testnet.

arboleya: The problem I see is that our tests should not be network-dependent.

Otherwise, we'd need to have several different hardcoded tests, one for each network (devnet, testnet, mainnet).

Makes sense?

Torres-ssf: @arboleya I see your point.

The motivation behind the tests is to spot changes related to the de/serialization logic. This is helpful because unlike `Create` and `Script` transactions, these TX types are not broadly tested within our tests.

We would need to update these IDs only for the testnet since I don't believe the serialization logic will be different between the Testnet / Mainnet. What do you think?

arboleya: What would be the downsides of creating these TXs on the fly and then reading and de/serializing them?

Torres-ssf: I reckon it would be forgetting to update these TXs types coders if any serialization logic happens.

We already had [tests](https://github.com/FuelLabs/fuels-ts/blob/f76aa5710134a72431b576572ef1db9737e0fd8f/packages/transactions/src/coders/transaction.test.ts#L333) in place but I only noticed they were false positives when I tried to query these TXs types using the SDK. Then I made this [fix](https://github.com/FuelLabs/fuels-ts/issues/2340).

Since the tests simply encode and decode them based on hardcoded values, they continue to pass even after we forgot to update the coders logic to use `V1`.

maschad (approved):

danielbate (approved):

packages/fuel-gauge/src/e2e-script.test.ts
arboleya: I'm not talking about purely encoding/decoding but submitting the TXs to the chain and then reading them back to avoid having hardcoded IDs. Are there any downsides to this approach?

Torres-ssf: @arboleya OK, I understand now.

This approach is not possible because the SDK does not support submitting these types of transactions

`Mint` transactions are generated by the block producer, so they cannot be submitted at all.

`Upgrade` and `Upload` transactions are _not_ used by users:

- `Upgrade Transaction`: Allows upgrading either consensus parameters or state transition function used by the network to produce future blocks.

- `Upload Transaction`: Intended to upload huge bytecode to upgrade the state transition of the network.

arboleya: Alright, thanks for the detailed explanation.

My concern still remains, though.

Imagine a (future) scenario where we need to test our stuff through different stages:
 1. devnet
 2. testnet
 3. mainnet

In such a scenario, the approach taken here won't work by design.

I will resolve this conversation for now, but that's something to keep in mind.

arboleya: @Torres-ssf @danielbate Should we add an issue to revisit this? The motivation is good, but we must ensure the solution doesn't create a new problem underneath, which may be the case with hardcoded TXs IDs.

danielbate: Our `e2e` tests are inherently bound to a version of the testnet though. And we pass in a wallet address/key that is environment dependent. IMO passing tx ids that are bound to a version is no different here.

arboleya: Agreed.

PR #2358, opened 2024-05-21
chore: refactor for performance improvements

maschad:


packages/abi-coder/src/ResolvedAbiType.ts
```diff
@@ -132,6 +134,13 @@ export class ResolvedAbiType {
   }
 
   private getArgSignaturePrefix(): string {
+    if (this.type.length > 1000) {
```
nedsalk, line 137: What's the reasoning behind this check?

packages/abi-coder/src/ResolvedAbiType.ts
```diff
@@ -168,6 +177,13 @@ export class ResolvedAbiType {
       return this.type;
     }
 
+    if (this.type.length > 1000) {
```
danielbate, line 180: What is the reasoning behind `1000`? We should probably take validation values from the [spec](https://docs.fuel.network/docs/specs/abi/fn-selector-encoding/).

.changeset/great-penguins-compete.md
```diff
@@ -0,0 +1,6 @@
+---
+"@internal/fuel-core": patch
+"@fuel-ts/abi-coder": patch
```
danielbate, line 3:
```suggestion
"@fuel-ts/abi-coder": minor
```

Should this be a breaking change now we are throwing for this condition?

packages/abi-coder/src/ResolvedAbiType.ts
arboleya: Was this resolved without an answer?

packages/abi-coder/src/ResolvedAbiType.ts
arboleya: Another one resolved without the reasoning.

Also, wasn't this the perfect use case for named constants/variables?

```ts
    if (this.type.length > myMeaningfulVariableName) {
```

Just trying to understand where all those requests usually come from.

.changeset/great-penguins-compete.md
arboleya: I'd say it's not a breaking change because the API is the same.

But it's worth the reflection, and more context would be helpful.

arboleya (commented):
Please consider adding a description for the PR, explaining why it is needed and what it solves.

packages/abi-coder/src/encoding/strategies/getCoderV1.ts
```diff
@@ -98,7 +98,15 @@ export const getCoder: GetCoderFn = (
   // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
   const components = resolvedAbiType.components!;
 
+  if (resolvedAbiType.type.length > 1000) {
```
danielbate, line 101: Should this also be `256` given this [change](https://github.com/FuelLabs/fuels-ts/pull/2358/commits/06d34c0d0d33cb78c2a1e48328582a4022b77e01).

maschad: Good catch

Dhaiwat10 (approved):

PR #2359, opened 2024-05-21
feat: make `typegen` factories an object instance

nedsalk:
`typegen` factories currently are classes with static methods which is a problem when wanting to pass the whole factory as an argument because e.g. passing a whole `TestContract__factory` as an argument and using it in the manner of `factoryArgument.deployContract` would yield an `undefined.deployContract` error because `TestContract__factory` is not an object instance. This PR is necessary to be able to pass the outputs of typegen to `launchTestNode` in #1356. It doesn't break anything for end consumers.


danielbate (approved):

Torres-ssf:
@nedsalk Does the described issue with Typegen factories indicate a design flaw, or is this change primarily to fit the needs of `launchTestNode`?

nedsalk:
> Does the described issue with Typegen factories indicate a design flaw, or is this change primarily to fit the needs of launchTestNode?

@Torres-ssf This change is primarily to fit the needs of `launchTestNode`. The design itself is okay as it follows `ContractFactory`. There is possibility for improvement here so that we don't need e.g. the `contract as unknown as ContractAbi` casts, but that's more related to the internal design of how we integrate typegen outputs with the rest of our codebase.

> I like this approach. Would it be worth having interfaces for these factories?

@petertonysmith94 No because these factories themselves follow the interface of the `ContractFactory` class. Ideally we'd be returning a `ContractFactory<TTypegenAbi>` and we wouldn't have the need for this `*__factory` object, but this is related to what I told @Torres-ssf above - an internal design restructuring.

PR #2364, opened 2024-05-21
chore: add test to verify `create fuels` template integrity

### LINKED ISSUE #2265: Add a 'live' automated testing suite for `create fuels`
We currently have some unit tests for `create fuels`, but these tests don't mimic a user running the command `npm create fuels` from their terminal.

As a part of #2170, we should add an automated testing suite that actually runs the `npm create fuels@pr-tag` on a given PR, lets the tool extract the template, and verifies that it is extracting all the files we would expect it to.

The verification of the template files working as intended will be handled by #2264, and the work of this issue and #2264 will eventually be merged together as #2170.

petertonysmith94:
Closes #2265

### DIFF
--- a/.changeset/giant-jobs-buy.md
+++ b/.changeset/giant-jobs-buy.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/account": patch
+---
+
+fix: add `predicateData` to predicate resources and inputs
--- a/packages/account/src/predicate/predicate.ts
+++ b/packages/account/src/predicate/predicate.ts
@@ -192,6 +192,7 @@ export class Predicate<TInputData extends InputValue[]> extends Account {
     return resources.map((resource) => ({
       ...resource,
       predicate: hexlify(this.bytes),
+      predicateData: hexlify(this.getPredicateData()),
     }));
   }
 
--- a/packages/account/src/providers/transaction-request/transaction-request.ts
+++ b/packages/account/src/providers/transaction-request/transaction-request.ts
@@ -27,7 +27,11 @@ import { normalizeJSON } from '../utils';
 import { getMaxGas, getMinGas } from '../utils/gas';
 
 import { NoWitnessAtIndexError } from './errors';
-import { isRequestInputResource } from './helpers';
+import {
+  getRequestInputResourceOwner,
+  isRequestInputResource,
+  isRequestInputResourceFromOwner,
+} from './helpers';
 import type {
   TransactionRequestInput,
   CoinTransactionRequestInput,
@@ -348,7 +352,7 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
    * @param coin - Coin resource.
    */
   addCoinInput(coin: Coin) {
-    const { assetId, owner, amount, id, predicate } = coin;
+    const { assetId, owner, amount, id, predicate, predicateData } = coin;
 
     let witnessIndex;
 
@@ -372,6 +376,7 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
       txPointer: '0x00000000000000000000000000000000',
       witnessIndex,
       predicate,
+      predicateData,
     };
 
     // Insert the Input
@@ -388,7 +393,7 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
    * @param message - Message resource.
    */
   addMessageInput(message: MessageCoin) {
-    const { recipient, sender, amount, predicate, nonce, assetId } = message;
+    const { recipient, sender, amount, predicate, nonce, assetId, predicateData } = message;
 
     let witnessIndex;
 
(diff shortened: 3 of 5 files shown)


.github/workflows/test.yaml
```diff
@@ -72,6 +72,37 @@ jobs:
           TEST_WALLET_ADDRESS: ${{ secrets.TEST_WALLET_ADDRESS }}
           FUEL_TESTNET_NETWORK_URL: ${{ secrets.FUEL_TESTNET_NETWORK_URL }}
 
+  integration:
```
Dhaiwat10, line 75: Might be worth renaming this to something like `create-fuels-cli-integration` ?

package.json
```diff
@@ -26,6 +26,7 @@
     "test:browser": "vitest --run --coverage --config vitest.browser.config.mts $(scripts/tests-find.sh --browser)",
     "test:browser:filter": "vitest --run --coverage --config vitest.browser.config.mts",
     "test:e2e": "vitest --run --config vitest.node.config.mts $(scripts/tests-find.sh --e2e)",
+    "test:integration": "vitest --run --config vitest.node.config.mts $(scripts/tests-find.sh --integration)",
```
Dhaiwat10, line 29: Might wanna rename this command to something more descriptive as well?

Dhaiwat10 (commented):
Great work mate @petertonysmith94 👏🏻

.github/workflows/test.yaml
```diff
@@ -72,6 +72,37 @@ jobs:
...
+      - name: Test Setup
+        uses: ./.github/actions/test-setup
+
+      - name: Publish to NPM
+        id: release
+        uses: ./.github/actions/pr-release
+        with:
+          npm-token: ${{ secrets.NPM_TOKEN }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          pr-number: ${{ github.event.pull_request.number }}
```
danielbate, line 96: Should we be cleaning up the version after a test?

.github/actions/pr-release/action.yaml
```diff
@@ -0,0 +1,41 @@
+name: "Publish PR to NPM"
+inputs:
+  npm-token:
+    description: "NPM token for authenticating to NPM registry"
+    required: true
+
+  github-token:
+    description: "GitHub token for authenticating to GitHub"
+    required: true
```
danielbate, line 9: Why are these passed in and not just taken from secrets?

.github/workflows/test.yaml
danielbate: Hmm with our [current classification](https://github.com/FuelLabs/fuels-ts/discussions/1491#discussioncomment-7800279) anything within `fuel-gauge` is an integration test also.

maschad (changes requested):
A few comments:

- I think the title of this PR should be `chore:  add test to verify create-fuels template`
- given that this PR doesn't necessarily add any functionality that we would use without https://github.com/FuelLabs/fuels-ts/pull/2278 I think it should be based off the `dp/create-fuels-template-e2e-tests` branch

- When running this locally the tests don't seem to pass for me.
Screenshot 2024-05-23 at 11 52 05 AM

maschad:
> As part of the overarching https://github.com/FuelLabs/fuels-ts/issues/2170: we will perform the merger of this PR and https://github.com/FuelLabs/fuels-ts/pull/2278 to get the full test. We believe they both serve value on their own, and can go out independently of each other as an increment of work which delivers value.
 
 Maybe I'm missing something but I am not sure what the use case for this test is outside of #2278

packages/create-fuels/test/e2e.test.ts
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * @group e2e
+ */
+describe('`create fuels` package integrity', () => {
+  let paths: ProjectPaths;
+  let shouldSkip = false;
+
+  beforeAll(() => {
+    if (!PUBLISHED_NPM_VERSION) {
+      log('Skipping live `create fuels` test');
```
nedsalk, line 28: It seems like this test would be skipped if the publishing fails because `PUBLISHED_NPM_VERSION` wouldn't be set, which is something that we don't want. Adding a check like `process.env.CI` would rule out that scenario:
```ts
if(!PUBLISHED_NPM_VERSION && !process.env.CI) {...} 
```

I don't know if the GitHub `CI` environment variable is available to node's `process.env` or you'd have to pass it.

.github/actions/pr-release/action.yaml
danielbate: TIL

danielbate (approved):

packages/create-fuels/test/e2e.test.ts
nedsalk: You're right, I didn't see this through completely.

nedsalk (approved):

.github/workflows/test.yaml
arboleya: Let me know when this goes to master so we can add it as a required workflow.

> Related [comment](https://github.com/FuelLabs/fuels-ts/pull/2278#discussion_r1613246021).

maschad (approved):

PR #2368, opened 2024-05-22
fix: `typegen` generic enums and `./common` import (predicates, scripts)

nedsalk:
I found this issue after running `pnpm tsc --noEmit` on typegen outputs of our `fuel-gauge` test suite.

### DIFF
--- a/packages/abi-typegen/src/runTypegen.test.ts
+++ b/packages/abi-typegen/src/runTypegen.test.ts
@@ -189,7 +189,7 @@ describe('runTypegen.js', () => {
     // validates execution was ok
     expect(error).toBeFalsy();
 
-    expect(stdoutWrite).toHaveBeenCalledTimes(4);
+    expect(stdoutWrite).toHaveBeenCalledTimes(5);
   });
 
   test('should raise error for non-existent Script BIN file', async () => {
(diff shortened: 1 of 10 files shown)


packages/abi-typegen/src/runTypegen.test.ts
```diff
@@ -189,7 +189,7 @@ describe('runTypegen.js', () => {
     // validates execution was ok
     expect(error).toBeFalsy();
 
-    expect(stdoutWrite).toHaveBeenCalledTimes(4);
+    expect(stdoutWrite).toHaveBeenCalledTimes(5);
```
nedsalk, line 192: Increased to 5 because now the `./common.d.ts` file gets generated as well and that's logged in the console.

nedsalk:
Converted to draft because I found an additional issue with `import ... from './common'

nedsalk:
@petertonysmith94 
1. Go to `master`,
2. Run `pnpm pretest --force`,
3. Run `pnpm tsc --noEmit packages/fuel-gauge/test/typegen/**/*.ts`

It will list out the problematic files. You'll find that the predicates and scripts that have types from `common.d.ts` are having errors because the import statement is missing.

Checkout this branch and do the same and it'll be solved.

danielbate (approved):

PR #2372, opened 2024-05-22
feat: support `forc build --json-abi-with-callpaths` flag

### LINKED ISSUE #2162: Support `forc build --json-abi-with-callpaths`
From [this rust-sdk issue](https://github.com/FuelLabs/fuels-rs/issues/881):
> Currently, type-paths are feature-gated behind the forc flag --json-abi-with-callpaths.
Once the TS SDK implements support for type paths, forc will make this behavior default.

Running `forc build` with the flag `--json-abi-with-callpaths` "outputs json abi with callpaths instead of names for struct and enums". We can take the sway program below as an example:
```rs
contract;

use std::address::Address;

struct CustomAddress {
    value: b256,
}

abi MyContract {
    fn std_address(address: Address) -> b256;
    fn custom_address(address: CustomAddress) -> b256;
}
```

Running `forc build` produces these types in the abi:
```json
[
  {
    "typeId": 0,
    "type": "b256",
    "components": null,
    "typeParameters": null
  },
  {
    "typeId": 1,
    "type": "struct Address",
    "components": [
      {
        "name": "bits",
        "type": 0,
        "typeArguments": null
      }
    ],
    "typeParameters": null
  },
  {
    "typeId": 2,
    "type": "struct CustomAddress",
    "components": [
      {
        "name": "value",
        "type": 0,
        "typeArguments": null
      }
    ],
    "typeParameters": null
  }
]
```

Whereas building it with `--json-abi-with-callpaths` produces this:
```json
[
  {
    "typeId": 0,
    "type": "b256",
    "components": null,
    "typeParameters": null
  },
  {
    "typeId": 1,
    "type": "struct CustomAddress",
    "components": [
      {
        "name": "value",
        "type": 0,
        "typeArguments": null
      }
    ],
    "typeParameters": null
  },
  {
    "typeId": 2,
    "type": "struct std::address::Address",
    "components": [
      {
        "name": "bits",
        "type": 0,
        "typeArguments": null
      }
    ],
    "typeParameters": null
  }
]
```

The second abi names the struct from the std library with its callpath. This is _very_ beneficial because we can use it to differentiate between sway standard library constructs and custom constructs and custom-tailor our typegen and abi-coder for a better developer experience.

For example, the generated types and functions based on the first abi would be
```ts
export type AddressInput = { bits: string };
export type AddressOutput = AddressInput;
export type CustomAddressInput = { value: string };
export type CustomAddressOutput = CustomAddressInput;

export class MyProjectAbi extends Contract {
  functions: {
    custom_address: InvokeFunction<[address: CustomAddressInput], string>;
    std_address: InvokeFunction<[address: AddressInput], string>;
  };
}
```

Whereas what we _could_ support with this flag is the following:
```ts
export type CustomAddressInput = { value: string };
export type CustomAddressOutput = CustomAddressInput;

export class MyProjectAbi extends Contract {
  functions: {
    custom_address: InvokeFunction<[address: CustomAddressInput], string>;
    std_address: InvokeFunction<[address: string], string>; // <--------------- Not object but string
  };
}
```
This way we can custom-tailor other std types like `EvmAddress`, [`Identity`](https://github.com/FuelLabs/fuels-ts/issues/1891), `ContractId`, etc.

Our typegen currently doesn't support this build flag because it generates the corresponding types like this:
```ts
export type std::address::AddressInput = { bits: string };
export type std::address::AddressOutput = std::address::AddressInput;
```

nedsalk:
This flag will become the default in `forc build` after we support it.

closes #2162


packages/abi-typegen/package.json
```diff
@@ -43,9 +43,10 @@
     "dist"
   ],
   "scripts": {
-    "pretest": "pnpm build:forc",
+    "pretest": "run-s build:forc-callpaths build:forc",
     "build": "tsup",
     "build:forc": "pnpm fuels-forc build -p test/fixtures/forc-projects --release",
+    "build:forc-callpaths": "pnpm fuels-forc build -p test/fixtures/forc-projects --json-abi-with-callpaths",
```
nedsalk, line 49: I'm using the fact that this generates outputs in the `debug` folder so that I can run [this](https://github.com/FuelLabs/fuels-ts/pull/2372/files#diff-7e78de98857b1c1d9877c6eec968b44cc4e968970a7eccbe1e64cc04ad381e53R17) test on both versions of the abi, expecting the same output in both cases.

packages/abi-typegen/src/templates/utils/formatEnums.ts
```diff
@@ -25,7 +25,8 @@ export function formatEnums(params: { types: IType[] }) {
         outputNativeValues,
         typeAnnotations,
       };
-    });
+    })
+    .sort((a, b) => (a.structName < b.structName ? -1 : 1));
```
nedsalk, line 29: I'm sorting enums and structs to get the same resulting `.d.ts` for when the `--json-abi-with-callpaths` flag is active and not. This is so that I can reuse the [test](https://github.com/FuelLabs/fuels-ts/pull/2372/files#diff-7e78de98857b1c1d9877c6eec968b44cc4e968970a7eccbe1e64cc04ad381e53R17).

packages/abi-typegen/test/fixtures/forc-projects/full/src/main.sw
```diff
@@ -1,5 +1,6 @@
 contract;
-
+use libs_for_testing::ExternalStruct;
+use libs_for_testing::ExternalEnum;
```
nedsalk, line 3: These types are used for non-std cases. For example, this type produces an abi type with the name `libs-for-testing::ExternalEnum`.

packages/abi-typegen/src/abi/types/EvmAddressType.ts
```diff
@@ -7,7 +7,7 @@ export class EvmAddressType extends AType implements IType {
 
   public name = 'evmAddress';
 
-  static MATCH_REGEX: RegExp = /^struct EvmAddress$/m;
+  static MATCH_REGEX: RegExp = /^struct (std::vm::evm::evm_address::)?EvmAddress$/m;
```
danielbate, line 10: Once this becomes default behaviour, will we also be removing the optional regexes?

packages/abi-typegen/test/fixtures/forc-projects/libs-for-testing/.gitignore
danielbate, line 1: Redundant file.

packages/abi-typegen/src/abi/types/EvmAddressType.ts
nedsalk: Yes, we'll remove them for standard library types because they have a known specifier, but the generic ones like [enums](https://github.com/FuelLabs/fuels-ts/pull/2372/files/d3c8361493ed474f22861bcebba7fa0f1b4aac43#diff-4c608f8490ef0e210424f72efec83bec40a2a143be25af884a3d2029760c2152R18) and [structs](https://github.com/FuelLabs/fuels-ts/pull/2372/files/d3c8361493ed474f22861bcebba7fa0f1b4aac43#diff-6ec5eb9d704ba8a14d891a6981cbabf6cf5572aa9c7b5076529208d0f3512c4fR15) will have to remain because they can be imported from external libraries.

danielbate (approved):

maschad (approved):

PR #2374, opened 2024-05-22
ci: added missing coverage for browser tests

petertonysmith94:
The coverage for browser tests was not being downloaded or merged - this change adds that.


.github/workflows/test.yaml
```diff
@@ -92,6 +92,12 @@ jobs:
           name: node
           path: coverage/environments
 
+      - name: Download Coverage Artifact for Browser Tests
```
arboleya, line 95: Is it also being considered and merged in the report?

maschad (approved):

Dhaiwat10 (approved):

.github/workflows/test.yaml
danielbate: I think it is because the `master` artefact contains only node tests, wheras this artefact contains both environments, so the coverage maps aren't lining up. Feel free to convert to draft and I can amend. We'll need to push up to `master` and have it generate the complete coverage artefact, but with some conditional logic so it only runs for specific PRs until everything is place. I did similar [here](https://github.com/FuelLabs/fuels-ts/pull/1310/files#diff-245392b692a50c38ecab4381b118862db514035c10983f3bd4f4b7f1f4be4692) / [here](https://github.com/FuelLabs/fuels-ts/pull/1676/files).

danielbate: That's the theory, I think we go for it, any issues we can rollback. We also have all the old artefacts stored.

Issue #2270, opened 2024-05-08
Insufficient Fee Error on `fuels` `0.82.0`


Torres-ssf:
A bug related to insufficient fee has been identified in the latest Beta-5 compatible release: [0.82.0](https://github.com/FuelLabs/fuels-ts/releases/tag/v0.82.0). 

We need to investigate further and fix this.

Perhaps a solution could be the same approach that was implemented during https://github.com/FuelLabs/fuels-ts/pull/2021.

The solution changes the value that is set to the `gasLimit` property before executing the dry-run to get the `gasUsed` value:

```ts
txRequest.gasLimit = maxGas
```

to

```ts
txRequest.gasLimit = maxGasPerTx.sub(maxGas)
```

Torres-ssf:
Closing this as I could not reproduce this error while executing contract calls on beta-5. 

Also, the implementation suggested here is already being used on beta-5

Issue #2212, opened 2024-04-30
function getTransactionCost return wrong gas amount 


nadiia-balaian:
### What version of fuels-ts are you using?

82

### Steps to Reproduce

call getTransactionCost and check what this function returns

### Expected Behavior

getTransactionCost returns valid amount of gas

### Actual Behavior

when I'm trying to make transaction and make request to calculate gas for it getTransactionCost, I got value that is not enough to send transaction 
when I try to send transaction with amount from the getTransactionCost I got error OutOfGas 

so we fixed this behaviour on our side with simple workaround  - we multiple amount from getTransactionCost for 2

nadiia-balaian:
> Hello @nadiia-balaian . Sorry about this error.
> 
> Could you provide us with some code snippets and more context on how this error occurred?
> 
> It would help a lot if we could replicate the exact scenario in which this error is happening on your side (executing the same contract call).


Hello @Torres-ssf 
So we tested later the same code and we found out that problem is not in this exact function.
![photo_2024-05-01_11-19-42](https://github.com/FuelLabs/fuels-ts/assets/109521832/5276ef8c-1a03-4ea1-bc68-c1d1290a5ea9).

Torres-ssf:
@nadiia-balaian It seems that the given print refers to the Rust SDK.

Are you experiencing the same error (OutOfGas) while using the TS SDK when calling the function `open_order` ?

Torres-ssf:
Closing this as I could not reproduce this error while executing contract calls on beta-5.

PR #2375, opened 2024-05-22
fix: gas and fee error on `0.82.0`

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/2270
- Close: https://github.com/FuelLabs/fuels-ts/issues/2212


arboleya:
Closing as is, but we should double-check the related issues.

PR #2376, opened 2024-05-23
feat: add new `node` command to `fuels` CLI

### LINKED ISSUE #2355: Adding new `node` command to `fuels` CLI
One may want to start a node in isolation, using _**the same configs**_ used by the `fuels dev`.

It should use the default configs or the user's configs if they specify them in `fuels.config.ts.`

Dhaiwat10:
Closes #2355


packages/fuels/src/cli.ts
```diff
@@ -66,6 +67,11 @@ export const configureCli = () => {
     .addOption(pathOption)
     .action(withConfig(command, Commands.dev, dev));
 
+  (command = program.command(Commands.node))
+    .description('Start a Fuel node')
```
danielbate, line 71:
```suggestion
    .description('Start a Fuel-Core node')
```

danielbate (approved):
🔥

packages/fuels/src/cli.ts
nedsalk: I suppose it would be `fuel-core` then.

Dhaiwat10: The description for the `dev` command says `Fuel`, so I went with that here as well.

packages/fuels/src/cli/commands/node/index.ts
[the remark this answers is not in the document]
Dhaiwat10: Do you think it's worth it? It only watches for any changes in the config file. I think that's something everyone would want, and is a good default convention that is used by projects like Next.js for their config files as well.

Dhaiwat10:
@petertonysmith94 added to readme: https://github.com/FuelLabs/fuels-ts/pull/2376/commits/643f9c2e1a3942abdf238eeca6d0d171d8ac1b72

packages/fuels/src/cli.ts
maschad: I think `Fuel` node would be more intelligible by the users of this command @danielbate , whilst we do make that distinction within the naming of the packages, there is only one implication of a _Fuel node_ to the best of my knowledge.

maschad (approved):
Great job on this!

danielbate (approved):

Issue #1820, opened 2024-02-29
Integrate `launchTestNode` with `fuel-gauge` 


arboleya:
The idea is to update tests to use an isolated node via `launchTestNode`.

Then, we can try running tests in parallel / concurrently again.

Past experiments have been successful in this regard.

----

Must be done after:
 - #1356 

And will probably fix:
 - #1755

arboleya:
@nedsalk It would be fantastic to have a draft PR for this, after:
 - #1356

Like a small POC proving the workings of the former and demoing it in the real world.

cc @danielbate

danielbate:
As part of this and with the merging of #2378, can we add the `@group browser` tag to tests also? The imports will need to be using the typegen paths.

PR #2378, opened 2024-05-23
chore: add browser testing infrastructure

### LINKED ISSUE #284: Create test to `browser compatibility`

### LINKED ISSUE #2178: Make `launchTestNode` utility compatible with browser testing
`launchTestNode` depends on `launchNode`, which depends on `node` and so cannot run in browser tests.

A possible solution could be to start a web server in the global test setup and have `launchNode` somehow interact with it, while keeping in mind that `launchNode` has to be able to run standalone as well.

nedsalk:
**Huge shoutout to @danielbate with whom I had multiple discussions and debugging sessions to get this working!**

---

The idea is to run an http server that has the responsibility of launching and killing `fuel-core` nodes and the functionality would be transparent to `launchTestNode`, meaning all the _@group node_ tests that use `launchTestNode` can be reused for `@group browser` tests as well.

This is achieved by setting an environment variable based on which the internals of `launchTestNode`/`setupTestProviderAndWallets` either use `launchNode` (in node tests) or send requests to the http server (browser tests).

I'll write documentation for the utility once we dogfood this in our whole test suite and verify everything works as expected. We might also put it behind a `fuels` cli command, but I'd also rather do that when we want to make this generally available.

closes #284, #2178


packages/account/src/test-utils/launchNode.ts
```diff
@@ -217,6 +216,8 @@ export const launchNode = async ({
       snapshotDirToUse = tempDir;
     }
 
+    const { spawn } = await import('child_process');
```
nedsalk, line 219: This dynamic import is necessary in order for `vitest` to not necessitate a polyfill for it.

packages/fuels/src/setup-launch-node-server.test.ts
```diff
@@ -0,0 +1,93 @@
+import { Provider } from '@fuel-ts/account';
+import { waitUntilUnreachable } from '@fuel-ts/utils/test-utils';
+import { spawn } from 'node:child_process';
+
+function startServer(
+  port: number = 0
+): Promise<{ serverUrl: string; killServer: () => void } & Disposable> {
```
maschad, line 7: We should probably define the return type as an interface 

```ts
interface ServerInfo extends Disposable {
  serverUrl: string;
  killServer: () => void;
}
```

packages/fuels/src/setup-launch-node-server.test.ts
```diff
@@ -0,0 +1,93 @@
...
+        serverUrl,
+        killServer,
+        [Symbol.dispose]: killServer,
+      });
+    });
+
+    cp.on('error', (err) => {
+      reject(err);
+    });
+  });
```
maschad, line 35: Suggestion: some cleanup on exit

```ts
cp.on('exit', (code) => {
      if (code !== 0) {
        reject(new Error(`Server process exited with code ${code}`));
      }
    });
 ```

File: packages/fuels/src/setup-launch-node-server.test.ts [hunk 4]
```diff
@@ -0,0 +1,93 @@
...
+    cp.stdout?.on('data', (chunk) => {
+      // first message is server url
+      const message: string[] = chunk.toString().split('\n');
+      const serverUrl = message[0].startsWith('http://') ? message[0] : '';
+      // teardown
+      resolve({
+        serverUrl,
+        killServer,
+        [Symbol.dispose]: killServer,
+      });
```
maschad, line 29: we should probably handle any potential errors that could arise with the `serverUrl`

```suggestion
    if (serverUrl) {
      resolve({
        serverUrl,
        killServer,
        [Symbol.dispose]: killServer,
      });
    } else {
      reject(new Error('Invalid server URL'));
    }
  ```

File: packages/fuels/src/setup-launch-node-server.ts [hunk 5]
```diff
@@ -0,0 +1,109 @@
+/* eslint-disable no-console */
+import type { LaunchNodeOptions, LaunchNodeResult } from '@fuel-ts/account/test-utils';
+import { launchNode } from '@fuel-ts/account/test-utils';
+import http from 'http';
+import type { AddressInfo } from 'net';
+
+process.setMaxListeners(Infinity);
+
+async function parseBody(req: http.IncomingMessage) {
+  return new Promise<string>((resolve, reject) => {
```
maschad, line 10: Add return types
```suggestion
async function parseBody(req: http.IncomingMessage): Promise<LaunchNodeOptions>  {
  return new Promise<LaunchNodeOptions>((resolve, reject) => {
```

packages/fuels/src/setup-launch-node-server.ts
```diff
@@ -0,0 +1,109 @@
...
+    const body: Buffer[] = [];
+    req.on('data', (chunk) => {
+      body.push(chunk);
+    });
+    req.on('end', () => {
+      resolve(JSON.parse(body.length === 0 ? '{}' : Buffer.concat(body).toString()));
```
maschad, line 16: This should be in a try/catch
```suggestion
 try {
       resolve(JSON.parse(body.length === 0 ? '{}' : Buffer.concat(body).toString()));
    } catch (err) {
        reject(err);
    }
    ```

File: packages/fuels/src/setup-launch-node-server.ts [hunk 7]
```diff
@@ -0,0 +1,109 @@
...
+
+function cleanupAllNodes() {
+  cleanupFns.forEach((fn) => fn());
+}
+
+const server = http.createServer(async (req, res) => {
+  res.setHeader('Access-Control-Allow-Origin', '*');
+
+  if (req.url === '/') {
+    const body = (await parseBody(req)) as LaunchNodeOptions;
```
maschad, line 32: I would wrap this in a try/catch as well just because `parseBody` could throw an error / reject the promise.

```suggestion
try {
      const body = await parseBody(req);

      const node = await launchNode({
        port: '0',
        ...body,
        fuelCorePath: 'fuels-core',
      });

      cleanupFns.set(node.url, () => {
        node.cleanup();
        cleanupFns.delete(node.url);
      });

      res.writeHead(200, { 'Content-Type': 'text/plain' });
      res.end(node.url);
    } catch (err) {
      res.writeHead(500, { 'Content-Type': 'text/plain' });
      res.end('Internal Server Error');
      console.error(err);
    }
    ```

File: vitest.global-browser-setup.ts [hunk 8]
```diff
@@ -0,0 +1,24 @@
...
+      // it will kill the server
+      resolve(() => {
+        // https://github.com/nodejs/node/issues/2098#issuecomment-169549789
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        process.kill(-cp.pid!);
+      });
+    });
+
+    cp.on('error', (err) => {
+      reject(err);
```
maschad, line 21: We probably still want to kill the server on error, we could extract the `process.kill` to a function

```ts
  const killServer = () => {
      if (cp.pid) {
        process.kill(-cp.pid);
      }
    };
   ```
   
  and then call it on `error` and `exit` scenarios
  
  ```ts
  cp.stdout?.on('data', () => {
      // Return teardown function to be called when tests finish
      // It will kill the server
      resolve(killServer);
    });

    cp.on('error', (err) => {
      // Ensure server is killed if there's an error
      killServer();
      reject(err);
    });

    cp.on('exit', (code) => {
      if (code !== 0) {
        reject(new Error(`Server process exited with code ${code}`));
      }
    });
  ```

**Review by maschad — CHANGES_REQUESTED** [review]
Great work on this! It's major benefits for our testing infra. Left a few comments.

File: packages/fuels/src/setup-launch-node-server.test.ts (hunk 4 above)
Comment (nedsalk) [reply packages/fuels/src/setup-launch-node-server.test.ts]: I simplified the logic a bit [here](https://github.com/FuelLabs/fuels-ts/pull/2378/commits/7a1ebbfe1a2944f5aebf3de9e75bd6003ba8b497#diff-fdddb3f57680c5f4a5d071b37425ff8f81fd98207894622482e7e9a637f2e7e0R28). Keep in mind that this is just a test helper and the tests will fail if the `serverUrl` is invalid in any way.

File: packages/fuels/src/setup-launch-node-server.test.ts (hunk 3 above)
Comment (nedsalk) [reply packages/fuels/src/setup-launch-node-server.test.ts:44]: [done](https://github.com/FuelLabs/fuels-ts/pull/2378/files#diff-fdddb3f57680c5f4a5d071b37425ff8f81fd98207894622482e7e9a637f2e7e0R390)

File: vitest.global-browser-setup.ts (hunk 8 above)
Comment (nedsalk) [reply vitest.global-browser-setup.ts:26]: https://github.com/FuelLabs/fuels-ts/pull/2378/commits/7a1ebbfe1a2944f5aebf3de9e75bd6003ba8b497

File: packages/fuels/src/setup-launch-node-server.ts (hunk 7 above)
Comment (nedsalk) [reply packages/fuels/src/setup-launch-node-server.ts]: https://github.com/FuelLabs/fuels-ts/pull/2378/commits/7a1ebbfe1a2944f5aebf3de9e75bd6003ba8b497

File: packages/fuels/src/setup-launch-node-server.ts (hunk 6 above)
Comment (nedsalk) [reply packages/fuels/src/setup-launch-node-server.ts]: https://github.com/FuelLabs/fuels-ts/pull/2378/commits/7a1ebbfe1a2944f5aebf3de9e75bd6003ba8b497

File: packages/fuels/src/setup-launch-node-server.ts (hunk 5 above)
Comment (nedsalk) [reply packages/fuels/src/setup-launch-node-server.ts]: https://github.com/FuelLabs/fuels-ts/pull/2378/commits/7a1ebbfe1a2944f5aebf3de9e75bd6003ba8b497

File: packages/fuels/src/setup-launch-node-server.test.ts (hunk 2 above)
Comment (nedsalk) [reply packages/fuels/src/setup-launch-node-server.test.ts]: https://github.com/FuelLabs/fuels-ts/pull/2378/commits/7a1ebbfe1a2944f5aebf3de9e75bd6003ba8b497

File: vitest.global-browser-setup.ts [hunk 9]
```diff
@@ -7,18 +7,29 @@ export default async function setup() {
...
+    };
+
+    cp.stdout?.on('data', () => {
+      // Return teardown function to be called when tests finish
+      // It will kill the server
+      resolve(killServer);
     });
 
     cp.on('error', (err) => {
+      // Ensure server is killed if there's an error
```
arboleya, line 24: Here are a couple more events that can be used (on parent process?) to detect non-clean exits.

Not sure which would be applicable here.

https://github.com/FuelLabs/fuels-ts/blob/bb5a123f46313f10d444d361fcfcf52e40f2430c/packages/account/src/test-utils/launchNode.ts#L284-L299

maschad (approved):
Awesome work!

danielbate (approved):
Great work 🚀 

I've also added a note to #1820 to fully incorporate the `@group browser` tag so we are fully browser testing in `fuel-gauge`

vitest.global-browser-setup.ts
nedsalk: Good that you commented on this! I analyzed killing a browser test suite mid-run and realized that it doesn't work as expected. I had to add an additional `/close-server` endpoint in https://github.com/FuelLabs/fuels-ts/pull/2378/commits/624f24b2e9f3065057a33b478fc9aa333b83c5de to make it work. The child process is run in a detached state to decouple it from the main process (to get away from polyfills) so killing it via this new endpoint was the easiest route - no shared `process` state and "I can kinda but not really kill the child process" shenanigans.

maschad (approved):

packages/fuels/src/setup-launch-node-server.ts
```diff
@@ -0,0 +1,142 @@
...
+    if (nodeUrl) {
+      const cleanupFn = cleanupFns.get(nodeUrl);
+      await cleanupFn?.();
+      res.end();
+    }
+  }
+});
+
+function closeServer() {
+  return new Promise<void>((resolve) => {
```
arboleya, line 71: Nit I.

(written via the UI, requires double-check)

```suggestion
const closeServer = (event: string) => (error?: unknown) => {
  console.log(event);
  if (error) {
    console.log(error);
  }
  return new Promise<void>((resolve) => {
```

packages/fuels/src/setup-launch-node-server.ts
```diff
@@ -0,0 +1,142 @@
...
+process.on('unhandledRejection', (reason) => {
+  console.log('unhandledRejection');
+  console.log(reason);
+
+  closeServer();
+});
+process.on('beforeExit', () => {
+  console.log('beforeExit');
+  closeServer();
+});
```
arboleya, line 142: Nit II.

(still, written via the UI)

```suggestion
process.on('exit', closeServer('exit'));
process.on('SIGINT', closeServer('SIGINT'));
process.on('SIGUSR1', closeServer('SIGUSR1'));
process.on('SIGUSR2', closeServer('SIGUSR2'));
process.on('uncaughtException', closeServer('uncaughtException'));
process.on('unhandledRejection', closeServer('unhandledRejection'));
process.on('beforeExit', closeServer('beforeExit'));
```

arboleya (dismissed):

packages/fuels/src/setup-launch-node-server.ts
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/2378/commits/9d281fd8a8f3d3e54005b9b6498d51f7749c22a3

packages/fuels/src/setup-launch-node-server.ts
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/2378/commits/9d281fd8a8f3d3e54005b9b6498d51f7749c22a3

packages/account/src/test-utils/setup-test-provider-and-wallets.ts
```diff
@@ -64,15 +64,33 @@ export async function setupTestProviderAndWallets({
     }
   );
 
-  const { cleanup, url } = await launchNode({
+  const launchNodeOptions = {
```
arboleya, line 67: Is there a type one can use for these options?

packages/account/src/test-utils/setup-test-provider-and-wallets.ts
```diff
@@ -64,15 +64,33 @@ export async function setupTestProviderAndWallets({
...
       defaultSnapshotConfigs,
       walletsConfig.apply(nodeOptions?.snapshotConfig)
     ),
     port: '0',
-  });
+  };
+
+  let cleanup: () => void;
+  let url: string;
+  if (process.env.LAUNCH_NODE_SERVER_PORT) {
```
arboleya, line 79: I believe the utility should work purely from arguments.

Env variables are counterintuitive and shouldn't be used in anything related to `testNodeLaunch`.

arboleya (approved):

packages/account/src/test-utils/setup-test-provider-and-wallets.ts
nedsalk: Used type in https://github.com/FuelLabs/fuels-ts/pull/2378/commits/c6005bee43e6cc1a76107a13cbafeec54f946dbe

packages/account/src/test-utils/setup-test-provider-and-wallets.ts
nedsalk: Moved to argument in https://github.com/FuelLabs/fuels-ts/pull/2378/commits/c6005bee43e6cc1a76107a13cbafeec54f946dbe. I didn't document the new `launchNodeServerPort` argument because the infrastructure added in the PR is still internal. When we decide to make it public, we can add proper documentation and showcase the bigger picture properly.

arboleya (approved):

maschad (approved):

Issue #2400, opened 2024-05-28
Upgrade `forc` to `0.60.0`


arboleya:
Release notes:
 - https://github.com/FuelLabs/sway/releases/tag/v0.60.0

PR #2401, opened 2024-05-28
chore: upgrade forc 0.60.0

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/2400


packages/abi-typegen/test/fixtures/templates/predicate-with-configurable/factory.hbs
```diff
@@ -78,7 +78,7 @@ const _abi = {
         "type": 2,
         "typeArguments": null
       },
-      "offset": 560
+      "offset": 1152
```
danielbate, line 81: How much bigger are the contract sizes? jw as the offset has doubled 😬

.changeset/six-bobcats-impress.md
```diff
@@ -0,0 +1,6 @@
+---
+"@fuel-ts/versions": patch
+"@internal/forc": patch
```
nedsalk, line 3: There shouldn't be a need for this package in the changesets anymore. We should ignore it.

nedsalk (approved):

packages/abi-typegen/test/fixtures/templates/predicate-with-configurable/factory.hbs
Torres-ssf: @danielbate How can I verify that?

danielbate: Before/after the file size of the bin file. Tbh we can't do anything about this so this won't be a blocker.

Torres-ssf: @danielbate Based on the example that I picked, it appears they've increased ~ 18%

PR #2404, opened 2024-05-28
fix: reading of sway types when loading `fuels.config.ts`

### LINKED ISSUE #2402: Typegen does not build types for a workspace if the program type starts with a comment
We recently added `fuels build` to `docs-snippets` so that we could start utilising typegen in tests. Not all tests are using it yet so it has not been a problem, however I have noticed that if a program types entry file starts with a comment it is not picked up by typegen. This is common with a lot of our `docs-snippets` program type files where we are using a snippet region:

https://github.com/FuelLabs/fuels-ts/blob/80e0c2361306ffaeb43f35a7b3db4f47706a4f9c/apps/docs-snippets/test/fixtures/forc-projects/sum-script/src/main.sw#L1-L11

nedsalk:
closes #2402


danielbate (changes requested):
Missing a changeset, but looking good otherwise

PR #2413, opened 2024-05-29
fix: `enum` size validation

danielbate:
Currently, we are validating the size of an `enum` against the [maximum](https://github.com/FuelLabs/fuels-ts/blob/5a6ca46ef46b856db780562b455dd9d742977404/packages/abi-coder/src/encoding/coders/EnumCoder.ts#L38-L41) size of a child value. If an `enum` contains both a `u8` and a `b512` and we pass the `u8` value, we'd expect this to throw for underflow. It is not, as we are incorrectly [validating](https://github.com/FuelLabs/fuels-ts/blob/5a6ca46ef46b856db780562b455dd9d742977404/packages/abi-coder/src/encoding/coders/EnumCoder.ts#L83) against the value size rather than the `encodedLength` which does not include the case length. 

This PR now correctly checks for `encodedLength`, as well as setting this value to the size of the smallest coder. We will then do another size check once we have checked the case key, so have the actual values expected length to validate against.

> **NOTE:**  `EnumCoder` is the only coder that calculates the `encodedLength` using `reduce` on the child coders, so this will not impact other coders.

### DIFF
--- a/packages/abi-coder/src/encoding/coders/EnumCoder.ts
+++ b/packages/abi-coder/src/encoding/coders/EnumCoder.ts
@@ -3,7 +3,7 @@ import { toNumber } from '@fuel-ts/math';
 import { concat } from '@fuel-ts/utils';
 import type { RequireExactlyOne } from 'type-fest';
 
-import { OPTION_CODER_TYPE, WORD_SIZE } from '../../utils/constants';
+import { OPTION_CODER_TYPE } from '../../utils/constants';
 import { hasNestedOption } from '../../utils/utilities';
 
 import type { TypesOfCoder } from './AbstractCoder';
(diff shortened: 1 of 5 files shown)


packages/abi-coder/src/encoding/coders/EnumCoder.ts
```diff
@@ -80,7 +80,7 @@ export class EnumCoder<TCoders extends Record<string, Coder>> extends Coder<
   }
 
   decode(data: Uint8Array, offset: number): [DecodedValueOf<TCoders>, number] {
-    if (this.#shouldValidateLength && data.length < this.#encodedValueSize) {
+    if (this.#shouldValidateLength && data.length < this.encodedLength) {
```
danielbate, line 83: Here we were not including the case length in the calculation. Meaning `#encodedValueSize` was `1` if the value was `u8`. An enum is always at least `WORD_SIZE` due to the case.

packages/abi-coder/src/encoding/coders/EnumCoder.ts
```diff
@@ -36,7 +36,7 @@ export class EnumCoder<TCoders extends Record<string, Coder>> extends Coder<
   constructor(name: string, coders: TCoders) {
     const caseIndexCoder = new BigNumberCoder('u64');
     const encodedValueSize = Object.values(coders).reduce(
-      (max, coder) => Math.max(max, coder.encodedLength),
+      (min, coder) => Math.min(min, coder.encodedLength),
```
danielbate, line 39: Use the minimum for now, we can check the actual length later once we've decoded the case.

arboleya (approved):

PR #2433, opened 2024-05-31
feat: emit the `build` and `deploy` events during `fuels dev`

### LINKED ISSUE #2423: Fire the `build` and `deploy` events during `fuels dev`
As of now, we don't fire the `build` and `deploy` events during an ongoing `fuels dev` process. 

We should fire them, as this would let people to write custom post-deployment and post-build logic even while using `fuels dev`. Right now, this is only possible if they use `fuels build` and `fuels deploy` separately.

Dhaiwat10:
Closes #2423

### DIFF
--- a/packages/fuels/src/cli/commands/dev/index.ts
+++ b/packages/fuels/src/cli/commands/dev/index.ts
@@ -3,7 +3,7 @@ import { watch } from 'chokidar';
 import { globSync } from 'glob';
 
 import { loadConfig } from '../../config/loadConfig';
-import type { FuelsConfig } from '../../types';
+import { Commands, type FuelsConfig } from '../../types';
 import { error, log } from '../../utils/logger';
 import { build } from '../build';
 import { deploy } from '../deploy';
@@ -18,7 +18,24 @@ export const closeAllFileHandlers = (handlers: FSWatcher[]) => {
 
 export const buildAndDeploy = async (config: FuelsConfig) => {
   await build(config);
-  return deploy(config);
+  config.onSuccess?.(
+    {
+      type: Commands.build,
+      data: null,
+    },
+    config
+  );
+
+  const deployedContracts = await deploy(config);
+  config.onSuccess?.(
+    {
+      type: Commands.deploy,
+      data: deployedContracts,
+    },
+    config
+  );
+
+  return deployedContracts;
 };
 
 export const getConfigFilepathsToWatch = (config: FuelsConfig) => {
--- a/.changeset/four-seas-stare.md
+++ b/.changeset/four-seas-stare.md
@@ -1,0 +1,5 @@
+---
+"fuels": patch
+---
+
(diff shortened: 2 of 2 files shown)


packages/fuels/src/cli/commands/dev/index.ts
```diff
@@ -18,7 +18,24 @@ export const closeAllFileHandlers = (handlers: FSWatcher[]) => {
...
 export const buildAndDeploy = async (config: FuelsConfig) => {
   await build(config);
-  return deploy(config);
+  config.onSuccess?.(
+    {
+      type: Commands.build,
+      data: null,
+    },
+    config
+  );
```
arboleya, line 27: Probably a good idea to move this inside the `build` and `deploy` methods.

Also, we should consider removing the `onSucces` in favor of `onBuild`, `onDeploy`, etc.

arboleya (changes requested):

packages/fuels/src/cli/commands/dev/index.ts
arboleya: What if we removed the `onSucces` in favor of `onBuild`, `onDeploy`, etc.?

Dhaiwat10: @arboleya that would cover the `build`, `deploy` and `dev` commands. Do we have more commands that use `onSuccess` to your knowledge?

arboleya: Only if we want to support the [other commands](https://github.com/FuelLabs/fuels-ts/tree/master/packages/fuels/src/cli/commands):
 - `init` / onInit?
 - `node` / onNode?

There could also be an `onDev` event that fires up every time a hot reload occurs, for example. So it would look like:

 1. `onBuild`
 1. `onDeploy`
 1. `onDev`

Dhaiwat10: I would be in favour of supporting all of them.

Dhaiwat10: And getting rid of `onSuccess` if we add all of these (?) @arboleya

arboleya: Sounds like a plan. 👌

arboleya: Wonderful. 🙌

arboleya: Yeah, I prefer the second option as well.

fuel-service-user:
✨ A PR has been created under the `rc-2433` tag on `fuels-wallet` repo.
https://github.com/FuelLabs/fuels-wallet/pull/1395

arboleya (approved):
Great work! 🎯

maschad (approved):

nedsalk (approved):

PR #2458, opened 2024-06-07
chore: hotfix receipt call

LuizAsFight:
created from https://github.com/FuelLabs/fuels-ts/pull/2407 which was created from `0.88.1`


LuizAsFight:
close in favor of https://github.com/FuelLabs/fuels-ts/pull/2479

Issue #2454, opened 2024-06-06
Remove `[skip ci]`  from dependabot workflow


maschad:
We should ensure that our dependabot CI jobs pass before skipping the CI in the changeset commit. We can achieve this by committing the changesets and allowing the CI to run afterwards.

nedsalk:
This is the user info set for the changesets PR: https://github.com/luizstacio/changesets-action/blob/6a46d2c28377497e64763ba5f1faeabbd3d88c33/src/gitUtils.ts#L4

PR #2488, opened 2024-06-10
build: update dependabot workflow to use REPO_TOKEN for changesets

maschad:
In our current workflow, we run the [validate changesets](https://github.com/FuelLabs/fuels-ts/blob/master/.github/workflows/pr-validate-changesets.yaml#L44-L45) independently of our [Tests workflow](https://github.com/FuelLabs/fuels-ts/blob/master/.github/workflows/test.yaml) on most occasions this is acceptable, but in the case of dependabot opening a PR, this means dependabot will create a changeset for a dep upgrade which _may_ cause breakages.

Also because of the [`skip_ci` tag ](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs) the resulting PR would remain in a _hanging_ state which made reviews more cumbersome. 

This PR seeks to rectify both issues by removing the `skip_ci` since the github actions bot [will not trigger a new workflow](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow) and ensuring that this workflow runs after the tests

Closes https://github.com/FuelLabs/fuels-ts/issues/2454


arboleya (commented):
Shouldn't `secrets.GITHUB_TOKEN` be replaced by `secrets.REPO_TOKEN` for using the `fuels-service-user`? cc @nedsalk

maschad:
> Shouldn't `secrets.GITHUB_TOKEN` be replaced by `secrets.REPO_TOKEN` for using the `fuels-service-user`? cc @nedsalk

I've opted against using the @fuel-service-user since we don't need to worry about the recursive workflow problem with github actions bot (see description for more details)

arboleya:
@chad IRRC, the problem is not with the recursive workflow but with the GitHub Action BOT, which may not trigger the workflows as expected. We can test it like this and create a follow-up PR if necessary.

arboleya (approved):

maschad:
> @maschad IRRC, the problem is not with the recursive workflow but with the GitHub Action BOT, which may not trigger the workflows as expected. We can test it like this and create a follow-up PR if necessary.

@arboleya I believe you are correct, the commits which have been [authored by dependabot](https://github.com/FuelLabs/fuels-ts/commits/master/?author=dependabot%5Bbot%5D) are not running the workflows when merged to master yet obviously those [authored by the fuel service user](https://github.com/FuelLabs/fuels-ts/commits/master/?author=fuel-service-user) are. 

 I've updated to use the `REPO_TOKEN` as that should elevate dependabot's permissions, we will need it to access the [secrets](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#accessing-secrets) though as @petertonysmith94 had suggested some time ago.

.github/workflows/pr-validate-changesets.yaml
```diff
@@ -6,6 +6,10 @@ on:
       - opened
       - edited
       - synchronize
+  workflow_run:
+    workflows: ["Tests"]
+    types:
+      - completed
```
nedsalk, line 12: This doesn't seem necessary anymore as the changeset commit push from `fuel-service-user` will cancel the initial runs.

maschad: Good catch, I had originally theorized we'd only want to validate the changesets once the tests have run but this will just prolong the workflow runs now.

nedsalk (approved):

arboleya (approved):

danielbate (approved):

PR #2489, opened 2024-06-10
chore: add integration test for `create fuels`

petertonysmith94:
- Closes: #2170 

---

The full `integration` tests (using published versions) will now only run on `master` and the latest changeset release PR (`changeset-release/master`). We will use the published `next` tag for running the tests against, and will no longer publish a new version for every test run.

### DIFF
--- a/.changeset/serious-dogs-wash.md
+++ b/.changeset/serious-dogs-wash.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/account": patch
+---
+
+chore: add browser testing infrastructure
--- a/.changeset/tender-coins-bathe.md
+++ b/.changeset/tender-coins-bathe.md
@@ -1,0 +1,6 @@
+---
+"@fuel-ts/abi-coder": patch
+"@fuel-ts/account": patch
+---
+
+build(deps-dev): bump prettier from 3.0.3 to 3.3.2
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,3 +11,4 @@ updates:
     commit-message:
       prefix: "build(deps):"
       prefix-development: "build(deps-dev):"
+    versioning-strategy: "increase"
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -1,13 +1,16 @@
 name: Pull Request Labeler
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, edited]
+
+permissions:
+  contents: read
+  pull-requests: write
 
 jobs:
   label_pull_request:
     runs-on: ubuntu-latest
-    if: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
     steps:
       - uses: fuxingloh/multi-labeler@v4
         with:
--- a/apps/docs-snippets/test/fixtures/abi/encode-and-decode.jsonc
+++ b/apps/docs-snippets/test/fixtures/abi/encode-and-decode.jsonc
@@ -6,8 +6,8 @@
       "typeId": 0,
       "type": "u32",
       "components": null,
-      "typeParameters": null
-    }
+      "typeParameters": null,
+    },
   ],
   "functions": [
     {
@@ -15,17 +15,17 @@
         {
           "name": "inputted_amount",
           "type": 0,
-          "typeArguments": null
-        }
+          "typeArguments": null,
+        },
       ],
       "name": "main",
       "output": {
         "name": "",
         "type": 0,
-        "typeArguments": null
+        "typeArguments": null,
       },
-      "attributes": null
-    }
+      "attributes": null,
+    },
   ],
   "loggedTypes": [],
   "messagesTypes": [],
@@ -35,11 +35,11 @@
       "configurableType": {
         "name": "",
         "type": 0,
-        "typeArguments": null
+        "typeArguments": null,
       },
-      "offset": 784
-    }
-  ]
+      "offset": 784,
+    },
+  ],
 }
 
 
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     "npm-run-all": "^4.1.5",
     "nyc": "^15.1.0",
     "open": "^8.4.0",
-    "prettier": "^3.0.3",
+    "prettier": "^3.3.2",
     "rimraf": "^3.0.2",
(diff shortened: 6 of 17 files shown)


arboleya (commented):
Conflicts to fix.

scripts/create-fuels-template-integration.sh
```diff
@@ -1,7 +1,64 @@
...
+if [ -d "$PROJECT_DIR" ]; then
+  echo "Removing existing project directory '$PROJECT_DIR'"
+  rm -rf $PROJECT_DIR
+fi
+pnpm create fuels@$PUBLISHED_NPM_VERSION $PROJECT_DIR --pnpm -cps --no-install
+
+echo "3. Intialise the project"
+cd $PROJECT_DIR
+pnpm add fuels@$PUBLISHED_NPM_VERSION > /dev/null 2>&1
+pnpm  --ignore-workspace install > /dev/null 2>&1
```
arboleya, line 40: Why do we have this?

Shouldn't it be automatic via `fuels@$PUBLISHED_NPM_VERSION`?

scripts/create-fuels-template-integration.sh
```diff
@@ -1,7 +1,64 @@
 #!/bin/bash
 
-(cd templates/nextjs && pnpm run fuels:dev) &
-cp templates/nextjs/.env.example templates/nextjs/.env.local
-(cd templates/nextjs && pnpm run dev) &
+# Debugging
+# PUBLISHED_NPM_VERSION="next"
```
maschad, line 4: Remove this once you've finished debugging

scripts/create-fuels-template-integration.sh
```diff
@@ -1,7 +1,64 @@
...
+
+# Versions
+FUEL_CORE_VERSION="0.28.0"
```
maschad, line 7: Will these have to be manually updated every upgrade? We should be able to retrieve it from https://github.com/FuelLabs/fuels-ts/blob/master/internal/fuel-core/VERSION#L0-L1 no?

scripts/create-fuels-template-integration.sh
arboleya: Hmm, I think I'm just making many high-level assumptions here.

I'll probably have to review it again.

scripts/create-fuels-template-integration.sh
maschad: I'd say remove it all together, given it's passed to the script as an env variable it's self explanatory. 
```suggestion

```

maschad: I don't have a strong opinion on it as it's relatively minor, but if we are going to keep it, then perhaps we should update the comment to say 

```suggestion
# The PUBLISHED_NPM_VERSION variable is required to create a new project with the current published `fuels` version
```

maschad (dismissed):
Nice work 🚀

.github/workflows/test-integration.yaml
```diff
@@ -0,0 +1,32 @@
+name: "Test Integration"
+
+on:
+  workflow_run:
+    workflow: [Release]
```
nedsalk, line 5: The `Release` workflow runs only on [push](https://github.com/FuelLabs/fuels-ts/blob/dc17b9c454fb025cd07840e79489afe346ff6114/.github/workflows/release.yaml#L3) to `master`/`release/*` branches, while in #2170 
> The idea is that every PR would publish a PR tag and have an install/assertion routine.

This would mean that we'd catch* regressions only _after_ merge to `master`. Am I missing something here?

*and only if we go to the actions tab and find that the workflow failed.

arboleya: Can't we just use `needs`?

https://github.com/FuelLabs/fuels-ts/blob/eefe0079eb253fcbf94d09a26c102d35366c696d/.github/workflows/test.yaml#L104

arboleya: Oh, I see; we are talking cross-workflows. I'm not sure, maybe not.

maschad (dismissed):

Torres-ssf (approved):

.github/workflows/test-integration.yaml
arboleya: That's perfect! 💯 

Feel free to resolve this conversation when it's done. (not sure if it is already)

arboleya (approved):

maschad (approved):

arboleya (approved):

arboleya (approved):

danielbate (approved):

.github/workflows/test.yaml
```diff
@@ -60,32 +59,36 @@ jobs:
...
+      - name: Run UI tests (on template)
+        run: sh ./scripts/tests-ui.sh
 
-      - name: Pretest
-        run: pnpm pretest
+      - name: Wait for release to succeed
+        if: github.ref_name == 'master' || github.head_ref == 'changesets-release/master'
+        uses: lewagon/wait-on-check-action@v1.3.4
+        with:
+          ref: ${{ github.ref }}
```
nedsalk, line 71: From the [environment variable docs](https://docs.github.com/en/actions/learn-github-actions/variables) for `GITHUB_REF`:
> 	The fully-formed ref of the branch or tag that triggered the workflow run. For workflows triggered by push, this is the branch or tag ref that was pushed. For workflows triggered by pull_request, this is the pull request merge branch. For workflows triggered by release, this is the release tag created. For other triggers, this is the branch or tag ref that triggered the workflow run. This is only set if a branch or tag is available for the event type. The ref given is fully-formed, meaning that for branches the format is refs/heads/<branch_name>, for pull requests it is refs/pull/<pr_number>/merge, and for tags it is refs/tags/<tag_name>. For example, refs/heads/feature-branch-1.

I think that this won't work as expected for changeset PRs because the step will be waiting for the `Release` workflow on `github.ref` which would be `refs/pull/<pr_number>/merge` and there won't be such a workflow running because `Release` is running only on `push` to `master` and `release/*` branches.

 I think that the value here should be hardcoded to `refs/heads/master`.

nedsalk (changes requested):

Torres-ssf (approved):

nedsalk (approved):

arboleya (approved):
:shipit:

danielbate (approved):

PR #2497, opened 2024-06-12
chore: add infrastructure to patch latest version

### LINKED ISSUE #2245: Implement a way to patch the last release
We recently added means for patching old releases, which was fantastic!

However, we still haven't covered one scenario: patching the _last_ release.

Imagine the scenario:

 1. We release `v0.80.0`
 1. One day later, we have a new changesets PR that will bump things to `v0.81.0`
 1. Before releasing `v0.81.0`, we find an issue and need to make a `v0.80.1` patch

This is the scenario we need to cover.

nedsalk:
This PR adds infrastructure to patch the _last_ release. Given that this workflow releases commits that are not on `master`, we'll have to manually apply those commits' changes to `master` if the released code doesn't already exist on it.

Besides the release of the package itself, the documentation also gets updated when this scenario occurs.

closes #2245


arboleya (commented):
Should we document the process as we did for patching old releases?

https://github.com/FuelLabs/fuels-ts/blob/269197f27f316557100998f42ace425d9d8441b1/CONTRIBUTING.md?plain=1#L240-L248

nedsalk:
@arboleya thanks for pointing it out! https://github.com/FuelLabs/fuels-ts/pull/2497/commits/acb4b447cbbd16437050e6a497e9464ed22a6033

There is room possible room for automation here with regards to the point
> - Update the package versions on `master` to match the released versions,

but I don't know if it's worth it. I could just comment out [this condition](https://github.com/FuelLabs/fuels-ts/blob/acb4b447cbbd16437050e6a497e9464ed22a6033/.github/workflows/pr-validate-changesets.yaml#L132) to have the step run on all PRs and we'd catch it right away. I'd just have to add an additional check to make sure that cases where the versions don't change are handled:
```bash
OLD_VERSION=$(sed -nE 's/^\s*"version": "(.*?)",$/\1/p' packages/fuels/package.json)
pnpm changeset version
VERSION=$(sed -nE 's/^\s*"version": "(.*?)",$/\1/p' packages/fuels/package.json)
git reset --hard

if [ "$OLD_VERSION" = "$VERSION" ]; then
  # the versions didn't change so we won't be releasing it anyways
  exit 0
fi

STATUS_CODE=$(curl -s -w '%{http_code}\n' "https://www.npmjs.com/package/fuels/v/$VERSION" | tail -n1)
if [[ $STATUS_CODE  != 404 ]]; then
  echo "Release for version $VERSION already exists or curl received an unexpected result (result is $STATUS_CODE). Exiting."
  exit 1
else
  exit 0
fi
```

.github/workflows/pr-validate-changesets.yaml
```diff
@@ -128,15 +128,21 @@ jobs:
           echo $CHANGES | grep -E 'patch' --silent && echo "Patch changes found." || (echo "No patch changes found." && exit 1)
           echo $CHANGES | grep -E 'minor|major' --silent && echo "Old releases can only be patched; no minor and major versions allowed." && exit 1 || echo "No minor nor major changes."
 
-      - name: Validate that there was no release for the next patch version
-        if: startsWith(github.base_ref, 'release/')
+      - name: Validate that there was no release for the next version
```
nedsalk, line 131: There is possible room for automation here with regards to the contributing guide's [guideline](https://github.com/FuelLabs/fuels-ts/pull/2497/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R260)

> - Update the package versions on `master` to match the released versions,

like automatically creating a new PR to update the versions, but I don't think it's worth it because `package.json`s might have been changed not only in their version but also in their dependencies and then manual work has to be done to fix that. The work to update the versions manually on `master` IMO isn't big enough to warrant automation.

Rather, I just made this step run on all PRs and we'll catch this version issue right away and manually fix it. The changes to this step reflect that logic.

CONTRIBUTING.md
```diff
@@ -244,7 +244,21 @@ The following example is for releasing a patch for `v0.69.0` -> `v0.69.1`.
 - Create PRs with base set to that release branch
   - When the PR is merged, a changeset PR is created
   - When the changeset PR is merged into the release branch, the next patch version is released and the commit is tagged (e.g. `v0.69.1`)
-- After release, delete the release branch from GitHub
+- After release, the release branch will be automatically deleted
```
arboleya, line 247: When patching the _latest release_, this last step could be to open a new PR using the same branch but against `master`. This would close the gap and re-pass the updates to the ongoing changesets PR (and upcoming release).

CONTRIBUTING.md
```diff
@@ -244,7 +244,21 @@ The following example is for releasing a patch for `v0.69.0` -> `v0.69.1`.
...
+
+# Patching latest release
+
+Imagine the scenario:
+
+1. We release `v0.80.0`
+1. One day later, we have a new changesets PR that will bump things to `v0.81.0`
+1. Before releasing `v0.81.0`, we find an issue and need to make a `v0.80.1` patch
```
arboleya, line 255: Would it work if the found issue needed a breaking change (minor bump) to be fixed?

CONTRIBUTING.md
```diff
@@ -244,7 +244,21 @@ The following example is for releasing a patch for `v0.69.0` -> `v0.69.1`.
...
+
+We'd follow the same approach as explained in the [Patching old releases](#patching-old-releases) section above, bearing in mind the following after the release:
+
+- Update the package versions on `master` to match the released versions,
+- If the released commits' functionality is not on `master`, you have to manually add it via a new PR to `master`
+- If the released commits' functionality is already on `master`, you don't have to do anything
```
arboleya, line 261: In this scenario, updating `master` should be mandatory, no?

Maybe the guide could explain everything til the end.

CONTRIBUTING.md
nedsalk: Currently not because of [this check](https://github.com/FuelLabs/fuels-ts/blob/e165e37307d65deda99f675723ea26470bf807d8/.github/workflows/pr-validate-changesets.yaml#L129). I can change the logic so that releases older than `latest` can only be patched, but `latest` can have a minor bump. What do you think?

CONTRIBUTING.md
nedsalk: I explained that the versions have to be updated in the first bullet point (meaning it's mandatory), but if the released functionality already exists on `master` then we don't have to do anything on that front. What else would you add/rephrase here?

CONTRIBUTING.md
nedsalk: Do you mean that we automate it in the CI or just explain it to the user to do it manually?
In both cases we'd have to have an exception to not delete the release branch when it's a patch to the latest release.

I touched upon automation in [this comment](https://github.com/FuelLabs/fuels-ts/pull/2497/files#r1636888919), but after a bit of investigation it doesn't seem like a particularly hard thing to automate. Initially I thought that it's not possible to open a PR when it's got merge conflicts with the base branch, but I just took a look on our repo with some old branches (e.g. trying to create a new PR from https://github.com/FuelLabs/fuels-ts/tree/aa/feat/new-abi-spec to `master`, which has merge conflicts) and it seems possible. Do you see any potential blockers to automate this?

arboleya: No, I don't. But if that's too much trouble, we could simply add some instructions.

 1. Clone your released tag branch with `git <...>`
    > If the branch was deleted, they'll have to recreate it.
 3. Create PR against `master`, etc.

CONTRIBUTING.md
arboleya: I think it's not obvious what "versions should be updated" means.

Is it saying that one must re-open their PR against `master`?

CONTRIBUTING.md
arboleya: That's alright.

.github/workflows/pr-validate-changesets.yaml
nedsalk: I have automated it [here](https://github.com/FuelLabs/fuels-ts/pull/2497/files#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05adaR174), it wasn't complex.

CONTRIBUTING.md
nedsalk: @arboleya I have automated it [here](https://github.com/FuelLabs/fuels-ts/pull/2497/files#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05adaR174) and updated the `CONTRIBUTING.md` file accordingly.

CONTRIBUTING.md
nedsalk: This part got changed because I automated the PR creation as mentioned in https://github.com/FuelLabs/fuels-ts/pull/2497#discussion_r1639851406.

CONTRIBUTING.md
nedsalk: This has been implemented [here](https://github.com/FuelLabs/fuels-ts/pull/2497/files#diff-900c14db30cf12f70161bd32e69363c032130443d4f052508d970d7d2fbfa472R124).

nedsalk:
@petertonysmith94 The _actual full workflow_ would be tested in the trenches after we merge it into `master` and release a new version of `master`. We'd then run this workflow on that `release/v0.xx.x` branch by doing a simple, non-breaking change and verifying that everything works as expected. :sweat_smile:  
However, I tested all the added logic in the CI in isolation, e.g. #2529 for PR creation and running the bash scripts in the CI on mock data to verify that there's no failure in them.

maschad (approved):

Torres-ssf (approved):

danielbate (approved):

arboleya (approved):
Firstly, thanks for addressing this final piece on our release process.

Having said that, I couldn't properly validate it. I mean no offense, but the changes in this PR are too convoluted and unreadable (hard to follow). If we add the fact that it is purely text (CI/YAML), the job becomes even more challenging. It is like validating code on a piece of paper.

I know others have approved it, but I wonder if they could accurately explain it back to you.

I'll go ahead and approve it as is so you can validate it yourself in practical terms. I suggest doing that in isolation at the beginning of your day so you can see it through and make any extra fixes on the spot if necessary.

nedsalk:
@arboleya you're right that these scripts aren't the most readable code. I have created #2581 where we can evolve the discussion and hopefully come to a better CI workflow codebase.

Issue #1795, opened 2024-02-23
Refactor the `ABI Coder` package


danielbate:
This issue should act as the epic with other issues spun off from here. 

The goals of this refactor are to:

- Reduce the coder touch points number of specific coder classes
- Decouple `encode` / `decode` and length validation logic
- Introduce a more functional approach (especially including any functionality covered by [private methods](https://github.com/FuelLabs/fuels-ts/blob/master/packages/abi-coder/src/coders/v1/enum.ts#L49))
- Improve error handling eminating from coders to enable #1552 
- Rethink the public entry points for `ABI Coder`
- Complete E2E testing for all supported types (_could_ be based on #2466)
- Deprecate `JsonAbiOld`
- Deprecate `UtxoIdCoder`

nedsalk:
If changes happen to coders so that they don't return the `offset` value anymore, we should also update the `Interface.decodeLog` and `Interface.decodeFunctionResult` methods accordingly. See #2514.

arboleya:
Closing in favor of:
 - https://github.com/FuelLabs/fuels-ts/issues/3031

PR #2514, opened 2024-06-13
feat!: remove offset from `decodeLog` and `decodeFunctionResult` methods

nedsalk:
`Interface.decodeLog` and `Interface.decodeFunctionResult` are top-level functions that don't need to return a data offset because you pass in _all_ the data and the returned offset is always the length of that data. If somebody wants to get the `offset` value after this change, they can just call `data.length`.

```ts
const theInterface = new Interface(...);

// previously
const [decodedLog, offset] = theInterface.decodeLog(data, logId);
// now
const decodedLog = theInterface.decodeLog(data, logId);

// previously
const [decodedFnResult, offset] = theInterface.decodeFunctionResult(fnName, data);
// now
const decodedFnResult = theInterface.decodeFunctionResult(fnName, data);

```

I tagged it as a `feat` because it improves the DX of the methods.

Supersedes #2308


fuel-service-user:
✨ A PR has been created under the `rc-2514` tag on `fuels-wallet` repo.
https://github.com/FuelLabs/fuels-wallet/pull/1374

arboleya (commented):
Are there downsides to removing the `offset` property?

Can someone be looking for the `offset` in the future?

nedsalk:
@arboleya the offset value is equal to `data.length` so people can always get it that way. I updated the description to be more explicit about this.

arboleya:
# Pattern

Could we be breaking the patterns already in use?

https://github.com/FuelLabs/fuels-ts/blob/e165e37307d65deda99f675723ea26470bf807d8/packages/transactions/src/coders/transaction.ts#L104-L133
Wether this is a good or a bad thing, I'll let other chime in.

```ts
let decoded;
let o = offset;

[decoded, o] = new BigNumberCoder('u64').decode(data, o);
const scriptGasLimit = decoded;

[decoded, o] = new B256Coder().decode(data, o);
const receiptsRoot = decoded;

// ..
```


# Exceptions

Also, it seems some Coders have non-obvious ways of computing the offset:

https://github.com/FuelLabs/fuels-ts/blob/e165e37307d65deda99f675723ea26470bf807d8/packages/abi-coder/src/encoding/coders/B256Coder.ts#L43

https://github.com/FuelLabs/fuels-ts/blob/e165e37307d65deda99f675723ea26470bf807d8/packages/abi-coder/src/encoding/coders/VecCoder.ts#L63-L69

# API Preferences

Regarding personal preference, I'd go with Objects instead of Arrays.

But only if we'd change the pattern everywhere, which is not the case.

```ts
const { decodedLog, offset } = theInterface.decodeLog(data, logId);
// vs
const [decodedLog, offset] = theInterface.decodeLog(data, logId);
```

# Conclusion

I'm not sure if the breaking change is worth it.

The change is subtle, but the latter hides information.

```ts
const [decodedLog] = theInterface.decodeLog(data, logId);
// vs
const decodedLog = theInterface.decodeLog(data, logId);
```

As for if this is a good or a bad thing, I'll let others chime in.

arboleya (commented):
# Pattern

Could we be breaking the patterns already in use?

https://github.com/FuelLabs/fuels-ts/blob/e165e37307d65deda99f675723ea26470bf807d8/packages/transactions/src/coders/transaction.ts#L104-L133


# Exceptions

Also, it seems some Coders have non-obvious ways of computing the offset:

https://github.com/FuelLabs/fuels-ts/blob/e165e37307d65deda99f675723ea26470bf807d8/packages/abi-coder/src/encoding/coders/B256Coder.ts#L43

https://github.com/FuelLabs/fuels-ts/blob/e165e37307d65deda99f675723ea26470bf807d8/packages/abi-coder/src/encoding/coders/VecCoder.ts#L63-L69

# API Preferences

Regarding personal preference, I'd go with Objects instead of Arrays.

But only if we'd change the pattern everywhere, which is not the case.

```ts
const { decodedLog, offset } = theInterface.decodeLog(data, logId);
// vs
const [decodedLog, offset] = theInterface.decodeLog(data, logId);
```

# Conclusion

I'm not sure if the breaking change is worth it.

The change is subtle, but the latter hides information.

```ts
const [decodedLog] = theInterface.decodeLog(data, logId);
// vs
const decodedLog = theInterface.decodeLog(data, logId);
```

As for if this is a good or a bad thing, I'll let others chime in.

nedsalk:
@arboleya you're right, introducing this would break the pattern present in other coders; I didn't think of that. If I remember correctly, there was talk on the offsite around not needing the offset value across the board on all our coders and that this would be considered in #1795. 

The argument for merging it right now is that this is a high level API relative to the coders and we can get the DX benefits of it right now because the offset value is never needed when using these methods. The counter-argument is that API consistency is more important and that it should be done in tandem with #1795.

In the end, I agree with your uncertainty
> I'm not sure if the breaking change is worth it.

and I'll close this PR. I've added a [comment](https://github.com/FuelLabs/fuels-ts/issues/1795#issuecomment-2172875295) on #1795 linking to this PR.

PR #2558, opened 2024-06-19
fix!: stop piping into `process.stdout/stderr` and use `console.log`

### LINKED ISSUE #2248: Stop silencing errors during `launchNode` startup
When using the `launchNode` utility, some node startup errors derived from improper configs are silenced.

This makes debugging issues tricky and prevents the utility from working as expected.

> Reported by @LuizAsFight

### LINKED ISSUE #2483: Stop silencing errors during `fuels init` command
I was starting a new TS workspace and ran `pnpm fuels init`, which kept failing to create any output.

Eventually I noticed an error was briefly flashing on my terminal before being erased. Using a screen recorder, I could see the following error was flashing:

```Error: required option '-w --workspace <path>' not specified```

First of all, [the docs](https://docs.fuel.network/docs/fuels-ts/fuels-cli/#next-step) should provide more information about the required flags.

Second, the CLI should make sure errors like this are visible to the user (without needing a screen recorder 😁).

### PR DESCRIPTION
**Pull request by nedsalk**
This PR substitutes `process.stdout` and `process.stderr` usage with `console.log` in our whole codebase because the logs would sometimes get removed once the process finishes.

- closes #2483
- closes #2248

# Breaking changes

Removed `debugEnabled` property from `launchNode` inputs, which was used to read from `stdout` which isn't used by `fuel-core` internally.

### CODE REVIEW

File: packages/fuels/src/cli/commands/init/index.ts [hunk 1]
```diff
@@ -1,21 +1,24 @@
...
+import type { Command, Option } from 'commander';
 import { existsSync, writeFileSync } from 'fs';
 import { globSync } from 'glob';
 import { join, relative, resolve } from 'path';
 
 import { renderFuelsConfigTemplate } from '../../templates/fuels.config';
 import { log } from '../../utils/logger';
 
+interface CommandWithInternals extends Command {
+  options: (Option & { conflictsWith: string[] })[];
```
nedsalk, line 10: I created this interface to be pragmatic and just take out what is needed without having to introduce a structural refactoring to get the `Option`s and their flags that are used [below](https://github.com/FuelLabs/fuels-ts/pull/2558/files#diff-1d2833f29f77c976e789bbe891b340cf95b11435b775736098d1887ed7d6ac14R53). If the property ever becomes private, our tests will fail and we can then do a refactor.

packages/fuels/src/cli/commands/init/index.ts
```diff
@@ -37,29 +40,41 @@ export function init(program: Command) {
...
+    const errorStringPadding = errorString.replaceAll(/./g, ' ');
+
+    const workspaceOption = (program as CommandWithInternals).options.find((o) => o.short === '-w')
+      ?.flags;
+    const optionsConflictingWithWorkspaceOption = (program as CommandWithInternals).options
+      .filter((o) => o.conflictsWith.includes('workspace'))
+      .map((o) => `'${o.flags}'`)
+      .join(', ');
+    // eslint-disable-next-line no-console
+    console.log(
```
nedsalk, line 53: I changed `process.stdout.write` to `console.log` and then the message appeared correctly on the terminal. When `process.stdout.write` is used, the message is shown and then removed right after `process.exit(1)` is called. To test this out, you can do the following:
```ts
process.stdout.write('whatever');
setTimeout(() => process.exit(1), 1000);
```

I couldn't pinpoint the reason for this behavior. It might be due to `process.stdout` being a `WritableStream` which is an object existing as long as `process` exists and the message gets removed when `process` is killed... I'm not sure.

arboleya: Great find about the `console.log` vs `process.stdout`!

I appreciate the attention to error handling. However, I'm in favor of just replacing `process.stdout.write()` calls with `console.log()` and calling it a day. 

I don't think the added complexity justifies itself; besides, option validation and error reporting should ideally stay the same across all CLI options, which is why it was simply mimicking how `Commander` reports these problems.

nedsalk: The original error message didn't consider the scenario that you can also pass `-c`/`-s`/`-p` and made it look like `-w` was the only way to go and that's why I went about it - for correctness' sake. I suppose the user would also see those flags via `pnpm fuels init --help` and via our documentation, though, so it's not _that_ big of an issue to leave the message as it was. I'm curious to hear what others think of this. @FuelLabs/sdk-ts

nedsalk: > could this be a wider problem?

It definitely could be. The `_exit` override in the linked PR seems unnecessary given that we can use `console.log/warn/debug` instead.

nedsalk:
@petertonysmith94 I can do that in this PR, I'll put it into draft mode until it's ready again.

arboleya:
@nedsalk Are you also resolving https://github.com/FuelLabs/fuels-ts/issues/2248 in this PR?

If yes, please remember to close it in the PR description.

packages/fuels/src/cli/commands/init/index.ts
nedsalk: > I appreciate the attention to error handling. However, I'm in favor of just replacing process.stdout.write() calls with console.log() and calling it a day.

I went with this to simplify the discussion since the PR increased in scope.

.fuel-core/configs/chainConfig.json
```diff
@@ -130,9 +130,21 @@
           "xor": 2,
           "xori": 2,
           "alocDependentCost": {
-            "LightOperation": {
-              "base": "2",
-              "unitsPerGas": "214"
+            "HeavyOperation": {
```
nedsalk, line 133: This was automatically added to the PR after I ran tests. @petertonysmith94

packages/account/src/test-utils/launchNode.ts
```diff
@@ -147,12 +145,11 @@ export const launchNode = async ({
   ip,
   port,
   args = [],
-  fuelCorePath = process.env.FUEL_CORE_PATH ?? undefined,
+  fuelCorePath = process.env.FUEL_CORE_PATH || undefined,
   loggingEnabled = true,
-  debugEnabled = false,
```
nedsalk, line 152: I removed `debugEnabled` because all the messages of `fuel-core` are being piped into `stderr`. I also took a look in the https://github.com/fuellabs/fuel-core source code and found piping into `stdout` (which was used with `debugEnabled`) only for their `fuel-core-keygen` utility. @arboleya Did you add this? Or maybe @Dhaiwat10? If yes, can you provide example usage of it?

arboleya: I think the `loggingEnabled` suffice; we just want a way of turning logs on/off, I guess.

maschad (approved):

nedsalk:
@petertonysmith94 
1. I purged it.
2. Destructuring `console.log` into `log` doesn't work with `vi.spyOn(console, 'log')`, unfortunately. We had this request multiple times previously but this was always the problem.
3. I applied the changes in my own commits - it was easier to track everything for me. Thanks!

nedsalk:
> Wonder if it's worth removing the eslint rule for no-console if that is the case? (dw for this ticket)

@petertonysmith94 I wouldn't do it because most of the time we don't want`console.log`s to remain as they're used for debugging.

arboleya (approved):

maschad (dismissed):

packages/fuels/src/cli/utils/logger.ts
```diff
@@ -23,9 +24,10 @@ export function debug(...data: unknown[]) {
 }
 
 export function error(...data: unknown[]) {
-  process.stderr.write(`${chalk.red(data.join(' '))}\n`);
+  // eslint-disable-next-line no-console
+  console.log(chalk.red(data.join(' ')));
 }
```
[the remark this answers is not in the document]
nedsalk: No, see more on that in https://github.com/FuelLabs/fuels-ts/pull/2558#issuecomment-2196551910.

Torres-ssf (dismissed):

packages/abi-typegen/src/cli.ts
```diff
@@ -49,7 +49,8 @@ export function runCliAction(options: ICliParams) {
       silent: !!silent,
     });
   } catch (err) {
-    process.stderr.write(`error: ${(<Error>err).message}\n`);
+    // eslint-disable-next-line no-console
+    console.log(`error: ${(<Error>err).message}\n`);
```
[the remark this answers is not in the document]
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/2558/commits/791805fcfad80a4ed22cfe8a6ed83f5326cc2d0f

packages/account/src/test-utils/launchNode.ts
```diff
@@ -138,20 +136,18 @@ function getFinalStateConfigJSON({ stateConfig, chainConfig }: SnapshotConfigs)
...
  * */
 // #endregion launchNode-launchNodeOptions
 export const launchNode = async ({
   ip,
   port,
   args = [],
-  fuelCorePath = process.env.FUEL_CORE_PATH ?? undefined,
+  fuelCorePath = process.env.FUEL_CORE_PATH || undefined,
   loggingEnabled = true,
-  debugEnabled = false,
```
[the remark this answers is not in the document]
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/2558/commits/5a572b9ffcdd756a94bf1ad82797fa98dc8a36ef

Torres-ssf (approved):

maschad (approved):

arboleya (approved):

Issue #2521, opened 2024-06-15
Add `UNKNOWN` error code and `rawError` property to `FuelError`


arboleya:
- Add `ErrorCode.UNKNOWN` for errors we haven't mapped yet (think of it as the default)
 - Add the full thrown error in the `FuelError.rawError` field so as not to lose info on the underlying error

cc @LuizAsFight @nedsalk

Issue #2470, opened 2024-06-07
Create `Unknown` Error Code


Torres-ssf:
We need to create the `Unknown` error code to be used by `FuelError` in cases where no errors can be properly mapped and the SDK cannot determine the error origin.

PR #2549, opened 2024-06-18
chore: replacing all generic errors with `FuelError`

mvares:
- Closes #2001

PR #2562, opened 2024-06-20
chore: add `UNKNOWN` error code

maschad:
- Closes https://github.com/FuelLabs/fuels-ts/issues/2470
- Closes https://github.com/FuelLabs/fuels-ts/issues/2521

This is the only specific use case I can currently see in the codebase. I'm not sure if any other error handling use cases apply at the moment @Torres-ssf


danielbate (approved):

apps/docs/src/guide/errors/error-codes.md
```diff
@@ -258,3 +258,9 @@ It could be that the passphrase is incorrect and/or the wallet does _not_ exist
 
 The hashing algorithm is currently locked, any subsequent attempts to register a new implementation will throw this error.
 The purpose of the lock function is to provide a way to ensure that the implementation of the specific hashing algorithm cannot be changed once it is locked. This can be useful in scenarios where you want to guarantee the integrity and consistency of the hashing function throughout your application.
+
+## `UNKNOWN_ERROR`
+
+In cases where the error code is not known, this error will be thrown.
```
arboleya, line 264:
```suggestion
In cases where the error hasn't been mapped yet, this code will be used.
```

apps/docs/src/guide/errors/error-codes.md
```diff
@@ -258,3 +258,9 @@ It could be that the passphrase is incorrect and/or the wallet does _not_ exist
...
+
+This may caused by a bug in the SDK or the Fuel Stack, please report the issue to the team.
```
arboleya, line 266:
```suggestion
If you believe you found a bug, please report the [issue](https://github.com/FuelLabs/fuels-ts/issues/new/choose) to the team.
```

packages/errors/src/error-codes.ts
```diff
@@ -103,4 +103,7 @@ export enum ErrorCode {
 
   // graphql
   STREAM_PARSING_ERROR = 'stream-parsing-error',
+
+  // Unknown
+  UNKNOWN_ERROR = 'unknown-error',
```
arboleya, line 108: It seems redundant to suffix the codes with `_ERROR.`

arboleya:
> I'm not entirely sure about the issue #2470.
> Potentially it's a duplicate of #2521?

@petertonysmith94 Almost, but not 100%.

The latter also requires the addition of the `rawError` property, which should be required for all `FuelError`s.

packages/account/src/providers/utils/extract-tx-error.ts
```diff
@@ -88,8 +88,10 @@ export const assembleRevertError = (
         break;
 
       default:
-        reason = 'unknown';
-        errorMessage = `The transaction reverted with an unknown reason: ${revertReceipt.val}`;
+        throw new FuelError(
+          ErrorCode.UNKNOWN_ERROR,
+          `The transaction reverted with an unknown reason: ${revertReceipt.val}`
+        );
```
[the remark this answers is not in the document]
maschad: Good point, I still believe this should be an `UNKNOWN_ERROR` but it is useful to include the metadata.

packages/errors/src/error-codes.ts
maschad: I see your point, I was following the syntax of the above error, but I can see that most error codes don't have it, should we remove it from the other error codes in a follow up issue?

packages/account/src/providers/utils/extract-tx-error.ts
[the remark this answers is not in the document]
Torres-ssf: I agree that `UNKNOWN_ERROR` can be used here as long as we use the `metadata`

packages/errors/src/error-codes.ts
arboleya: I'd say yes, but that can be left out of this PR.

The `UNKNOWN` is simple and works fine, others may require pondering/renaming.

maschad: [ffa6470](https://github.com/FuelLabs/fuels-ts/pull/2562/commits/ffa64701a31bde6686bcfc5cf61894457508b4f2)

packages/account/src/providers/utils/extract-tx-error.ts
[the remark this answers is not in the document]
maschad: [ffa6470](https://github.com/FuelLabs/fuels-ts/pull/2562/commits/ffa64701a31bde6686bcfc5cf61894457508b4f2)

apps/docs/src/guide/errors/error-codes.md
```diff
@@ -258,3 +258,9 @@ It could be that the passphrase is incorrect and/or the wallet does _not_ exist
 
 The hashing algorithm is currently locked, any subsequent attempts to register a new implementation will throw this error.
 The purpose of the lock function is to provide a way to ensure that the implementation of the specific hashing algorithm cannot be changed once it is locked. This can be useful in scenarios where you want to guarantee the integrity and consistency of the hashing function throughout your application.
+
+## `UNKNOWN_ERROR`
```
arboleya, line 262: You forgot to rename this.

```suggestion
## `UNKNOWN`
```

arboleya (changes requested):

apps/docs/src/guide/errors/error-codes.md
maschad: Thanks!

Torres-ssf (dismissed):

packages/account/src/providers/utils/extract-tx-error.ts
```diff
@@ -88,12 +95,21 @@ export const assembleRevertError = (
         break;
 
       default:
-        reason = 'unknown';
-        errorMessage = `The transaction reverted with an unknown reason: ${revertReceipt.val}`;
+        throw new FuelError(
+          ErrorCode.UNKNOWN,
```
arboleya, line 99: Wouldn't this be best served with a `TRANSACTION_REVERTED`?

maschad: We currently don't have a `TRANSACTION_REVERTED` error code but it may be worthwhile discussing if there are other similar instances.

I opted for `UNKNOWN` because we are currently unable to determine why the script reverted in this instance. There was a previous discussion in the [above thread](https://github.com/FuelLabs/fuels-ts/pull/2562#discussion_r1647985538)

packages/account/src/providers/utils/extract-tx-error.ts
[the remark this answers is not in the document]
arboleya: Initially, I suggested adding this `UNKNOWN` code only for 100% untreated/unmapped errors. Those like this:

```ts
try {
  // anything that may throw
} catch (rawError) {
  throw new FuelError(
    ErrorCode.UNKNOWN,
    'Check `metadata` or `rawError` for more info.',
    { rawError },
  });
}
```

This already considered adding the `rawError` property to `FuelError`, of course.

But I'm okay with whatever the majority chooses.

packages/account/src/providers/utils/extract-tx-error.ts
arboleya: I added my thoughts to the linked conversation.

arboleya:
> The latter also requires the addition of the `rawError` property, which should be required for all `FuelError`s.

@maschad Since it's a simple addition (optional property), I think it could be done in this PR, if you don't mind.

This way others will be able to use it alongside the `UNKNOWN` error code (see [this](https://github.com/FuelLabs/fuels-ts/pull/2549#issuecomment-2183321805) and [this](https://github.com/FuelLabs/fuels-ts/pull/2549#issuecomment-2183358427)).

maschad:
> > The latter also requires the addition of the `rawError` property, which should be required for all `FuelError`s.
> 
> @maschad Since it's a simple addition (optional property, #2521), I think it could be done in this PR, if you don't mind.
> 
> This way others will be able to use it alongside the `UNKNOWN` error code (see [this](https://github.com/FuelLabs/fuels-ts/pull/2549#issuecomment-2183321805) and [this](https://github.com/FuelLabs/fuels-ts/pull/2549#issuecomment-2183358427)).

Good idea, will do.

packages/account/src/providers/utils/extract-tx-error.ts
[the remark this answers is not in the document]
maschad: In this particular case there wouldn't be anything to include in the `rawError` property given the metadata is populated, so perhaps we should use `SCRIPT_REVERTED` or this would be an exception for an `UNKNOWN` , I lean towards the latter because even though the script is reverted, it's for an `UNKNOWN` reason.

arboleya (approved):

Torres-ssf (dismissed):

packages/errors/src/fuel-error.ts
```diff
@@ -32,15 +33,21 @@ export class FuelError extends Error {
...
+    code: ErrorCode,
+    message: string,
+    metadata: Record<string, unknown> = {},
+    rawError: unknown = {}
+  ) {
     super(message);
     this.code = code;
     this.name = 'FuelError';
     this.metadata = metadata;
+    this.rawError = rawError;
```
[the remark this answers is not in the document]
maschad: [`a50553e` (#2562)](https://github.com/FuelLabs/fuels-ts/pull/2562/commits/a50553ee51a8a894d31c7d0e89dbb0b741246d9b)

Torres-ssf (approved):

arboleya (approved):

PR #2577, opened 2024-06-21
chore: add permissions for pr labeller

petertonysmith94:
Fixing mistake from #2576


arboleya (dismissed):

maschad (changes requested):
Two things: 

- Why hasn't the label been automagically added to this PR?
- Shouldn't we need to add:

```yaml
permissions:
  pull-requests: write 
```

according to the [usage docs](https://github.com/fuxingloh/multi-labeler#usage)

arboleya:
> Why hasn't the label been automagically added to this PR?

@maschad New workflows must first go to `master` before they can run, no?

maschad:
> > Why hasn't the label been automagically added to this PR?
> 
> @maschad New workflows must first go to `master` before they can run, no?

My tendency was to think so as well, but I did note when this job was [introduced](https://github.com/FuelLabs/fuels-ts/pull/2383) the automation was applied. It may be because the current commit still has the conditional though. 

Screenshot 2024-06-21 at 5 02 27 PM

maschad (approved):

arboleya (approved):

danielbate (approved):

PR #2578, opened 2024-06-21
chore: add preview link for docs changes

### LINKED ISSUE #2455: Add preview link for PRs that change docs (using vitepress)

msensys:
This PR provides a preview link for pull requests that change the docs. That offers several benefits as:

- A draft view of the documentation changes included in the PR
- An automatically deployed version using `GitHub Pages`

Closes #2455

### DIFF
--- a/.github/workflows/docs-preview.yaml
+++ b/.github/workflows/docs-preview.yaml
@@ -1,0 +1,61 @@
+name: Docs Preview on PR
+
+on:
+  pull_request:
+    paths:
+      - "apps/docs/**"
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: pages
+  cancel-in-progress: false
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Install PNPM
+        uses: pnpm/action-setup@v3
+        with:
+          version: 8
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: pnpm
+
+      - name: Setup Pages
+        uses: actions/configure-pages@v4
+
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Build with VitePress
+        run: pnpm docs:build
+
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v3
+        with:
+          path: apps/docs/.vitepress/dist/
+
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    needs: build
+    runs-on: ubuntu-latest
+    name: Deploy
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v4
--- a/.changeset/little-ligers-agree.md
+++ b/.changeset/little-ligers-agree.md
@@ -1,0 +1,4 @@
+---
+---
+
+chore: add preview link for docs changes


.github/workflows/docs-preview.yaml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: pnpm
+
+      - name: Setup Pages
+        uses: actions/configure-pages@v4
+
+      - name: Install dependencies
```
msensys, line 40: I don't know if this makes sense here

msensys, line 40: I would like to know if this step makes sense since we should only run `pnpm install` in `apps/docs`

.github/workflows/docs-preview.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Install PNPM
         uses: pnpm/action-setup@v3
         with:
-          version: 8
+          version: 9.0.5
```
msensys, line 29: Maybe there is a better way to declare the `version` from `step`

msensys:
Before merging the PR, someone must grant some permissions to the repository related to `Github Pages` .

.github/workflows/docs-preview.yaml
```diff
@@ -0,0 +1,61 @@
...
+        with:
+          path: apps/docs/.vitepress/dist/
+
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    needs: build
+    runs-on: ubuntu-latest
+    name: Deploy
```
msensys, line 61: I think that `name: Deploy` sounds unnecessary

arboleya (commented):
We should also post the preview link as a comment on the PR.

Here's one example of how coverage report does it:

 - https://github.com/FuelLabs/fuels-ts/blob/e2259ed5186c9909f324f20342b8ce259cee685f/.github/workflows/test.yaml#L154-L162

.github/workflows/docs-preview.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Install PNPM
+        uses: pnpm/action-setup@v4
+        with:
+          version: 9.0.5
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: pnpm
```
danielbate, line 34: We have a `ci-setup` action that could be used here.

.github/workflows/docs-preview.yaml
```diff
@@ -0,0 +1,76 @@
...
+        uses: actions/deploy-pages@v4
+
+  comment:
+    runs-on: ubuntu-latest
+    needs: deploy
+    steps:
+      - name: Comment PR
+        uses: thollander/actions-comment-pull-request@v2
+        with:
+          message: "Here is the preview link: ${{ needs.deploy.outputs.page_url }}"
```
danielbate, line 74:
```suggestion
          message: "Documentation preview: ${{ needs.deploy.outputs.page_url }}"
```

.github/workflows/docs-preview.yaml
```diff
@@ -0,0 +1,76 @@
...
+    outputs: 
+      page_url: ${{ steps.deployment.outputs.page_url }}
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v4
+
+  comment:
```
danielbate, line 67: IMO this could all be a single job as it's running serially?

.github/workflows/docs-preview.yaml
```diff
@@ -0,0 +1,76 @@
...
+
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Build project
+        run: pnpm build
+
+      - name: Build with VitePress
+        working-directory: apps/docs
+        run: pnpm build
```
danielbate, line 47: The previous build will also have built `docs`.

danielbate (commented):
Great work, couple comments

.github/workflows/docs-preview.yaml
msensys: Hmm, now I will use this! Thanks

.github/workflows/docs-preview.yaml
msensys: Thinking well... We could put this task (commenting on PR) as a step from the `deployment` job. Thanks for the comments!

.github/workflows/docs-preview.yaml
msensys: Sincerely, I didn't understand your point, but I think you're talking about the two builds that the CI does.

msensys:
> Will this override our existing GitHub pages deployment? https://fuellabs.github.io/fuels-ts/

Probably yes, but we can see some way to create pages based on `PR number` or `contributor name`. (It's something we can analyze)

mvares:
> It looks like currently GH pages doesn't support preview deployment and from the [discussion](https://github.com/orgs/community/discussions/7730), it doesn't look remotely close.
> 
> I have hesitations about deploying to our primary GH page, and potentially deploying to Vercel is a better approach. What are other peoples thoughts?
> 
> cc: @FuelLabs/sdk-ts

I agree with you. Maybe it may be interesting to try something with Vercel (By the way, vercel already has a `pre-deployment` feature)

Torres-ssf:
> It looks like currently GH pages doesn't support preview deployment and from the [discussion](https://github.com/orgs/community/discussions/7730), it doesn't look remotely close.
> 
> I have hesitations about deploying to our primary GH page, and potentially deploying to Vercel is a better approach. What are other peoples thoughts?
> 
> cc: @FuelLabs/sdk-ts

@petertonysmith94 Agreed. We should use a different strategy here. Vercel seems to be a good suggestion.

mvares:
> @petertonysmith94 Agreed. We should use a different strategy here. Vercel seems to be a good suggestion.

@Torres-ssf, considering that we are likely already planning a different strategy here, I think this PR is now beyond their scope. Should this PR be closed?

Issue #2561, opened 2024-06-20
Make `deployContract` method a non-blocking call


Torres-ssf:
The [deployContract](https://github.com/FuelLabs/fuels-ts/blob/dc17b9c454fb025cd07840e79489afe346ff6114/packages/contract/src/contract-factory.ts#L134) method in the `ContractFactory` currently submits the transaction and waits for it to be processed. This synchronous approach can lead to significant delays, as transaction processing times in a blockchain environment often take several seconds.

Having this method always act as a blocking call is impractical for real-world blockchain environments as it forces developers to wait for transaction completion before proceeding with other tasks.

The `deployContract` method should only wait for the TX to be processed if the user explicitly chooses to do it. This non-blocking approach should be the default behavior, as blocking the UI is not what users usually desire.

> similar to https://github.com/FuelLabs/fuels-ts/issues/2492

arboleya:
Are we talking about adding the `awaitExecution` flag as a new parameter?

https://github.com/FuelLabs/fuels-ts/blob/dc17b9c454fb025cd07840e79489afe346ff6114/packages/contract/src/contract-factory.ts#L164

Torres-ssf:
@arboleya Yes, we will also need to change the method's response type.

Currently, it returns an instance of the `Contract` class, ready to be used. This is only possible because the method waits for the transaction to be processed.

The ideal approach would be to have the `deployContract` method simply submit the transaction and return something that can be used to fetch the transaction, such as its ID or an instance of `TransactionResponse`. This way, dApp applications wouldn't get blocked when submitting the TX, and they would receive the transaction ID as a response, making it possible to attach hooks to fetch its status.

PR #2597, opened 2024-06-24
feat!: made `deployContract` a non-blocking call

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/2561

# Release notes

In this release, we:
 - Made `deployContract` a non-blocking call

# Summary

This PR refactors the `deployContract` method to be non-blocking. As a result, the method no longer returns the contract instance directly. Instead, it returns an object containing the `transactionId` and a `waitForResult` function. This change improves the flexibility and control over contract deployment but introduces a breaking change that requires updates to existing implementations.

# Breaking Changes

The `deployContract` method no longer returns the contract instance directly. Instead, it returns an object containing the `transactionId` , the `contractId`, and a `waitForResult` function.

```ts
// before
const factory = new ContractFactory(contractByteCode, contractAbi, wallet);

const contract = await factory.deployContract();

const { value } = await contract.functions.xyz().call();
```

```ts
// after
const factory = new ContractFactory(contractByteCode, contractAbi, wallet);

const { waitForResult, transactionId, contractId } = await factory.deployContract();

const { contract, transactionResult } = await waitForResult();

const { value } = await contract.functions.xyz().call();
```

# Checklist


packages/contract/src/contract-factory.ts
```diff
@@ -132,39 +132,35 @@ export default class ContractFactory {
...
-    } else {
-      transactionRequest.maxFee = txCost.maxFee;
-    }
+  /**
+   * Deploys a contract asynchronously.
+   *
+   * @param deployContractOptions - The options for deploying the contract (optional).
+   * @returns An object containing the deployed contract and the transaction response.
+   */
+  async deployContractAsync(deployContractOptions: DeployContractOptions = {}) {
```
Torres-ssf, line 151: There is a third option, which is to use TypeScript generics to infer if `awaitExecution` was specified, and then make `deployContract` have two possible return types. This was suggested by @nedsalk:

```ts
async deployContractNedim<
  AwaitExecution extends boolean = false,
  R = AwaitExecution extends true
    ? Contract
    : { transactionResponse: TransactionResponse; contract: Contract },
>(
  deployContractOptions: DeployContractOptions = {},
  awaitExecution?: AwaitExecution
): Promise<R> {
  if (!this.account) {
    throw new FuelError(ErrorCode.ACCOUNT_REQUIRED, 'Cannot deploy Contract without an account.');
  }
  const { configurableConstants } = deployContractOptions;
  if (configurableConstants) {
    this.setConfigurableConstants(configurableConstants);
  }
  const { contractId, transactionRequest } = this.createTransactionRequest(deployContractOptions);
  const txCost = await this.account.provider.getTransactionCost(transactionRequest);
  const { maxFee: setMaxFee } = deployContractOptions;
  if (isDefined(setMaxFee)) {
    if (txCost.maxFee.gt(setMaxFee)) {
      throw new FuelError(
        ErrorCode.MAX_FEE_TOO_LOW,
        `Max fee '${deployContractOptions?.maxFee}' is lower than the required: '${txCost.maxFee}'.`
      );
    }
  } else {
    transactionRequest.maxFee = txCost.maxFee;
  }
  await this.account.fund(transactionRequest, txCost);
  const transactionResponse = await this.account.sendTransaction(transactionRequest, {
    awaitExecution,
  });

  const contract = new Contract(contractId, this.interface, this.account);

  return (awaitExecution === true ? contract : { contract, transactionResponse }) as R;
}
```

apps/docs/src/guide/contracts/deploying-contracts.md
```diff
@@ -27,11 +27,23 @@ Load the contract bytecode and JSON ABI, generated from the Sway source, into th
 
 ## 4. Deploying the Contract
 
-Initialize a [`ContractFactory`](../../api/Contract/ContractFactory.md) with the bytecode, ABI, and wallet. Deploy the contract and use its methods.
+To deploy the contract we can instantiate the [`ContractFactory`](../../api/Contract/ContractFactory.md) with the bytecode, ABI, and wallet. The we can call `deployContract` method. This method returns a promise that will resolve to a `Contract` instance, ready to be used to interact with the deployed smart contract.
```
danielbate, line 30:
```suggestion
To deploy the contract we can instantiate the [`ContractFactory`](../../api/Contract/ContractFactory.md) with the bytecode, ABI, and wallet. Then we can call the `deployContract` method. This method returns a promise that will resolve to a `Contract` instance which provides methods to interact with the deployed smart contract.
```

apps/docs/src/guide/contracts/deploying-contracts.md
```diff
@@ -27,11 +27,23 @@ Load the contract bytecode and JSON ABI, generated from the Sway source, into th
...
 <<< @/../../docs-snippets/src/guide/contracts/deploying-contracts.test.ts#contract-setup-3{ts:line-numbers}
 
-## 5. Executing a Contract Call
+The `deployContract` method will only resolve after the deploy contract transaction has been processed and the smart contract is deployed on the blockchain. As a result, this call can take several seconds to resolve and will block any code execution that follows.
+
+If this situation is inappropriate for your dApp, you can use the `deployContractAsync` method.
+
+## 5. Deploying the Contract Asynchronously
+
+In some cases, you may not want to wait for the deploy contract transaction to finish processing. In these instances, you can use `deployContractAsync`. This method resolves as soon as the transaction to deploy the contract is submitted and returns two objects: a `TransactionResponse` and a `Contract` instance.
```
danielbate, line 40: I'd probably make the point about the `waitForResult` here as that seems like the most important point here, and the key distinction between the 2 methods.

packages/contract/src/contract-factory.ts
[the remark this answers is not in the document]
danielbate: Do we think async should be the default behaviour given most people will be interacting with testnet and will have the time lapse? Just wondering whether it warrants the breaking change.

packages/contract/src/contract-factory.ts
```diff
@@ -202,4 +198,45 @@ export default class ContractFactory {
       );
     }
   }
+
+  private getAccount(): Account {
+    if (!this.account) {
+      throw new FuelError(ErrorCode.ACCOUNT_REQUIRED, 'Cannot deploy Contract without account.');
```
danielbate, line 204:
```suggestion
      throw new FuelError(ErrorCode.ACCOUNT_REQUIRED, 'Account not assigned to contract.');
```

Made more generic.

danielbate (commented):
Nice one, couple comments

packages/contract/src/contract-factory.ts
nedsalk: The reasoning behind this is
- we don't need to duplicate logic in a second `deployContractAsync` method when they're both doing the same thing,
- `awaitExecution` already exists as a parameter in `Provider.sendTransaction` and `Account.sendTransaction` so we'd be following the same pattern

Besides this, regardless of the approach taken, we need to apply it to typegen as well. Do we want to do it in this PR or in a subsequent one?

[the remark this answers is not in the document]
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2597/commits/7eabf88aac7e64313a329bebb027f10db3114dc9

apps/docs/src/guide/contracts/deploying-contracts.md
Torres-ssf: @danielbate The docs are slightly different now https://github.com/FuelLabs/fuels-ts/pull/2597/commits/7c800e1786393467ffb92433a7a331d6ca38d4cb

apps/docs/src/guide/contracts/deploying-contracts.md
Torres-ssf: @danielbate The docs are slightly different now https://github.com/FuelLabs/fuels-ts/pull/2597/commits/7c800e1786393467ffb92433a7a331d6ca38d4cb

.changeset/giant-ads-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/contract": patch
+---
+
+feat: add a non-blocking call for deploy contract
```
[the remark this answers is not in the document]
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2597/commits/5626f9a39c04ac531bf9974794b452618c457951

fuel-service-user:
✨ A PR has been created under the `rc-2597` tag on `fuels-wallet` repo.
https://github.com/FuelLabs/fuels-wallet/pull/1393

packages/fuel-gauge/src/contract-factory.test.ts
```diff
@@ -85,16 +85,33 @@ describe('Contract Factory', () => {
...
     expect(callResult).toMatchObject({
       receipts: expect.arrayContaining([expect.any(Object)]),
     });
   });
 
+  it('can deploy a contract asynchronously without waiting for the TX to be processed', async () => {
+    const factory = await createContractFactory();
+
+    const { transactionResponse, contract } = await factory.deployContract();
+
```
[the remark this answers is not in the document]
Torres-ssf: @petertonysmith94 An error will be thrown because the contract was not deployed yet, therefore, it does not exist.

.changeset/giant-ads-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/contract": patch
```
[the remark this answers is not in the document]
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2597/commits/5626f9a39c04ac531bf9974794b452618c457951

Torres-ssf:
> Do you think it's worthwhile using Typegen consistently in our documentation - over using `ContractFactory` ([example](https://github.com/FuelLabs/fuels-ts/pull/2597/files#diff-f09392838abb9aa9ccbab9942dbbebc7cb1ae40c65fb3904255daa80f923f04f))?
> 
> (This can be a separate issue - just posing the question)

@petertonysmith94 Good question. Since both approaches are valid we should have examples for each one. But at the same time, I think that having 2 different approaches can be confusing for new users. What do think?

danielbate:
@petertonysmith94 @Torres-ssf just weighing in that I would love to see us favouring Typegen consistently for `fuel-gauge` and snippets as it's essential for browser testing, and for users sake. I've added it as a requirement to #1820. Then we could have contract factory having it's own doc snippet.

arboleya:
@danielbate @petertonysmith94 I lost a bit of context here, but I'd say we should use the `fuels` CLI instead, which internally will use `typegen` anyway. This will close the cycle of dogfooding our stuff in our test suites as much as possible.

packages/fuel-gauge/src/contract-factory.test.ts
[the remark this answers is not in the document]
nedsalk: Worthy of mentioning is that if the `awaitExecution = true` argument is passed, then there is no need to call `waitForResult`. Whoever uses the non-blocking `awaitExecution = false` should accept the nature of blockchain and do a `waitForResult` before running contract calls.

packages/abi-typegen/src/templates/contract/factory.hbs
```diff
@@ -1,7 +1,7 @@
 {{header}}
 
 import { Interface, Contract, ContractFactory } from "fuels";
-import type { Provider, Account, AbstractAddress, BytesLike, DeployContractOptions, StorageSlot } from "fuels";
+import type { Provider, Account, AbstractAddress, BytesLike, DeployContractOptions, StorageSlot, TransactionResponse, DeployContractResult } from "fuels";
```
nedsalk, line 4: I don't see the `TransactionResponse` import used anywhere in the template. We should remove it if it's not used.

packages/abi-typegen/test/fixtures/templates/contract/factory.hbs
```diff
@@ -10,7 +10,7 @@
 */
 
 import { Interface, Contract, ContractFactory } from "fuels";
-import type { Provider, Account, AbstractAddress, BytesLike, DeployContractOptions, StorageSlot } from "fuels";
+import type { Provider, Account, AbstractAddress, BytesLike, DeployContractOptions, StorageSlot, TransactionResponse, DeployContractResult } from "fuels";
```
nedsalk, line 13: I don't see the `TransactionResponse` import used anywhere in the template. We should remove it if it's not used.

packages/fuel-gauge/src/contract-factory.test.ts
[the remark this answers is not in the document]
arboleya: I understand the rationale and that this is a delicate issue.

Still, I suspect this error case can be counterintuitive.

apps/docs/src/guide/contracts/deploying-contracts.md
```diff
@@ -27,10 +27,16 @@ Load the contract bytecode and JSON ABI, generated from the Sway source, into th
 
 ## 4. Deploying the Contract
 
-Initialize a [`ContractFactory`](../../api/Contract/ContractFactory.md) with the bytecode, ABI, and wallet. Deploy the contract and use its methods.
+To deploy the contract, instantiate the [`ContractFactory`](../../api/Contract/ContractFactory.md) with the bytecode, ABI, and wallet. Then, call the `deployContract` method. This method resolves as soon as the transaction to deploy the contract is submitted and returns two objects: a `TransactionResponse` and a `Contract` instance.
```
arboleya, line 30:
```suggestion
To deploy the contract, instantiate the [`ContractFactory`](../../api/Contract/ContractFactory.md) with the bytecode, ABI, and wallet. Then, call the `deployContract` method. This method resolves when the transaction to deploy the contract is submitted, returning two objects: a `TransactionResponse` and a `Contract` instance.
```

apps/docs/src/guide/contracts/deploying-contracts.md
```diff
@@ -27,10 +27,16 @@ Load the contract bytecode and JSON ABI, generated from the Sway source, into th
...
 
 <<< @/../../docs-snippets/src/guide/contracts/deploying-contracts.test.ts#contract-setup-3{ts:line-numbers}
 
+**Important**: The contract instance can only be safely used after the promise from `transactionResponse.waitForResult` is resolved. To avoid blocking your code, attach this promise to a hook or listener that will use the contract only after it is fully deployed.
```
arboleya, line 34:
```suggestion
**Important**: The contract instance can only be safely used after the promise from `transactionResponse.waitForResult` is resolved. To avoid blocking your code, attach this promise to a hook or listener that awaits it to ensure you don't use the contract before its deployment finishes.
```

packages/contract/src/contract-factory.ts
```diff
@@ -131,40 +142,27 @@ export default class ContractFactory {
...
-    }
-
-    const { contractId, transactionRequest } = this.createTransactionRequest(deployContractOptions);
-
-    const txCost = await this.account.provider.getTransactionCost(transactionRequest);
+  /**
+   * Deploys a contract and returns the transaction response and the deployed contract instance.
+   * @template T - The type of the contract to be deployed.
+   * @param deployContractOptions - The options for deploying the contract.
+   * @returns A promise that resolves to the transaction response and the deployed contract instance.
```
arboleya, line 149:
```suggestion
   * @returns A promise that resolves the transaction response and the deployed contract instance.
```

arboleya (commented):
I left some small [optional] nits; pick what you want.

Overall, I'm still thinking if this approach is intuitive enough.

danielbate:
> I'd say we should use the fuels CLI instead, which internally will use typegen anyway

@arboleya agreed. Right now we are [building with the fuels CLI](https://github.com/FuelLabs/fuels-ts/blob/6184bc9cd0982360b880d4cf4c2f112def1bd70b/apps/docs-snippets/package.json#L8), so [like we do](https://github.com/FuelLabs/fuels-ts/blob/6184bc9cd0982360b880d4cf4c2f112def1bd70b/packages/fuel-gauge/src/call-test-contract.test.ts#L6-L7) in a few places in `fuel-gauge`, I would like to see us utilising those factories rather than importing the `ContractFactory`.

packages/fuel-gauge/src/contract-factory.test.ts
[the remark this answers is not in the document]
nedsalk: > Still, I suspect this error case can be counterintuitive.

If we want to still _not_ fail a contract call and not have this error case, we could couple deployment to contract calls and wait for deployment to finish before running the actual contract call:
```ts
// mock internals of a contract call
await this.deploymentTransactionResponse?.waitForResult(); 
return call();
```

Then we can just explain in the documentation that, for `awaitExecution=false`, the first contract call might take a bit longer because the contract might not have been deployed yet.

packages/abi-typegen/test/fixtures/templates/contract/factory.hbs
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2597/commits/f528fac6c55fed7c9ad4d6264439de416c59d9b3

packages/abi-typegen/src/templates/contract/factory.hbs
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2597/commits/f528fac6c55fed7c9ad4d6264439de416c59d9b3

nedsalk (dismissed):
lgtm

packages/fuel-gauge/src/contract-factory.test.ts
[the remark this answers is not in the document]
maschad: @Torres-ssf what would be the error type thrown if one calls the contract before it's deployed? I think a better path would be to document that error so that consumers can error handle appropriately and retry when necessary.

packages/fuels/src/cli/commands/deploy/deployContract.ts
```diff
@@ -25,6 +25,10 @@ export async function deployContract(
   const abi = JSON.parse(readFileSync(abiPath, 'utf-8'));
   const contractFactory = new ContractFactory(bytecode, abi, wallet);
 
-  const contract = await contractFactory.deployContract(deployConfig);
+  const { contract } = await contractFactory.deployContract({
+    awaitExecution: true,
+    ...deployConfig,
+  });
+
```
arboleya, line 32: Have you considered splitting the functionality?

I think it could bring some more cohesion and make it easier to follow. 

```ts
// breaking change
const { txResponse, contractId } = contractFactory.deployContract(config);

// new method
const { contract } = await contractFactory.deployAndAwaitForContract(config);
```

To avoid breaking changes, the opposite could be done:

```ts
// same as is
const { contract } = contractFactory.deployContract(config);

// new method
const { txResponse, contractId } = contractFactory.deployContractForLater(config);
```

packages/fuel-gauge/src/contract-factory.test.ts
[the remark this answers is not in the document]
arboleya: We could consider [splitting the functionality](https://github.com/FuelLabs/fuels-ts/pull/2597#discussion_r1658507262).

nedsalk (approved):

packages/fuels/src/cli/commands/deploy/deployContract.ts
arboleya: In the current approach, one must wait on a promise that doesn't contain the value they're waiting for.

The value (`Contract` instance) is outside the promise.

arboleya: For combining the Promise and the `Contract` instance, another option could be:

```ts
// breaking change
const { contractId, waitForDeployment } = contractFactory.deployContract(config);

// new method
const { contract } = await waitForDeployment();
```

Torres-ssf: @arboleya Using a secondary method was my initial approach: https://github.com/FuelLabs/fuels-ts/pull/2597/commits/5f6ffdd15a0b9689eed09f6821a1e55934ee007b

@FuelLabs/sdk-ts How do you feel about this proposal?

```ts
// same as is
const { contract } = contractFactory.deployContract(config);

// new method
const { txResponse, contractId } = await contractFactory.deployContractForLater(config);
```

arboleya: I didn't have the context of your first approach, sorry. I like the idea of a second method!

Even on the non-blocking route, we can offer a `contract` instance at the end.

I just want the instance to come from the Promise I'm waiting for:

```ts
const { awaitForDeployment } = await contractFactory.deployForLater(config);
const { contract } = await awaitForDeployment();
```

We could also return the other properties, like `txReponse`, `contractId` etc.:

```ts
const {
  txReponse,
  contractId,
  awaitForDeployment,
} = await contractFactory.deployForLater(config);
```

And it would still be non-breaking.

maschad: > I just want the instance to come from the Promise I'm waiting for:
```ts
const { awaitForDeployment } = await contractFactory.deployForLater(config);
const { contract } = await awaitForDeployment();
```

I think there is credence to @arboleya suggestion because we would still be enabling non-blocking call, and it is also poka-yoked to prevent users from calling a contract instance that is not yet deployed. 

In the [current state](https://github.com/FuelLabs/fuels-ts/pull/2597/files#diff-a8671e1aac0508f14642db9f3313cc1297d41afed42f68e12448c8973de3989fR32-R36) of the PR, if I call  `factory.deployContract` with `awaitExecution: false`, the `contract` instance returned _may_ throw an error that I will have to handle in my code. I think we should steer consumers away from that code path.

danielbate: I agree with @maschad sentiment. However given the current design, I can't see a way of doing this without making _many_ changes as we can't do this at just the `Contract` level, this would need to be drilled to the invocation scopes. I think the value for users is worth the change. However can we better defend against the breaking change by making default behaviour `awaitExecution: true`?

Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2597/commits/c0b06bbc64a49daaae493b3a83cdfc78b56de7df

packages/fuel-gauge/src/contract-factory.test.ts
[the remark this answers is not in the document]
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2597/commits/c0b06bbc64a49daaae493b3a83cdfc78b56de7df

arboleya (dismissed):

maschad (commented):
Looks good, just one minor nit.

maschad (dismissed):
Awesome job on this @Torres-ssf

arboleya (dismissed):

danielbate (dismissed):

arboleya (dismissed):

apps/docs/src/guide/contracts/deploying-contracts.md
```diff
@@ -27,14 +27,18 @@ Load the contract bytecode and JSON ABI, generated from the Sway source, into th
 
 ## 4. Deploying the Contract
 
-Initialize a [`ContractFactory`](../../api/Contract/ContractFactory.md) with the bytecode, ABI, and wallet. Deploy the contract and use its methods.
+To deploy the contract, instantiate the [`ContractFactory`](../../api/Contract/ContractFactory.md) with the bytecode, ABI, and wallet. Then, call the `deployContract` method. This call resolves as soon as the transaction to deploy the contract is submitted and returns two items: the `transactionId` and a `waitForResult` function.
```
petertonysmith94, line 30:
```suggestion
To deploy the contract, instantiate the [`ContractFactory`](../../api/Contract/ContractFactory.md) with the bytecode, ABI, and wallet. Then, call the `deployContract` method. This call resolves as soon as the transaction to deploy the contract is submitted and returns three items: the `contractId`, the `transactionId` and a `waitForResult` function.
```

petertonysmith94 (dismissed):
Small suggestion - rest looks good to me!

petertonysmith94 (approved):
💚

nedsalk (approved):

Issue #2056, opened 2024-04-15
Add docs related to the optional caching UTXOs logic


Torres-ssf:
This issue is a reminder that we should add docs information explaining the optional [caching UTXOs logic](https://github.com/FuelLabs/fuels-ts/blob/917a57d9efddc9376e7a1ecbe722c8eeda184523/packages/account/src/providers/provider.ts#L356-L357), if we decide to maintain this functionality.

More info here: https://github.com/FuelLabs/fuels-ts/issues/1515

PR #2598, opened 2024-06-24
docs: add `Provider` doc section for `cacheUtxo`

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/2056
- Close: https://github.com/FuelLabs/fuels-ts/issues/1515


apps/docs/src/guide/provider/provider-options.md
```diff
@@ -39,3 +39,25 @@ Provide a custom `fetch` function that'll replace the default fetch call.
...
+
+However, if a second transaction is created before the first one is processed, there is a chance of using the same UTXO(s) for both transactions. This happens because the UTXO(s) used in the first transaction are still unspent until the transaction is fully processed.
+
+If the second transaction attempts to use the same UTXO(s) that the first transaction has already spent, it will result in the following error:
+
+```console
+Transaction is not inserted. UTXO does not exist: 0xf5...
+```
+
+This error indicates that the UTXO(s) used by the second transaction no longer exist, as they were already spent by the first transaction.
```
maschad, line 55:
```suggestion
This error indicates that the UTXO(s) used by the second transaction no longer exists, as it was already spent by the first transaction
```

apps/docs/src/guide/provider/provider-options.md
```diff
@@ -39,3 +39,25 @@ Provide a custom `fetch` function that'll replace the default fetch call.
...
+
+To prevent this issue, you can use the `cacheUtxo` flag. This flag sets a TTL (Time-To-Live) for caching UTXO(s) used in a transaction, preventing them from being reused in subsequent transactions within the specified time.
```
maschad, line 57:
```suggestion
To prevent this issue, you can use the `cacheUtxo` flag. This flag sets a TTL (Time-To-Live) for caching a UTXO(s) used in a transaction, preventing it from being reused in subsequent transactions within the specified time.
```

apps/docs/src/guide/provider/provider-options.md
```diff
@@ -39,3 +39,25 @@ Provide a custom `fetch` function that'll replace the default fetch call.
...
+
+<<< @/../../docs-snippets/src/guide/provider/provider.test.ts#options-cache-utxo{ts:line-numbers}
+
+It's important to note that you can only submit multiple transactions from the same account without waiting for the previous transactions to complete if your account has multiple UTXOs available. If you only have one UTXO, the first transaction will spend it, and any remaining amount will be converted into a new UTXO with a different ID.
```
maschad, line 61: I'm not fully clear on what you are trying to convey here, is my suggestion in alignment with your sentiment?

```suggestion
**Note**:

If you would like to submit multiple transactions without waiting for each transactions to complete, those transactions **must** be from the same account. If you only have one UTXO, the first transaction will spend it, and any remaining amount will be converted into a new UTXO with a different ID.
```

maschad (changes requested):
Nice job on this 🚀 , left some suggestions.

apps/docs/src/guide/provider/provider-options.md
Torres-ssf: @maschad My version seems to be correct already as it matches the plural "UTXO(s)" with "them."

apps/docs/src/guide/provider/provider-options.md
Torres-ssf: @maschad this version has inaccurate information: `those transactions **must** be from the same account`.

Isn't the information here clear enough?

apps/docs/src/guide/provider/provider-options.md
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2598/commits/0941eaa790b3355ab8aa0a6911fdb6c8da051b6b

apps/docs/src/guide/provider/provider-options.md
maschad: So apparently there [isn't consensus](https://www.grammarphobia.com/blog/2015/04/parenthetical-plurals.html) on parenthetical plurals, I had always made the assumption they were treated as singular. So I suppose this can work.

apps/docs/src/guide/provider/provider-options.md
maschad: I'm not understanding what is being said here, are you suggesting that:

a) If I want to submit multiple transactions AND I don't want to wait on each transaction to be completed, I must have multiple UTXOs available from the same account.

b) I must submit multiple transactions from the same account without waiting for the previous transactions to be completed if I have multiple UTXOs available.

Torres-ssf: @maschad Option A

maschad: Thanks for clarifying, I may be one of the few to misinterpret, but the wording I would suggest would be

```suggestion
**Note:**

If you would like to submit multiple transactions without waiting on each transaction to be completed, your account must have multiple UTXOs available. If you only have one UTXO, the first transaction will spend it, and any remaining amount will be converted into a new UTXO with a different ID.
```

Feel free to ignore the suggestion but the current wording isn't immediately obvious to me

apps/docs/src/guide/provider/provider-options.md
```diff
@@ -39,3 +39,27 @@ Provide a custom `fetch` function that'll replace the default fetch call.
 _Note: If defined, `requestMiddleware`, `timeout` and `retryOptions` are applied to this custom `fetch` function as well._
 
 <<< @/../../docs-snippets/src/guide/provider/provider.test.ts#options-fetch{ts:line-numbers}
+
+### `cacheUtxo`
+
+When using the SDK, it may be necessary to submit multiple transactions from the same account in a short period. In such cases, the SDK creates and funds these transactions, then submits them to the node.
+
+However, if a second transaction is created before the first one is processed, there is a chance of using the same UTXO(s) for both transactions. This happens because the UTXO(s) used in the first transaction are still unspent until the transaction is fully processed.
+
```
danielbate, line 48: Might be beneficial to include a link above the [UTXO architecture](https://docs.fuel.network/docs/fuels-ts/wallets/checking-balances/#understanding-the-utxo-model) docs.

danielbate (commented):
First pass, couple comments and questions.

1. If I have one utxo and am submitting multiple transactions, `cacheUtxo` has no effect and in fact we risk the second tx failing but we would be able to fund both txs initially?
2. Why would you not want to `cacheUtxo`s?

maschad (dismissed):

apps/docs/src/guide/provider/provider-options.md
```diff
@@ -39,3 +39,27 @@ Provide a custom `fetch` function that'll replace the default fetch call.
...
+This error indicates that the UTXO(s) used by the second transaction no longer exist, as the first transaction already spent them.
+
+To prevent this issue, you can use the `cacheUtxo` flag. This flag sets a TTL (Time-To-Live) for caching UTXO(s) used in a transaction, preventing them from being reused in subsequent transactions within the specified time.
+
+<<< @/../../docs-snippets/src/guide/provider/provider.test.ts#options-cache-utxo{ts:line-numbers}
+
+**Note:**
+
+If you would like to submit multiple transactions without waiting on each transaction to be completed, your account must have multiple UTXOs available. If you only have one UTXO, the first transaction will spend it, and any remaining amount will be converted into a new UTXO with a different ID.
```
arboleya, line 63: Small detail.

```suggestion
If you would like to submit multiple transactions without waiting for each transaction to be completed, your account must have multiple UTXOs available. If you only have one UTXO, the first transaction will spend it, and any remaining amount will be converted into a new UTXO with a different ID.
```

arboleya (commented):
I'll think more about this before coming back to it.

At first glance, this flag looks confusing.

apps/docs/src/guide/provider/provider-options.md
arboleya: Unrelated question outside of the scope of this PR, but should this [UTXO architecture](https://docs.fuel.network/docs/fuels-ts/wallets/checking-balances/#understanding-the-utxo-model) be its own thing?

I mean, moved outside of `Wallet` > `Checking Balances`.

Torres-ssf: @arboleya https://github.com/FuelLabs/fuels-ts/pull/2624

maschad (dismissed):

Torres-ssf:
@danielbate 
> First pass, couple comments and questions.
> 
> 1. If I have one utxo and am submitting multiple transactions, `cacheUtxo` has no effect, and in fact we risk the second tx failing but we would be able to fund both txs initially?

If you have just one UTXO and you fund and submit 2 TXs simultaneously:

- without `cacheUtxo`: The first TX to be processed would work, and the second one would fail if the error related to a non-existent UTXO

- with `cacheUtxos`: The first one will be funded normally, but now this UTXO is cached, then when funding the second one an error will thrown related to not having enough balance to fit the target.

When you have just one UTXO, you cannot send multiple TXs without waiting for the first one to complete. It will always fail

> 2. Why would you not want to `cacheUtxo`s?

The only reason that comes to mind would be in a situation where your first TX failed, and then your UTXO that was not being spent is now unusable until the cache for it got whipped.

Also, the error message is different. Maybe you have just one UTXO and since the cache is being used by default, and you are completely unaware of it, if you try to send 2 transactions in a row the error message would be `not enough balance to fit the target` instead of `UTXO not existent`.

This flag makes it possible for users that own multiple UTXOs to submit multiple TXs in a row, without waiting for each one to be processed, and without running into the error UTXO non-existent.

This error will likely happen, as the TXs are going to be funded automatically, and TX1 and TX2 can be funded with the same UTXO(s). Then, when TX1 is processed, that specific UTXO will be spent, and TX2 will fail.

danielbate (approved):
Thanks for the detailed explanation :shipit:

maschad (approved):

Issue #2523, opened 2024-06-15
Consider reviewing GH Issue templates


arboleya:
The current templates are not always helpful:
https://github.com/FuelLabs/fuels-ts/tree/master/.github/ISSUE_TEMPLATE

We should consider reviewing and simplifying them.

We could use a blank page or something straightforward if time is short.

maschad:
I certainly think for creating bugs, the current [bug template](https://github.com/FuelLabs/fuels-ts/blob/master/.github/ISSUE_TEMPLATE/bug.yaml could be amended to include [Stackblitz templates](https://github.com/FuelLabs/fuels-ts/issues/1927) so that we can have easier reproducibility

PR #2623, opened 2024-06-26
chore: add the triage label for new issues

### LINKED ISSUE #2557: Add `triage` label to new issue templates
Externally created issues should be automatically assigned a `triage` label so that we can apply [our process to prioritize them](https://github.com/FuelLabs/fuels-ts/issues/2555)

mvares:
- Closes #2557

### DIFF
--- a/.changeset/lemon-tomatoes-sing.md
+++ b/.changeset/lemon-tomatoes-sing.md
@@ -1,0 +1,4 @@
+---
+---
+
+chore: add the triage label for new issues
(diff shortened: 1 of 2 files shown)


mvares:
> My preference would be to do this via the [issue templates](https://github.com/FuelLabs/fuels-ts/tree/master/.github/ISSUE_TEMPLATE) rather than a workflow? Could tie in nicely to #2523.

Hmm, this seems interesting. Maybe this approach closes the scope only for issues.

Good catch, @danielbate!

Issue #449, opened 2022-08-09
Add to docs a section to explain ethers vs fuels (explanation of UTXO transactions)


luizstacio:

danielbate:
+1 for an encompassing `ethers` -> `fuels` migration guide.

PR #2624, opened 2024-06-26
docs: add doc section for `UTXO` model

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/449


maschad (approved):

danielbate (approved):
Great work Sergio, one comment around the UTXO flow example. Also, do we think this should have it's own section or could be included in transactions to keep the docs structure slim? Approving though as I don't think either of these things should block the PR.

apps/docs/src/guide/the-utxo-model/index.md
```diff
@@ -0,0 +1,34 @@
...
+Suppose you have the following UTXOs in your account:
+
+- $10 UTXO
+- $20 UTXO
+
+You want to send $15 to someone, and the transaction fee is $1. Here's what happens:
+
+- The $20 UTXO is selected to fund the $15 transaction and cover the $1 fee.
+- The transaction is completed and the $20 UTXO is spent.
+- A new $15 UTXO is generated to the recipient, and a new $4 UTXO (change) is created and added to your account.
```
danielbate, line 34: Approved but intended to comment here.

Should we have a snippet showing the flow?
1. Get balance and show UTXOS of different amounts for same asset ID
2. Transfer
3. Spend the higher UTXO
4. Get balance showing new UTXO

nedsalk (approved):

apps/docs/src/guide/the-utxo-model/index.md
Torres-ssf: @danielbate I thought about doing this but I was afraid it would just add more noise. Do you think this doc page will benefit from a more detailed example like this?

Torres-ssf: I think this is good for now. Let's if we receive any feedback regarding this doc page in the future. We can then update it feel necessary

PR #2639, opened 2024-06-27
ci: dependabot changeset push to use `REPO_TOKEN`

### LINKED ISSUE #2559: Investigate dependabot workflow runs hanging
When the [changesets are committed](https://github.com/FuelLabs/fuels-ts/blob/master/.github/workflows/pr-validate-changesets.yaml#L39-L47) by github bot in the dependabot workflow, https://github.com/FuelLabs/fuels-ts/pull/2488 was supposed to have addressed this as well but it seems it didn't,  the CI still hangs as can be seen in https://github.com/FuelLabs/fuels-ts/pull/2554
Screenshot 2024-06-19 at 11 36 11 AM

This isn't a blocker for CI runs as once the branch is updated the workflows [re-run](https://github.com/FuelLabs/fuels-ts/pull/2528)

nedsalk:
I believe that this is the same issue that plagued the changeset PR, which was solved in #1848.

- closes #2559


maschad:
Good find, to test this I wonder if we could fork this repo, create a test secret `REPO_TOKEN`, then modify the [dependabot interval](https://github.com/FuelLabs/fuels-ts/blob/master/.github/dependabot.yml#L6) to force it to open a PR and then see if the workflow runs there?

nedsalk:
@maschad I'm fairly certain that this is the root cause because this solution solved the same problem in #1848. IMO we don't need to go out of our way to test it; merging and verifying the behavior live is okay in this case because, even if this PR is a wrong solution, the workflow itself doesn't block anybody from work nor is it of crucial, "we must get this right the first time" importance.

danielbate (approved):

arboleya (approved):

Issue #2362, opened 2024-05-21
Proposal to Simplify `Provider.getTransactionCost`


Torres-ssf:
We should consider refactoring [Provider.getTransactionCost](https://github.com/FuelLabs/fuels-ts/blob/7c08593037e65ce6256c6cc63a8d0a2fe2f6c0b8/packages/account/src/providers/provider.ts#L1064) by removing the responsibility of adding fake funds to the `TransactionRequest` before the dry-run to get the `gasUsed`. 

This change will delegate this task to the caller of `getTransactionCost`, therefore reducing unnecessary parameter flags.

The current implementation of `getTransactionCost` handles multiple tasks, including adding fake funds for transaction requests. 

By offloading the fake fund's addition to the caller, we can eliminate the need for specific parameter flags associated with the funding step:

1. [resourcesOwner](https://github.com/FuelLabs/fuels-ts/blob/7c08593037e65ce6256c6cc63a8d0a2fe2f6c0b8/packages/account/src/providers/provider.ts#L1066): Represents the `Account` instance that owns the fake resources and it is also used to validate if it is a predicate, which is required for populating the `predicate` and `predicateData` for the fake resources.
2. [quantitiesToContract](https://github.com/FuelLabs/fuels-ts/blob/7c08593037e65ce6256c6cc63a8d0a2fe2f6c0b8/packages/account/src/providers/provider.ts#L1066): This includes the `amount` and `assetId` of a coin that could be forwarded to a contract.

mvares:
hey @petertonysmith94, is this issue still on your radar? 🦅

PR #2643, opened 2024-06-28
chore!: refactored the `getTransactionCost` method

petertonysmith94:
- Closes: #2362

# Release notes

In this release, we:

- Refactored functionality of `Provider.getTransactionCost` to `Account.getTransactionCost`
- Deprecated the `Provider.getResourcesForTransaction` method

# Summary

- Refactored functionality for `Provider.getTransactionCost` to `Account.getTransactionCost`:
  - Separate concerns of estimating and fake funding a transaction.
  - Reducing unnecessary parameters, leading to a cleaner API.
- Deprecated the `Provider.getResourcesForTransaction` method:
  -  Lack of use-cases and deemed not required.

# Breaking Changes

- Refactored functionality for `Provider.getTransactionCost` to `Account.getTransactionCost` **and** changed estimation parameter from `quantitiesToContract` to `quantities`.

```ts
// Before
const provider = Provider.create(...);
const account = Wallet.generate({ ... }) || new Predicate(...);
const quantities: Array<CoinQuantityLike> = [
  { amount: 1000, assetId: provider.getBaseAssetId() }
];

provider.getTransactionCost(txRequest, {
  resourceOwner: account,
  quantitiesToContract: quantities,
})
```

```ts
// After
const provider = Provider.create(...);
const account = Wallet.generate({ ... }) || new Predicate(...);
const quantities: Array<CoinQuantityLike> = [
  { amount: 1000, assetId: provider.getBaseAssetId() }
];

account.getTransactionCost(txRequest, {
  quantities
});
```

- Deprecation of the `Provider.getResourcesForTransaction` method.

# Checklist


packages/account/src/account.ts
```diff
@@ -569,6 +572,90 @@ export class Account extends AbstractAccount {
...
+
+    /**
+     * The fake utxos added above can be from a predicate
+     * If the resources owner is a predicate,
+     * we need to populate the resources with the predicate's data
+     * so that predicate estimation can happen.
+     */
+    if (this && 'populateTransactionPredicateData' in this) {
+      (this as unknown as Predicate<[]>).populateTransactionPredicateData(txRequestClone);
+    }
```
Torres-ssf, line 607: By funding the TX with `Account.generateFakeResources` instead of `txRequestClone.fundWithFakeUtxos`, we could remove this check enterely.

packages/account/src/account.ts
```diff
@@ -569,6 +572,90 @@ export class Account extends AbstractAccount {
...
+   *
+   * @param transactionRequestLike - transaction request to populate resources for.
+   * @param quantitiesToContract - quantities for the contract (optional).
+   *
+   * @returns a promise resolving to the required quantities for the transaction.
+   */
+  async getResourcesForTransaction(
+    transactionRequestLike: TransactionRequestLike,
+    quantitiesToContract: CoinQuantity[] = []
+  ) {
```
Torres-ssf, line 631: What is the motivation for moving this method to the Account class?

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -216,22 +223,53 @@ export class BaseInvocationScope<TReturn = any> {
...
+
+    /**
+     * The fake utxos added above can be from a predicate
+     * If the resources owner is a predicate,
+     * we need to populate the resources with the predicate's data
+     * so that predicate estimation can happen.
+     */
+    if (this.program.account && 'populateTransactionPredicateData' in this.program.account) {
+      (this.program.account as unknown as Predicate<[]>).populateTransactionPredicateData(request);
+    }
```
Torres-ssf, line 262: If the `Account.getTransactionCost` is already doing the same thing we have here, why do we need this code here?

Torres-ssf (commented):
Hey @petertonysmith94 I understand that this is a draft yet, just leave a couple of questions here to better understand your motivations

packages/account/src/account.ts
Torres-ssf: @petertonysmith94 what was the error?

packages/account/src/account.ts
Torres-ssf: @petertonysmith94 I was just curious about this. I am under the impression that we can remove `getResourcesForTransaction` entirely.

packages/program/src/functions/base-invocation-scope.ts
Torres-ssf: The `getTransactionCost` here can be called by any of the `dryRun` methods so if the account is not defined, we can simply generate a new one using the provider. This method is only used for estimation purposes.

packages/account/src/account.ts
petertonysmith94: @Torres-ssf encountered a couple issues:

- `fundWithFakeUtxos` adds an input for the `baseAssetId` (if no present), or updates the existing input [quantity](https://github.com/FuelLabs/fuels-ts/blob/3b27bac26463b39b8201db96147ec4ef091d3de0/packages/account/src/providers/transaction-request/transaction-request.ts#L596) for the estimate. I believe this is be why I'm experiencing the `InsufficientInputAmount` related errors.

- I also attempted to use the `generateFakeResources` alongside the `fundWithFakeUtxos` and removing the `populateTransactionPredicateData`. However, also encountered some issues with the correct `generateFakeResources` not being used via inheritance.

Might be beneficial for us to have a chat over huddle - as I'd like to get these breaking changes over the line ASAP.

Torres-ssf: @petertonysmith94 Sure, I've been trying to mimic this issue without success.

Maybe it will be better to discuss it over a huddle

Torres-ssf: @petertonysmith94 Sorry for the delay here but it took me a while to figure out what the error was.

The problem is that you forget to add an amount to fund the TX using the base asset, to cover the TX fee for the dry-run

```ts
  async getTransactionCost(
    { signatureCallback, quantitiesToContract = [] }: TransactionCostParams = {}
  ): Promise<TransactionCost> {
    const txRequestClone = clone(transactionRequestify(transactionRequestLike));
    const baseAssetId = this.provider.getBaseAssetId();

    // Fund with fake UTXOs to avoid not enough funds error
    // Getting coin quantities from amounts being transferred
    const coinOutputsQuantities = txRequestClone.getCoinOutputsQuantities();
    // Combining coin quantities from amounts being transferred and forwarding to contracts
    const requiredQuantities = mergeQuantities(coinOutputsQuantities, quantitiesToContract);
    const resources = this.generateFakeResources(
      mergeQuantities(
        [...requiredQuantities],
        [{ assetId: baseAssetId, amount: bn('100000000000000000') }]
      )
    );

    txRequestClone.addResources(resources);

    const txCost = await this.provider.getTransactionCost(txRequestClone, {
      signatureCallback,
    });

    return {
      ...txCost,
      requiredQuantities,
    };
  }
```

petertonysmith94: 33d9b1b346cae412d901fda51a66451755477f1e

packages/account/src/account.ts
```diff
@@ -500,6 +502,45 @@ export class Account extends AbstractAccount {
...
+    // Fund with fake UTXOs to avoid not enough funds error
+    // Getting coin quantities from amounts being transferred
+    const coinOutputsQuantities = txRequestClone.getCoinOutputsQuantities();
+    // Combining coin quantities from amounts being transferred and forwarding to contracts
+    const requiredQuantities = mergeQuantities(coinOutputsQuantities, quantitiesToContract);
+    const resources = this.generateFakeResources(
+      mergeQuantities(
+        [...requiredQuantities],
+        [{ assetId: baseAssetId, amount: bn('100000000000000000') }]
+      )
```
Torres-ssf, line 522: @petertonysmith94 Maybe we should make it clear here that this quantity in `baseAssetId` is being added to pay for the TX fee at the `dryRun`

packages/account/src/account.ts
```diff
@@ -500,6 +502,45 @@ export class Account extends AbstractAccount {
...
+    );
+    txRequestClone.addResources(resources);
+
+    /**
+     * Estimate predicates gasUsed
+     */
+    // Remove gasLimit to avoid gasLimit when estimating predicates
+    if (isScriptTransaction) {
+      txRequestClone.gasLimit = bn(0);
+    }
```
Torres-ssf, line 532: @petertonysmith94 Perhaps this should be addressed on `Provider.getTransactionCost`

.changeset/short-sheep-divide.md
```diff
@@ -0,0 +1,8 @@
+---
+"@fuel-ts/account": minor
+"@fuel-ts/interfaces": patch
+"@fuel-ts/contract": patch
+"@fuel-ts/program": patch
```
Torres-ssf, line 5: @petertonysmith94 Shouldn't be `minor` for everyone here?  

The packages `contract`, and `program` are using the `account` package, and `interfaces` is being used by `program`?

Torres-ssf (changes requested):
@petertonysmith94 Great job 👏

packages/account/src/account.ts
petertonysmith94: a14330cf21d98d3c9b5f0dab5f7914b0122fa704

packages/account/src/account.ts
petertonysmith94: f6f66789fc41f704493de07a6f125d1de69f2f40

.changeset/short-sheep-divide.md
petertonysmith94: The rationale behind this was that the `@fuel-ts/account` is the only package with breaking changes.

Changed: b9058e0812bb123d3f1dd1be1bae4eadbaddd9e8

Torres-ssf:
@ petertonysmith94 We still have some tests failing here.

Please let me know if you are having a hard time addressing them.

packages/account/src/providers/provider.test.ts
```diff
@@ -1075,13 +1075,15 @@ Supported fuel-core version: ${mock.supportedVersion}.`
   // TODO: validate if this test still makes sense
   it.skip('should ensure estimated fee values on getTransactionCost are never 0', async () => {
```
maschad, line 1076: I think this test should still be run, otherwise if this test is being skipped what's the purpose of these changes?

packages/account/src/providers/provider.ts
```diff
@@ -1131,40 +1123,18 @@ Supported fuel-core version: ${supportedVersion}.`
    */
   async getTransactionCost(
     transactionRequestLike: TransactionRequestLike,
-    { resourcesOwner, signatureCallback, quantitiesToContract = [] }: TransactionCostParams = {}
-  ): Promise<TransactionCost> {
+    { signatureCallback }: TransactionCostParams = {}
+  ): Promise<Omit<TransactionCost, 'requiredQuantities'>> {
```
maschad, line 1127: Given these changes are already breaking, why not just remove the `requiredQuantities` from the `TransactionCost` type?

packages/account/src/providers/provider.test.ts
petertonysmith94: > TODO: validate if this test still makes sense

With the following todo - I am unsure about this test.

From what I can see, the `minFee` will never be `0` as we add `1` after the `calculateGasFee`:
https://github.com/FuelLabs/fuels-ts/blob/8776b79986b8c47b34a9954278c3d9eafd504c09/packages/account/src/providers/provider.ts#L1024-L1029

Do you think this test is still valid?

packages/account/src/providers/provider.ts
petertonysmith94: The `requiredQuantities` is required for the `TransactionCost` result in the `Account`. I've chosen to share this type between the `Provider.getTransactionCost` and `Account.getTransactionCost` methods, hence the omission here.

packages/account/src/providers/provider.test.ts
maschad: That's correct but I think the test ensures that code remains as such, so I think it's valid. But I don't have the full context.

arboleya (commented):
Conflicts to fix.

packages/account/src/providers/provider.test.ts
Torres-ssf: @petertonysmith94 This test can be skipped for now. We can unskip this after https://github.com/FuelLabs/fuels-ts/issues/2813 is solved.

If we cannot use a gas price of `0`, the `maxFee` will never be `0`.

> From what I can see, the minFee will never be 0 as we add 1 after the calculateGasFee:

@petertonysmith94 you are correct, since `calculateGasFee` is always adding `1`, even when the `gasPrice` is `0`, the `maxFee` | `minFee` will never be `0`. 

But this can serve as validation that we are always adding `1` to `maxFee` and `minFee`.

petertonysmith94: @Torres-ssf I think I'll wait for the merge of #2814 - may as well resolve this as one. Unless you'd advise differently?

packages/account/src/account.ts
```diff
@@ -500,6 +501,35 @@ export class Account extends AbstractAccount {
     return this.sendTransaction(request);
   }
 
+  async getTransactionCost(
+    transactionRequestLike: TransactionRequestLike,
+    { signatureCallback, quantitiesToContract = [] }: TransactionCostParams = {}
```
Torres-ssf, line 506: The property name `quantitiesToContract` could use a renaming as it is not always used as quantities being forwarded to contracts. 

The Account method [withdrawToBaseLayer](https://github.com/FuelLabs/fuels-ts/blob/f3453b9ea669575e3cf97ee50512bc9039845e0c/packages/account/src/account.ts#L464) uses the property to inform the amount of the base asset that should be withdrawn to the base layer.

This is not a hard requirement but in my opinion, since we are already refactoring and breaking things, we could rename this property. What do you think?

Torres-ssf (commented):
@petertonysmith94 I feel like we need to update this doc [page](https://github.com/FuelLabs/fuels-ts/blob/master/apps/docs/src/guide/contracts/cost-estimation.md) to say that the `getTransactionCost` method is provided by the Account class instead of the Provider class, what do you think?

packages/account/src/providers/provider.test.ts
Torres-ssf: @petertonysmith94 That is ok. We can wait for https://github.com/FuelLabs/fuels-ts/pull/2814

packages/account/src/account.ts
petertonysmith94: I agree, however, I wasn't sure of a suitable name. What about just `quantities`?

Torres-ssf: Sure, this one is good as any

packages/account/src/providers/provider.test.ts
petertonysmith94: Will resolve this after this has been merged:
- https://github.com/FuelLabs/fuels-ts/pull/2814

packages/account/src/account.ts
petertonysmith94: - Fixed docs: 126107d7d6e69190c946a4ef20c2fc1b00537802
- Updated parameter: 4c81e4b4c0ebd11fddd1c3e3f326645b498534d9

Torres-ssf (approved):

maschad (approved):
Nice job!

arboleya (approved):

PR #2648, opened 2024-06-29
feat: drafting `fuels` entry point

arboleya:
Draft based on:
 - https://github.com/FuelLabs/fuels-ts/issues/2490

```bash
./api
├── fuels.ts # ...................... Sugar API
├── index.ts # ...................... Overloaded `fuels()` entry point
└── snippets # ...................... Code snippets (for new API)
    └── typegend # .................. Mocked Files
```

Related PR combining this PR with the revised code snippets proposal:
 - https://github.com/FuelLabs/fuels-ts/pull/2651

# TL;DR

https://github.com/FuelLabs/fuels-ts/blob/7f3d64bc30c71f5b42b936b182c3d0dd61d1fcf1/packages/fuels/src/api/snippets/minimal.ts#L8-L10

https://github.com/FuelLabs/fuels-ts/blob/7f3d64bc30c71f5b42b936b182c3d0dd61d1fcf1/packages/fuels/src/api/snippets/minimal.ts#L14-L16

# All in One

https://github.com/FuelLabs/fuels-ts/blob/7f3d64bc30c71f5b42b936b182c3d0dd61d1fcf1/packages/fuels/src/api/snippets/allinone.ts#L8-L18

# Modes

## 1. Promise

https://github.com/FuelLabs/fuels-ts/blob/7f3d64bc30c71f5b42b936b182c3d0dd61d1fcf1/packages/fuels/src/api/snippets/contract.ts#L29-L34

## 2. Callback

https://github.com/FuelLabs/fuels-ts/blob/7f3d64bc30c71f5b42b936b182c3d0dd61d1fcf1/packages/fuels/src/api/snippets/contract.ts#L21-L24

## 3. Custom instance

https://github.com/FuelLabs/fuels-ts/blob/7f3d64bc30c71f5b42b936b182c3d0dd61d1fcf1/packages/fuels/src/api/snippets/contract.ts#L39-L43

# Modules

## 1. Wallet

https://github.com/FuelLabs/fuels-ts/blob/7f3d64bc30c71f5b42b936b182c3d0dd61d1fcf1/packages/fuels/src/api/snippets/wallet.ts#L14-L15

## 2. Predicate

https://github.com/FuelLabs/fuels-ts/blob/7f3d64bc30c71f5b42b936b182c3d0dd61d1fcf1/packages/fuels/src/api/snippets/predicate.ts#L18-L20

## 3. Contract

[...]

Issue #2399, opened 2024-05-28
Implement new setup for code snippets


arboleya:
# Pooblem

In the past, we had a couple of issues involving code snippets:
 - https://github.com/FuelLabs/fuels-ts/issues/1960

# Examples

They can be missing imports, have invalid syntax, etc. We must improve our handling of these to ensure that the code chunks we provide in the docs are 100% reliable.

 - https://docs.fuel.network/docs/fuels-ts/getting-started/connecting-to-a-local-node/
 - https://docs.fuel.network/docs/fuels-ts/getting-started/connecting-to-testnet/

![comment](https://github.com/FuelLabs/fuels-ts/assets/26660/79066fbe-3c80-46d0-a22a-86ec7d776c2a)

![comment2](https://github.com/FuelLabs/fuels-ts/assets/26660/0e0f222f-38f4-4066-bfe0-0a929ee2e8d8)
</details>

---

# Propoal

I outlined a proposal for it in this discussion — https://github.com/FuelLabs/fuels-ts/discussions/2159#discussioncomment-9219003.

Here's a quick PoC for it — https://github.com/arboleya/docs-widgets.

arboleya:
Blocked by:
 - https://github.com/FuelLabs/fuels-ts/issues/2712

Issue #1927, opened 2024-03-22
Create a Stackblitz dev container for `create-fuels`


Dhaiwat10:
It could be worth having a Stackblitz template for the create-fuels template app. This will be a fullstack Fuel dev container that works fully in the browser.

We should mention this in our docs if we create one such container, as well as using it in the [bug issue template](https://github.com/FuelLabs/fuels-ts/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=bug.yaml).

maschad:
We will probably want to have this running against a Testnet fuel node, so I can re-address it once that's live.

danielbate:
We can unblock this now due to devnet.

arboleya:
Ideally, we should be able to run `fuel-core` and `forc` node on [Stackblitz](https://stackblitz.com/). 🙃 

Simple things:

```ts
// package.json
{
  "devDependencies": {
    "@fuels/forc": "0.60.0",
    "@fuels/fuel-core": "0.30.0",
  }
}
```

If we'd cross-publish `fuelup`:

```ts
// package.json
{
  "scripts": {
    "install": "fuelup default testnet",
  },
  "devDependencies": {
    "@fuels/fuelup": "0.26.0"
  }
}
```

```toml
# fuel-toolchain.toml
[toolchain]
channel = "testnet"

[components]
forc = "0.60.0"
fuel-core = "0.30.0"
```

We should consider approximating the Rust and NodeJs ecosystems.

Cross-publishing some crates to NPM could be a start. 🙂

arboleya:
Yes, we should consider approximating the Rust and NodeJs ecosystems.

Cross-publishing crates to NPM could be a good start. 🙂

Just like it happens with `fuel-asm` ([source](https://github.com/FuelLabs/fuel-vm/blob/56d7deb66a70964d2b22de80299585f2b5d36fab/.github/workflows/ci.yml#L298-L325)).

danielbate:
Marking as blocked as this topic is being investigated by product.

arboleya:
Let's close for now.

PR #2651, opened 2024-06-30
feat: new infra for code snippets

arboleya:
Draft based on:
 - https://github.com/FuelLabs/fuels-ts/issues/2399

Designed after:
 - https://github.com/FuelLabs/fuels-ts/pull/2648

In short, this is a valid code snippet:

https://github.com/FuelLabs/fuels-ts/blob/d0158cd585772c9c55d743bb795471ddc7bb8fc1/apps/docs-snippets2/src/wallet/getBalance.ts#L1-L9


apps/docs-snippets2/src/wallet/getBalance.test.ts
```diff
@@ -0,0 +1,10 @@
+import { main } from './getBalance.wrapped';
+
+/**
+ * @group node
+ * @group browser
+ */
+test('it works', async () => {
+  const { balances } = await main();
+  expect(balances).toBeTruthy();
+});
```
arboleya, line 10: How to use `launchTestNode` without polluting the [code snippet](https://github.com/FuelLabs/fuels-ts/pull/2651/files#diff-410c18e6bfa8e46676076161046c8a19c44c71c3736ea45b56578606eb0cd3a4) below?

apps/docs-snippets2/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "module": "ES2022",
```
arboleya, line 4: Had to use `"module": "ES2022"` to enable top-level `await`.

apps/docs-snippets2/src/wallet/getBalance.ts
```diff
@@ -0,0 +1,9 @@
+import { fuels, TESTNET_NETWORK_URL } from 'fuels/api';
+
+const MY_WALLET_ADDRESS = '0x767caf5b08eba21c561078a4d5be09bbd7f16b9eca22699a61f1edd9e456126f';
+
+const { wallet } = await fuels(TESTNET_NETWORK_URL);
+
+const balances = await wallet(MY_WALLET_ADDRESS).getBalances();
+
+console.log({ balances });
```
arboleya, line 9: How to use `launchTestNode` without polluting the code snippet?

arboleya: The Stackblitz topic runs [independently](https://github.com/FuelLabs/fuels-ts/issues/1927#issuecomment-2198814457) of this one, but at one point, I think we'd want that.

arboleya: cc @danielbate @nedsalk

apps/docs-snippets2/scripts/wrap-snippets.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const wrapper = `export const main = async (): Promise<any | any[]> => {
+  const logs: any[] = [];
+  const console = { log (...args: any[]) { logs.push(args) }};
+  // ———>>>
+  %SNIPPET%
+  // <<<———
+  const singleCall = logs.length === 1 && logs[0].length === 1;
+  return singleCall ? logs[0][0] : logs;
+}`.replace(/^\s{4}/gm, '  ');
```
arboleya, line 14: All code snippets are wrapped inside a `main` function and saved to a `.wrapped.ts` equivalent.

The params for all `console.log()` calls inside the snippet are collected and can be later asserted.

apps/docs-snippets2/src/introduction/connecting-to-localnode.test.ts
```diff
@@ -0,0 +1,11 @@
+import { main } from './connecting-to-testnet.wrapped';
```
arboleya, line 1: In tests, wrapped (generated) snippets should be imported instead of the original ones.

apps/docs-snippets2/src/introduction/connecting-to-localnode.test.ts
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * @group node
+ * @group browser
+ */
+test('it works', async () => {
+  const { balances } = await main();
+  expect(balances).toBeTruthy();
+  expect(balances).toBeInstanceOf(Array);
```
arboleya, line 10: Then, the (wrapped) snippet can be tested from outside.

apps/docs-snippets2/src/introduction/connecting-to-testnet.ts
```diff
@@ -0,0 +1,14 @@
...
+
+// Create a provider using the `TESTNET_NETWORK_URL` constant
+const provider = await Provider.create(TESTNET_NETWORK_URL);
+
+// Then use it when creating a wallet
+const wallet = Wallet.fromAddress(ADDRESS, provider);
+const balances = await wallet.getBalances();
+
+console.log({ balances });
+// #endregion main
```
arboleya, line 14: And the snippet remains clean, without unrelated code.

arboleya:
snippeets-gist

Torres-ssf (commented):
This is really nice @arboleya 👍 

I've noticed that for these 2 new snippets, the anchor link `see code in context` is not working anymore.

I wonder if we should remove this functionality entirely, as it does not work with the integrated docs.

arboleya:
@Torres-ssf I think we should keep the `see code in context` button and also add it to the Docs Hub.

> cc @matt-user @calldelegation @petertonysmith94

Torres-ssf (approved):

apps/docs-snippets2/src/introduction/connecting-to-localnode.ts
```diff
@@ -0,0 +1,14 @@
+// #region main
```
nedsalk, line 1: Do we still want to remain with the `// #region xxx ... #endregion xxx` pattern? It seems to me that a change in how the snippets are read would remove the need for regions and we'd go from 
```
<<< @/../../docs-snippets2/src/introduction/connecting-to-localnode.ts#main{ts:line-numbers}
```
to
```
<<< @/../../docs-snippets2/src/introduction/connecting-to-localnode.ts
```

We can even update it to be a bit nicer:
```
code-snippet@/../../docs-snippets2/src/introduction/connecting-to-localnode.ts
```

arboleya: Yes, you are correct, but after further thought, I concluded it could be helpful to keep it for situations where the snippets are inevitably big and one wants to present it in consecutive slices.

nedsalk: Okay, then we can also use it to mitigate the "should we show the `console.log` in the snippet or not" issue by putting the `console.log` after `#endregion main`.

danielbate: IMO we keep the pattern, we've currently got it in places in `src` and in `.sw` files where the entire file wouldn't be practical.

danielbate (commented):
This is a really nice, creative solution to the problem. However, unless I am missing something obvious, why can't we do something like this? As this solution feels a bit like in `src` testing however across multiple files? 🤔 

```ts
// #region main
import { Provider, Wallet, LOCAL_NETWORK_URL } from 'fuels';

const ADDRESS = '0x767caf5b08eba21c561078a4d5be09bbd7f16b9eca22699a61f1edd9e456126f';

// Create a provider using the `LOCAL_NETWORK_URL` constant
const provider = await Provider.create(LOCAL_NETWORK_URL);

// Then use it when creating a wallet
const wallet = Wallet.fromAddress(ADDRESS, provider);
const balances = await wallet.getBalances();
// #endregion main

/**
 * @group node
 * @group browser
 */
test('it works', async () => {
  const someAsyncImport = await import('fuels');
  const someValueToAssert = someAsyncImport.bn(1);

  expect(balances).toBeTruthy();
  expect(balances).toBeInstanceOf(Array);
  expect(someValueToAssert.toString()).toBe('0x01');
});
```
Also been through the solution with @nedsalk and we came up with a possible solution for `launchTestNode` and overriding `fuels` imports without polluting the snippet-related imports:

```ts
const overrideFuelsImports = async () => {
  const { launchTestNode } = await import('fuels/test-utils');
  const someAsyncImport = await import('fuels');

  const { provider, cleanup } = await launchTestNode();

  Object.defineProperty(someAsyncImport, 'LOCAL_NETWORK_URL', { value: provider.url });

  return {
    [Symbol.dispose]: cleanup,
  };
};
```

The full snippet could then look like so:

```ts
const overrideFuelsImports = await import ('fuels/test-utils');
using overridden = await overrideFuelsImports(/** things to override */);

// #region main
import { Provider, Wallet, LOCAL_NETWORK_URL } from 'fuels';

console.log('MY URL,', LOCAL_NETWORK_URL);

const ADDRESS = '0x767caf5b08eba21c561078a4d5be09bbd7f16b9eca22699a61f1edd9e456126f';

// Create a provider using the `LOCAL_NETWORK_URL` constant
const provider = await Provider.create(LOCAL_NETWORK_URL);

// Then use it when creating a wallet
const wallet = Wallet.fromAddress(ADDRESS, provider);
const balances = await wallet.getBalances();
// #endregion main

/**
 * @group node
 * @group browser
 */
test('it works', async () => {
  const someAsyncImport = await import('fuels');
  const someValueToAssert = someAsyncImport.bn(1);

  expect(balances).toBeTruthy();
  expect(balances).toBeInstanceOf(Array);
  expect(someValueToAssert.toNumber()).toBe(1);
});
```

arboleya:
@danielbate Having tests together pollutes the snippet file regardless of the portion we show in the docs.

The center of this proposal is keeping the snippets **pristine**, which is why all the work is done _around_ it.

We will soon have the `see code in context` button in the Docs Hub, another reason for the above.

I also see value in running snippets as is, especially when getting to StackBlitz.

```bash
$ tsx my-snippet.ts
# stuff gets logged to the console
```

danielbate (dismissed):
Thankyou for the clarification, that perspective makes sense. I think it's important to know that this solution isn't just trying to be a more feature complete version of the previous solution. But actually we are treating the snippet now as a shippable piece of code, and the docs is a consumer of that. We could have a range of transformers for these snippets to serve different purposes, StackBlitz, `create-fuels` UI builder even. LGTM 👍🏻

nedsalk (dismissed):

arboleya:
> I've noticed that for these 2 new snippets, the anchor link see code in context is not working anymore.

@Torres-ssf Forgot about this. Any tips?

> [..] but actually we are treating the snippet now as a shippable piece of code, and the docs is a consumer of that.

@danielbate Yep, that's a good way of wording it.

This PR is still missing the `launchTestNode` integration.

Torres-ssf:
> > I've noticed that for these 2 new snippets, the anchor link see code in context is not working anymore.
> 
> @Torres-ssf Forgot about this. Any tips?
> 

@arboleya I believe we can leave it as is for now. The `see code in context` button currently works only in developer mode, as our documentation is linked to the integrated documentation. 

We should focus on making this button functional on the docs hubs, keeping the new snippet solution in mind as we proceed.

apps/docs-snippets2/scripts/wrapper-fn.ts
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+export const main = async (localNetworkUrl?: string): Promise<any | any[]> => {
```
arboleya, line 4: The wrapper method can optionally receive `localNetworkUrl` from outside.

If informed, that overrides `LOCAL_NETWORK_URL`.

arboleya (commented):
@danielbate @nedsalk Still overriding `LOCAL_NETWORK_URL`, but differently.

snippet

Torres-ssf (dismissed):

arboleya:
The code below results in these:
 - [Transaction Parameters](https://fuels-pv7rmz9mw-fuel-labs.vercel.app/guide/transactions/transaction-parameters.html)
 - [TransactionParameters - Full Example](https://fuels-pv7rmz9mw-fuel-labs.vercel.app/guide/transactions/transaction-parameters.html#full-example) (something we could start doing more)

---

The test node launcher is injected at the wrapper level and bring the snippet to life.

 1. Snippets uses stubbed env files
 1. Snippets are wrapped with node launcher
 1. Stubbed env vars are overridden with test node launcher info
 1. Tests run against wrapped snippets


snippets

.changeset/config.json
```diff
@@ -17,6 +17,7 @@
     "demo-node-esm",
     "demo-typegen",
     "@fuel-ts/docs-snippets",
+    "@fuel-ts/docs-snippets2",
```
petertonysmith94, line 20: Why do we prefix the `docs-snippets` with `@fuel-ts/`? I thought we only did this for packages?

Dhaiwat10:
Really neat PR! What's the status of this one? Should we review it, or wait before it's marked as ready?

Torres-ssf (approved):

apps/docs-snippets2/src/introduction/connecting-to-localnode.ts
```diff
@@ -0,0 +1,14 @@
...
+import { LOCAL_NETWORK_URL } from '../env';
+
+const ADDRESS = '0x767caf5b08eba21c561078a4d5be09bbd7f16b9eca22699a61f1edd9e456126f';
+
+const provider = await Provider.create(LOCAL_NETWORK_URL);
+const wallet = Wallet.fromAddress(ADDRESS, provider);
+
+const { balances } = await wallet.getBalances();
+
+console.log(`Balances should be an array`, balances instanceof Array);
```
maschad, line 13: @arboleya I had actually thought the `console.assert` pattern was more expressive as it offered a pass/fail debug experience. What was the rationale for the change to `console.log` ?

maschad (commented):
Looks good to me but I left a comment re: the `console.log` 

Should we create a separate issue to eventually remove the `doc-snippets` folder as we port all the snippets over?

apps/docs-snippets2/src/introduction/connecting-to-localnode.ts
arboleya: I saw no advantage in using `assert()` over `log()`, since both require the same [patching](https://github.com/FuelLabs/fuels-ts/blob/254bb621e2b9f4835cc726d7e394b15cf36ce25f/apps/docs-snippets2/scripts/test-template.ts#L20-L27). Also, native `log()` always logs something, while `assert()` logs only failures. Ultimately, going with `log()` also goes well with the initial idea of removing testing concepts from snippets.

In fact, I'd argue `log` is more _~expressive~_ natural (clean) for a code snippet. For example, this snippet you commented on would look much better like this while still passing the same `!!truthy` validation.

```ts
console.log(`Balances`, balances);
```


Issue added:
 - https://github.com/FuelLabs/fuels-ts/issues/3205

maschad: Good points.

maschad (approved):

maschad (approved):
Awesome work on this btw @arboleya  🚀  🥳  I know this has been the result of a lot of discussion within the team to get to this point, so great to see the end result in sight.

petertonysmith94 (dismissed):

.changeset/config.json
arboleya: 01c6d10cd371a6d7fafa5f4875943ad3239b8490

Torres-ssf (approved):

Torres-ssf (approved):

maschad (approved):

danielbate (approved):

Issue #2457, opened 2024-06-06
Update integration tests to use `launchTestNode` 


nedsalk:

arboleya:
Duplicate of:
 - https://github.com/FuelLabs/fuels-ts/issues/1820

PR #2670, opened 2024-07-01
fix: flaky predicate test

nedsalk:
This one has been failing in recent CI runs so I fixed it before #2457 to reduce flakiness.


Torres-ssf:
Hey @nedsalk,

Do you know if this PR is still valid or should we close it?

nedsalk:
@Torres-ssf it's valid and ready for review again, I fixed the problem. The amount required for the transfer from the predicate to happen was too small so I had to increase the `amountToPredicate` value quite a bit. Previously, this predicate was funded from multiple tests and thus was passing in some places and sometimes failing in other.

arboleya (approved):

nedsalk:
Okay, so apparently me removing the seeding of this predicate address in this file's tests resulted in other tests failing because now _they_ don't have enough funds :cold_sweat: :hot_face: :cold_face:

I'm putting it into draft again.

nedsalk:
#2457 is around the corner and I'll close this in favor of that PR.

Issue #2361, opened 2024-05-21
Implement Transaction Queue Support


Torres-ssf:
The TS SDK should support queuing of transactions. This feature would allow developers to queue multiple transactions and ensure they are executed sequentially. For example, once transaction X is completed, transaction X+1 should automatically start, and so on. (i.e. once X tx is done, do X + 1, etc).

This feature is particularly useful in scenarios where a series of dependent transactions need to be processed in a specific order. It simplifies the workflow for developers who otherwise would have to manually handle the sequencing and execution of transactions.

> Extracted from: https://github.com/FuelLabs/fuels-ts/issues/161

nedsalk:
Would this cover using the outputs of one transaction as inputs for the next?

Torres-ssf:
@nedsalk I believe it would, but the funding of transaction X + 1 should happen after transaction X is processed (In case X + 1 needs to use UTXOs created by X). 

Since this approach is designed to wait for X to be processed in order to submit X + 1, any output produced from X (UTXOs or even a deployed contract) will be usable on X + 1.

maschad:
I think we need should consider the various scenarios that could arise by introducing a queue as this is more complex than stated in the issue. I've [opened a discussion](https://github.com/FuelLabs/fuels-ts/discussions/2678) to address this, we can then re-adjust the issue based on the outcome of that discussion.

danielbate:
Closing for now so all comms can move to #2678, we can reopen this should we have a resolution from that discussion.

PR #2672, opened 2024-07-01
feat!: send multiples transactions to the network

mvares:
- Closes: #2361


packages/account/src/account.ts
```diff
@@ -620,12 +620,46 @@ export class Account extends AbstractAccount {
...
+
+      if (!response) {
+        throw new FuelError(
+          ErrorCode.TRANSACTION_FAILED,
+          'Transaction failed during batch processing. Aborting further transactions.'
+        );
+      }
+    }
+
+    // TODO: decide whether we should return an array of successful transactions or non
```
mvares, line 660: @Torres-ssf, if a X was processed, but X + 1 wasn't processed, is it possible to revert X?

Torres-ssf: @mvares No, we cannot revert a TX that was already processed. 

For this one needs to return an array of successful/reverted transactions

maschad:
Thanks for taking this on @mvares , I think there needs to be [more discussion](https://github.com/FuelLabs/fuels-ts/discussions/2678) around the issue first before beginning any work on it as it's quite complex.

Issue #2492, opened 2024-06-10
Make `call` method a non-blocking call


guimroque:
### Motivation

as we are a multi-sig wallet, the contract call will not be sent instantly. people need to sign it first. for that we need the fuel sdk to add support for contract call allow to don't wait for execution

this is a huge UX blocker in our flow, making the DAPP get stuck waiting for the transaction to complete because we can't opt out of waiting for execution

Torres-ssf:
Hey @guimroque. Indeed when calling `BaseInvocationScope .call` it is not possible to not wait for the TX execution.

However, there is a way around this. The following example executes a contract function `increment_counter`.

```ts
const transactionRequest = await contract.functions
          .increment_counter(1)
          .txParams({ gasLimit: 200_000, maxFee: 150_000 })
          .fundWithRequiredCoins()
          
const transactionResponse = await wallet.sendTransaction(transactionRequest, { awaitExecution: false })
```

I wonder if this approach solves your problem and unblocks you.

PR #2692, opened 2024-07-03
feat!: implement non-blocking contract call

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/2492

# Release notes

In this release, we:
- Made `BaseInvocationScope.call` a non-blocking call

# Summary

This PR updates the `call` method for better efficiency. Previously, `call` was a blocking method waiting for the transaction's execution. Now, `call` resolves immediately after the transaction is submitted and returns a callback, `waitForResult`, which resolves when the transaction is processed.

# Breaking Changes

The `call` method in the `BaseInvocationScope` class no longer waits for transaction execution, making it non-blocking. This change affects how transaction responses are handled.

```ts
// before
const { logs, value, transactionResult } = await contract.functions.xyz().call()
```

```ts
// after
const { transactionId, waitForResult } = await contract.functions.xyz().call();

const { logs, value, transactionResult } = await waitForResult();
```

# Checklist


.changeset/spicy-trains-provide.md
```diff
@@ -0,0 +1,6 @@
+---
+"@fuel-ts/contract": minor
+"@fuel-ts/program": minor
+---
+
+feat!: implement non-blocking contract call
```
arboleya, line 6: The breaking changes in this one seem very problematic and will affect all our users without exception. Have you considered adding a new method, as in #2597?

Torres-ssf: @arboleya I've modified the `call` method to be non-blocking. I believe this breaking change ensures that users are aware of the update and can adjust their projects accordingly. 

This modification reflects the natural behavior of a blockchain environment, where transaction processing can take several seconds, especially during network congestion. This approach is also consistent with how other popular web3 libraries handle transaction submissions.

To accommodate users who prefer a call that will resolve only after the transaction is processed, I've implemented a new method, `callAndWait`, which retains the previous behavior of the `call` method.

Using a flag within the same method to switch between blocking and non-blocking behaviors would result in different response types, potentially leading to unexpected results. Having two distinct methods avoids this issue and simplifies the interface, in my opinion.

After reflecting on this, I believe it makes sense to modify https://github.com/FuelLabs/fuels-ts/pull/2597 to mirror this approach. Where, `deployContract` would be a non-blocking call, and a new method, `deployAndWait`, would serve as the blocking call, resolving only when the contract is deployed.

danielbate: So the current playoff is what we consider the expected behaviour vs scale of the breaking change. Firstly, I do agree with separate methods. Given the impact, it makes sense to make the breakage as obvious as possible. 

Is there any other case for making the blocking call the default behaviour of `call`/`deployContract` other than it being the natural behaviour of a blockchain environment (which is a strong case)?

Because currently edging on the side of caution may make sense given current timescales and rollout. These methods could remain blocking and we have `callAsync`/`deployContractAsync` for non-blocking. Users that want this change can be made aware of it. 

Then in `1.0.0`, we could move to `call`/`deployContract` being blocking and `callSync`/`deployContractSync` (or `...andWait`). Given the major upgrade such a change would be expected.

apps/docs/src/guide/contracts/methods.md
```diff
@@ -24,12 +24,20 @@ A funded wallet it's required:
 
 ## `call`
 
-The `call` method should be used to submit a real contract call transaction to the node.
+The `call` method submits a real contract call transaction to the node. It resolves as soon as the transaction is submitted and returns a `transactionId` along with a `waitForResult` callback to wait for the transaction execution.
```
danielbate, line 27:
```suggestion
The `call` method submits a real contract call transaction to the node. It resolves as soon as the transaction is submitted and returns a `transactionId` along with a `waitForResult` callback to wait for the transaction execution.

As is the natural behaviour of blockchains, transactions may take a few seconds before they are resolved on chain. Therefore this method is useful where you want to submit a transaction but do not want it's result to block further processing of your application.
```

Some reasoning / use-case I think would be beneficial here.

apps/docs/src/guide/contracts/methods.md
```diff
@@ -24,12 +24,20 @@ A funded wallet it's required:
...
 
 Real resources are consumed, and any operations executed by the contract function will be processed on the blockchain.
 
 <<< @/../../docs-snippets/src/guide/contracts/interacting-with-contracts.test.ts#interacting-with-contracts-4{ts:line-numbers}
 
+## `callAndWait`
+
+The `callAndWait` method also submits a real contract call transaction, but unlike `call`, it resolves only when the transaction has finished processing.
```
danielbate, line 35:
```suggestion
The `callAndWait` method also submits a real contract call transaction, but unlike `call`, it resolves only when the transaction has finished processing. This is useful for when you want to wait for the result of your transaction, before performing any further functionality in your application.
```

Torres-ssf: I've modified this part https://github.com/FuelLabs/fuels-ts/pull/2692/commits/536e87ba3f46fc7ba84e3d0b360a5a4101819d76

.changeset/spicy-trains-provide.md
Torres-ssf: @danielbate I agree with your concerns, and they indeed make sense.

However, I see this as an opportunity to fix the behavior that was initially implemented using a non-ideal approach.

I believe the `BaseInvocationScope` and the `ContractFactory` should have been implemented with a similar logic to the TransactionResponse class, which from the beginning, was never a blocking call. Users need to call `waitForResult` when they want to wait until the transaction is processed.

Now that the main-net has not been released yet, this is the perfect time to make this breaking change, as it makes complete sense in the blockchain environment. This way, all users will be aware of it and can update their code accordingly.

apps/docs/src/guide/contracts/methods.md
Torres-ssf: I've modified this part https://github.com/FuelLabs/fuels-ts/commit/536e87ba3f46fc7ba84e3d0b360a5a4101819d76

petertonysmith94 (approved):
The conformity for `waitForResult` shared between the [`deployContract`](https://github.com/FuelLabs/fuels-ts/pull/2597) and `call` methods makes me happy.

.changeset/spicy-trains-provide.md
danielbate: I guess it's now or`1.0.0`. I like the parity with #2597 so I will go with you on this one. Tricky to understand the impact this will have but I assume pretty significant. Let's make sure it's clearly communicated in the release notes. They may need some manual intervention.

danielbate (dismissed):

arboleya (approved):

maschad (dismissed):
Awesome job on this 🚀

maschad (approved):

petertonysmith94 (approved):

arboleya (approved):

nedsalk (approved):

PR #2718, opened 2024-07-07
fix!: `launchNode.cleanup` not killing node in last test of test group

### LINKED ISSUE #2712: Node started via `launchTestNode` isn't closed in tests
# Repro

Create a test with the snippet below and execute it.

```console
pnpm test:filter launching.test.ts
```

The test finishes and exits, but `fuel-core` continues running in the background.

```ts
// launching.test.ts
import { sleep } from 'fuels';
import { launchTestNode } from 'fuels/test-utils';

test('it works', async () => {
  console.log('before');
  using node = await launchTestNode();
  await sleep(2500);
  console.log('after', node.provider.url);
});
```

# Repro 2

I also tried it with `const` + `cleanup`; nothing changed.

```ts
import { sleep } from 'fuels';
import { launchTestNode } from 'fuels/test-utils';

test('it works', async () => {
  console.log('before');
  const node = await launchTestNode();
  await sleep(2500);
  console.log('after', node.provider.url);
  node.cleanup();
});
```

# Repro 3

However, it works if I transform this into a regular file (not a test) and execute it.

```ts
import { sleep } from 'fuels';
import { launchTestNode } from 'fuels/test-utils';

function main() {
  console.log('before');
  using node = await launchTestNode();
  await sleep(2500);
  console.log('after', node.provider.url);
};

main.catch(console.error);
```

# Conclusion

The utility fails where it will be primarily used: in tests.

nedsalk:
- Closes #2712 

# Summary

Launching a `fuel-core` node via `launchNode` and killing it via the `cleanup` function wouldn't kill the _last_  node in a test group because the test runner would exit _before_ the node was killed and the node wouldn't get killed. The other nodes would get killed because there was enough time for the killing to happen before the last test finishes. Launching the nodes in a detached state and killing them via `process.kill(-child.pid)` did the trick.

Besides this, multiple cleanup cases have been ensured to work as a result of the discussion in https://github.com/FuelLabs/fuels-ts/pull/2718#discussion_r1667759801.

# Breaking Changes

The `killNode` and `KillNodeParams` functionality has been internalized and the method and interface have been deleted so they're no longer exported.  It's marked as a breaking change for pedantic reasons and there shouldn't really be any affected users given that they kill nodes via `cleanup` which is unchanged, so no migration notes are necessary.

# Checklist


packages/account/src/test-utils/launchNode-singular-test.test.ts
```diff
@@ -0,0 +1,20 @@
...
+ * The test runner creates a test environment per file,
+ * which we can use to isolate the faulty behavior.
+ */
+/**
+ * @group node
+ */
+describe('launchNode-singular-test', () => {
+  let killedNodeUrl = '';
+  afterAll(async () => {
+    await expect(fetch(killedNodeUrl)).rejects.toThrow('fetch failed');
```
maschad, line 13: You'll need to update this since the url is empty.
```suggestion
  let killedNodeUrl = '/deadnode';
  afterAll(async () => {
    await expect(fetch(killedNodeUrl)).rejects.toThrow('Failed to parse URL from deadnode');
```

packages/account/src/test-utils/launchNode.ts
```diff
@@ -3,13 +3,11 @@ import { randomBytes } from '@fuel-ts/crypto';
 import { FuelError } from '@fuel-ts/errors';
 import type { SnapshotConfigs } from '@fuel-ts/utils';
 import { defaultConsensusKey, hexlify, defaultSnapshotConfigs } from '@fuel-ts/utils';
-import type { ChildProcessWithoutNullStreams } from 'child_process';
 import { randomUUID } from 'crypto';
 import { existsSync, mkdirSync, rmSync, writeFileSync } from 'fs';
 import os from 'os';
 import path from 'path';
 import { getPortPromise } from 'portfinder';
-import treeKill from 'tree-kill';
```
maschad, line 12: I note the `tree-kill` dep is still in the [fuels `package.json`](https://github.com/FuelLabs/fuels-ts/blob/638eae3c3bea11ad2f016c1757e2e014e89d5c2a/packages/fuels/package.json#L87) but is it being used there? If not we should remove it there as well as in the [knip.json](https://github.com/FuelLabs/fuels-ts/blob/638eae3c3bea11ad2f016c1757e2e014e89d5c2a/.knip.json#L31)

packages/account/src/test-utils/launchNode.ts
```diff
@@ -242,13 +213,26 @@ export const launchNode = async ({
...
+      // Remove all the listeners we've added.
+      child.stderr.removeAllListeners();
+
+      // Remove the temporary folder and all its contents.
+      if (existsSync(tempDir)) {
+        rmSync(tempDir, { recursive: true });
+      }
+
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      process.kill(-child.pid!);
```
maschad, line 235: We probably should log an error if we are unable to kill the process

```suggestion
if (child.pid !== undefined) {
  process.kill(-child.pid);
} else {
  console.error('No PID available for the child process, unable to kill launchTestNode');
}
```

maschad (changes requested):
Nice job on this, ran it locally and `pgrep fuel-core` comes back clean 😄  (As opposed to master which will report a pid) 

Left some comments

packages/account/src/test-utils/launchNode-singular-test.test.ts
nedsalk: The `killedNodeUrl` value is updated in the test below and `fetch failed` is the error message thrown when a connection cannot be established.

packages/account/src/test-utils/launchNode.ts
nedsalk: From the [docs](https://nodejs.org/api/child_process.html#subprocesspid) of `child.pid`:
> Returns the process identifier (PID) of the child process. If the child process fails to spawn due to errors, then the value is undefined and error is emitted.

This will never be `undefined` because `cleanup` will only be returned and be callable if the `child` has started successfully and the node started properly ([logic](https://github.com/FuelLabs/fuels-ts/blob/aeb7ff8bc9d6fc04cbadd063cdb7ab84b8682900/packages/account/src/test-utils/launchNode.ts#L251)). That's why I think there's no need for this type of proposed handling.

packages/account/src/test-utils/launchNode.ts
nedsalk: Done in the last commits.

packages/account/src/test-utils/launchNode-singular-test.test.ts
nedsalk: Okay, I just saw the failing tests :smile: I'm investigating.

Torres-ssf (dismissed):

arboleya (commented):
> The `killNode` function and the `KillNodeParams` interface have been internalized

Why?

nedsalk:
@arboleya Upon removing `tree-kill` I realized that there is no need for the the `killNode` abstraction given that the killing now happens via `process.kill` exclusively so I internalized it [here](https://github.com/FuelLabs/fuels-ts/pull/2718/files#diff-b8f07cac51692505e88e065262fd373293a373ed5f6196eb13ec811e37aee136R220) and simplified the [related test](https://github.com/FuelLabs/fuels-ts/pull/2718/files#diff-4a69805e7f6998db762f2618f4d54da8e4a4c6d69cbf8ba323b2e9b816096b64R131).

packages/account/src/test-utils/launchNode.ts
maschad: I agree with your suggestion at face value but I also feel that there's nothing to lose in having that over disabling the eslint rule unnecessarily, it looks less clean having the disabled rule. Worst case scenario is there is an actually an unforeseen error that is logged.

maschad:
> @arboleya Upon removing `tree-kill` I realized that there is no need for the the `killNode` abstraction given that the killing now happens via `process.kill` exclusively so I internalized it [here](https://github.com/FuelLabs/fuels-ts/pull/2718/files#diff-b8f07cac51692505e88e065262fd373293a373ed5f6196eb13ec811e37aee136R220) and simplified the [related test](https://github.com/FuelLabs/fuels-ts/pull/2718/files#diff-4a69805e7f6998db762f2618f4d54da8e4a4c6d69cbf8ba323b2e9b816096b64R131).

I don't think this should be a breaking change given that `killNodeParams` was being exported solely for the purposes of our own internal consumption, corroborate that with the fact that it's not documented.

nedsalk:
@maschad I'm fine both marking it or not marking it as a breaking change.

arboleya:
> [..] Upon removing `tree-kill` I realized that there is no need for the the `killNode` abstraction

@nedsalk Ok, so you internalized the **_functionality_** and deleted the **_interface/method_**. 👍

packages/account/src/test-utils/launchNode.ts
arboleya: @nedsalk Should we consider the scenario where the `child_process` starts successfully but terminates itself due to some internal error before the `cleanup` is called?

In this case, a try/catch could prevent a dirty exit.

nedsalk:
@arboleya that's better phrasing, I updated the description.

packages/account/src/test-utils/launchNode.ts
nedsalk: @arboleya I have added `'error'` and `'exit'` event handlers in https://github.com/FuelLabs/fuels-ts/pull/2718/commits/1babf7068d5c19f0b23eda05a664c72f5f49ddc9, and with it all termination modes are covered.

maschad: It seems as though those changes are now causing the node to not be killed unfortunately 😓

nedsalk: @maschad it seems to be something related to #2717. When I run the tests with `fuel-core 0.30.0` the nodes are killed, whereas with `fuel-core 0.31.0` they remain alive.

nedsalk: @maschad  fixed in https://github.com/FuelLabs/fuels-ts/pull/2718/commits/823c33a83fd0474ead05e18889fc6578f8f7ec50.

arboleya: @nedsalk Have you tried what I said to see if the events will catch a termination happening from outside?

nedsalk: You mentioned the `child_process` terminating itself, which is caught by the `'exit'` or `'error'` events. I have added a test case for external termination in https://github.com/FuelLabs/fuels-ts/pull/2718/commits/ebcf55c6aa0440a851ddabd6c828b4e64813bacc and it works.

arboleya: The test kills the child process from inside [of NodeJS env].

I suspect you'll get an error if it's killed from the outside [before the `cleanup` is called].

nedsalk: Locally running the test below and running `kill -- -the_logged_pid` in a separate terminal passes the test, meaning the temporary directory got deleted.

```ts
  test(
    'external killing of node runs side-effect cleanup',
    async () => {
      const mkdirSyncSpy = vi.spyOn(fsMod, 'mkdirSync');

      const { pid } = await launchNode({ loggingEnabled: false });
      console.log(pid);
      expect(mkdirSyncSpy).toHaveBeenCalledTimes(1);
      const tempDirPath = mkdirSyncSpy.mock.calls[0][0];
      expect(tempDirPath).toBeTruthy();
      console.log(tempDirPath);
      // wait until cleanup finishes (done via events)
      await sleep(15000);
      expect(fsMod.existsSync(tempDirPath)).toBeFalsy();
    },
    { timeout: 100000 }
  );
```

arboleya: I didn't understand why you used a 16+ mins timeout on the test.

https://github.com/FuelLabs/fuels-ts/blob/ebcf55c6aa0440a851ddabd6c828b4e64813bacc/packages/account/src/test-utils/launchNode.test.ts#L218

---

I tried this repro on my end:

https://github.com/FuelLabs/fuels-ts/assets/26660/d9c4fd54-a992-4fa4-81db-9266d68dd801

nedsalk: > I didn't understand why you used a 16+ mins timeout on the test.

Just an arbitrary number that doesn't interfere with the `sleep` call because the default is 5s. I wanted to give myself enough time to kill the node in a different terminal before the test finishes.

> I tried this repro on my end:

Okay, I understand now what you wanted. I updated the logic and added a test to cover that in https://github.com/FuelLabs/fuels-ts/pull/2718/commits/dc647cba6f1b97b2b5893aa351c289e73c4b6898.

maschad (approved):
Nice job!

packages/account/src/test-utils/launchNode.ts
arboleya: It's not about _me_, and given the task we're putting this utility through and planning for its future within our very own test suite (parallelism?, maybe even concurrency?), no amount of care and stability seems too much. You've even handled the specific error code. Well done, thanks for the attention to detail. 👏 🎯

arboleya (approved):

packages/account/src/test-utils/launchNode.ts
arboleya: Out of curiosity, did you manage to trigger this case scenario from within the tests?

Or do the events always catch it from the perspective?

nedsalk: > Well done, thanks for the attention to detail. 👏 🎯

Thanks! And thanks for the in-depth review!

> Out of curiosity, did you manage to trigger this case scenario from within the tests?

This scenario is covered by [this test](https://github.com/FuelLabs/fuels-ts/commit/e7abaeed48483675abe3c913397c8045a3e5e54a#diff-4a69805e7f6998db762f2618f4d54da8e4a4c6d69cbf8ba323b2e9b816096b64R216).

> Or do the events always catch the termination from this perspective?

Yes, the events are called whenever the child is terminated, be it from within the `process` or outside as you did in the video above.

PR #2734, opened 2024-07-08
docs: remove repetitive words

ChengenH:
remove repetitive words


Torres-ssf (changes requested):
Hey @ChengenH thanks for your contribution.

We still need to add an empty changeset just like it is mentioned [here](https://github.com/FuelLabs/fuels-ts/blob/master/CONTRIBUTING.md#steps-to-pr)

```terminal
pnpm changeset add --empty
```

PR #2756, opened 2024-07-11
fix: `launchTestNode` multiple contracts type inference

### LINKED ISSUE #2754: `launchTestNode` should infer contracts sequentially
`launchTestNode` currently returns contracts as a discriminated union:
```ts
      contractsConfigs: [
        { deployer: AdvancedLoggingAbi__factory, bytecode: '' },
        { deployer: CallTestContractAbi__factory, bytecode: '' },
      ],
    });

    const {
      contracts: [c1, c2], // c1 and c2 are both AdvancedLoggingAbi | CallTestContractAbi
    } = launched;
```
whereas it should be inferred based on the order of the passed-in contract factories so that `c1` is `AdvancedLoggingAbi` and `c2` is `CallTestContractAbi`.

nedsalk:
- closes #2754

# Summary

Contracts inferred from an array of factories were inferred as a discriminated union of all factory contract returns instead of matching the sequence of contract factories in the `contractConfigs` array:
```ts
  contractsConfigs: [
    { deployer: AdvancedLoggingAbi__factory, bytecode: '' },
    { deployer: CallTestContractAbi__factory, bytecode: '' },
  ],
});

const {
  contracts: [c1, c2],
} = launched;

// before: c1 and c2 are both AdvancedLoggingAbi | CallTestContractAbi
// after: c1 is AdvancedLoggingAbi, c2 is CallTestContractAbi
```

# Checklist

### DIFF
--- a/packages/contract/src/test-utils/launch-test-node.ts
+++ b/packages/contract/src/test-utils/launch-test-node.ts
@@ -130,14 +130,12 @@ function getWalletForDeployment(config: DeployContractConfig, wallets: WalletUnl
   return wallets[config.walletIndex];
 }
 
-export async function launchTestNode<const TFactories extends DeployContractConfig[]>(
-  {
-    providerOptions = {},
-    walletsConfig = {},
-    nodeOptions = {},
-    contractsConfigs,
-  }: Partial<LaunchTestNodeOptions<TFactories>> = {} as const
-): Promise<LaunchTestNodeReturn<TFactories>> {
+export async function launchTestNode<const TFactories extends DeployContractConfig[]>({
+  providerOptions = {},
+  walletsConfig = {},
+  nodeOptions = {},
+  contractsConfigs,
+}: Partial<LaunchTestNodeOptions<TFactories>> = {}): Promise<LaunchTestNodeReturn<TFactories>> {
   const snapshotConfig = getChainSnapshot(nodeOptions);
   const args = getFuelCoreArgs(nodeOptions);
   const { provider, wallets, cleanup } = await setupTestProviderAndWallets({
--- a/packages/contract/src/test-utils/launch-test-node.ts
+++ b/packages/contract/src/test-utils/launch-test-node.ts
@@ -130,12 +130,14 @@ function getWalletForDeployment(config: DeployContractConfig, wallets: WalletUnl
   return wallets[config.walletIndex];
 }
 
-export async function launchTestNode<TFactories extends DeployContractConfig[]>({
-  providerOptions = {},
-  walletsConfig = {},
-  nodeOptions = {},
-  contractsConfigs,
-}: Partial<LaunchTestNodeOptions<TFactories>> = {}): Promise<LaunchTestNodeReturn<TFactories>> {
+export async function launchTestNode<const TFactories extends DeployContractConfig[]>(
+  {
+    providerOptions = {},
+    walletsConfig = {},
+    nodeOptions = {},
+    contractsConfigs,
+  }: Partial<LaunchTestNodeOptions<TFactories>> = {} as const
+): Promise<LaunchTestNodeReturn<TFactories>> {
   const snapshotConfig = getChainSnapshot(nodeOptions);
   const args = getFuelCoreArgs(nodeOptions);
   const { provider, wallets, cleanup } = await setupTestProviderAndWallets({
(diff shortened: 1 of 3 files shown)


packages/fuel-gauge/src/launch-test-node-types.test.ts
```diff
@@ -0,0 +1,25 @@
+import { launchTestNode } from 'fuels/test-utils';
+
+import type { AdvancedLoggingAbi, CallTestContractAbi } from '../test/typegen';
+import { AdvancedLoggingAbi__factory, CallTestContractAbi__factory } from '../test/typegen';
+import AdvancedLoggingBytecode from '../test/typegen/contracts/AdvancedLoggingAbi.hex';
+import CallTestContractBytecode from '../test/typegen/contracts/CallTestContractAbi.hex';
+
```
maschad, line 7:
```suggestion
    /**
	 * @group node
	 * @group browser
    */
```

maschad (commented):
Nice job on this, just one missing change

packages/fuel-gauge/src/launch-test-node-types.test.ts
```diff
@@ -0,0 +1,29 @@
...
+        { deployer: AdvancedLoggingAbi__factory, bytecode: AdvancedLoggingBytecode },
+        { deployer: CallTestContractAbi__factory, bytecode: CallTestContractBytecode },
+        { deployer: AdvancedLoggingAbi__factory, bytecode: AdvancedLoggingBytecode },
+      ],
+    });
+    const {
+      contracts: [c1, c2, c3],
+    } = launched;
+
+    expectTypeOf(c1).toMatchTypeOf<AdvancedLoggingAbi>();
```
nedsalk, line 25: Apparently these tests pass regardless of the expected type matching the real type or not. I'm currently trying to get them to fail. If @FuelLabs/sdk-ts think that we should merge it for the immediate benefit of the PR and that I investigate it in another PR, I'm down for that as well as holding the PR until the tests actually fail when they should.

Torres-ssf: @nedsalk This is ok for now. I've tested this locally and the types are being properly inferred.

Screenshot 2024-07-11 at 12 26 33
Screenshot 2024-07-11 at 12 26 50

Torres-ssf (approved):
@nedsalk Are you using the `AdvancedLogging` contract 2 times intentionally on the test case?

packages/fuel-gauge/src/launch-test-node-types.test.ts
maschad: I don't believe there is a way to actually unit test this because Type inferencing is done statically and types don't exist at runtime. I suggest we remove this test because the proof this works is done by the language server

petertonysmith94: @nedsalk I'm getting TS errors if it isn't the correct type - IMO this is sufficient :)

image

petertonysmith94 (approved):

maschad (commented):
I think the test should be removed but otherwise approve of this fix.

nedsalk:
> Are you using the AdvancedLogging contract 2 times intentionally on the test case?

There's no real intention behind it, it could've been some other contract as well. It's important the first is different from the second, the third is sugar on top.

packages/fuel-gauge/src/launch-test-node-types.test.ts
nedsalk: > Upon reconsideration, I don't believe there is a way to actually unit test this because Type inferencing is done statically and types don't exist at runtime. I suggest we remove this test because the proof this works is done by the language server

`vitest` has type-level testing capabilities, I just need to figure out how to use it. Apparently it's still experimental as well. I know of a different way to have these tests by using `tsc` directly, but I'll first try to use `vitest`.

arboleya: Awesome!

packages/contract/src/test-utils/launch-test-node.ts
```diff
@@ -130,7 +130,7 @@ function getWalletForDeployment(config: DeployContractConfig, wallets: WalletUnl
   return wallets[config.walletIndex];
 }
 
-export async function launchTestNode<TFactories extends DeployContractConfig[]>({
+export async function launchTestNode<const TFactories extends DeployContractConfig[]>({
```
arboleya, line 133: All of this because of a missing `const`? 🙃

arboleya (approved):

packages/contract/src/test-utils/launch-test-node.ts
nedsalk: Yeah, a standard array of multiple factories `[F1, F2, F3, F1]` is inferred as `(F1 | F2 | F3)[]`, whereas a `const` array is inferred as `[F1, F2, F3, F1]`.

PR #2759, opened 2024-07-12
fix: verification of `fuel-core` node killing

nedsalk:
# Summary

This PR addresses the issue reported in https://github.com/FuelLabs/fuels-ts/issues/1755#issuecomment-2224906626.

I have removed the flaky test and opted for verifying the logic via mocks. I have verified that spawning a `fuel-core` process group with the `detached:true` argument and killing the process group via `process.kill(-child.pid)` works and the mocks are verifying that the code is doing that. 
I initially wanted to ensure that all `fuel-core` nodes are killed in the `tests-ci.sh` script by verifying that no `fuel-core` processes are running after `pnpm test` is run, but the team suggested this approach as being enough and that this external checking is not necessary.

# Checklist


danielbate (commented):
Whilst your in the area, can we make this a browser test also?

nedsalk:
@danielbate No, this is testing the real implementation of `launchNode` and we can't mock out the `node` dependencies. Browser tests will benefit from it, though, because their http server is using `launchNode` under the hood as well.

petertonysmith94:
> This is probably because I initially didn't give the node time to gracefully shut down.

Is this idea of the test that it doesn't gracefully shutdown or have I got the wrong end of the stick?

nedsalk:
@petertonysmith94 The idea is explained in the description of #2718. Given that a graceful shutdown can take time, I added the `sleep(1000)` here. If it still doesn't shut down, then it'd probably be a bug in `cleanup`.

Torres-ssf:
@nedsalk Can we update the PR's description to something like:

_This PR addresses the issue reported in [fuels-ts issue #1755](https://github.com/FuelLabs/fuels-ts/issues/1755#issuecomment-2224906626). The failure was reproduced by adding { repeats: 100 } to the test. The flakiness was resolved by introducing `await sleep(1000)`, which allows the node to gracefully shut down._

Torres-ssf:
@nedsalk Sorry for the lack of context here, but why does only this test suite require the `await sleep(1000)`? 

Why is this not needed in the other test suites?

arboleya:
> [..] but why does only this test suite require the `await sleep(1000)`?

Yes, I second this.

I am always hesitant about fixing flakiness with timeouts because the latter is usually the cause of the former.

Timeouts often seem like a quick symptomatic solution to a potentially unclear underlying cause.

danielbate (commented):
@nedsalk thinking out loud here, could block scoping inside the test have a better desired effect, given that using the hook seems to be flaky and that is the expected usage of the disosable.

```ts
  test('synchronous cleanup kills node before test runner exits', async () => {
    const killedNode = {
      url: '',
    };

    {
      const { cleanup, url } = await launchNode({ loggingEnabled: false });
      killedNode.url = url;
      cleanup();
    }

    await expect(fetch(killedNode.url)).rejects.toThrow('fetch failed');
  });
```

packages/fuel-gauge/src/transaction-response.test.ts
```diff
@@ -113,10 +118,12 @@ describe('TransactionResponse', () => {
   });
 
   it('should ensure getTransactionSummary fetchs a transaction and assembles transaction summary', async () => {
-    const { ip, port } = await launchNode({
-      args: ['--poa-instant', 'false', '--poa-interval-period', '1s'],
+    using launched = await launchTestNode({
```
nedsalk, line 121: This test was missing a `cleanup` call at the end so I switched to using `launchTestNode` right away.

nedsalk:
@FuelLabs/sdk-ts I have approached this problem from a different angle and explained my reasoning in the PR's description.

nedsalk:
@danielbate The problem is that `cleanup` is synchronous whereas process killing is by it's nature asynchronous because it's done by the OS. The only 100% bulletproof way to _ensure_ that after `cleanup` is called the node is killed is to make `cleanup` asynchronous and await some OS signal. I opted out of that because it would force us to use the async disposable pattern as explained [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management), which is IMO prohibitive for the DX:
```ts
// vs
await using launched = await launchTestNode();
```

maschad (dismissed):
Nice job

danielbate (approved):
Much better than `await using ...`

petertonysmith94 (dismissed):

packages/account/src/test-utils/launchNode.test.ts
```diff
@@ -38,14 +38,36 @@ describe('launchNode', () => {
...
     cleanup();
 
+    expect(killSpy).toHaveBeenCalledTimes(1);
+    // adding a minus prefix kills the process group
+    expect(killSpy).toHaveBeenCalledWith(-pid);
+
     await waitUntilUnreachable(url);
   });
 
+  /**
```
arboleya, line 55: I wonder if both tests shouldn't be just one, mocking both `kill` and `spawn` methods simultaneously and ensuring both are called when as expected.

packages/account/src/test-utils/launchNode.test.ts
```diff
@@ -38,14 +38,36 @@ describe('launchNode', () => {
...
+   * Spawning the child process in a detached state
+   * Results in the OS assigning a process group to the child.
+   * Combining that with `process.kill(-pid)`,
+   * which sends a "kill process group" signal to the OS,
+   * ensures that the node will be killed.
+   */
+  it('spawns the fuel-core node in a detached state', async () => {
```
arboleya, line 62: I wonder if both tests shouldn't be just one, mocking both kill and spawn methods simultaneously and ensuring both are called when as expected.

arboleya (dismissed):

packages/account/src/test-utils/launchNode.test.ts
nedsalk: You are right. https://github.com/FuelLabs/fuels-ts/pull/2759/commits/aa4f52453974d0172bd39f74d76af37e91e0c126

danielbate (approved):

petertonysmith94 (approved):

maschad (approved):

Issue #2346, opened 2024-05-20
Merge `abi-coder` and `abi-typegen` packages


nedsalk:
Relevant discussion: https://github.com/FuelLabs/fuels-ts/discussions/1517

arboleya:
Closing in favor of:
- https://github.com/FuelLabs/fuels-ts/issues/3031
- https://github.com/FuelLabs/fuels-ts/issues/3032

PR #2777, opened 2024-07-16
feat!: improve `()` and `Option<T>` type handling

### LINKED ISSUE #2148: Improve `empty` type for ABI Coder and Typegen
After merge of #2122, typegen removes empty types from function input arguments, which corresponds to the runtime behavior of the abi-coder, which ignores empty inputs. This led to unexpected function signatures. Based on the abi below,
```rs
abi MyContract {
   fn types_value_then_empty(x: u8, y: ()) -> ();
   fn types_value_then_empty_then_value(x: u8, y: (), z: u8) -> ();
}
```
these are the generated functions (in usage):
```ts
contract.functions.empty_then_value(35).call();
contract.functions.value_then_empty_then_value(35, 35).call();
```

This is odd and unexpected and should be like this:
```ts
contract.functions.empty_then_value(null, 35).call();
contract.functions.value_then_empty_then_value(35, null, 35).call();
```

To implement this, we need to change both the abi-coder to accept `null` (or `undefined`, which I prefer) for empty types as well as the typegen to match this new runtime behavior.

### LINKED ISSUE #2532: Implement `VoidCoder`
In sway, 2 types produce a void `()` value. These are `Option::None()` and native enums. Both encode to just the case bytes, meaning the value itself is empty.

Internally this is picked up by the `TupleCoder` due to the bracket regex match, and as it has no children it correctly encodes to empty bytes. However to anyone trying to build this out themselves without using the `AbiCoder` would have to do this:

```ts
const nativeCoder = new EnumCoder('Native', {
    One: new TupleCoder([]),
    Two: new TupleCoder([]),
    Three: new TupleCoder([]),
});
```

This seems pretty hacky and not immediately obvious to anyone attempting to debug option and native enum problems. We should provide a `VoidCoder` that does an exact match on the void type for this scenario to provide a better experience for working with void values.

This is a feature but more an implementation detail of our internal encoding, as the `VoidCoder` itself should not be exported from `@fuel-ts/abi-coder`.

### LINKED ISSUE #2788: Make `Option<x>` optional with typegen
For a simple sway program:
```sway
contract;

abi MyProgram {
    fn optional_input(input: Option<u8>) -> u8;
}

impl MyProgram for Contract {
    fn optional_input(input: Option<u8>) -> u8 {
        1
    }
}
```

Typegen generates the function type
```ts
optional_input: InvokeFunction<[input: Option<BigNumberish>], number>;
```
which mandates providing an input even though it's optional in sway. The generated type should be
```ts
optional_input: InvokeFunction<[input?: Option<BigNumberish>], number>;
```

Note that this might become tricky once multiple inputs are introduced where some are optional and some are not, akin to what is mentioned in #2148.

petertonysmith94:
- Closes #2532
- Closes #2148
- Closes #2788 

# Release notes

In this release, we:

- Improved the type handling for `()` and `Option<T>` types.

# Summary

- Our types for functions are now like for like against their Sway counterparts.
- Made the `Option<T>` property optional.
- Added the `VoidCoder`, enabling encoding and decoding void types.

# Breaking Changes

- `()` and `Option<T>` Sway types are now either required or optional, dependent on where the argument appears in the function arguments.

  Given these Sway functions:
  ```sway
  fn type_then_void_then_type(x: u8, y: (), z: u8) -> ()
  fn type_then_void_then_void(x: u8, y: (), z: ()) -> ()
  
  fn type_then_option_then_type(x: u8, y: Option<u8>, z: u8) -> ()
  fn type_then_option_then_option(x: u8, y: Option<u8>, z: Option<u8>) -> ()
  ```
  
  This is what changes:
  ```ts
  // before
  contract.functions.type_then_void_then_type(42, 43)
  contract.functions.type_then_void_then_void(42) // Unchanged
  
  contract.functions.type_then_option_then_type(42, undefined, 43)
  contract.functions.type_then_option_then_option(42, undefined, undefined)
  ```

  ```ts
  // after
  contract.functions.type_then_void_then_type(42, undefined, 43)
  contract.functions.type_then_void_then_void(42) // Unchanged 

  contract.functions.type_then_option_then_type(42, undefined, 43)
  contract.functions.type_then_option_then_option(42)
  ```

# Checklist

- [X] I **_added_** — `tests` to prove my changes
- [X] I **_described_** — all breaking changes and the Migration Guide


packages/fuel-gauge/src/void.test.ts
```diff
@@ -0,0 +1,49 @@
...
+import { VoidAbi__factory } from '../test/typegen';
+import type { NativeEnumInput } from '../test/typegen/contracts/VoidAbi';
+import VoidAbiHex from '../test/typegen/contracts/VoidAbi.hex';
+import type { Option } from '../test/typegen/contracts/common';
+
+/**
+ * @group node
+ */
+describe('Void Tests', () => {
+  const contractsConfigs: DeployContractConfig[] = [
```
petertonysmith94, line 13:
```suggestion
  const contractsConfigs = [
```

Breaks the type inference

petertonysmith94:
I am putting this up for review early (skipped `scripts` and `predicates`) as I'm not 💯 on these [breaking changes](https://github.com/FuelLabs/fuels-ts/issues/2148):

- It could become a massive change for end users.
- I've seen many projects not using typegen - which would make these changes difficult to fix quickly.

Torres-ssf (commented):
@petertonysmith94 Why do we need to inform the `undefined`?

I can understand for cases like this:

```ts
contract.functions.type_then_void_then_type(42, undefined, 43)
```

But why it is also required for cases like this? 

```ts
contract.functions.type_then_void(42, undefined)
```

petertonysmith94:
@Torres-ssf I agree that the following does seem off:

```ts
contract.functions.type_then_void(42, undefined)
```

I'll put this back in a draft and implement these changes now.

packages/abi-typegen/test/fixtures/templates/contract/dts.hbs
```diff
@@ -109,8 +109,8 @@ export class MyContractAbi extends Contract {
     types_b512: InvokeFunction<[x: string], string>;
     types_bool: InvokeFunction<[x: boolean], boolean>;
     types_bytes: InvokeFunction<[x: Bytes], Bytes>;
-    types_empty: InvokeFunction<[], void>;
-    types_empty_then_value: InvokeFunction<[y: BigNumberish], void>;
```
nedsalk, line 113: I think `void` as a return is nicer than `undefined` and it's implied in the mind of a developer that a `void` return is `undefined` anyways.

petertonysmith94: b3d704821f90640105688c2888259693938bfbf0

petertonysmith94:
This work is close to being ready, however, aware that the following issue will take precedence:
- https://github.com/FuelLabs/fuels-ts/pull/2826

Final todos:
- Tests for predicates
- Tests for scripts

packages/abi-typegen/src/utils/getMandatoryInputs.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import { EmptyType } from '../abi/types/EmptyType';
+import { OptionType } from '../abi/types/OptionType';
+import type { IType } from '../types/interfaces/IType';
+import type { JsonAbiArgument } from '../types/interfaces/JsonAbi';
+
+import { findType } from './findType';
+
+export type ArgumentWithMetadata<TArg extends JsonAbiArgument = JsonAbiArgument> = TArg & {
+  isOptional: boolean;
```
nedsalk, line 11: A `getMandatoryInputs` returning an `isOptional` field is counterintuitive. I think `getFunctionInputs` is a better name.

petertonysmith94: I was planning on changing - just didn't have a good name yet.

petertonysmith94: b27050190e2a155cc104b248374b3e9fbe600fec

nedsalk:
In the breaking changes section you wrote:
> Empty type parameters are now required in every function call (as `undefined`).

I suggest rewriting this as:
`()` and `Option<T>` sway types are now either required or optional, depending on the circumstances.

packages/abi-coder/src/utils/getFunctionInputs.ts
```diff
@@ -0,0 +1,35 @@
...
+export type ArgumentWithMetadata<TArg extends JsonAbiArgument = JsonAbiArgument> = TArg & {
+  isOptional: boolean;
+};
+
+export const getFunctionInputs = (params: {
+  jsonAbi: JsonAbi;
+  inputs: readonly JsonAbiArgument[];
+}): Array<ArgumentWithMetadata> => {
+  let inMandatoryRegion = false;
+  const inputs = clone(params.inputs);
```
nedsalk, line 17: you can use `structuredClone` instead of `clone` here and remove the [dependency](https://github.com/FuelLabs/fuels-ts/pull/2777/files#diff-ae8b97dabdf446b7f229ac907def90a770ef59693a48fee298fdd926ab353d3dR36). I'm fine with this remaining `clone`, however, because I believe we are using `clone` in places where `structuredClone` can't be used because it doesn't clone functions, and so removing it or staying with it doesn't affect the `fuels` package.

pnpm-lock.yaml
danielbate, line 1: Lock file seems to have updated many packages other than `ramda`, was this intentional?

nedsalk (approved):

pnpm-lock.yaml
petertonysmith94: Should be fixed now, not 100% sure on the CPU changes [here](https://github.com/FuelLabs/fuels-ts/pull/2777/files#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL6553).

nedsalk (dismissed):

packages/abi-coder/src/utils/getFunctionInputs.ts
```diff
@@ -0,0 +1,33 @@
+import type { JsonAbi, JsonAbiArgument } from '../types/JsonAbi';
+
+import { OPTION_REGEX, VOID_TYPE } from './constants';
+import { findTypeById } from './json-abi';
+
+export type ArgumentWithMetadata<TArg extends JsonAbiArgument = JsonAbiArgument> = TArg & {
+  isOptional: boolean;
+};
+
+export const getFunctionInputs = (params: {
```
maschad, line 10: A few optimization suggestions I've noticed: 

- Deep copies can be quite expensive, especially on more complex objects, better to work with the original inputs
- reverses the array twice (.reverse() at the start and end).
-  creates multiple intermediate arrays (`reverse()`, `map` and `reverse()` again)

```suggestion
export const getFunctionInputs = ({
  jsonAbi,
  inputs,
}: {
  jsonAbi: JsonAbi;
  inputs: readonly JsonAbiArgument[];
}): ArgumentWithMetadata[] => {
  let isMandatory = false;

  return inputs.reduceRight((result, input) => {
    const type = findTypeById(jsonAbi, input.type);
    isMandatory = isMandatory || (type.type !== VOID_TYPE && !OPTION_REGEX.test(type.type));

    return [{ ...input, isOptional: !isMandatory }, ...result];
  }, [] as ArgumentWithMetadata[]);
 ```

File: packages/abi-typegen/src/utils/getFunctionInputs.test.ts [hunk 6]
```diff
@@ -0,0 +1,134 @@
...
+  ['option (release)', optionAbiType],
+  ['option (debug)', debugOptionAbiType],
+];
+
+const types: Array<IType> = [nonEmptyType, voidAbiType, optionAbiType, debugOptionAbiType].map(
+  (rawAbiType) => makeType({ rawAbiType })
+);
+
+/**
+ * @group node
```
maschad, line 46: Should be browser compatible 

```suggestion
 * @group node
 * @group browser
```

packages/abi-typegen/src/utils/getFunctionInputs.ts
```diff
@@ -0,0 +1,38 @@
...
+import type { IType } from '../types/interfaces/IType';
+import type { JsonAbiArgument } from '../types/interfaces/JsonAbi';
+
+import { findType } from './findType';
+
+export type ArgumentWithMetadata<TArg extends JsonAbiArgument = JsonAbiArgument> = TArg & {
+  isOptional: boolean;
+};
+
+export const getFunctionInputs = (params: {
```
maschad, line 12: Same applies here , but why do we need to [redefine](https://github.com/FuelLabs/fuels-ts/pull/2777/files#diff-07695b2096aef52c7df838b7bdf2485bc032260a1d529ee5cd3fbff48e00c675R10) this exact same function?

packages/fuel-gauge/src/void.test.ts
```diff
@@ -0,0 +1,83 @@
+import { launchTestNode } from 'fuels/test-utils';
+
+import { VoidAbi__factory } from '../test/typegen';
+import type { NativeEnumInput } from '../test/typegen/contracts/VoidAbi';
+import VoidAbiHex from '../test/typegen/contracts/VoidAbi.hex';
+import type { Option } from '../test/typegen/contracts/common';
+
+/**
+ * @group node
```
maschad, line 9: Should be browser compatible
```suggestion
 * @group node
 * @group browser
```

maschad (changes requested):
Nice work on this 🚀    I've left some comments in line.

apps/docs-snippets/src/guide/types/options.test.ts
```diff
@@ -1,19 +1,24 @@
...
+import bytecode from '../../../test/typegen/contracts/SumOptionU8Abi.hex';
+
+function setupContract() {
+  return launchTestNode({
+    contractsConfigs: [{ deployer: SumOptionU8Abi__factory, bytecode }],
+  });
+}
 
 /**
  * @group node
```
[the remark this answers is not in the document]
petertonysmith94: 2f5b4bb7399a268b9d736ed73bddda15403d0fc2

packages/abi-coder/src/encoding/coders/VoidCoder.test.ts
```diff
@@ -0,0 +1,37 @@
+import { VoidCoder } from './VoidCoder';
+
+/**
+ * @group node
```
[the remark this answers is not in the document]
petertonysmith94: 2f5b4bb7399a268b9d736ed73bddda15403d0fc2

packages/abi-coder/src/utils/getFunctionInputs.test.ts
```diff
@@ -0,0 +1,137 @@
...
+  encoding: '1',
+  types: [nonEmptyType, voidAbiType, optionAbiType, debugOptionAbiType],
+  functions: [],
+  loggedTypes: [],
+  messagesTypes: [],
+  configurables: [],
+};
+
+/**
+ * @group node
```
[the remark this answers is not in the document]
petertonysmith94: 2f5b4bb7399a268b9d736ed73bddda15403d0fc2

packages/abi-typegen/src/utils/getFunctionInputs.test.ts
petertonysmith94: 2f5b4bb7399a268b9d736ed73bddda15403d0fc2

packages/fuel-gauge/src/void.test.ts
petertonysmith94: 2f5b4bb7399a268b9d736ed73bddda15403d0fc2

packages/abi-coder/src/utils/getFunctionInputs.ts
petertonysmith94: Thank you for the optimizations 🙏🏼 

71a1309b87050e391251fef84fe2cd57ded1afe6

packages/abi-typegen/src/utils/getFunctionInputs.ts
petertonysmith94: 71a1309b87050e391251fef84fe2cd57ded1afe6

I redefined a similar function (with different inputs) due to the nature of two packages that are very closely linked `abi-coder` and `abi-typegen`. Once we merge these two packages (https://github.com/FuelLabs/fuels-ts/issues/2346), these two methods would sit together.

packages/abi-coder/src/FunctionFragment.ts
```diff
@@ -66,17 +56,23 @@ export class FunctionFragment<
...
-    const nonEmptyInputs = findNonEmptyInputs(this.jsonAbi, this.jsonFn.inputs);
+    if (values.length < mandatoryInputLength) {
+      throw new FuelError(
+        ErrorCode.ABI_TYPES_AND_VALUES_MISMATCH,
+        `Invalid number of arguments. Expected a minimum of ${mandatoryInputLength} arguments, received ${values.length}`
+      );
+    }
 
-    if (Array.isArray(values) && nonEmptyInputs.length !== values.length) {
+    const shallowCopyValues = values.slice();
```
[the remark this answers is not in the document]
petertonysmith94: 1c32d0f385cd71145bac37ac645afd7f794ab1ea

I continued to use a `slice` as it appears more performant than a spread ([source](https://measurethat.net/Benchmarks/Show/2667/0/arrayprototypeslice-vs-spread-operator)) :)

petertonysmith94:
> > ```ts
> > // before
> > contract.functions.type_then_void_then_type(42, 43)
> > contract.functions.type_then_option(42, undefined)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```ts
> > // after
> > contract.functions.type_then_void_then_type(42, undefined, 43)
> > contract.functions.type_then_option(42)
> > ```
> 
> Context question, how come we are treating these two things differently? I'd expect to pass `()` and `Option::None` in the same way?

They are treated the same - maybe the example that I provided is confusing - I've updated the example, please let me know if that clears it up.

Torres-ssf (approved):

danielbate:
@petertonysmith94 thankyou for clearing it up. Makes sense, we are treating them the same, it is actually dependent on their position in the function arguments that had thrown me off.

petertonysmith94:
> @petertonysmith94 thankyou for clearing it up. Makes sense, we are treating them the same, it is actually dependent on their position in the function arguments that had thrown me off.

Correct - do you think the breaking change description need work?

maschad (approved):
Awesome!

packages/abi-coder/src/FunctionFragment.ts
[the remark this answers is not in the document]
maschad: Good catch!

maschad (approved):

nedsalk (approved):

danielbate:
> () and Option<T> Sway types are now either required or optional, depending on the circumstances.

`()` and `Option<T>` Sway types are now either required or optional, dependent on where the argument appears in the function arguments.

@petertonysmith94 change for breaking changes maybe?

danielbate (approved):
Clean 🧹

Issue #2763, opened 2024-07-12
Empty operation being returned in txSummary


LuizAsFight:
### What version of fuels-ts are you using?

0.92.0

### Steps to Reproduce

do a contract call with an account that only have message to fund the transaction (to do that you can receive from the bridge as funds bridged will be message in fuel side)

### Expected Behavior

Should show operations just fine

### Actual Behavior

It's doesn't return any operations

### Code

I believe this part is the problem: https://github.com/FuelLabs/fuels-ts/blob/master/packages/account/src/providers/transaction-summary/input.ts#L42

this `TODO`  looks like it's not making sense because a message won't have assetId at all, as [mentioned in the spec](https://docs.fuel.network/docs/specs/tx-format/input/#inputmessage) on amount field: "Amount of base asset coins sent with message.".

when it has amount, it's a base asset message
when it hasn't it can be anything, depending on `data` field

I think on this case it should find by amount instead:
```
const messageInput = messageInputs.find(
    ({ amount }) => !!amount && amount.gt(0)
  );
```

PR #2782, opened 2024-07-17
fix: return correct operations from coin and message inputs

maschad:
- Closes https://github.com/FuelLabs/fuels-ts/issues/2763

- Related to  https://github.com/FuelLabs/fuels-wallet/pull/1419

# Release notes

In this release, we:

- Fixed `getInputFromAssetId` so it returns the correct `InputCoin` and `InputMessage`

# Summary

At the moment [`getContractCallOperations`](https://github.com/FuelLabs/fuels-ts/blob/master/packages/account/src/providers/transaction-summary/operations.ts#L235) is currently returning empty operations on wallets that only have an `InputMessage` but no `InputCoin` , this is because `getInputFromAssetId` function was only returning `messageInputs` with the _32ZeroBytes_ hex, which of course excludes any input from a contract call that has a separate `assetId` such as a predicate call.

# Notes

The best way to verify that this fix works is to use this branches build with the [fuel-wallet](https://github.com/FuelLabs/fuels-wallet) and test it against the connectors. https://github.com/FuelLabs/fuels-wallet/issues/1415 goes into greater detail on how to reproduce the scenarios.

# Checklist

- [x] I've tested this on https://connectors.fuel.network/ with the [fuel-wallet](https://github.com/FuelLabs/fuels-wallet/pull/1419)


packages/account/src/providers/transaction-summary/input.ts
```diff
@@ -37,10 +37,7 @@ export function getInputFromAssetId(inputs: Input[], assetId: string) {
   const coinInputs = getInputsCoin(inputs);
   const messageInputs = getInputsMessage(inputs);
   const coinInput = coinInputs.find((i) => i.assetId === assetId);
-  // TODO: should include assetId in InputMessage as well. for now we're mocking ETH
-  const messageInput = messageInputs.find(
-    (_) => assetId === '0x0000000000000000000000000000000000000000000000000000000000000000'
-  );
+  const messageInput = messageInputs.find(({ amount }) => !!amount && amount.gt(0));
```
danielbate, line 40: Do we not need to validate the base asset ID here?

petertonysmith94: [`InputMessage`](https://docs.fuel.network/docs/specs/tx-format/input/#inputmessage) doesn't have an asset id to validate against.

packages/account/src/providers/transaction-summary/input.test.ts
```diff
@@ -120,4 +121,18 @@ describe('transaction-summary/input', () => {
       MOCK_INPUT_MESSAGE
     );
   });
+
+  it('should ensure getInputFromAssetId returns input message if the coin input amount is 0', () => {
```
petertonysmith94, line 125: I don't believe this description is describing the behaviour that is actually happening, as the reason the message input is selected is because the `assetId` of `MOCK_INPUT_COIN` does not equal the `ASSET_A`.

petertonysmith94 (commented):
> Do a contract call with an account that only have message to fund the transaction (to do that you can receive from the bridge as funds bridged will be message in fuel side)

Would it be worth adding a fuel-gauge test around this use-case?

.changeset/thin-pumpkins-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/account": patch
+---
+
+fix: ensure `getInputFromAssetId` returns correct `InputMessage`
```
arboleya, line 5: I've been suggesting [shorter release notes](https://github.com/FuelLabs/fuels-ts/pull/2796#discussion_r1683652796) in some PRs.

I think you could almost swap the two, see what you think.

----

In this release, we:
 - Fixed the `getInputFromAssetId` method used in transaction summary operations

# Summary

Fixed an issue where `getInputFromAssetId` was returning incorrect `InputMessage`s based on base asset ids, resulting in empty operations being returned in the txn summaries.

packages/account/src/providers/transaction-summary/input.ts
danielbate: I think this is because it will always be for the base asset ID, as the message is used for withdrawals in the base asset i.e. bridging. That check did work in the past when the base asset was hard coded.

maschad: It turns out the the [ReceiptCall's](https://docs.fuel.network/docs/specs/abi/receipts/#call-receipt)  `baseAssetId`  is always `0x0000000000000000000000000000000000000000000000000000000000000000` but this doesn't provide the correct `InputMessage` for predicate / contract calls **AND** also for withdrawals, therefore I opted to use the `amount` field as a delineating factor.

.changeset/thin-pumpkins-sell.md
maschad: I've updated the description now based on [a9cb052](https://github.com/FuelLabs/fuels-ts/pull/2782/commits/a9cb0521cc414be8e7104ed5163fea48a3796e08) let me know your thoughts

packages/account/src/providers/transaction-summary/input.ts
```diff
@@ -33,16 +34,23 @@ export function getInputsContract(inputs: Input[]) {
...
+export function getRelevantInputs(inputs: Input[], assetId: string = ZeroBytes32) {
   const coinInputs = getInputsCoin(inputs);
   const messageInputs = getInputsMessage(inputs);
-  const coinInput = coinInputs.find((i) => i.assetId === assetId);
-  // TODO: should include assetId in InputMessage as well. for now we're mocking ETH
-  const messageInput = messageInputs.find(
-    (_) => assetId === '0x0000000000000000000000000000000000000000000000000000000000000000'
-  );
+  let coinInput = coinInputs.find((i) => i.assetId === assetId);
+  // #TODO: There are times when the ReceiptCall's baseAssetId doesn't match the CoinInput's assetId
```
maschad, line 41: In my experience testing this with the [connectors](https://connectors.fuel.network/), I've found that the [ReceiptCall](https://docs.fuel.network/docs/specs/abi/receipts/#call-receipt) 's `baseAssetId` was the zero 32-byte default. This doesn't seem to be accurate though given the predicate one may be interacting with would have a different `baseAssetId`. This method needs to be flexible enough to handle multiple assets within multiple `CoinInput` messages.

packages/account/src/providers/transaction-summary/input.test.ts
maschad: [`63e827d` (#2782)](https://github.com/FuelLabs/fuels-ts/pull/2782/commits/63e827df7cf85317ae9a9017edbe7d22fe54c845)

.changeset/thin-pumpkins-sell.md
Torres-ssf: @maschad How about?

_made `getInputFromAssetId` return correct `baseAssetId` for a given `InputMessage`_

maschad: Well @Torres-ssf it also applies to `InputCoins` as well, unfortunately the 72 character limitation isn't as expressive for cases like this 😢

Torres-ssf: @maschad So `getInputFromAssetId` was returning the wrong `baseAssetId` not only for `InputMessage` but for `InputCoin` as well?

maschad: Not exactly, but it was returning the `coinMessage` or `coinInput` as _undefined_ based on the incorrect `baseAssetId` being sent in some situations, I described in more detail [here](https://github.com/FuelLabs/fuels-ts/pull/2782#discussion_r1690388439)

arboleya: My main concern/suggestion was to make the `release notes` shorter:

Some that would fit like this:
option-1


Instead of this:
option-2

packages/account/src/providers/transaction-summary/operations.ts
```diff
@@ -252,7 +257,12 @@ export function getContractCallOperations({
     if (contractInput) {
       const newCallOps = contractCallReceipts.reduce((prevContractCallOps, receipt) => {
         if (receipt.to === contractInput.contractID) {
-          const input = getInputFromAssetId(inputs, receipt.assetId);
+          let input;
+          if (receipt.assetId !== ZeroBytes32) {
+            input = getRelevantInputs(inputs, receipt.assetId);
+          } else {
+            input = getRelevantInputs(inputs);
+          }
```
LuizAsFight, line 265: what is the purpose of this conditional, if the method `getRelevantInputs` will anyway use ZeroBytes32 as default value?

packages/account/src/providers/transaction-summary/input.ts
```diff
@@ -33,16 +34,23 @@ export function getInputsContract(inputs: Input[]) {
 }
 
 /** @hidden */
-export function getInputFromAssetId(inputs: Input[], assetId: string) {
+export function getRelevantInputs(inputs: Input[], assetId: string = ZeroBytes32) {
```
LuizAsFight, line 37: the previous name was better imo:
1- it was more explicit, this method exists to return the input funding a specific asset to the transaction
2- getRelevantInputs is too generic. "relevant" to what?

packages/account/src/providers/transaction-summary/input.ts
```diff
@@ -33,16 +34,23 @@ export function getInputsContract(inputs: Input[]) {
...
-  // TODO: should include assetId in InputMessage as well. for now we're mocking ETH
-  const messageInput = messageInputs.find(
-    (_) => assetId === '0x0000000000000000000000000000000000000000000000000000000000000000'
-  );
+  let coinInput = coinInputs.find((i) => i.assetId === assetId);
+  // #TODO: There are times when the ReceiptCall's baseAssetId doesn't match the CoinInput's assetId
+  // In this case, we should return the last input in the CoinInput array
+  if (!coinInput && coinInputs.length > 0) {
+    coinInput = coinInputs[coinInputs.length - 1];
+  }
```
LuizAsFight, line 45: I don't think this exception should be handled here. this function should purely return the correct input for the param asset id.

If the contract call receipt is the problem, the exception should be handled there before executing this function `getInputFromAssetId`

LuizAsFight (changes requested):

packages/account/src/providers/transaction-summary/operations.ts
Torres-ssf, line 265: 2. The `assetId` property for the call receipt will always be `0x0000...` unless amounts are being forwarded to the contract call.

This part of the code tries to identify the account address responsible for the contract call by checking the first input with the same `assetId` as the amounts forwarded to the contract call.

However, if no amount is being forwarded and the `assetId` is a zeroed hex, we need to look for the first input with the `assetId` of the base asset.

So we can change this to:
```ts
if (receipt.to === contractInput.contractID) {
  const assetId = receipt.assetId === ZeroBytes32 ? baseAssetId : receipt.assetId;
  const input = getInputFromAssetId(inputs, assetId, baseAssetId);
```

packages/account/src/providers/transaction-summary/input.ts
```diff
@@ -33,16 +34,23 @@ export function getInputsContract(inputs: Input[]) {
...
+
+  const messageInput = messageInputs.find(({ amount }) => !!amount && amount.gt(0));
+
+  if (coinInput) {
+    return coinInput;
+  }
 
-  return coinInput || messageInput;
+  return messageInput;
 }
```
Torres-ssf, line 54: 3. It is better to change the name of this helper later on when we decide to refactor all these helpers.

This helper was created to extract the first input from the input array with the provided `assetId`. We can only return an `InputMessage` here if the provided `assetId` is the base asset ID

So we can change it to this (feel free to change the logic but the behavior needs to be the same)
```ts
export function getInputFromAssetId(inputs: Input[], assetId: string, isBaseAsset = false) {
  const coinInputs = getInputsCoin(inputs);
  const coinInput = coinInputs.find((i) => i.assetId === assetId);

  let messageInput: InputMessage | undefined;
  if (isBaseAsset) {
    messageInput = getInputsMessage(inputs)?.[0];
  }

  return coinInput || messageInput;
}
```

Torres-ssf (changes requested):
Seems like there is misleading information in this issue description. After investigating this, I could verify that the Contract Call operation only works when the contract call transaction was funded with a message.

When funded the contract call with a UTXO results in the contract call operation not being identified.

1. We need to add the `baseAssetId` to the [parameters](https://github.com/FuelLabs/fuels-ts/blob/3818684d204c6cd2c37c9bce9955d7f67b26bc6c/packages/account/src/providers/transaction-summary/operations.ts#L249) of `getContractCallOperations`

```ts
Pick<GetOperationParams, 'abiMap' | 'maxInputs' | 'baseAssetId'> &
```

packages/account/src/providers/transaction-summary/operations.ts
maschad: > 2. The assetId property for the call receipt will always be 0x0000... unless amounts are being forwarded to the contract call.

Right, I had alluded to this [here](https://github.com/FuelLabs/fuels-ts/pull/2782#discussion_r1690388439) but thanks for the clarification. I see [this issue](https://github.com/FuelLabs/fuel-core/issues/1941) has been opened to provide a more accurate response for such a scenario. In the interim I'll use this suggestion.

packages/account/src/providers/transaction-summary/input.ts
maschad: 7004319a6

packages/account/src/providers/transaction-summary/input.ts
maschad: 7004319a6

packages/account/src/providers/transaction-summary/operations.ts
maschad: 7004319a6

.changeset/thin-pumpkins-sell.md
maschad: 7004319a6

Torres-ssf:
> I think simulating this with the current infrastructure in fuel-gauge would be cumbersome, `launchTestNode` only allows for seeding with `InputCoin`s as opposed to `InputMessage`s and not varying flexibility with `baseAssetId`s
> 
> That being said I do think this is an appropriate use case for an [e2e test](https://github.com/FuelLabs/fuels-ts/blob/master/packages/fuel-gauge/src/e2e-script.test.ts) with the testnet but we would need to find a way to _mock_ bridged assets

launchTestNode supports generating messages. 
```ts
    const testMessage = new TestMessage({ amount: 100 });

    using launched = await launchTestNode({ walletsConfig: { messages: [testMessage] } });
```

I am unsure if it supports seeding [MessageCoin](https://github.com/FuelLabs/fuels-ts/blob/560c81908375f08ed19ec9649f89e477b7034a1d/packages/account/src/providers/message.ts#L21) (spendable resource) or [Message](https://github.com/FuelLabs/fuels-ts/blob/560c81908375f08ed19ec9649f89e477b7034a1d/packages/account/src/providers/message.ts#L10). If it does not support seeding `MessageCoin`, we should fix that.
> cc @nedsalk

nedsalk:
> I am unsure if it supports seeding [MessageCoin](https://github.com/FuelLabs/fuels-ts/blob/560c81908375f08ed19ec9649f89e477b7034a1d/packages/account/src/providers/message.ts#L21) (spendable resource) or [Message](https://github.com/FuelLabs/fuels-ts/blob/560c81908375f08ed19ec9649f89e477b7034a1d/packages/account/src/providers/message.ts#L10). If it does not support seeding MessageCoin, we should fix that.

@Torres-ssf It puts messages in the stateConfig.messages array and associates them with the wallets. I tried running `getResourcesToSpend` on a wallet with test messages while simultaneously excluding all the coins and it fails with `not enough coins to fit the target`. It's unclear to me how the message would be converted to a `MessageCoin`.

Torres-ssf:
> @Torres-ssf It puts messages in the stateConfig.messages array and associates them with the wallets. I tried running `getResourcesToSpend` on a wallet with test messages while simultaneously excluding all the coins and it fails with `not enough coins to fit the target`. It's unclear to me how the message would be converted to a `MessageCoin`.

@nedsalk I've just validate that it is possible to generate spendable messages. The difference between the 2 is that a `MessageCoin` has no data:

```ts
const spendableMessage = new TestMessage({ amount: 100_000, data: '' });
``` 
> cc @maschad

maschad:
> ```ts
> const testMessage = new TestMessage({ amount: 100 });
> 
>     using launched = await launchTestNode({ walletsConfig: { messages: [testMessage] } });
> ```

Thanks for the confirmation @Torres-ssf I was unaware that's how a spendable message could be seeded. I've added a corresponding test in 2a5abeae1  I had to remove the zero `amountPerCoin` constraint as well so that we could simulate the scenario in the issue.

packages/fuel-gauge/src/transaction-summary.test.ts
```diff
@@ -591,5 +591,45 @@ describe('TransactionSummary', () => {
         recipients: allRecipients,
       });
     });
+
+    it.only('should ensure that transfer operations are assembled correctly if only seeded with a MessageInput (SPENDABLE MESSAGE)', async () => {
```
petertonysmith94, line 595:
```suggestion
    it('should ensure that transfer operations are assembled correctly if only seeded with a MessageInput (SPENDABLE MESSAGE)', async () => {
```

packages/account/src/providers/transaction-summary/operations.ts
```diff
@@ -239,9 +239,10 @@ export function getContractCallOperations({
   abiMap,
   rawPayload,
   maxInputs,
+  baseAssetId,
```
petertonysmith94, line 242: Would this inclusion make this a breaking change?

Torres-ssf: @petertonysmith94 No, this helper is intended for internal use only.

packages/fuel-gauge/src/transaction-summary.test.ts
```diff
@@ -591,5 +591,45 @@ describe('TransactionSummary', () => {
         recipients: allRecipients,
       });
     });
+
+    it('should ensure that transfer operations are assembled correctly if only seeded with a MessageInput (SPENDABLE MESSAGE)', async () => {
+      const testMessage = new TestMessage({ amount: 1000000, data: '' });
```
Torres-ssf, line 596: @maschad Do you think it makes sense also to validate if this operation is being returned as expected when funded with UTXOs only? When investigating this issue at my side, using master, I could see that this operation was not working ( at least not always ) when the TX was funded with UTXOs.

maschad: Maybe I am misinterpreting your suggestion @Torres-ssf but isn't that what the preceding [two tests](https://github.com/FuelLabs/fuels-ts/blob/mc/fix/ensure-correct-input-message-returned/packages/fuel-gauge/src/transaction-summary.test.ts#L448-L593) are testing for?

Torres-ssf (approved):

packages/account/src/providers/transaction-summary/input.ts
```diff
@@ -33,16 +33,32 @@ export function getInputsContract(inputs: Input[]) {
...
+
+/** @hidden */
+function findMessageInput(inputs: Input[]): InputMessage | undefined {
+  return getInputsMessage(inputs)?.[0];
+}
+/** @hidden */
+export function getInputFromAssetId(
+  inputs: Input[],
+  assetId: string,
+  isBaseAsset = false
```
LuizAsFight, line 49: make sense to receive `baseAssetId` , and do the comparison `assetId === baseAssetId` inside of this method?  

would be more smooth as normally the operations already have the `baseAssetId` avoiding the dev needing to provide `assetId === baseAssetId` everytime using this method

maschad: This method is used in [`getWithdrawFromFuelOperations`](https://github.com/FuelLabs/fuels-ts/blob/mc/fix/ensure-correct-input-message-returned/packages/account/src/providers/transaction-summary/operations.ts#L201) as well which doesn't have access to an `assetId` to compare with the `baseAssetId` so I don't think that is possible.

petertonysmith94: In this instance used in [`getWithdrawFromFuelOperations`](https://github.com/FuelLabs/fuels-ts/blob/mc/fix/ensure-correct-input-message-returned/packages/account/src/providers/transaction-summary/operations.ts#L201), could we not just pass through the `baseAssetId` as both arguments?

```ts
const input = getInputFromAssetId(inputs, baseAssetId, baseAssetId);
```

maschad: I don't see how passing it that way offers better ergonomics or readability, passing the same parameter to a function twice is generally bad practice. I think the current signature conveys what's happening more clearly.

petertonysmith94 (approved):

danielbate (approved):
👏🏻

packages/account/src/providers/transaction-summary/input.ts
Torres-ssf: @LuizAsFight, from my understanding, the helper `getInputFromAssetId` is designed to find the first resource input ([InputCoin](https://github.com/FuelLabs/fuels-ts/blob/4bc03879babf40df16a58c17e15b46093e519032/packages/transactions/src/coders/input.ts#L19) or [InputMessage](https://github.com/FuelLabs/fuels-ts/blob/4bc03879babf40df16a58c17e15b46093e519032/packages/transactions/src/coders/input.ts#L207)) based on a given asset ID.

Since all `InputMessage` are base asset ID resources, this helper can only return an `InputMessage` if the given asset ID is the base asset ID.

Therefore, without validation, there is no way to know if the given asset ID is the base asset ID, because an `InputMessage` does not have the property `assetId`, unlike `InputCoin`.

Please let me know if I am missing something here

LuizAsFight (approved):

PR #2579, opened 2024-06-21
chore: integrate `launchTestNode` into current test suites

maschad:
- Closes https://github.com/FuelLabs/fuels-ts/issues/1820
- Closes https://github.com/FuelLabs/fuels-ts/issues/2081

# Summary

This removes the `getSetupContract` ,`generateTestWallet` and `seedTestWallet` utilities in favour of the `launchTestNode` utility in the `fuel-gauge` test suite.

Builds on the work done in https://github.com/FuelLabs/fuels-ts/pull/2378/files to add browser compatibility for the `fuel-gauge` test suite.

This also refactors `launchTestNode` to accommodate  for scripts, predicates, seeding wallets with multiple wallets and different asset types, and mocking more provider methods.

# Checklist

- [x] I **_added_** — `tests` to prove my changes
- [ ] I **_updated_** — all the necessary `docs`
- [x] I **_reviewed_** — the entire PR myself, using the GitHub UI
- [ ] I **_described_** — all breaking changes and the Migration Guide

# Notes

Running 

```
pnpm test:filter packages/fuel-gauge/src/  
```

Should result in all tests passing.

Screenshot 2024-07-11 at 4 17 20 PM

Then running 

```
prgrep fuel-core
```

should return clean as there are no processes running.

PR #2789, opened 2024-07-18
chore: improve `fuel-gauge` tests

nedsalk:
# Summary

Merges into #2579.

I didn't want to pollute #2579 with lots of comments so I grouped all change requests into this PR. 

Due to other more pressing matters, I did not cover all the places where typegen predicate and script factories could have been used.

I think the rest of the `fuel-gauge` code base should be left to be analyzed some other time because I already found a lot issues to work on when I integrated these factories and more might be lying in there, so one should take time to analyze each problem and create an appropriate issue for it. #2579 shouldn't be slowed down by that analysis.

# Checklist


packages/fuel-gauge/src/predicate-conditional-inputs.test.ts
```diff
@@ -22,11 +22,8 @@ describe('PredicateConditionalInputs', () => {
...
     const amountToTransfer = 1000;
 
-    const predicate = new Predicate({
-      bytecode: PredicateConditionalInputsAbi__factory.bin,
-      abi: PredicateConditionalInputsAbi__factory.abi,
-      provider,
-      configurableConstants: { MAKER: aliceWallet.address.toB256() },
+    const predicate = PredicateConditionalInputsAbi__factory.createInstance(provider, undefined, {
+      MAKER: aliceWallet.address.toB256(),
     });
```
petertonysmith94, line 27: Not a now issue - but what do you think of this DX?

Maybe an options argument might be nicer

```ts
PredicateConditionalInputsAbi__factory.createInstance(provider, {
  configurables: {
    MAKER: aliceWallet.address.toB256(),
  }
})
```

packages/fuel-gauge/src/doc-examples.test.ts
```diff
@@ -277,67 +277,14 @@ describe('Doc Examples', () => {
...
-            type: 0,
-            typeArguments: null,
-          },
-          attributes: null,
-        },
-      ],
-      loggedTypes: [],
-      configurables: [],
-      messagesTypes: [],
-    };
```
petertonysmith94, line 329: Do you know the original rational behind having the ABI here in the first place?

(Not an issue, just intrigued)

packages/fuel-gauge/src/coverage-contract.test.ts
```diff
@@ -523,12 +512,12 @@ describe('Coverage Contract', () => {
       .echo_struct_vector_last([
         {
           foo: 1,
-          bar: 11337n,
+          bar: 11337,
```
petertonysmith94, line 515:
```suggestion
          bar: 11337n,
```

Why are we removing the BigInt expression here?

packages/fuel-gauge/src/coverage-contract.test.ts
```diff
@@ -523,12 +512,12 @@ describe('Coverage Contract', () => {
...
           baz: '123456789',
         },
         {
           foo: 2,
-          bar: 21337n,
+          bar: 21337,
```
petertonysmith94, line 520:
```suggestion
          bar: 21337n,
```

Why are we removing the BigInt expression here?

petertonysmith94 (commented):
Looks good to me

packages/fuel-gauge/src/coverage-contract.test.ts
nedsalk: Because we're getting compilation errors that `bigint` can't be passed as `BigNumberish`.

packages/fuel-gauge/src/coverage-contract.test.ts
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/2789#discussion_r1682778529

packages/fuel-gauge/src/predicate-conditional-inputs.test.ts
nedsalk: I don't know, really. A GH discussion on this would be nice to flesh it out.

packages/fuel-gauge/src/payable-annotation.test.ts
```diff
@@ -39,7 +39,7 @@ test("don't allow sending coins to non-payable functions", async () => {
   const baseAssetId = contract.provider.getBaseAssetId();
 
   // This should fail because the function is not payable
-  await expect(async () =>
+  await expect(() =>
```
maschad, line 42: Perhaps to make this more intuitive we can refactor as such

```suggestion
expect(async () => {
    const tx = await contract.functions
      .non_payable()
      .callParams({
        forward: {
          amount: bn(100),
          assetId: baseAssetId,
        },
      })
      .call();

    await tx.waitForResult();
  }).rejects.toThrowError(
    `The target function non_payable cannot accept forwarded funds as it's not marked as 'payable'.`
  );
   ```

maschad (approved):
Thanks for the work here @nedsalk , appreciate you helping me get this over the line.

packages/fuel-gauge/src/predicate-conditional-inputs.test.ts
petertonysmith94: https://github.com/FuelLabs/fuels-ts/discussions/2792

packages/fuel-gauge/src/payable-annotation.test.ts
arboleya: The `await` should not have been removed.

maschad: Why is that? If we are awaiting the contract call isn't it the same result?

PR #2814, opened 2024-07-22
fix: allow setting min gas price in `launchNode` args

### LINKED ISSUE #2813: Make `launchNode` Support Min Gas Price
The [launchNode](https://github.com/FuelLabs/fuels-ts/blob/f3453b9ea669575e3cf97ee50512bc9039845e0c/packages/account/src/test-utils/launchNode.ts#L115) test utility does not support setting the `--min-gas-price` node configuration correctly. 

The issue arises because when the `--min-gas-price` flag is manually specified in the args array, it is not removed from the default flags. As a result, the flag is included twice, leading to an error.

### Intended use:
```ts
  launchNode({
    args: ['--min-gas-price', '0'],
  });
```

### Result Error
```
FuelError: error: the argument '--min-gas-price <MIN_GAS_PRICE>' cannot be used multiple times
```

mvares:
- Closes #2813 

# Summary
In the previous solution, the `minGasPrice` was already defined by default. This PR fixes it by basing it on the `params`.
# Checklist

### DIFF
--- a/.changeset/mighty-crews-vanish.md
+++ b/.changeset/mighty-crews-vanish.md
@@ -1,4 +1,4 @@
 ---
 ---
 
-fix: support minimum gas price in args
+fix: allow setting min gas price in `launchNode` args
--- a/.changeset/mighty-crews-vanish.md
+++ b/.changeset/mighty-crews-vanish.md
@@ -1,0 +1,4 @@
+---
+---
+
+fix: support minimum gas price in args
--- a/packages/account/src/test-utils/launchNode.ts
+++ b/packages/account/src/test-utils/launchNode.ts
@@ -129,6 +129,8 @@ export const launchNode = async ({
       '--consensus-key',
       '--db-type',
       '--poa-instant',
+      '--min-gas-price',
+      '--native-executor-version',
     ]);
 
     const snapshotDir = getFlagValueFromArgs(args, '--snapshot');
--- a/packages/account/src/test-utils/launchNode.ts
+++ b/packages/account/src/test-utils/launchNode.ts
@@ -142,6 +142,8 @@ export const launchNode = async ({
 
     const nativeExecutorVersion = getFlagValueFromArgs(args, '--native-executor-version') || '0';
 
+    const minGasPrice = getFlagValueFromArgs(args, '--min-gas-price') || '1';
+
     // This string is logged by the client when the node has successfully started. We use it to know when to resolve.
     const graphQLStartSubstring = 'Binding GraphQL provider to';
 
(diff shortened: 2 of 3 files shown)


petertonysmith94 (approved):

packages/account/src/test-utils/launchNode.ts
petertonysmith94, line 1: Could we exclude `--min-gas-price` and `--native-executor-version` from the [`remainingArgs`](https://github.com/FuelLabs/fuels-ts/pull/2814/files#diff-b8f07cac51692505e88e065262fd373293a373ed5f6196eb13ec811e37aee136R131)?

mvares: cc @Torres-ssf

Torres-ssf (changes requested):
@mvares Thanks for this contribution 👍  

We are almost there, we only need to remove the `--min-gas-price` from the informed args. 

This is required based on your solution, otherwise the `--min-gas-price` will still be used twice.

You simply need to edit the call to `extractRemainingArgs` at the `launchNode` function.

```ts
    const remainingArgs = extractRemainingArgs(args, [
      '--snapshot',
      '--consensus-key',
      '--db-type',
      '--poa-instant',
      '--min-gas-price', // <- needs to be added here
      '--native-executor-version', // Please can you also add this missing flag?
    ]);
```

packages/account/src/test-utils/launchNode.ts
mvares: https://github.com/FuelLabs/fuels-ts/pull/2814/commits/1a22d73cf018c8650a61f0a78fc694d1bab57702

mvares:
Sorry. I still need to edit something.

Torres-ssf (changes requested):
@mvares It would be good if we could also add a test to validate the changes made here.

Let's unskip and change this test case [here](https://github.com/FuelLabs/fuels-ts/blob/f3453b9ea669575e3cf97ee50512bc9039845e0c/packages/account/src/providers/provider.test.ts#L1076-L1090) to:

```ts
  it('should ensure estimated fee values on getTransactionCost are never 0', async () => {
    using launched = await setupTestProviderAndWallets({
      nodeOptions: {
        args: ['--min-gas-price', '0'],
      },
    });
    const { provider } = launched;
    const request = new ScriptTransactionRequest();

    const { minFee, maxFee, gasPrice } = await provider.getTransactionCost(request);

    expect(gasPrice.eq(0)).toBeTruthy();

    expect(maxFee.eq(0)).not.toBeTruthy();
    expect(minFee.eq(0)).not.toBeTruthy();
  });
```

mvares:
:shipit:

.changeset/mighty-crews-vanish.md
```diff
@@ -0,0 +1,4 @@
+---
+---
+
+fix: support minimum gas price in args
```
Torres-ssf, line 4: fix: allow setting min gas price in `launchNode` args

Torres-ssf (changes requested):
@mvares This is the last one

.changeset/mighty-crews-vanish.md
mvares: https://github.com/FuelLabs/fuels-ts/pull/2814/commits/81eafa3c89827cfa498b0a93404d350ed1bbdf00

Torres-ssf (approved):

arboleya (approved):

nedsalk (approved):

Issue #2790, opened 2024-07-18
Make configurable properties optional in typegen outputs


nedsalk:
Configurable constants are by their nature optional. You can define one, multiple or all of them for your program. 
If we take a [predicate with configurables](https://github.com/FuelLabs/fuels-ts/blob/master/packages/fuel-gauge/test/fixtures/forc-projects/predicate-with-configurable/src/main.sw) as an example:
```sway
predicate;

configurable {
    FEE: u8 = 10,
    ADDRESS: b256 = 0x38966262edb5997574be45f94c665aedb41a1663f5b0528e765f355086eebf96,
}

fn main(fee: u8, address: b256) -> bool {
    FEE == fee && address == ADDRESS
}
```
The (relevant) generated typegen output is:
```ts
export type PredicateWithConfigurableAbiConfigurables = {
  FEE: BigNumberish;
  ADDRESS: string;
};

export const PredicateWithConfigurableAbi__factory = { ... }
```
In usage, when wanting to define _only_ the `FEE` configurable you'd get a compilation error:
```ts
const predicate = PredicateWithConfigurableAbi__factory.createInstance(
  provider,
  [myFee, myAddress ],
  { FEE: 234 } // <----------- complains that ADDRESS cannot be `undefined`
);
```

## The fix

To fix this, the generated typegen output must be `Partial`:
```ts
// before
export type PredicateWithConfigurableAbiConfigurables = {
  FEE: BigNumberish;
  ADDRESS: string;
};

// after
export type PredicateWithConfigurableAbiConfigurables = Partial<{
  FEE: BigNumberish;
  ADDRESS: string;
}>;
```

Issue #2787, opened 2024-07-18
Add validation around `programType` in relevant places


nedsalk:
https://github.com/FuelLabs/fuel-specs/pull/596 introduced the [`programType`](https://github.com/FuelLabs/fuel-specs/pull/596/files#diff-29c836429beedee654a72c37d1f883c1e5a0084d468e1446c61da13d7f88783dR37) property on the abi. Once this is part of `forc` outputs, we can use it to validate ABIs in the constructors of `Contract`, `ContractFactory`, `Predicate`, `Script`, and anywhere else that it's used.

arboleya:
Duplicate of:
 - https://github.com/FuelLabs/fuels-ts/issues/989

arboleya:
@nedsalk Please, feel free to add extra notes to the original issue's description.

Issue #2686, opened 2024-07-03
Integrate typegen outputs and programs via type generic parameters


nedsalk:
The end goal is to have `typegen` only output types and integrate those types with the real classes via generic type parameters. `typegen` currently outputs code which casts programs _as_ the types it generates. For example, for a simple counter contract which looks like this:
```sway
abi Counter {
    #[storage(read)]
    fn get_count() -> u64;

    #[storage(write, read)]
    fn increment_count(amount: u64) -> u64;
}
```

The generated outputs are `CounterAbi.d.ts` and `CounterAbi__factory.ts`, respectively:
```ts
// CounterAbi.d.ts
interface CounterAbiInterface extends Interface {
  functions: {
    get_count: FunctionFragment;
    increment_count: FunctionFragment;
  };
}

export class CounterAbi extends Contract {
  interface: CounterAbiInterface;
  functions: {
    get_count: InvokeFunction<[], BN>;
    increment_count: InvokeFunction<[amount: BigNumberish], BN>;
  };
}
```

```ts
// CounterAbi__factory.ts
export const CounterAbi__factory = {
  abi: _abi,

  storageSlots: _storageSlots,

  createInterface(): CounterAbiInterface {
    return new Interface(_abi) as unknown as CounterAbiInterface
  },

  connect(
    id: string | AbstractAddress,
    accountOrProvider: Account | Provider
  ): CounterAbi {
    return new Contract(id, _abi, accountOrProvider) as unknown as CounterAbi
  },

  async deployContract(
    bytecode: BytesLike,
    wallet: Account,
    options: DeployContractOptions = {}
  ): Promise<CounterAbi> {
    const factory = new ContractFactory(bytecode, _abi, wallet);

    const contract = await factory.deployContract({
      storageSlots: _storageSlots,
      ...options,
    });

    return contract as unknown as CounterAbi;
  }
}
```

The idea is for `CounterAbi.d.ts` to contain an appropriate type representation of the ABI which would be passed to `Interface<TAbi>` and `Contract<TAbi>` classes, removing the need for the `as unknown as CounterAbi` casts. Below is an idea of the updated `CounterAbi.d.ts` and `CounterAbi__factory.ts` contents, respectively:

```ts
// CounterAbi.d.ts
export interface CounterAbi {
  functions: {
    get_count: {
      inputs: [];
      output: BN;
    };
    increment_count: {
      inputs: [amount: BigNumberish];
      output: BN;
    };
  };
[...]

Issue #989, opened 2023-05-08
Typegen - Account for program types automatically


danielbate:
Currently the default behaviour of the typegen CLI is to generate types for the contract program type:

https://github.com/FuelLabs/fuels-ts/blob/49c103b4952ce3d68efb28e3f8ba9c7d33260021/packages/abi-typegen/src/cli.ts#L16-L34

We have had a couple of questions on the forum suggesting that they would expect typegen default behaviour to account for predicates:
- https://forum.fuel.network/t/how-to-build-predicate-ad-create-abi-for-predicate/2467/4
- https://forum.fuel.network/t/cant-create-predicate-instance-following-use-type-tutorial-in-ts-sdk/2490/6

So should we adjust typegen to account for all program types? Program type can be inferred from the ABI [`programType`](https://github.com/FuelLabs/fuels-ts/blob/7af2130c12e969e8bc1fa709a8371d77c12037fe/packages/abi-coder/src/types/JsonAbiNew.ts#L8) property. The parameters could remain and act as a filter for when a single program type is required.

arboleya:
**[TL;DR]** Currently, ABIs have no information about the program type.

----

We touched on this here:
 - https://github.com/FuelLabs/fuels-ts/pull/750#discussion_r1121555229

And also here:
 - https://github.com/FuelLabs/sway/issues/4244#issuecomment-1466671639

----

Maybe we should have an explicit issue for this on the Sway repo?

arboleya:
@nedsalk Should we address this together with the new abi spec PR?
 - https://github.com/FuelLabs/fuels-ts/pull/2747

nedsalk:
@arboleya we can, but I'd do it in a separate PR that can be based off of the "new ABI format" one. Also, I believe that this can bring about some breaking changes like removing `contracts`/`scripts`/`predicates` from the fuels cli if we increase the scope to cover the cli.

arboleya:
Makes sense, better to do it in isolation.

petertonysmith94:
This is being implemented in [TS-600](https://linear.app/fuel-network/issue/TS-600/abi-refactor-gen)

danielbate:
Resolved by #3032

PR #2815, opened 2024-07-22
fix: make properties of configurables optional in typegen

Torres-ssf:
- Closes https://github.com/FuelLabs/fuels-ts/issues/2790

# Summary

This PR updates the typegen template to ensure that all configurable properties are optional.

# Checklist


arboleya (dismissed):

packages/fuel-gauge/src/policies.test.ts
```diff
@@ -157,8 +157,8 @@ describe('Policies', () => {
     } = launched;
 
     const factory = new ContractFactory(
-      ScriptMainArgsAbi__factory.bin,
-      ScriptMainArgsAbi__factory.abi,
+      PayableAnnotationAbiHex,
```
maschad, line 160: Why were these changed to `PayableAnnotation`  abi?

packages/fuel-gauge/src/predicate/predicate-configurables.test.ts
```diff
@@ -198,6 +198,50 @@ describe('Predicate', () => {
...
+        wallets: [wallet],
+      } = launched;
+
+      const configurableConstants = {
+        ADDRESS: getRandomB256(),
+      };
+
+      const amountToTransfer = 300;
+
+      const predicate = new Predicate({
```
maschad, line 215: I think given we are testing out the typegen for this use case we should probably use the `createInstance` syntax to demonstrate the type system is accommodating the partiality.  e.g.

```ts
const predicate = PredicateWithConfigurableAbi__factory.createInstance(
  provider,
  [myFee, myAddress ],
  { FEE: 234 }
);
```

maschad (changes requested):
Looks good, left two comments

packages/fuel-gauge/src/policies.test.ts
petertonysmith94: I assume because the `ScriptMainArgs` is not a contract and `PayableAnnotation` is.

maschad: That's valid 😅 , perhaps it should be a script instance, although `ContractFactory` was used from this [test's inception](https://github.com/FuelLabs/fuels-ts/pull/1437/files#diff-0c6d881187bb1fdc9547309f7ee2830612689983980df36e3ec61be1224670b2R95), perhaps it was an oversight then as well. @Torres-ssf thoughts?

Torres-ssf: @maschad I am using `PayableAnnotation` because the `ScriptMainArgs` ABI and bytes were being incorrectly used as a contract ABI and bytes.

maschad: I suppose it doesn't really matter whether this is contract or a script for the purposes of this test, which is why it was passing before.

packages/fuel-gauge/src/predicate/predicate-configurables.test.ts
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2815/commits/0242f3bd6ae7a9bd9bf455ddfc07c6052113cc62

maschad (dismissed):
Nice

packages/fuel-gauge/src/policies.test.ts
Torres-ssf: @maschad The tests were passing because in both examples, the contract was never being deployed.

petertonysmith94 (dismissed):

packages/fuel-gauge/src/policies.test.ts
nedsalk: This will become impossible once #2787 / #989 is done.

nedsalk (approved):

.changeset/tasty-fireants-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/abi-typegen": patch
+---
+
+fix: make Predicate configurables option in typegen
```
nedsalk, line 5:
```suggestion
fix: make configurables properties optional in typegen
```

nedsalk (changes requested):
This PR currently only solves the problem for predicates, but #2790 is a general issue that includes scripts and contracts as well, so they should be covered in this PR also.

.changeset/tasty-fireants-switch.md
Torres-ssf: @nedsalk Thanks for your suggestion. 

I am having a hard time understanding how this issue extends to Contracts and Scripts. 

Could you please provide a detailed outline of the error and an example of a scenario where this issue occurs for Contracts and Scripts?

nedsalk: For contracts, you would have to change [this line](https://github.com/FuelLabs/fuels-ts/blob/8c5171d947291814c6bca0334d72379e51167167/packages/abi-typegen/src/templates/contract/dts.hbs#L44) to include the `Partial<...>` as you did for the predicates. The relevant script file to change is [this one](https://github.com/FuelLabs/fuels-ts/blob/8c5171d947291814c6bca0334d72379e51167167/packages/abi-typegen/src/templates/script/factory.hbs#L47). Upon changing it, you'll find that some tests are failing and you'll have to do a change like you did [here](https://github.com/FuelLabs/fuels-ts/pull/2815/files#diff-50b9975b1c89e23c847d731fb5c02edaac999b537242e2056e29451e904cae3fR19).

Torres-ssf: @nedsalk Ok so for Scripts and Contracts this only results in error if you use the generated configurable type.

That is why I did not understand the request.

The error is flagged for predicates as soon as you try to instantiate the predicate with a partial configurable.

But for Scripts and Contracts it only happens if you type safe the configurable object:

```ts
const configurable: ConfigurableContractAbiConfigurables {
   ...
}
```

nedsalk: Okay, so configurables for contracts and scripts aren't taken into account in the generated factory code. This is one more reason to go for #2686 which will make it easy to add this type safety.

nedsalk (approved):

PR #2796, opened 2024-07-18
feat!: deploy contract validation

danielbate:
- Relates to #2741

# Release notes

In this release, we:

- Limited deployment to contracts up to 100kb

# Summary

This is the predecessor to the large contract deployment support in #2741. It introduces two changes, one to `throw` when deploying contracts greater than 100 KB in size via the `ContractFactory`. Additionally a change in terminology from `invalid` -> `unsupported` tx types to better describe tx support for blob txs.

# Breaking Changes

`ErrorCode.INVALID_TRANSACTION_TYPE` was migrated to `ErrorCode.UNSUPPORTED_TRANSACTION_TYPE`.

```ts
// before
const code = ErrorCode.INVALID_TRANSACTION_TYPE;
```

```ts
// after
const code = ErrorCode.UNSUPPORTED_TRANSACTION_TYPE;
```

# Checklist

- [x] I **_added_** — `tests` to prove my changes
- [x] I **_updated_** — all the necessary `docs`
- [x] I **_reviewed_** — the entire PR myself, using the GitHub UI
- [x] I **_described_** — all breaking changes and the Migration Guide

PR #2824, opened 2024-07-23
feat!: prettify `typegen` api

arboleya:
- Closes #2686
 - Closes #2707

# Release notes

In this release, we:
- Made `typegen`d API more straightforward and ergonomic

# Summary

Parts of our great experience are shipped through the Typegend files, which contain part of the boilerplate necessary to make some things work. This PR revisits and trims these templates nicely to support a more ergonomic and intuitive API.

All main Typegen templates have changed:
 1. `Contract`
 1. `ContractFactory` ⁽¹⁾
 1. `Script`
 1. `Predicate`

> ⁽¹⁾ _Contract `bytecode` lives [inside](https://github.com/FuelLabs/fuels-ts/pull/2824/files#diff-0dcc8736e901872c3f59316d826b990d4fcbacef951f8093c2b8b6dfec055f55R8) the `Factory` class now, no more `.hex.ts` files_

To follow these changes, we also modified:
 - `launchTestNode` utility
 - `Predicate` class

## Special thanks ❤️ 

Goes to @petertonysmith94, for the rigorous reviews and additional API prettification via:
 - https://github.com/FuelLabs/fuels-ts/pull/2876.

Thanks also to @nedsalk for all the involvement, not letting this fade out, the original issues, reviews, and also for:
 - https://github.com/FuelLabs/fuels-ts/pull/2873

Releasing this will undoubtedly pave the way for other good things to happen! 

# Breaking Changes

## `Predicate` class

 - `Predicate` class constructor parameters renamed: `inputData` > `data`

```ts
// before
import { Predicate } from 'fuels';

const predicate = new Predicate({
  ...unchangedParameters,
  inputData,
});
```

```ts
// after
import { Predicate } from 'fuels';

const predicate = new Predicate({
  ...unchangedParameters,
  data,
});
```

- Typegen extended/generated `Predicate` now accepts a single parameter for initialization

```ts
// before
import { TestPredicateAbi__factory } from './typegend';

TestPredicateAbi__factory.createInstance(provider, data, configurableConstants);
```

```ts
// after
import { TestPredicate } from './typegen';

new TestPredicate({
  provider,
  data,
  configurableConstants
});
```

## `launchTestNode` utility

 - Renamed `contractsConfigs[].deployer` to  `contractsConfigs[].factory`
 - Removed `contractsConfigs[].bytecode` and `.hex.ts` file

The bytecode is now saved within the factory class, so you don't have to deal with it.

```ts
// before
import { TokenAbi__factory } from './typegend';
[...]

PR #2827, opened 2024-07-24
feat!: `fuel-core@0.32.1` and large contract deployments

### LINKED ISSUE #2741: Support new Blob TX type and deployment in chunks
> [!IMPORTANT]
> - Tasks are prioritized in order.
> - It is possible to address each one in isolation with a different PR.
> - Only the third and last one can close this issue; the rest should only mention it.

Tasks:
 - [x] 1) Change TX error message from `invalid` to `unsupported` ([source](https://github.com/FuelLabs/fuels-ts/blob/25587d4aeb4d33f2fc47b03719dc799d47a04a32/packages/transactions/src/coders/transaction.ts#L629-L634))
 - [x] 2) Throw error prohibiting deployment of >100kb contracts ([source](https://github.com/FuelLabs/fuels-ts/blob/25587d4aeb4d33f2fc47b03719dc799d47a04a32/packages/contract/src/contract-factory.ts#L149))
 - [x] 3) Implement support for >100kb contracts
    - [x] Support blob TX type (requirement for contract chunk deployments)
    - [x] Deploy contract validation for contracts that exceed the max contract size (taken from consensus parameters)
    - [x] Chunk and deploy large contracts via blob tx and store contract IDs
    - [x] Generate loader contract 
    - [x] Call contract functions via the loader contract 

---

> [!NOTE]
> This is the rationale for the above.

Options analysis:
 1. Keep everything as is for <=100k contracts; throws for bigger ones
    - ✅ Keeps browser compatibility
    - ✅ Doesn't block mainnet
    - ⚠️ >100kb contracts must be deployed via `forc deploy`
 1. Implement new [chunk] deployment mechanism
    - ✅ Keeps browser compatibility
    - ⛔ Blocks mainnet
 1. Use `forc deploy` under the hood
    - ⛔ Breaks browser compatibility
    - ⛔ Blocks mainnet

Conclusions:
 1. Clear, immediate winner.
 1. May still be considered.
 1. Probably a no-go.

---

Related PRs:
 - https://github.com/FuelLabs/fuel-vm/pull/780
 - https://github.com/FuelLabs/fuel-core/pull/1988
 - https://github.com/FuelLabs/sway/pull/6250

---

Specs:
 - https://github.com/FuelLabs/fuel-specs/issues/589

---

Image

### LINKED ISSUE #2895: Upgrade to `fuel-core@0.32.0`
[Release notes](https://github.com/FuelLabs/fuel-core/pull/2053)

danielbate:
- Closes #2895 
- Closes #2741

# Release notes

In this release, we:

- Support for `fuel-core@0.32.1`
- Implemented deploying large contracts using blob transactions

# Summary

- Introduced a `BlobTransactionRequest` to support the submission of blob transactions, a type of stateless tx whose purpose is just to facilitate arbitrary data in the VM.
- Added the `ContractFactory.deployAsBlobTx()` function to deploy contracts via a loader script whose bytecode is greater than the accepted max contract size for a chain. A new attribute has been added to the `DeployContractOptions`, `chunkSizeOverride` for managing fluctuating transaction sizes when calculating the max chunk size for a blob tx.
- The `ContractFactory.deploy()` can now deploy contracts of any size, it chooses the most appropriate deployment method dependent on the contract size.
- Introduced a `ContractFactory.deployAsCreateTx()` for deployments where a user wants to enforce a single create tx, which will now validate contract size against a consensus parameter rather than the constant in #2796.
- Introduced a loader script which takes blob IDs and coverts them into a deployable loader contract, based on instructions from the ASM package.

# Breaking Changes

`MAX_CONTRACT_SIZE` is no longer exported, it should now be fetched from the chain.

```ts
// before
import { MAX_CONTRACT_SIZE } from 'fuels';
```

```ts
// after
import { Provider, FUEL_NETWORK_URL } from 'fuels'; 

const { consensusParameters } = provider.getChain();
const maxContractSize = consensusParameters.contractParameters.contractMaxSize.toNumber();
```

# Checklist


packages/account/src/providers/provider.ts
```diff
@@ -951,7 +950,7 @@ Supported fuel-core version: ${supportedVersion}.`
         const hasMissingOutputs =
           missingOutputVariables.length > 0 || missingOutputContractIds.length > 0;
         const request = allRequests[requestIdx];
-        if (hasMissingOutputs && request?.type === TransactionType.Script) {
+        if (hasMissingOutputs && request instanceof ScriptTransactionRequest) {
```
Torres-ssf, line 953: @danielbate why using `instanceof` for this check? 

I have encountered situations where `instanceof` returned false when it should have returned true in similar contexts. Specifically, I've seen this issue when validating if a `TransactionRequest` is a `ScriptTransactionRequest`.

ref: https://github.com/FuelLabs/fuels-ts/pull/1702

danielbate: I'd wrongly assumed that `... === TransactionType.Script` didn't pass down types in block scope. Tbf @petertonysmith94 had also warned me.

Will change 👍🏻

packages/account/src/providers/fuel-core-schema.graphql
```diff
@@ -387,7 +389,6 @@ type GasCosts {
   divi: U64!
   ecr1: U64!
   eck1: U64!
-  ed19: U64!
```
petertonysmith94, line 390: Should this be updated in the [chainConfig](https://github.com/FuelLabs/fuels-ts/blob/fd95a20d87038b30aa1d03fc356648027606ff57/.fuel-core/configs/chainConfig.json#L60)?

packages/account/src/providers/transaction-response/transaction-response.ts
```diff
@@ -228,6 +228,13 @@ export class TransactionResponse {
 
     for await (const { statusChange } of subscription) {
       if (statusChange.type === 'SqueezedOutStatus') {
+        if (/BlobIdAlreadyUploaded/i.exec(statusChange.reason)) {
```
petertonysmith94, line 231: Favour `indexOf` over regex for performance.

```suggestion
        if (statusChange.reason.toLowerCase().indexOf('blobidalreadyuploaded') > -1) {
```

or (if it doesn't need to case insensitive)

```suggestion
        if (statusChange.reason.indexOf('BlobIdAlreadyUploaded') > -1) {
```

packages/account/src/providers/utils/gas.ts
```diff
@@ -2,6 +2,7 @@ import { bn } from '@fuel-ts/math';
 import type { BN, BNInput } from '@fuel-ts/math';
 import { ReceiptType, type Input } from '@fuel-ts/transactions';
 import { arrayify } from '@fuel-ts/utils';
+import { blob } from 'stream/consumers';
```
petertonysmith94, line 5:
```suggestion
```

Unused

packages/contract/src/contract-factory.ts
```diff
@@ -154,18 +200,135 @@ export default class ContractFactory {
...
+
+      // Chunks must be byte aligned
+      if (chunk.length % WORD_SIZE !== 0) {
+        const paddingLength = chunk.length % WORD_SIZE;
+        const paddedChunk = concat([chunk, new Uint8Array(paddingLength)]);
+        chunks.push({ id: index, size: paddedChunk.length, bytecode: paddedChunk });
+        // eslint-disable-next-line no-continue
+        continue;
+      }
+      chunks.push({ id: index, size: chunk.length, bytecode: chunk });
```
petertonysmith94, line 276:
```suggestion
      let chunk = this.bytecode.slice(offset, offset + chunkSize);
      let length = chunk.length

      // Chunks must be byte aligned
      if (chunk.length % WORD_SIZE !== 0) {
        length = chunk.length % WORD_SIZE;
        chunk = concat([chunk, new Uint8Array(length)]);
      }
      chunks.push({ id: index, size: chunk.length, bytecode: chunk });
```

packages/contract/src/contract-factory.ts
```diff
@@ -154,18 +200,135 @@ export default class ContractFactory {
...
+          // eslint-disable-next-line no-continue
+          continue;
+        }
+
+        throw new FuelError(ErrorCode.TRANSACTION_FAILED, 'Failed to deploy contract chunk');
+      }
+
+      if (!result.status || result.status !== TransactionStatus.success) {
+        throw new FuelError(ErrorCode.TRANSACTION_FAILED, 'Failed to deploy contract chunk');
+      }
```
petertonysmith94, line 314: Probably not one for now - but I assume there are some conditions where we'd want to retry the transaction rather than just fail?

packages/account/src/providers/fuel-core-schema.graphql
danielbate: This field has changed a few times since Thursday last week, it's now settled on being a dependent cost field.

.fuel-core/configs/stateConfig.json
```diff
@@ -469,6 +469,7 @@
     }
   ],
   "contracts": [],
```
nedsalk, line 471: I missed this field, would be interesting to see how it could be integrated with `launchTestNode`. Here's an [example](https://github.com/FuelLabs/fuel-core/blob/c2f5a3a64c0e792ef5390346ce0264696c95ea21/crates/chain-config/src/config/snapshots/fuel_core_chain_config__config__state__tests__snapshot_contract_with_state.snap#L8) of a state config with this array filled.

Torres-ssf: @nedsalk, I also reflected on this configuration yesterday. 

I believe we can use it to start the node with the contract already deployed, rather than deploying it afterward.

packages/fuel-gauge/src/contract-factory.test.ts
```diff
@@ -249,21 +269,43 @@ describe('Contract Factory', () => {
...
+        cacheUtxo: -1,
+      },
+    });
+
+    const {
+      wallets: [wallet],
+    } = launched;
+    const factory = new ContractFactory(largeContractHex, LargeContractAbi__factory.abi, wallet);
+
+    const deploy = await factory.deployContractLoader<LargeContractAbi>();
```
nedsalk, line 301: What is the reasoning for having two deployment methods based on the contract size? At first impression, it seems to me that this should be transparent to the user and that all of this should be under the `deploy` method: if the contract is less than 100KB, deploy as usual; if it's more, deploy via blobs.

packages/contract/src/contract-factory.ts
```diff
@@ -145,6 +157,38 @@ export default class ContractFactory {
...
+    const { blobBytecode } = options;
+    return new BlobTransactionRequest({
+      blobId: hash(blobBytecode),
+      witnessIndex: 0,
+      witnesses: [blobBytecode],
+      ...options,
+    });
+  }
+
+  async fundTransactionRequest(request: TransactionRequest, options: DeployContractOptions = {}) {
```
nedsalk, line 170:
```suggestion
  private async fundTransactionRequest(request: TransactionRequest, options: DeployContractOptions = {}) {
```

packages/contract/src/contract-factory.ts
```diff
@@ -154,18 +198,141 @@ export default class ContractFactory {
...
+  async deployContractLoader<TContract extends Contract = Contract>(
+    deployContractOptions: DeployContractOptions = {}
+  ): Promise<DeployContractResult<TContract>> {
     const account = this.getAccount();
+    const { consensusParameters } = account.provider.getChain();
+    const maxContractSize = consensusParameters.contractParameters.contractMaxSize.toNumber();
 
-    const transactionResponse = await account.sendTransaction(transactionRequest);
+    // Ensure the max contract size is byte aligned before chunking
+    if (maxContractSize % WORD_SIZE !== 0) {
```
nedsalk, line 240: Why is this check our concern? Shouldn't it be the responsibility of the network/node runners to have this be word size aligned?

packages/contract/src/contract-factory.ts
```diff
@@ -154,18 +198,141 @@ export default class ContractFactory {
...
+      // Chunks must be byte aligned
+      if (chunk.length % WORD_SIZE !== 0) {
+        const paddingLength = chunk.length % WORD_SIZE;
+        const paddedChunk = concat([chunk, new Uint8Array(paddingLength)]);
+        chunks.push({ id: index, size: paddedChunk.length, bytecode: paddedChunk });
+        // eslint-disable-next-line no-continue
+        continue;
+      }
+      chunks.push({ id: index, size: chunk.length, bytecode: chunk });
+    }
```
nedsalk, line 273: I suggest putting this in a `private chunkBytecode` method.

packages/contract/src/contract-factory.ts
```diff
@@ -154,18 +198,141 @@ export default class ContractFactory {
...
+
+    const uploadedBlobIds: string[] = [];
+
+    // Deploy the chunks as blob txs
+    for (const { id, bytecode } of chunks) {
```
nedsalk, line 278: 1. Could all of this chunk deployment be done via a single multicall?
2. Is there a way to warn the user that they can't deploy the contract because they don't have enough funds? I imagine that it would be straightforward if it was a multicall.

.fuel-core/configs/stateConfig.json
arboleya: I think state and balance are good candidates, but I'd argue that we should not move everything and continue deploying as we do to dogfood the [deployment functionality](https://github.com/FuelLabs/fuels-ts/issues/2878#issuecomment-2265617725).

packages/contract/src/contract-factory.ts
danielbate: Agreed. However, muck alike other bytcode position issues, if there is an alignment issue it's likely to fail for a different reason (e.g. blob ID not found) which isn't helpful. I'd be okay with removing this though now that things are in the right place.

packages/contract/src/contract-factory.ts
danielbate: 1. Not sure but I'll investigate.
2. As above not sure re multicall, but not difficult to estimate blobs + create tx anyway, will implement.

packages/fuel-gauge/src/contract-factory.test.ts
danielbate: I had this thought also, however the experiences will be different and maybe that warrants a separate function / flow.

a) Contract < max size, calls `deployContract` and we submit a single create tx. Contract ID can be inferred directly from the bytecode that the user passed to it.

b) Contract > max size, calls `deployLoaderContract`, we submit multiple blob txs and then a create tx. Contract ID cannot be inferred until we have the blob IDs, and has to be via the loader bytecode rather than the bytecode the user passed.

Consider that a lot of our users will be using browser wallets that requires manual intervention to approve transactions, for a 2 mb contract under testnet conditions this could mean approving 7 transactions.

IMO the flow differences warrants a different functions. The RS team also went for a [different function](https://github.com/FuelLabs/fuels-rs/pull/1472), I'll find their reasoning also.

cc @FuelLabs/sdk-ts

arboleya: @nedsalk Yes, the original idea was to have it working transparently under the hood.

@danielbate I wonder if delegating the contract sizing to the user would be an improvement.

I like the idea of a smart[er] `deploy` method internally calling `deploySingle` or `deployMulti`.

If we want to give users the choice, we could also export all three.

packages/contract/src/contract-factory.ts
arboleya: Good ideas to look into.

packages/account/src/providers/provider.ts
```diff
@@ -846,7 +850,7 @@ Supported fuel-core version: ${supportedVersion}.`
       const hasMissingOutputs =
         missingOutputVariables.length !== 0 || missingOutputContractIds.length !== 0;
 
-      if (hasMissingOutputs) {
+      if (hasMissingOutputs && isTransactionTypeScript(transactionRequest)) {
```
maschad, line 853: There's a lot of repetitions of `isTransactionTypeScript(transactionRequest)` throughout this function, perhaps we could just set that as a constant.

packages/account/src/providers/provider.ts
```diff
@@ -1109,17 +1113,16 @@ Supported fuel-core version: ${supportedVersion}.`
...
     // Remove gasLimit to avoid gasLimit when estimating predicates
-    if (isScriptTransaction) {
+    if (isTransactionTypeScript(txRequestClone)) {
       txRequestClone.gasLimit = bn(0);
     }
 
     const signedRequest = clone(txRequestClone) as ScriptTransactionRequest;
     let addedSignatures = 0;
-    if (signatureCallback && isScriptTransaction) {
+    if (signatureCallback && isTransactionTypeScript(signedRequest)) {
```
maschad, line 1125: If this was [cast](https://github.com/FuelLabs/fuels-ts/pull/2827/files#diff-69a6670d0426c74935ed2830ce552e7df2379c2d112a582306e57569fcd17029R1123) to a `ScriptTransactionRequest` the latter half of this check should be redundant.

maschad (commented):
First pass, nice work so far :)

packages/account/src/providers/provider.ts
danielbate: Good spot, I actually think the cast is incorrect here.

packages/fuel-gauge/src/contract-factory.test.ts
danielbate: @segfault-magnet shares the sentiment that the flow change, and deployment of different bytecode that the user passes warrants a separate, more transparent flow.

He also mentions as contract size is chain configurable, users may see a difference in behaviour when testing locally / testnet vs mainnet. 

I'm going to keep the functions separate and exported seperately, but I'll introduce the _deployMulti_ function at the top of `deployContract` with a size check. And we can continue to discuss which way we'd like to go.

> @danielbate I wonder if delegating the contract sizing to the user would be an improvement.

I'll definitely implement something of this ilk.

arboleya: Apologies, I expressed myself poorly. 🙂 I meant the opposite—delegating the task of checking contract size and deciding on method A or B to the user doesn't seem like an improvement.

Although I understand the flow is different, I'm still on the original train of thought—about internalizing this if/else and guiding the user from there with clear errors in case they happen. Maybe I'm missing the actual downsides? 

Perhaps having all the options + plus a happy path with smart defaults could work well.

At the very least, we must provide utilities around contract size checks.

- `deploy` (if/else —> A|B happy path)
  - `deploySingle`
  - `deployMulti`
- `getContractSize(bytecode)`
- `getContractMaxSizeLimit()`

internal/fuel-core/VERSION
```diff
@@ -1 +1 @@
-0.31.0
+git:release/0.32.0
```
danielbate, line 1: Need to use the actual version once it's released.

packages/utils/src/utils/defaultSnapshotConfigs.ts
```diff
@@ -4,6 +4,7 @@ import stateConfigJson from './defaultSnapshots/stateConfig.json';
 import type { SnapshotConfigs } from './types';
 
 export const defaultSnapshotConfigs: SnapshotConfigs = {
+  // @ts-expect-error Getting type errors as we can't gen the fuel schema for v4 gas costs
```
danielbate, line 7: Need to gen the `v4` gas costs schema

packages/fuel-gauge/src/contract-factory.test.ts
danielbate: Understood. Currently I've implemented `deploy` (A|B), `deployContract` (ensures no breaking change) and `deployContractAsBlobs`. The [docs](https://github.com/FuelLabs/fuels-ts/blob/db/feat/blob-contract-deploys/apps/docs/src/guide/contracts/deploying-contracts.md) should give better context now.

packages/contract/src/contract-factory.ts
```diff
@@ -213,6 +357,16 @@ export default class ContractFactory {
     }
   }
 
+  /**
+   * Helper function to get the size of the contract's bytecode. Useful for informing
+   * which deployment method to use.
+   *
+   * @returns The size of the contract's bytecode.
+   */
+  getBytecodeSize(): number {
```
nedsalk, line 366: IMO there's no need for this method as `this.bytecode.length` / `myContractFactory.bytecode.length` is straightforward.

apps/docs/src/guide/contracts/deploying-contracts.md
```diff
@@ -7,40 +7,44 @@
 
 # Deploying Contracts
 
-This guide walks you through deploying a contract using the SDK, covering loading contract artifacts, initializing a contract factory, and deploying the contract.
+Deploying contracts using the SDK is handled by the `ContractFactory`. The process involves collecting the contract artifacts, initializing the contract factory, and deploying the contract.
```
maschad, line 10:
```suggestion
To deploy a contract using the SDK, you can use the `ContractFactory`. This process involves collecting the contract artifacts, initializing the contract factory, and deploying the contract.
```

packages/account/src/providers/provider.ts
maschad: I think you may have resolved this by accident @danielbate because I'm noit seeing any changes here

packages/fuel-gauge/src/contract-factory.test.ts
```diff
@@ -249,21 +258,146 @@ describe('Contract Factory', () => {
...
+    await expectToThrowFuelError(
+      () => factory.deployContractAsBlobs<LargeContractAbi>({ chunkSizeTolerance }),
+      new FuelError(
+        ErrorCode.INVALID_CHUNK_SIZE_TOLERANCE,
+        'Chunk size tolerance must be between 0 and 1'
+      )
+    );
+  });
+
+  it('deploys a small contract via blobs', async () => {
```
maschad, line 345: out of curiosity why would someone want to do this?

packages/contract/src/loader/loader-script.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  // Bytes for the Blob Ids
+  const blobIdBytes = concat(blobIds.map((b) => arrayify(b)));
+
+  // There are 2 main steps:
+  // 1. Load the blob contents into memory
+  // 2. Jump to the beginning of the memory where the blobs were loaded
+  // After that the execution continues normally with the loaded contract reading our
+  // prepared fn selector and jumps to the selected contract method.
+  const instructionBytes = new InstructionSet(
```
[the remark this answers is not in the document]
danielbate: There is no docs if that's what you are looking for? This is the RS SDKs design.

packages/fuel-gauge/src/contract-factory.test.ts
```diff
@@ -249,21 +258,146 @@ describe('Contract Factory', () => {
...
+    const { value } = await call.waitForResult();
+    expect(value.toNumber()).toBe(1001);
+  });
+
+  it('deploys large contracts via blobs [padded]', async () => {
+    using launched = await launchTestNode({
+      providerOptions: {
+        cacheUtxo: -1,
+      },
+    });
```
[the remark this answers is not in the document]
danielbate: If a transaction fails because a blob ID has already been used, the UTXO is still cached. This is being investigated in #2875.

packages/contract/src/contract-factory.ts
nedsalk: > 1. Could all of this chunk deployment be done via a single multicall?

I think not, because the [`max_size`](https://github.com/FuelLabs/fuels-ts/blob/db/feat/blob-contract-deploys/.fuel-core/configs/chainConfig.json#L11) of a transaction is the same as [`contract_max_size`](https://github.com/FuelLabs/fuels-ts/blob/aac2745f25dfeb9ee8e1ebc83c11f2268ff94ef4/.fuel-core/configs/chainConfig.json#L31) which we're [using](https://github.com/FuelLabs/fuels-ts/blob/36358fa37ca2bd032c88e0baa02ddab732a682a3/packages/contract/src/contract-factory.ts#L202) to determine if we'll deploy as blobs or not. A multicall transaction with all the blobs combined in it would suffer the same problem a `TransactionType.Create` transaction has and which we're mitigating with this PR.

 > 2. Is there a way to warn the user that they can't deploy the full contract because they don't have enough funds to cover all the blobs and contract loader transactions? 

This is still valid and should be done. We don't want to spend the user's funds on blob transactions if they can't deploy the whole contract.

danielbate: Apologies I hadn't completely looking into multicall yet but thankyou for the clarification. Agreed on the second point, I'll implement that today.

apps/docs/src/guide/contracts/deploying-contracts.md
```diff
@@ -7,40 +7,44 @@
 
 # Deploying Contracts
 
-This guide walks you through deploying a contract using the SDK, covering loading contract artifacts, initializing a contract factory, and deploying the contract.
+To deploy a contract using the SDK, you can use the `ContractFactory`. This process involves collecting the contract artifacts, initializing the contract factory, and deploying the contract.
 
-> **Note:** The maximum contract deployment size supported by the SDK is 100 KB. The SDK will `throw` an [error](../errors/index.md) for contracts larger than this size.
+The SDK utilizes two different deployment processes. Either simply using a single create transaction to deploy the entire contract bytecode, or, by splitting the contract into multiple chunks, deploying them as blobs (on chain data accessible to the VM) and then generating a contract from the associated blob IDs. That generated contract is then deployed as a create transaction.
```
nedsalk, line 12:
```suggestion
The SDK utilizes two different deployment processes, depending on the contract's size being less than or equal to the `contract_max_size` in the chain's consensus parameters. It either uses a single create transaction to deploy the entire contract bytecode, or it splits the contract bytecode into multiple chunks, deploys them as blobs (on chain data accessible to the VM), and then generates a contract from the associated blob IDs. That generated contract is then deployed as a create transaction.
```

apps/docs/src/guide/contracts/deploying-contracts.md
```diff
@@ -7,40 +7,44 @@
...
-## 1. Obtaining Contract Artifacts
+The `ContractFactory` offers the following methods for the different processes:
 
-After writing a contract in Sway and compiling it with `forc build` (<a :href="indexUrl" target="_blank" rel="noreferrer">read more</a> on how to work with Sway), you will obtain two important artifacts: the compiled binary file and the JSON ABI file. These files are required for deploying a contract using the SDK.
+- `deploy` for deploying contacts of any size (will automatically choose the appropriate deployment process).
+- `deployContract` for deploying the entire contract bytecode in a single create transaction.
+- `deployContractAsBlobs` for deploying the contract in chunks as blobs, and then deploying the contract as a create transaction.
 
-## 2. Setting up the SDK Environment
+> **Note:** Due to the nature of blob deployments, both `deploy` and `deployContractAsBlobs` may require multiple transactions to deploy the contract.
```
nedsalk, line 20:
```suggestion
> **Note:** If the contract is deployed via blob deployments, multiple transactions will be required to deploy the contract.
```

apps/docs/src/guide/contracts/deploying-contracts.md
```diff
@@ -7,40 +7,44 @@
...
 
-Before deploying a contract, set up the necessary environment by importing the required SDK components and initializing a wallet and a provider.
+The deployment process used by the `ContractFactory` is dependent on the size of the contract you are attempting to deploy. The threshold for the contract size is dictated by the chain. You can find the maximum contract size supported by the chain by querying the chains consensus parameters.
 
-<<< @/../../docs-snippets/src/guide/contracts/deploying-contracts.test.ts#contract-setup-1{ts:line-numbers}
+<<< @/../../docs-snippets/src/guide/contracts/deploying-contracts.test.ts#get-contract-max-size{ts:line-numbers}
```
nedsalk, line 25:
```suggestion
The SDK utilizes two different deployment processes, depending on the contract's size. The threshold for the contract size is dictated by the chain size and can be queried like this:

<<< @/../../docs-snippets/src/guide/contracts/deploying-contracts.test.ts#get-contract-max-size{ts:line-numbers}

It either uses a single create transaction to deploy the entire contract bytecode, or it splits the contract bytecode into multiple chunks, deploys them as blobs (on chain data accessible to the VM), and then generates a contract from the associated blob IDs. That generated contract is then deployed as a create transaction.

The `ContractFactory` offers the following methods for the different processes:

- `deploy` for deploying contacts of any size (will automatically choose the appropriate deployment process).
- `deployContract` for deploying the entire contract bytecode in a single create transaction.
- `deployContractAsBlobs` for deploying the contract in chunks as blobs, and then deploying the contract as a create transaction.

> **Note:** Due to the nature of blob deployments, both `deploy` and `deployContractAsBlobs` may require multiple transactions to deploy the contract.
```

packages/contract/src/loader/loader-script.ts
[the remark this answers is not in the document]
petertonysmith94: I assume you've copied these instructions from somewhere? Could we add a comment with the git reference?

packages/contract/src/contract-factory.ts
```diff
@@ -145,27 +152,162 @@ export default class ContractFactory {
...
+    const blobIds = chunks.map((c) => c.blobId as string);
+    const loaderBytecode = getLoaderInstructions(blobIds);
+
+    // Deploy the loader contract via create tx
+    const { contractId, transactionRequest: createRequest } = this.createTransactionRequest({
+      bytecode: loaderBytecode,
+      ...deployContractOptions,
+    });
+    await this.fundTransactionRequest(createRequest, deployContractOptions);
+    const transactionResponse = await account.sendTransaction(createRequest);
```
nedsalk, line 310: Referencing #2678 here as an example of a dependent transaction.

packages/fuel-gauge/src/contract-factory.test.ts
```diff
@@ -245,21 +253,146 @@ describe('Contract Factory', () => {
...
+
+    const {
+      wallets: [wallet],
+    } = launched;
+    const factory = new ContractFactory(LargeContractFactory.bytecode, LargeContract.abi, wallet);
```
arboleya, line 283: Now you can use typegen factory classes directly:

```suggestion
    const factory = new LargeContractFactory(wallet);
```

packages/utils/src/utils/defaultSnapshotConfigs.ts
Torres-ssf: @danielbate Was this fixed at the `fuel-core` graphql schema on `0.32.0`? 

It is still unclear to me why they did not add new values for the `GasCostsVersion` enum

packages/fuel-gauge/src/e2e-script.test.ts
```diff
@@ -76,6 +77,28 @@ describe.each(selectedNetworks)('Live Script Test', (selectedNetwork) => {
     wallet = new WalletUnlocked(privateKey, provider);
   });
 
+  it('can deploy a large contract to a live Fuel Node', async () => {
```
danielbate, line 80: @petertonysmith94 I've added an e2e for validation but tbh I'm not sure whether we should merge it with this. It seems unnecessarily costly and the only additional proving it gives is to validate blob deploys. With the same logic we should be e2e testing everything. Our current e2es prove compatibility with the live network for a subset of functionality and i think that's enough.

petertonysmith94: @danielbate I agree - doesn't seem wise to run on every e2e test run. Thanks for proving the change against the network though, that's valuable in itself.

danielbate: Note to remove the e2e before resolving this conversation.

packages/contract/src/contract-factory.ts
```diff
@@ -146,26 +154,175 @@ export default class ContractFactory {
...
+      };
+    });
+
+    // Check the account can afford to deploy all chunks
+    const costs = [];
+    for (const { transactionRequest } of chunks) {
+      const { minFee } = await account.getTransactionCost(transactionRequest);
+      costs.push(minFee);
+    }
+    const totalCost = costs.reduce((acc, cost) => acc.add(cost), bn(0));
```
Torres-ssf, line 275: @danielbate we can calculate the `minFee` directly here without calling `getTransactionCost` since we do not need to dry run it in order to get the gas used. 

I've also filtered the already estimated blob IDs just like already happens when the blob TXs are being submitted. Let me know if that makes sense. 

By doing this we could get a value close between the estimated and real fee values:

```terminal
total estimated fee: 557841

total payed in fee:  559354
```

```suggestion
    let totalCost = bn(0);
    const chainInfo = account.provider.getChain();
    const gasPrice = await account.provider.estimateGasPrice(10);
    const estimatedBlobIds: string[] = [];

    for (const { transactionRequest, blobId } of chunks) {
      if (!estimatedBlobIds.includes(blobId)) {
        const minGas = transactionRequest.calculateMinGas(chainInfo);
        const minFee = calculateGasFee({
          gasPrice,
          gas: minGas,
          priceFactor: chainInfo.consensusParameters.feeParameters.gasPriceFactor,
          tip: transactionRequest.tip,
        }).add(1);

        totalCost = totalCost.add(minFee);
        estimatedBlobIds.push(blobId);
      }
    }
```

danielbate: Great suggestion mate, and good point on filtering the already estimated blobId, makes sense.

packages/contract/src/contract-factory.ts
```diff
@@ -146,26 +155,188 @@ export default class ContractFactory {
...
+
+  /**
+   * Chunks and deploys a contract via a loader contract. Suitable for deploying contracts larger than the max contract size.
+   *
+   * @param deployOptions - Options for deploying the contract.
+   * @returns A promise that resolves to the deployed contract instance.
+   */
+  async deployContractAsBlobs<TContract extends Contract = Contract>(
+    deployOptions: DeployContractOptions = {
+      chunkSizeTolerance: CHUNK_SIZE_TOLERANCE,
```
maschad, line 250: @danielbate pardon my ignorance, but I'm having difficulty determining how I could derive an appropriate `chunkSizeTolerance` programmatically, if the is tied to fee fluctuations. Perhaps I am not understanding the context of how to properly estimate that.

Would it be possible to determine the optimal chunk size based on the account's current balance? I wonder if we could iteratively determine a cost for the user and thus avoid the need for a user provided value and also increase the probability that we only deploy in chunk sizes that will fall within the available balance.

@Torres-ssf made a great recommendation [here](https://github.com/FuelLabs/fuels-ts/pull/2827#discussion_r1707915203) which made me think it seems possible.

```suggestion
  const chainInfo = account.provider.getChain();
  const gasPrice = await account.provider.estimateGasPrice(10);

  // Function to estimate cost for a given chunk size
  const estimateCost = (chunkSize: number) => {
    const chunks = getContractChunks(this.bytecode, chunkSize);
    let totalCost = bn(0);
    const estimatedBlobIds: string[] = [];

    for (const chunk of chunks) {
      const transactionRequest = this.blobTransactionRequest({ bytecode: chunk.bytecode });
      if (!estimatedBlobIds.includes(transactionRequest.blobId)) {
        const minGas = transactionRequest.calculateMinGas(chainInfo);
        const minFee = calculateGasFee({
          gasPrice,
          gas: minGas,
          priceFactor: chainInfo.consensusParameters.feeParameters.gasPriceFactor,
          tip: transactionRequest.tip,
        }).add(1);
        totalCost = totalCost.add(minFee);
        estimatedBlobIds.push(transactionRequest.blobId);
      }
    }

    return { totalCost, chunkCount: chunks.length };
  };

  //search for optimal chunk size
  let minSize = WORD_SIZE; // Minimum chunk size
  let maxSize = this.getMaxChunkSize(); // Maximum possible chunk size
  let optimalSize = maxSize;
  let optimalCost = estimateCost(maxSize);

  while (minSize <= maxSize) {
    const midSize = Math.floor((minSize + maxSize) / 2);
    const { totalCost, chunkCount } = estimateCost(midSize);

    if (totalCost.lte(await account.getBalance())) {
      optimalSize = midSize;
      optimalCost = { totalCost, chunkCount };
      minSize = midSize + 1;
    } else {
      maxSize = midSize - 1;
    }
  }
```

Thoughts? @Torres-ssf @danielbate

danielbate: No good suggestion mate, I think there are improvements that can be made to the chunk optimisation tbh, potentially as follow up, but I'll take a look at this today. @Torres-ssf suggestion has been added 👍🏻 

Additionally, re chunk size tolerance, there are some parameters that are out of our control when coming to the chunk size at the time of estimation.  I should've also mentioned that the contract size consensus param isn't necessarily the upper limit. We also have a request size limit for HTTP requests against the server. Contract size is configurable by the chain and therefore could be set higher than the request size limit. Therefore we need some amount of user configurable tolerance for unknowns at the time of estimation.

packages/fuel-gauge/src/contract-factory.test.ts
danielbate: The issue here is the typegen factory only exposes the A|B deploy method (which I agree with). But in some of these tests I want to force the create or blob deploy method.

apps/docs/src/guide/contracts/deploying-contracts.md
nedsalk: Note that this is just a take on the docs that flows better for me. Pick and choose what you will from it.

danielbate: Sorry @nedsalk I haven't missed this. Will compare today.

packages/transactions/src/coders/transaction.ts
```diff
@@ -568,20 +569,122 @@ export class TransactionUploadCoder extends Coder<TransactionUpload, Transaction
...
+  /** Number of inputs (u16) */
+  inputsCount: number;
+
+  /** Number of outputs (u16) */
+  outputsCount: number;
+
+  /** Number of witnesses (u16) */
+  witnessesCount: number;
+
+  /** List of policies, sorted by PolicyType. */
```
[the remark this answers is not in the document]
danielbate: I've kept this doc the same as all other places where we use the `policies` property. It's sorted in the coder, policy type is a bitwise field.

packages/contract/src/contract-factory.ts
danielbate: Point 2 has now been implemented.

packages/contract/src/contract-factory.ts
danielbate: Implemented in `packages/contract/src/loader/utils.ts`.

packages/account/src/providers/provider.ts
danielbate: In `estimateTxDependencies` I can only see one call to `isTransactionTypeScript` and one for `isTransactionTypeCreate`? I can only see one function in this class where we could reduce the repetition in `getTransactionCost`, but we clone the request so still need to cast `ScriptTransactionRequest`, I'd rather just call the function twice? Let me know if you think I'm missing something.

apps/docs/src/guide/contracts/deploying-contracts.md
danielbate: It's nice 👍🏻

packages/contract/package.json
```diff
@@ -45,16 +45,18 @@
     "@fuel-ts/errors": "workspace:*",
     "@fuel-ts/hasher": "workspace:*",
     "@fuel-ts/interfaces": "workspace:^",
+    "@fuel-ts/math": "^0.93.0",
```
petertonysmith94, line 48:
```suggestion
    "@fuel-ts/math": "workspace:^",
```

packages/contract/src/contract-factory.ts
```diff
@@ -109,7 +117,7 @@ export default class ContractFactory {
    * @param deployOptions - Options for deploying the contract.
    * @returns The CreateTransactionRequest object for deploying the contract.
    */
-  createTransactionRequest(deployOptions?: DeployContractOptions) {
+  createTransactionRequest(deployOptions?: DeployContractOptions & { bytecode?: Uint8Array }) {
```
petertonysmith94, line 120:
```suggestion
  createTransactionRequest(deployOptions?: DeployContractOptions & { bytecode?: BytesLike }) {
```

packages/utils/src/utils/types.ts
petertonysmith94, line 1: I believe `ed19` is now an operation.

packages/contract/src/contract-factory.ts
maschad: Thanks for the clarification @danielbate 

I'm definitely open to this being a follow-up for sure, as I think the current implementation satisfies the current requirements of getting larger contracts deployed. 

I am still wrapping my mind around how one would derive a reasonable tolerance though, based on what you're saying, my understanding is that the upper limit is the max of the request size and the contract size, but shouldn't both of those parameters be retrievable at compile time? 

Perhaps we can have a call to discuss.

danielbate: So we have the transaction limit and contract limit, both retrievable via the consensus params (I actually hadn't accounted for tx limit so thats's now been added). However graphQL also has a maximum configurable request body size to contend with which will also return a `413` which I'm not sure we can retrieve.

We've also experienced some flakiness in devnet and the manual override has been really helpful.

But it's definitely worth the call.

packages/utils/src/utils/defaultSnapshotConfigs.ts
danielbate: It was not fixed however I've managed to manually fix the types so we no longer need the `@ts-expect-error`. I'll ask about the enum.

packages/contract/src/contract-factory.ts
danielbate: Going for fail fast at the moment, what scenarios do you think?

.github/workflows/test.yaml
```diff
@@ -103,9 +103,9 @@ jobs:
       - name: Run Isolated Tests
         run: pnpm test:e2e
         env:
-          TESTNET_WALLET_PVT_KEY: ${{ secrets.TESTNET_WALLET_PVT_KEY }}
+          # TESTNET_WALLET_PVT_KEY: ${{ secrets.TESTNET_WALLET_PVT_KEY }}
```
danielbate, line 106: Note to restore testnet and disable devnet once it is supported in both environments.

danielbate:
Marking as blocked whilst the devnet network is upgraded to support blob transactions. All tests are passing apart from the devnet/testnet `e2e`. All essential conversations have been resolved, I have a conversation with @maschad later on around improvements to chunk size estimation. Apart from that I think all issues have been resolved. Only additional thing I am going to investigate is the schema does not fetch `v4` gas costs, this has been manually included, but not a blocker.

danielbate:
Just awaiting [this](https://github.com/FuelLabs/fuel-core/pull/2067) being released to devnet.

danielbate:
Alternatively, @nedsalk suggested reverting #2820 and using `submitAndAwait` but looks like the [fuel core fix](https://github.com/FuelLabs/fuel-core/pull/2067) is moving forwards.

.github/workflows/test.yaml
danielbate: devnet e2e passing so enabling testnet

danielbate:
Devnet e2e is passing so I'm enabling tesnet e2e, we await the release to testnet.

.changeset/thirty-steaks-return.md
```diff
@@ -0,0 +1,11 @@
+---
+"@fuel-ts/transactions": minor
+"@internal/fuel-core": minor
+"@fuel-ts/contract": minor
+"@fuel-ts/versions": minor
+"@fuel-ts/account": minor
+"@fuel-ts/errors": minor
+"@fuel-ts/utils": minor
```
arboleya, line 8: Sorting and adding `errors` package.

```suggestion
"@fuel-ts/account": minor
"@fuel-ts/contract": minor
"@fuel-ts/errors": minor
"@fuel-ts/program": minor
"@fuel-ts/transactions": minor
"@fuel-ts/utils": minor
"@fuel-ts/versions": minor
"@internal/fuel-core": minor
```

arboleya (commented):
Is this ready for a final review?

Asking because it's marked as a draft.

danielbate:
@arboleya let's do final reviews but I'll leave the e2e in place until testnet supports `0.32.1` and will remove it then. Will pick up any additional unresolved conversations in a follow-up issue.

packages/fuel-gauge/src/e2e-script.test.ts
petertonysmith94: Is it worth just skipping - rather than removing entirely?

apps/docs-snippets/src/guide/contracts/deploying-contracts.test.ts
```diff
@@ -1,65 +1,108 @@
...
+    const abi = EchoValues.abi;
+    const bytecode = EchoValuesFactory.bytecode;
+
+    // #region setup
+    // #import { Provider, TESTNET_NETWORK_URL, Wallet, ContractFactory };
+    // #context import { WALLET_PVT_KEY } from 'path/to/my/env/file';
+    // #context import { bytecode, abi } from 'path/to/typegen/outputs';
+    const provider = await Provider.create(TESTNET_NETWORK_URL);
+    const wallet = Wallet.fromPrivateKey(WALLET_PVT_KEY, provider);
+    const factory = new ContractFactory(bytecode, abi, wallet);
```
petertonysmith94, line 46: Do you think we should favour showcasing typegen contract setup here?

apps/docs-snippets/src/guide/contracts/deploying-contracts.test.ts
```diff
@@ -1,65 +1,108 @@
...
+    const abi = EchoValues.abi;
+    const bytecode = EchoValuesFactory.bytecode;
+
+    // #region blobs
+    // #import { Provider, TESTNET_NETWORK_URL, Wallet, ContractFactory };
+    // #context import { WALLET_PVT_KEY } from 'path/to/my/env/file';
+    // #context import { bytecode, abi } from 'path/to/typegen/outputs';
+    const provider = await Provider.create(TESTNET_NETWORK_URL);
+    const wallet = Wallet.fromPrivateKey(WALLET_PVT_KEY, provider);
+    const factory = new ContractFactory(bytecode, abi, wallet);
```
petertonysmith94, line 88: Do you think we should favour showcasing typegen contract setup here?

apps/docs/src/guide/contracts/deploying-contracts.md
```diff
@@ -7,40 +7,54 @@
...
+In the above guide we use the recommended `deploy` method. If you are working with a contract that is too large to be deployed in a single transaction, then the SDK will chunk the contract for you and submit it as blobs, to then be accessed later by a create transaction. This process is handled by the `deployContractAsBlobs` method, also available on the `ContractFactory` should you want to use that directly.
 
-<<< @/../../docs-snippets/src/guide/contracts/deploying-contracts.test.ts#contract-setup-5{ts:line-numbers}
+<<< @/../../docs-snippets/src/guide/contracts/deploying-contracts.test.ts#blobs{ts:line-numbers}
 
-For a more comprehensive TypeScript-backed Fuel usage, learn how to [generate types from ABI](../fuels-cli/generating-types.md)
+In the above example, we also pass a `chunkSizeOverride` option to the deployment method. The SDK will attempt to chunk the contract to the most optimal about, however the transaction size can fluctuate and you can also be limited by request size limits against the node. By default we set an of 5%, however you can adjust this to suit your needs and ensure the transaction passes. It must be set to a value between 0 and 1.
```
petertonysmith94, line 60:
```suggestion
In the above example, we also pass a `chunkSizeOverride` option to the deployment method. The SDK will attempt to chunk the contract to the most optimal amount, however, the transaction size can fluctuate and you can also be limited by request size limits against the node. By default, we set a value for `chunkSizeOverride` of 0.05 (5%), however you can adjust this to suit your needs and ensure the transaction passes. It must be set to a value between 0 and 1.
```

packages/contract/src/contract-factory.ts
```diff
@@ -1,26 +1,33 @@
...
-  MAX_CONTRACT_SIZE,
-  getContractId,
-  getContractStorageRoot,
-  hexlifyWithPrefix,
-} from './util';
+import { getLoaderInstructions, getContractChunks } from './loader';
+import { getContractId, getContractStorageRoot, hexlifyWithPrefix } from './util';
+
+/** Amount of percentage override for chunk sizes in blob transactions */
+export const CHUNK_SIZE_OVERRIDE = 0.05;
```
petertonysmith94, line 30:
```suggestion
const CHUNK_SIZE_OVERRIDE = 0.05;
```

IMO, we don't need to export this constant.

packages/contract/src/contract-factory.ts
```diff
@@ -235,28 +408,64 @@ export default class ContractFactory {
...
+      ...options,
+    });
+  }
+
+  /**
+   * Get the maximum chunk size for deploying a contract by chunks.
+   */
+  private getMaxChunkSize(
+    deployOptions: DeployContractOptions,
+    chunkSizeOverride: number = CHUNK_SIZE_OVERRIDE
```
petertonysmith94, line 440: I'm not entirely convinced on the naming of this variable as override gives the impression that we're overriding an entire value. However, we're using this override to calculate an offset multiplier for the contract size threshold.

`CHUNK_SIZE_THRESHOLD` - this would be defaulted to `95%`, so in essence, the result of `override`:
```ts
const override = CHUNK_SIZE_THRESHOLD = (1 - chunkSizeOverride);
```

I am not entirely convinced with this naming either, but I'd do away with the inverse option (`1 - chunkSizeOverride`).

packages/fuel-gauge/src/contract-factory.test.ts
```diff
@@ -245,21 +253,232 @@ describe('Contract Factory', () => {
...
+  it('deploys a contract via blobs [configurables]', async () => {
+    using launched = await launchTestNode();
+
+    const {
+      wallets: [wallet],
+    } = launched;
+
+    const factory = new ContractFactory(
+      ConfigurableContractFactory.bytecode,
+      ConfigurableContract.abi,
```
petertonysmith94, line 426: Do we want to be testing that the `LargeContract` can be deployed with configurables?

packages/fuel-gauge/src/contract-factory.test.ts
```diff
@@ -245,21 +253,232 @@ describe('Contract Factory', () => {
...
+  it('deploys a contract via blobs [storage]', async () => {
+    using launched = await launchTestNode();
+
+    const {
+      wallets: [wallet],
+    } = launched;
+
+    const factory = new ContractFactory(
+      StorageTestContractFactory.bytecode,
+      StorageTestContract.abi,
```
petertonysmith94, line 466: Do we want to be testing that the `LargeContract` can be deployed with storage?

apps/docs-snippets/src/guide/contracts/deploying-contracts.test.ts
```diff
@@ -1,65 +1,110 @@
...
+    // #import { Provider, TESTNET_NETWORK_URL, Wallet, ContractFactory };
+    // #context import { WALLET_PVT_KEY } from 'path/to/my/env/file';
+    // #context import { bytecode, abi } from 'path/to/typegen/outputs';
+
+    const provider = await Provider.create(TESTNET_NETWORK_URL);
+    const wallet = Wallet.fromPrivateKey(WALLET_PVT_KEY, provider);
+    const factory = new ContractFactory(bytecode, abi, wallet);
+
+    // Deploy the contract as blobs
+    const { waitForResult, contractId, transactionId } = await factory.deployContractAsBlobs({
```
Dhaiwat10, line 93: Have we added a test that covers the functionality of the `deploy` function that automatically decides between a 'normal' deployment and a blob tx depending on the contract's size?

packages/fuel-gauge/src/contract-factory.test.ts
danielbate: IMO as we are enforcing blob deployment via `deployContractAsBlobs` it doesn't make much difference which contract is used. The configurable contract bytecode will still be configured and deployed as blob, then accessed by the loader contract which takes the blob ID. The only difference will be in the number of blobs, the loader contract loops through, the logic remains the same.

packages/fuel-gauge/src/contract-factory.test.ts
danielbate: As [above](https://github.com/FuelLabs/fuels-ts/pull/2827#discussion_r1713534682), the only difference is the number of blobs. This ensures that storage persists when deploying as blobs.

packages/contract/src/contract-factory.ts
danielbate: Removed the inverse and changed to `chunkSizeMultiplier`.

apps/docs-snippets/src/guide/contracts/deploying-contracts.test.ts
danielbate: Covered in the `fuel-gauge` contract factory tests but I've also added [this](https://github.com/FuelLabs/fuels-ts/pull/2827/commits/4cb5ff6ee44c396dcff9d638c8d58fbd432c5ff3) to check bytecode len.

packages/contract/src/contract-factory.ts
maschad: After discussing with @danielbate we decided we will plan for this in the future

https://github.com/FuelLabs/fuels-ts/issues/2917

packages/contract/src/contract-factory.ts
```diff
@@ -146,26 +155,190 @@ export default class ContractFactory {
...
+    // Deploy the loader contract via create tx
+    const { contractId, transactionRequest: createRequest } = this.createTransactionRequest({
+      bytecode: loaderBytecode,
+      ...deployOptions,
+    });
+    await this.fundTransactionRequest(createRequest, deployOptions);
+    const transactionResponse = await account.sendTransaction(createRequest);
 
     const waitForResult = async () => {
       const transactionResult = await transactionResponse.waitForResult<TransactionType.Create>();
```
nedsalk, line 344: Are you going to be putting the sending of blob transactions + loader contract into `waitForResult` as we discussed? 
For others, the idea is that deploying a large contract should behave the same as deploying a small contract in the sense that the end user shouldn't be blocked by transaction execution. For small contracts, we deploy the contract in a single transaction which the user can await with `waitForResult`. Based on the current implementation for large contracts, the user would be blocked by the sending and execution of all blob transactions and their call of `waitForResult` would be just on the last loader contract transaction. Ideally the `waitForResult` would also encompass the blob transactions as well so that, from the perspective of the user, it all behaves as a single transaction.

danielbate:
@maschad has also spun up #2917 for follow ups, should anyone find any issues that don't warrant blocking this work.

packages/contract/src/contract-factory.ts
danielbate: I've already added it as a follow point to #2917. I took a brief look the other day but had an issue with the loader contract `sendTransaction` call. Will try get some more time today.

arboleya: @nedsalk Good catch. This is a DX must follow up with the rest.

@danielbate If we can, we should ship this DX change with this PR.

maschad (dismissed):
Awesome job on this @danielbate

danielbate:
e2e tests on testnet are [succeeding](https://github.com/FuelLabs/fuels-ts/actions/runs/10373201280/job/28719459008?pr=2827#step:5:21) so marking this as unblocked and will be ready for merge.

packages/account/src/account.test.ts
```diff
@@ -307,13 +308,11 @@ describe('Account', () => {
     const response = await sender.transfer(receiver.address, 1, baseAssetId, {
       gasLimit: 10_000,
     });
-    await response.wait();
```
maschad, line 310: Why have all the balance checks in these tests been removed? They don't seem relevant to this PR

packages/contract/src/contract-factory.ts
danielbate: Correct, the whole deployment will fail. Any blobs already uploaded can be reused without additional cost though. I'll add blob retries to #2917.

packages/contract/src/contract-factory.ts
danielbate: Given testnet is ready I'm going to do it as follow up, I've got a PR but it's not perfect and requires some discussion.

packages/account/src/account.test.ts
danielbate: cc @Torres-ssf

Torres-ssf: @maschad The related tests were failing because the balance was being compared to pre-defined values. Since some gas instructions values changed, the fee changed, resulting in the test failing.

I've simplified all these tests as the balance validation part does not add any relevance to the test case.

packages/contract/src/contract-factory.ts
```diff
@@ -146,26 +155,190 @@ export default class ContractFactory {
...
+      : this.deployContract(deployOptions);
+  }
+
+  /**
+   * Deploy a contract with the specified options.
+   *
+   * @param deployOptions - Options for deploying the contract.
+   * @returns A promise that resolves to the deployed contract instance.
+   */
+  async deployContract<TContract extends Contract = Contract>(
```
arboleya, line 213: One of the breaking changes introduced in #2824 was the renaming of:
 - `deployContract()` to `deploy()`

The suffix is redundant given the class is called `ContractFactory`.

This PR reintroduces that method [name], but now it works differently. Aside from any confusion this can cause, I believe we should not reintroduce the `Contract` suffix to these method names. Did you explore other variations? Perhaps a middle ground:

 - `deploy()`
 - `deployAsXyz()`
 - `deployAsBlobs()`

arboleya: Nit - should we say `Blobs` or `Blob` (singular) as in "`Blob` TX" — a TX of the `Blob` type.

Just flushing ideas:
 - `deploy()`
 - `deployAsCreateTx()`
 - `deployAsBlobTx()`

danielbate: Agreed on singular. It's a little incorrect as we still submit a final create tx, but I think they are the best options so far.  Will give it till the morning to let others weigh in.

packages/account/src/account.test.ts
maschad: Thanks for clarifying

maschad (approved):
Great job @Torres-ssf and @danielbate

packages/contract/src/contract-factory.ts
arboleya: Would this require introducing new breaking changes later?

packages/contract/src/contract-factory.ts
petertonysmith94: > Nit - should we say `Blobs` or `Blob` (singular) as in "`Blob` TX" — a TX of the `Blob` type.
> 
> Just flushing ideas:
> 
> * `deploy()`
> * `deployAsCreateTx()`
> * `deployAsBlobTx()`

This seems more descriptive to me.

danielbate: SGTM, will change.

packages/contract/src/contract-factory.ts
nedsalk: @arboleya It doesn't seem like it would require breaking changes because we'd just put the logic inside of the `waitForResult` that's already returned from the method.

danielbate: My current problem is with returning the transaction ID for the create tx. Putting everything inside `waitForResult` means that we don't that ID until we've completed all the blob TXs and then funded the create TX. I should have some time this afternoon. You're welcome to take a look if you have time @nedsalk. I've got a draft in #2929.

nedsalk: @danielbate I took a look and, from what it looks like, this is a problem we won't be able to solve without partially (or fully?) implementing #2678. We'd have to know upfront which UTXOs the blob TXs will spend and the new UTXOs they'll create up until the last blob TX's UTXOs which we'd then use for the create TX - something like that. Right now the only thing we can do is document for the users that deploying contracts as blobs might hang a bit longer on the `await`.

danielbate: I'd considered making transaction ID a promise but obviously a breaking change. For now I'll add additional docs. Thanks for taking a look.

petertonysmith94 (dismissed):

arboleya (approved):

petertonysmith94 (approved):

Torres-ssf (approved):

nedsalk (approved):

PR #2851, opened 2024-07-29
chore: add infra to support multiple `create fuels` templates

### LINKED ISSUE #2837: Add support for multiple templates in `create fuels`
Currently, we have only one template using NextJS.

Before we can add more templates, we need the basic infra to support N templates.

Then we'll be able to:

```bash
npm create fuels -- --template nextjs
npm create fuels -- --template <template-name>
```

The default version should continue working as it is now, using NextJS:

```bash
npm create fuels
```

This issue is just about pluralizing the template options but adds nothing new.

Follow-up issue:
 - https://github.com/FuelLabs/fuels-ts/issues/2838

Dhaiwat10:
- Closes #2837 

# Summary

This PR adds a new flag to the `create fuels` CLI which will let users pass in a template name of their choice like so:

```
npm create fuels  -- --template nextjs
```

Do note that this PR does not add any new templates, but rather the infrastructure to support multiple of them in the future.

# Checklist


packages/create-fuels/src/cli.ts
```diff
@@ -36,20 +37,36 @@ function writeEnvFile(envFilePath: string) {
   writeFileSync(envFilePath, newFileContents);
 }
 
+const doesTemplateExist = (templateName: Template): boolean => templates.has(templateName);
+
 export const runScaffoldCli = async ({
   program,
-  templateName = 'nextjs',
+  explicitTemplateName, // Only to be used for testing. The actual template name is passed in as a CLI arg.
```
petertonysmith94, line 44: Could we not pass this through the args for testing?

packages/create-fuels/src/cli.ts
```diff
@@ -36,20 +37,36 @@ function writeEnvFile(envFilePath: string) {
...
+
+  if (!doesTemplateExist(templateOfChoice) && checkIfTemplateExists) {
+    error(`Template '${templateOfChoice}' does not exist.`);
+    log();
+    log('Available templates:');
+    for (const template of templates) {
+      log(`  - ${template}`);
+    }
+    process.exit(1);
+  }
```
petertonysmith94, line 66: Instead of `checkIfTemplateExists` for testing purposes, could we not mock this function in the test?

packages/create-fuels/src/cli.ts
```diff
@@ -86,7 +103,7 @@ export const runScaffoldCli = async ({
 
   await mkdir(projectPath);
 
-  const templateDir = join(__dirname, '..', 'templates', templateName);
+  const templateDir = join(__dirname, '..', 'templates', templateOfChoice);
```
petertonysmith94, line 106: Maybe a future improvement, but we could make the bootstrapping based [logic](https://github.com/FuelLabs/fuels-ts/pull/2851/files#diff-a1150dd0573a07b2a93609a4fc4fb055d41c138716e3c27515bf25cc23537787R106-R137) live in the template.

packages/create-fuels/src/lib/setupProgram.ts
```diff
@@ -22,6 +26,7 @@ export const setupProgram = () => {
     .option('--bun', 'Use bun to install dependencies')
     .option('--verbose', 'Enable verbose logging')
     .option('--no-install', 'Do not install dependencies')
+    .option('--template <template>', 'Specify a template to use')
```
petertonysmith94, line 29: Could we add a `defaultValue` here?

```suggestion
    .option('--template <template>', 'Specify a template to use', 'nextjs')
```

packages/create-fuels/src/cli.ts
arboleya: Exactly, we must keep the source free from test code exceptions.

packages/create-fuels/src/cli.ts
arboleya: Yes, some split of responsibility may be necessary.

Perhaps this can be better evaluated during the Vite addition.

petertonysmith94: I would agree, tackle it when we have an example at hand.

packages/create-fuels/src/lib/setupProgram.ts
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/2851/commits/4f1720a14e2b16bc59b6d9c14f634e8d4aade15b

packages/create-fuels/src/cli.ts
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/2851/commits/5e4e9dfb92db96615a8632072d3c6fdcbf7e5c70

packages/create-fuels/src/cli.ts
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/2851/commits/b6c15e941032840637832d82f29eaf6b7b25d5dc

packages/create-fuels/test/cli.test.ts
```diff
@@ -16,11 +19,20 @@ import { generateArgv } from './utils/generateArgs';
 import { mockLogger } from './utils/mockLogger';
 import { filterOriginalTemplateFiles, getAllFiles } from './utils/templateFiles';
 
+vi.mock('../src/lib/doesTemplateExist', async () => {
```
petertonysmith94, line 22: Did you try `spyOn` this method?

Nitty but I quite like the [`mockAll`](https://github.com/FuelLabs/fuels-ts/blob/6eb24007258ed946c177369cba648b5bd6ad2535/packages/fuels/src/cli/commands/deploy/index.test.ts#L16-L29) function in tests that we seem to be adhering to.

packages/create-fuels/test/utils/generateArgs.ts
```diff
@@ -1,18 +1,26 @@
-export const generateArgs = (projectName?: string, packageManager: string = 'pnpm'): string[] => {
+export const generateArgs = (
+  projectName?: string,
+  packageManager: string = 'pnpm',
+  template?: string
+): string[] => {
```
petertonysmith94, line 5: May be worth making this a single parameter with an object, just to avoid confusion in tests

petertonysmith94 (approved):
Nitty comments :)

danielbate (approved):

Torres-ssf (approved):

packages/create-fuels/test/cli.test.ts
Dhaiwat10: @petertonysmith94 `spyOn` only lets you _inspect_ a function and not actually mock it if I'm not wrong. In this case, I wanted to actually mock the implementation/return values

arboleya: The `vi.mock` call is hoisted, and in most cases that's not what you want.

https://github.com/FuelLabs/fuels-ts/blob/fe06f0de2ebc4cdbc83614b085568db1f0b45eae/packages/create-fuels/src/lib/tryInstallFuelUp.test.ts#L36-L38

packages/create-fuels/test/utils/generateArgs.ts
arboleya: ➕ for named params.

packages/create-fuels/test/cli.test.ts
Dhaiwat10: Oh I wasn't aware of the hoisting @arboleya, thanks for the heads-up. In that case, maybe we should use `vi.fn().mockImplementation` then

According to Vitest docs, `spyOn` will not let you change the behaviour of any functions: https://vitest.dev/guide/mocking.html#functions 

I remember that in a recent PR, I spotted that some of our mock implementations were using this pattern and as a result of that, the functions weren't being mocked at all. This might be the case for the example you have quoted as well

packages/create-fuels/test/utils/generateArgs.ts
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/2851/commits/3eac3a1fce009ce7564eb7c0e5ce789aa5447b7b

packages/create-fuels/test/cli.test.ts
arboleya: To mock, you need to call `spyOn` followed by `mockImplementation`.

Dhaiwat10: Done. Apologies for the confusion

Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/2851/commits/cf0ab4c845f88df3dad60db2ce3eb2a3f1f3071f

petertonysmith94 (approved):

nedsalk (approved):

danielbate (approved):

Issue #2828, opened 2024-07-24
Stop using Sparse Merkle Tree


arboleya:
We may have `fuel-core` removing the use of SMT.

I know we have a package:
 - [fuels-ts/tree/master/packages/merkle](https://github.com/FuelLabs/fuels-ts/tree/master/packages/merkle)

Which was ported from:
 - [fuel-vm/tree/master/fuel-merkle](https://github.com/FuelLabs/fuel-vm/tree/master/fuel-merkle)

A quick search led me here. I'm not sure about the state of this port, but it can be a good starting point to look into.

https://github.com/FuelLabs/fuels-ts/blob/b1dbe42361e52af77f124ec8c633e86885a790be/packages/account/src/predicate/predicate.ts#L65

---

Related:
 - https://github.com/FuelLabs/fuel-core/issues/2042

Torres-ssf:
As mentioned in the [fuel-core issue](https://github.com/FuelLabs/fuel-core/issues/2042), the use of SMT has only been removed from the `balanceRoot` and `stateRoot` fields, which are part of the [OutputContract](https://docs.fuel.network/docs/specs/tx-format/output/#outputcontract).

These fields are exclusively set by the Client/VM. In the TypeScript SDK, we simply de/serialize these values from a processed transaction. Therefore, we can safely close this issue.

PR #2852, opened 2024-07-29
chore: remove use of sparse merkle tree

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/2828

# Release notes

In this release, we:

- To be defined

# Summary

The implementation and use of the SMT (Sparse Merkle Tree) were removed. This aligns with the changes proposed [here](https://github.com/FuelLabs/fuel-core/issues/2042). At the SDK level, this change affects only the calculation of the [contract ID](https://docs.fuel.network/docs/specs/identifiers/contract-id/), as the contract storage slots root was previously calculated using SMT. Since no alternative tree structure will replace SMT at this time, the contract storage slots root will now have the value of an empty tree node.

# Checklist


Torres-ssf:
This is blocked by the next `fuel-core` release

Torres-ssf:
Closing this due to a misunderstanding on my part. The use of SMT has only been removed from the `balanceRoot` and `stateRoot` fields, which are part of the [OutputContract](https://docs.fuel.network/docs/specs/tx-format/output/#outputcontract).

These fields are exclusively set by the Client/VM. In the TypeScript SDK, we simply de/serialize these values from a processed transaction.

PR #2854, opened 2024-07-29
feat: improve typegen bytecode compression

### LINKED ISSUE #1462: Typegen - Improve `bytecode` compression
Today, we read binaries after `forc build` and `hexilify` them.

The `hex` string is then embedded within the `typegen` generated Typescript factory class.

However, @luizstacio reported achieving a 70% size reduction using the method below:
 1. Read binaries
 1. Convert to Byte Array
 1. Gzip the Byte Array
 1. Encode the result using Base64

The reverse must be done for decoding, and the reduction is _considerable_.

We should look into that.

Refs:
 - https://www.npmjs.com/package/pako
 - https://www.npmjs.com/package/fflate

Dhaiwat10:
Closes #1462 

# Release notes

In this release, we:

- Improved the bytecode compression for Typegen, making files 40%+ smaller

# Summary

This PR introduces a compression strategy for typegen contract artifacts. We are achieving this by gzipping and base64-encoding the bytecode before we pack it into the typegen file that is shipped to our users.

# Checklist


packages/utils/src/utils/bytecode.ts
```diff
@@ -0,0 +1,20 @@
+import type { BytesLike } from '@fuel-ts/interfaces';
+import { gzipSync, gunzipSync } from 'fflate';
```
arboleya, line 2: Will this `fflate` lib work in browser envs?

How much kbs does it add to the final bundle?

Dhaiwat10: Yes, I have ensured this by running the unit tests for these functions in the browser environment as well.

arboleya: Great.

Do you know how much kbs it adds to the final bundle?

It may be valuable information.

Dhaiwat10: I'm not sure, investigating.

packages/abi-typegen/src/utils/assembleContracts.test.ts
```diff
@@ -7,36 +7,65 @@ import { ProgramTypeEnum } from '../types/enums/ProgramTypeEnum';
...
-/**
- * @group node
- */
-describe('assembleContracts.ts', () => {
-  function mockAllDeps() {
-    vi.resetAllMocks();
-
-    const renderCommonTemplate = vi
-      .spyOn(renderCommonTemplateMod, 'renderCommonTemplate')
-      .mockReturnValue('');
```
Dhaiwat10, line 19: These mocks were not working as intended. They were still resulting in the tests using the actual functions instead of a mock implementation. I had to fix this as these tests were now failing after me adding some logic to one of the functions

packages/utils/src/utils/bytecode.ts
Dhaiwat10: Bundle size comparisons:

Before: 
```
308K	dist/main.js
 92K	dist/main.js.gz
 ```
 
After:
```
316K	dist/main.js
 96K	dist/main.js.gz
 ```

nedsalk: We'll be adding 4 kB to _all_ applications using our sdk, regardless if they're deploying contracts or not. Given that contracts are deployed once and then just interacted with in a production app, it seems to me that it's wiser to leave things as they are. It's the users' fault if they add the contract bytecode into their apps even though the contract is already deployed and they can interact with it via just its id and abi.

arboleya: In the case of contracts, the bytecode is optional.

But it is required for scripts and predicates.

Torres-ssf (approved):

packages/utils/src/utils/bytecode.ts
arboleya: Should we consider using native modules for this task?

NodeJS refs:
 - https://stackoverflow.com/a/63007331

Browser refs:
 - https://stackoverflow.com/a/247261
 - https://stackoverflow.com/a/68829631

This API needs to be built like the crypto package, in a hybrid way targeting NodeJS + Browser, using native modules of each. I guess this could reduce the size increase to almost zero.

Dhaiwat10: @arboleya that sounds fair. Do you reckon we'll need to create a new package for this then since we'll need two different bundles for the two environments? Or should I just tweak the `utils` package?

arboleya: I guess you could keep the utility in the `utils` package and configure it like the `crypto` so you can have individual entry points for NodeJS and the Browser. cc @Torres-ssf

Dhaiwat10: @arboleya I looked into this and while I found a straightforward solution for node, unfortunately there doesn't seem to be any way to do the compression in browser using Native APIs without it being async. This won't work for our use case.

We could re-write the whole algorithm ourselves just for the browser but that can turn out to be tricky. It's not exactly straightforward as it involves two or three different 'steps', and even then our custom rewrite might not turn out to be that efficient.

I think we should take the 4kb hit and go with the external library.

packages/abi-typegen/src/utils/assembleContracts.test.ts
arboleya: This is weird. 👀

Dhaiwat10:
Closing in favor of #2897

Issue #596, opened 2022-11-12
Fix `forc-bin` scripts


arboleya:
This change here breaks this script since the field `forcVersion` doesn't have a comma in the end.

_Originally posted by @AlicanC in https://github.com/FuelLabs/fuels-ts/pull/543#discussion_r1020557192_

PR #2856, opened 2024-07-29
feat!: adding `abi` transpiler

### LINKED ISSUE #2313: Adapt ABI format to new specification
Spec PR:
 - https://github.com/FuelLabs/fuel-specs/pull/596

Dependent on:
 - https://github.com/FuelLabs/sway/issues/5951

arboleya:
- Closes #2313

# Release notes

In this release, we:
- Upgraded `forc` to [v0.63.0](https://github.com/FuelLabs/sway/releases/tag/v0.63.0) & [v0.63.1](https://github.com/FuelLabs/sway/releases/tag/v0.63.1)
- Added support for the new ABI spec _([#596](https://github.com/FuelLabs/fuel-specs/pull/596), [#599](https://github.com/FuelLabs/fuel-specs/pull/599))_

# Summary

Given the amount of structural changes we had in the new ABI and the impact it will have on the codebase, and provided the fact we have a lot of technical debt to clear regarding ABI `coder` and `typegen` packages ([#1795](https://github.com/FuelLabs/fuels-ts/issues/1795), [#2346](https://github.com/FuelLabs/fuels-ts/issues/2346), [#2686](https://github.com/FuelLabs/fuels-ts/issues/2686)), this proposal aim at releasing the pressure of the moment by transpiling the old format into the new one, while still:

 - Supporting the new ABI spec
 - Introducing the **_least_** amount of changes possible
 - Keeping all tests and coverage intact

Transpilation occurs internally between public and private APIs.

- Transpilation is an internal detail on the `Interface` level
- Typegen includes the new abi format in its generated code but doesn't transpile it
- All public-facing APIs use the new spec ABI under `JsonAbi`, and the old one is available as `JsonAbiOld`
- `AbiCoder` is no longer exported, and its methods moved to `Interface` via `encodeType`/`decodeType` (https://github.com/FuelLabs/fuels-ts/pull/2862/commits/ee734b988e614a2025fc5c353408fc40ba702482)
- We removed the `Interface.findTypeById` method in https://github.com/FuelLabs/fuels-ts/pull/2862/commits/157bdba5e1d6b06427db4fdd3f4826904b5fe97e because the concept of a type in the new ABI differs from the old one. Whoever wants to get a specific concrete or metadata type can do so directly from the ABI they have on the `Interface.jsonAbi` property

## Special thanks ❤️ 

Huge shoutout to @nedsalk, who was crucial for this work and ensured **_everything_** around this was working as expected everywhere and finalized all the heavy lifting around public vs private APIs via:
 - #2862

Also, to @petertonysmith94, for the thorough reviews and additional bug-fixing via:
 - https://github.com/FuelLabs/fuels-ts/pull/2870

# Breaking Changes

## New ABI spec

The SDK now adheres to the new specs introduced via:
 - https://github.com/FuelLabs/fuel-specs/pull/596
 - https://github.com/FuelLabs/fuel-specs/pull/599

Check these out to understand all its changes.

## `AbiCoder` is no longer available to import

The class `AbiCoder` is no longer exported, and the way to do encoding and decoding of specific types is now via the `Interface.encodeType` and `Interface.decodeType` methods:

```ts
// before
const abi = yourAbi;
const functionArg = abi.functions.inputs[0];

const encoded = AbiCoder.encode(abi, functionArg, valueToEncode);
const decoded = AbiCoder.decode(abi, functionArg, valueToDecode, 0);
```

```ts
// after
import { Interface } from 'fuels';

const abi = yourAbi;
const functionArg = abi.functions.inputs[0];

const abiInterface = new Interface(abi);

const encoded = abiInterface.encodeType(functionArg.concreteTypeId, valueToEncode);
const decoded = abiInterface.decodeType(functionArg.concreteTypeId, valueToDecode);
```

## `Interface.findTypeById` has been removed

Previously, you could get a type from the ABI via the `Interface.findTypeById`. This method was removed after introducing the new abi specification because the concept of a _type_ has been split into concrete and metadata types. If you want a specific type, you can get it directly from the ABI.

```ts
// before
const abiInterface = new Interface(abi);

// internally this method searched the abi types:
// abi.types.find(t => t.typeId === id);
const type = abiInterface.findTypeById(id);
```

```ts
// after
import { Interface } from 'fuels';

// search the types on the abi directly
const concreteType = abi.concreteTypes.find(ct => ct.concreteTypeId === id);
const metadataType = abiInterface.jsonAbi.metadataTypes.find(mt => mt.metadataTypeId === id);
```

## `JsonAbiArgument` is no longer exported

The `JsonAbiArgument` type isn't part of the new ABI spec _([#596](https://github.com/FuelLabs/fuel-specs/pull/596), [#599](https://github.com/FuelLabs/fuel-specs/pull/599))_ as such so we stopped exporting it. Its closest equivalent now would be a concrete type because it fully defines a type.

```ts
// before
const arg: JsonAbiArgument = {...};
```

```ts
// after
import { Interface } from 'fuels';

type ConcreteType = JsonAbi["concreteTypes"][number]
const arg: ConcreteType = {...};
```

# Checklist


internal/forc/VERSION
```diff
@@ -1 +1 @@
-0.62.0
+git:esdrubal/abi_changes
```
arboleya, line 1: > [!IMPORTANT]
> Replace the branch name with the soon-to-be-released sway version.

packages/abi-coder/src/encoding/strategies/getCoderV1.ts
```diff
@@ -122,21 +123,26 @@ export const getCoder: GetCoderFn = (
...
 
-  const enumMatch = enumRegEx.exec(resolvedAbiType.type)?.groups;
+  const enumMatch = enumRegEx.test(resolvedAbiType.type);
   if (enumMatch) {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const [name] = resolvedAbiType.type.match(lastNameRegExMatch)!;
+
     const coders = getCoders(components, { getCoder });
 
     const isOptionEnum = resolvedAbiType.type === OPTION_CODER_TYPE;
```
petertonysmith94, line 141:
```suggestion
    const isOptionEnum = resolvedAbiType.type.test(optionCoderRegex);
```

Do we want to be testing against the regex here?

packages/abi-typegen/src/runTypegen.ts
```diff
@@ -52,10 +53,13 @@ export function runTypegen(params: IGenerateFilesParams) {
...
+    const raw = readFileSync(filepath, 'utf-8');
+    const transpiled = JSON.stringify(transpileAbi(JSON.parse(raw)), null, 2);
     const abi: IFile = {
       path: filepath,
-      contents: readFileSync(filepath, 'utf-8'),
+      contents: transpiled,
     };
+
     return abi;
   });
```
petertonysmith94, line 64: Might it be worth adding a try/catch around this operation?

packages/abi-coder/src/utils/constants.ts
```diff
@@ -10,18 +10,24 @@ export const RAW_PTR_CODER_TYPE = 'raw untyped ptr';
...
-export const B512_CODER_TYPE = 'struct B512';
-export const OPTION_CODER_TYPE = 'enum Option';
-export const VEC_CODER_TYPE = 'struct Vec';
-export const BYTES_CODER_TYPE = 'struct Bytes';
-export const STD_STRING_CODER_TYPE = 'struct String';
+export const B512_CODER_TYPE = 'struct std::b512::B512';
+export const OPTION_CODER_TYPE = 'enum std::option::Option';
+export const VEC_CODER_TYPE = 'struct std::vec::Vec';
+export const BYTES_CODER_TYPE = 'struct std::bytes::Bytes';
+export const STD_STRING_CODER_TYPE = 'struct std::string::String';
```
petertonysmith94, line 17: Are these type formats the same for both `debug` and `release`?

packages/abi-coder/src/encoding/strategies/getCoderV1.ts
arboleya: No, at this point we are only interested in the standard `Option`.

packages/abi-coder/src/utils/constants.ts
arboleya: Yes, because now `--json-abi-with-callpaths` flag became the default behavior.

packages/abi-typegen/src/runTypegen.ts
arboleya: We could, but we never had it in place, and it always worked. It may be redundant and overly cautious since filepaths are extracted [from the disk](https://github.com/FuelLabs/fuels-ts/pull/2856/files/3b62ef47a32129ae8b6a9d8296435bb1109051c3#diff-1c28ec9246fc42304c8c01640e13f3a55f4fa8b7cb8672c36f936a4e08344da5R42) right before reading these files, so they will always exist.

petertonysmith94: Completely understandable.

packages/abi-coder/src/utils/constants.ts
petertonysmith94: Good to know - thanks.

Do we also want to update the associated typegen `swayType` (e.g. [B512Type](https://github.com/FuelLabs/fuels-ts/blob/3b62ef47a32129ae8b6a9d8296435bb1109051c3/packages/abi-typegen/src/abi/types/B512Type.ts#L4-L5))?

arboleya: Rephrasing: Type formats have never differed between `debug` and `release`, only when using the `--json-abi-with-callpaths` flag, which has now become the default behavior.

petertonysmith94:
All tests are passing for me locally.

petertonysmith94 (approved):

petertonysmith94 (commented):
Just remembered the connectors and the ABI related functions. Is this going to cause problems with integrating in the connectors?

https://github.com/FuelLabs/fuels-ts/blob/6793b4b634799f153f8a01a160705eab9d850b19/packages/account/src/connectors/fuel-connector.ts#L69-L77

arboleya:
@petertonysmith94 I don't have context - can you please elaborate on the issue you're concerned about?

packages/abi-coder/src/encoding/coders/OptionCoder.ts
```diff
@@ -25,7 +25,7 @@ export class OptionCoder<TCoders extends Record<string, Coder>> extends EnumCode
   }
 
   private toOption(output?: DecodedValueOf<TCoders>): Option<unknown> {
-    if (output && 'Some' in output) {
+    if (output && Object.hasOwn(output, 'Some')) {
```
petertonysmith94, line 28: I believe that the following check is causing the issues.
https://github.com/FuelLabs/fuels-ts/blob/45f3ef2bac2999f876e1390f0670ac7df5737f8b/packages/abi-coder/src/encoding/coders/EnumCoder.ts#L45-L47

We've changed constant value for `OPTION_CODER_TYPE` to `enum std::option::Option` (in this [PR](https://github.com/FuelLabs/fuels-ts/pull/2856)).

https://github.com/FuelLabs/fuels-ts/blob/45f3ef2bac2999f876e1390f0670ac7df5737f8b/packages/abi-coder/src/utils/constants.ts#L14

The [type](https://github.com/FuelLabs/fuels-ts/blob/929ac2d33d53254d2e7fa4201955686f4e17d540/packages/abi-coder/src/encoding/coders/AbstractCoder.ts#L34) of the coder has remained the same (`enum Option`) as we're only pulling off the [last name](https://github.com/FuelLabs/fuels-ts/blob/45f3ef2bac2999f876e1390f0670ac7df5737f8b/packages/abi-coder/src/encoding/strategies/getCoderV1.ts#L130-L150) from the ABI type. This leads to the `isNativeEnum` method now returns `true` rather than `false` in the instance of an `Option`. This then leads to `None` being returned as the decoded value, rather than the correct result of `{ None: undefined }`.

---

We could change the [`lastNameRegExMatch`](https://github.com/FuelLabs/fuels-ts/blob/45f3ef2bac2999f876e1390f0670ac7df5737f8b/packages/abi-coder/src/utils/constants.ts#L29) to obtain the full namespace type so for case of an Option, it would get `std::option::Option`.

```ts
export const lastNameRegExMatch = /([^\s]+)$/m;
```

We could also make two constants:

```ts
const `RAW_ABI_OPTION_TYPE` = `enum std::option::Option`; // Can be used in the ABI
const `OPTION_CODER_TYPE` = `enum Option`;
```

petertonysmith94:
> @petertonysmith94 I don't have context - can you please elaborate on the issue you're concerned about?

@arboleya Upon reflection of #2862 I realised this is a non-issue. The new format will be public facing and the old format will be internal.

packages/abi-coder/src/encoding/coders/OptionCoder.ts
arboleya: Got it. You can try it out and see how it goes.

We can circle back later.

arboleya: Thanks for the fix via:
 - https://github.com/FuelLabs/fuels-ts/pull/2870

.changeset/odd-horses-cheer.md
```diff
@@ -0,0 +1,9 @@
+---
+"@fuel-ts/abi-typegen": patch
+"@fuel-ts/abi-coder": patch
+"@fuel-ts/versions": patch
+"@fuel-ts/utils": patch
+"@internal/forc": patch
+---
+
+feat: adding `abi` transpiler
```
arboleya, line 9:
```suggestion
feat!: adding `abi` transpiler
```

.changeset/odd-horses-cheer.md
```diff
@@ -0,0 +1,9 @@
+---
+"@fuel-ts/abi-typegen": patch
+"@fuel-ts/abi-coder": patch
+"@fuel-ts/versions": patch
+"@fuel-ts/utils": patch
+"@internal/forc": patch
```
arboleya, line 6:
```suggestion
"@fuel-ts/abi-typegen": minor
"@fuel-ts/abi-coder": minor
"@fuel-ts/versions": minor
"@fuel-ts/utils": minor
"@internal/forc": minor
```

nedsalk:
This PR will also add support for `forc 0.63.0` so you can add that to the release notes.

internal/forc/VERSION
arboleya: > [!NOTE]
> This PR will also upgrade `forc` to the upcoming version `v0.63.0`.
> 
> **TODO**: Add mention to release notes when replacing the branch with the version.

packages/abi-coder/src/encoding/coders/OptionCoder.test.ts
```diff
@@ -7,7 +7,7 @@ import { VoidCoder } from './VoidCoder';
  * @group browser
  */
 describe('OptionCoder', () => {
-  const coder = new OptionCoder('Option', {
+  const coder = new OptionCoder('std::option::Option', {
```
danielbate, line 10: Is `std::option::Option` representing the name or type here. I've always found the combination of both name and type for the type coder classes a bit confusing, we should probably settle on one. Probably a point for #1795.

packages/utils/src/utils/transpileAbi.ts
```diff
@@ -0,0 +1,152 @@
...
+    encoding: abi.encodingVersion,
+    types,
+    functions,
+    loggedTypes,
+    messagesTypes: abi.messagesTypes,
+    configurables,
+  };
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return transpiled as any;
```
danielbate, line 151: Can this now be cast as `JsonABI`?

danielbate (commented):
Nice approach

packages/abi-coder/src/encoding/coders/OptionCoder.test.ts
arboleya: Not sure I follow your point, but for our context, I guess it's both because callpaths are now enabled by default.

Good to have as a checkpoint on #1795, as you said.

packages/utils/src/utils/transpileAbi.ts
arboleya: Yep, this script was initially a typeless `.mjs`.

However, the new ABI interface is in another PR and if we were to apply typings to the transpiler, it should go there:
 - https://github.com/FuelLabs/fuels-ts/pull/2862

packages/abi-coder/src/encoding/coders/OptionCoder.test.ts
nedsalk: After working with this `name` parameter in #2826 I realized that it does nothing and can be deleted.

packages/utils/src/utils/transpileAbi.ts
nedsalk: In that PR I just cast it wherever I use it:
```ts
const transpiled = transpileAbi(jsonAbi) as JsonAbi
```

The file has `// @ts-nocheck` in it and removing it brings about a lot of compilation errors. There wouldn't be any real benefit to fully typing it because we won't be using it anywhere else nor touching it except to delete it, so I'd just leave it as it is.

arboleya: Yep, fine by me.

packages/abi-coder/src/encoding/strategies/getCoderV1.ts
```diff
@@ -126,21 +127,26 @@ export const getCoder: GetCoderFn = (
     return new VecCoder(itemCoder as Coder);
   }
 
-  const structMatch = structRegEx.exec(resolvedAbiType.type)?.groups;
+  const structMatch = structRegEx.test(resolvedAbiType.type);
   if (structMatch) {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
maschad, line 132: Just as a suggestion to avoid the _no-non-null assertion_ rule 

```ts
function getFirstMatch(str: string, regex: RegExp): string | null {
  const match = str.match(regex);
  return match ? match[0] : null;
}

const structMatch = structRegEx.test(resolvedAbiType.type);
if (structMatch) {
  const name = getFirstMatch(resolvedAbiType.type, fullNameRegExMatch);
  if (name) {
    const coders = getCoders(components, { getCoder });
    return new StructCoder(name, coders);
  }
}
```

packages/abi-typegen/src/utils/transpile-abi.ts
```diff
@@ -0,0 +1,154 @@
+/* eslint-disable no-restricted-globals */
```
maschad, line 1: Duplicate files like this are all the more reason for https://github.com/FuelLabs/fuels-ts/issues/2346 which I think should be a part of the checklist in https://github.com/FuelLabs/fuels-ts/issues/2493

packages/create-fuels/test/cli.test.ts
```diff
@@ -19,7 +19,7 @@ import { filterOriginalTemplateFiles, getAllFiles } from './utils/templateFiles'
 /**
  * @group node
  */
-describe('CLI', () => {
+describe('CLI', { timeout: 15_000 }, () => {
```
maschad, line 22: Why did the timeout need to be increased here?

maschad (commented):
Nice work on this @arboleya 🚀  @nedsalk and @petertonysmith94  I've left some suggestions. 

This is perhaps tangential to this PR but my primary concern is exporting a class called `Interface` which is a reserved word and I feel it can easily lead to confusion for consumers, I would prefer it was called `AbiCoderInterface` , but 

```ts
import { Interface } from 'fuels';
``` 
feels a bit odd to me as a consumer, especially if I am working with multiple interfaces from other libraries.

Torres-ssf (approved):
@arboleya Awesome job 🚀

packages/abi-typegen/src/utils/transpile-abi.ts
maschad: Noted in [here](https://github.com/FuelLabs/fuels-ts/issues/2493#issuecomment-2269946371)

packages/abi-coder/src/encoding/strategies/getCoderV1.ts
arboleya: For accessing array indexes conditionally, you can simply:

```ts
resolvedAbiType.type.match(fullNameRegExMatch)?.[0]
```

https://github.com/FuelLabs/fuels-ts/pull/2856/commits/5621ccfd433f58e73f9188fb03f05088223ca57b

packages/abi-coder/src/encoding/strategies/getCoderV1.ts
```diff
@@ -126,21 +127,26 @@ export const getCoder: GetCoderFn = (
...
     const coders = getCoders(components, { getCoder });
-    return new StructCoder(structMatch.name, coders);
+    return new StructCoder(name, coders);
   }
 
-  const enumMatch = enumRegEx.exec(resolvedAbiType.type)?.groups;
+  const enumMatch = enumRegEx.test(resolvedAbiType.type);
   if (enumMatch) {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const [name] = resolvedAbiType.type.match(fullNameRegExMatch)!;
```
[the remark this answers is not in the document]
arboleya: https://github.com/FuelLabs/fuels-ts/pull/2856/commits/5621ccfd433f58e73f9188fb03f05088223ca57b

packages/abi-coder/src/utils/transpile-abi.ts
```diff
@@ -0,0 +1,154 @@
+/* eslint-disable no-restricted-globals */
```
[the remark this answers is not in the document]
arboleya: I appreciate the attempt; as you may guess, it was a `.mjs` file at its inception. 🙂 

The correct solution, typing-wise, would be to strictly plug in the old/new ABI interfaces. We discussed that in https://github.com/FuelLabs/fuels-ts/pull/2862 and agreed it wasn't worth the trouble. This is temporary by design, works fine, and any extra work will go to waste. That's the same reason this file has no tests. Everything else _is_ the test for it.

packages/abi-typegen/src/types/interfaces/JsonAbiNew.ts
```diff
@@ -0,0 +1,101 @@
+/**
+ * Types for Fuel JSON ABI Format as defined on:
+ * https://github.com/FuelLabs/fuel-specs/blob/master/src/abi/json-abi-format.md
+ */
+export interface JsonAbi {
+  readonly specVersion: string;
+  readonly encodingVersion: string;
+  readonly programType: string;
```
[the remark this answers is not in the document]
arboleya: It makes sense, but nothing uses this yet, and it goes a bit beyond the scope of this PR.

Ultimately, it will probably be a shared `enum ProgramType` between all-things ABI.

packages/create-fuels/test/cli.test.ts
arboleya: I was getting timeouts consistently.

.changeset/odd-horses-cheer.md
```diff
@@ -0,0 +1,9 @@
+---
+"@fuel-ts/abi-typegen": minor
+"@fuel-ts/abi-coder": minor
+"@fuel-ts/versions": minor
+"@fuel-ts/utils": minor
+"@internal/forc": minor
+---
```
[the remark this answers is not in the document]
arboleya: Good catch.

arboleya:
@maschad I agree with the sentiment about the `Interface` class name, but that's for the refactoring.

maschad (dismissed):
Nice work

.changeset/odd-horses-cheer.md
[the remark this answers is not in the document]
arboleya: https://github.com/FuelLabs/fuels-ts/pull/2856/commits/1e4996f8be8685fbaad4757333f0d11713328110

maschad (approved):

packages/abi-coder/src/types/JsonAbiNew.ts
```diff
@@ -0,0 +1,101 @@
...
+  readonly loggedTypes: readonly LoggedType[];
+  readonly messagesTypes: readonly MessageType[];
+  readonly configurables: readonly Configurable[];
+}
+
+export interface ConcreteType {
+  readonly type: string;
+  readonly concreteTypeId: string;
+  readonly metadataTypeId?: number;
+  readonly typeArguments?: readonly string[];
```
sdankel, line 21: Is this the same as the old `typeParameters`?

packages/abi-coder/src/types/JsonAbiNew.ts
```diff
@@ -0,0 +1,101 @@
...
+  readonly concreteTypes: readonly ConcreteType[];
+  readonly metadataTypes: readonly MetadataType[];
+  readonly functions: readonly AbiFunction[];
+  readonly loggedTypes: readonly LoggedType[];
+  readonly messagesTypes: readonly MessageType[];
+  readonly configurables: readonly Configurable[];
+}
+
+export interface ConcreteType {
+  readonly type: string;
```
sdankel, line 18: Is this the human readable name with the full path? 

It would be nice if there were a built in helper function for getting the name without the path (e.g. using a custom type rather than string).

packages/abi-coder/src/types/JsonAbiNew.ts
```diff
@@ -0,0 +1,101 @@
...
+  readonly components?: readonly Component[];
+  readonly typeParameters?: readonly number[];
+}
+
+export interface Component extends TypeArgument {
+  readonly name: string;
+}
+
+export interface TypeArgument {
+  readonly typeId: number | string; // the type metadata declaration ID or type concrete declaration hash based ID of the type of the component.
```
sdankel, line 36: How do we know which array to look for the `typeId` in? If it's a number, it's the `metadataTypes` array?

packages/abi-coder/src/types/JsonAbiNew.ts
nedsalk: @sdankel I suggest you check out the spec changes _([#596](https://github.com/FuelLabs/fuel-specs/pull/596), [#599](https://github.com/FuelLabs/fuel-specs/pull/599))_ to get an understanding of the new format.

To answer the question, no it's not because `typeParameters` contained a list of used generic types whereas the `typeArguments` here are a list of concrete types that will be replacing the type parameters of the linked metadata type.

packages/abi-coder/src/types/JsonAbiNew.ts
nedsalk: Yes, if it's a number then it's in the `metadataTypes` array, and if it's a string then it's in the `concreteTypes` array.

Are you doing some parsing of the abi on your own in some applications hence why you're asking the questions? If yes, I'd like to see where it's done and why because ideally the `Interface` should be sufficient for all abi-related requirements.

packages/abi-coder/src/types/JsonAbiNew.ts
nedsalk: Yes, this would be e.g. `struct std:vec::Vec<b256>`. The linked metadata type (via `metadataTypeId`) contains the name of the type: `struct std::vec::Vec`. Why would you want to get the name without the path? What additional abi parsing are you doing that isn't covered by `Interface`?

packages/abi-coder/src/types/JsonAbiNew.ts
```diff
@@ -0,0 +1,101 @@
...
+  readonly typeArguments?: readonly TypeArgument[];
+}
+
+export interface AbiFunction {
+  readonly name: string;
+  readonly inputs: readonly AbiFunctionInput[];
+  readonly output: string;
+  readonly attributes: readonly AbiFunctionAttribute[] | null;
```
petertonysmith94, line 44:
```suggestion
  readonly attributes?: readonly AbiFunctionAttribute[];
```

The specification mentions "an optional array of attributes". Should this be optional and not nullable?

packages/abi-typegen/src/types/interfaces/JsonAbiNew.ts
```diff
@@ -0,0 +1,101 @@
...
+export interface TypeArgument {
+  readonly typeId: number | string; // the type metadata declaration ID or type concrete declaration hash based ID of the type of the component.
+  readonly typeArguments?: readonly TypeArgument[];
+}
+
+export interface AbiFunction {
+  readonly name: string;
+  readonly inputs: readonly AbiFunctionInput[];
+  readonly output: string;
+  readonly attributes: readonly AbiFunctionAttribute[] | null;
```
petertonysmith94, line 44:
```suggestion
  readonly attributes?: readonly AbiFunctionAttribute[];
```

The specification mentions "an optional array of attributes". Should this be optional and not nullable?

packages/abi-coder/src/types/JsonAbiNew.ts
petertonysmith94: @nedsalk what do you think?

apps/create-fuels-counter-guide/sway-programs/Forc.toml
```diff
@@ -1,2 +1,5 @@
 [workspace]
 members = ["contract", "predicate", "script"]
+
+[patch.'https://github.com/fuellabs/sway']
+std = { git = "https://github.com/fuellabs/sway", branch = "master" }
```
arboleya, line 5: > [!IMPORTANT]
> Remove all patches before release.

internal/forc/VERSION
```diff
@@ -1 +1 @@
-0.62.0
+git:master
```
arboleya, line 1: > [!IMPORTANT]
> Replace it with the upcoming `forc` version.

packages/abi-coder/src/types/JsonAbiNew.ts
arboleya: @petertonysmith94 No, because the property is either an array or `null` - never `undefined`.

nedsalk: @arboleya thanks for investigating it.

.changeset/odd-horses-cheer.md
```diff
@@ -0,0 +1,12 @@
...
+"@fuel-ts/abi-coder": minor
+"@fuel-ts/abi-typegen": minor
+"@fuel-ts/account": minor
+"@fuel-ts/program": minor
+"@fuel-ts/script": minor
+"@fuel-ts/versions": minor
+"@internal/forc": minor
+"create-fuels": minor
+---
```
petertonysmith94, line 10:
```suggestion
---
"@fuel-ts/abi-coder": minor
"@fuel-ts/abi-typegen": minor
"@fuel-ts/versions": minor
"@internal/forc": minor
"create-fuels": minor
---
```

I believe only test files have changed in these packages.

nedsalk (dismissed):

petertonysmith94 (approved):
🚀 🚢

Torres-ssf (approved):

danielbate (approved):
:shipit:

danielbate (approved):

Torres-ssf (approved):

packages/abi-coder/src/types/JsonAbiNew.ts
sdankel: Sorry I missed this. Yes I have to do all of this in my application. See [here](https://github.com/FuelLabs/sway-playground/blob/6ae1b33ce1e4bcf589f2acfa25fb89c827b8e5ac/app/src/features/interact/components/ContractInterface.tsx#L27)

I'm curious why this is the hash for concrete types rather than the array index.

This new interface in the SDK is pretty complex.

packages/abi-coder/src/types/JsonAbiNew.ts
```diff
@@ -0,0 +1,101 @@
...
+  readonly concreteTypeId: string;
+  readonly metadataTypeId?: number;
+  readonly typeArguments?: readonly string[];
+}
+
+export interface MetadataType {
+  readonly type: string;
+  readonly metadataTypeId: number;
+  readonly components?: readonly Component[];
+  readonly typeParameters?: readonly number[];
```
sdankel, line 28: Is this an array of concrete type indexes or metadata type indexes? It's hard to tell which array this refers to.

packages/abi-coder/src/types/JsonAbiNew.ts
sdankel: Okay, so is this an array of concrete type hashes? It would be much easier to use if the SDK didn't expose the hashes at all and was consistent about using array indexes everywhere.

It would be nice also if the SDK would decode the hashes.

packages/abi-coder/src/types/JsonAbiNew.ts
```diff
@@ -0,0 +1,101 @@
...
+  readonly functions: readonly AbiFunction[];
+  readonly loggedTypes: readonly LoggedType[];
+  readonly messagesTypes: readonly MessageType[];
+  readonly configurables: readonly Configurable[];
+}
+
+export interface ConcreteType {
+  readonly type: string;
+  readonly concreteTypeId: string;
+  readonly metadataTypeId?: number;
```
sdankel, line 20: Per the [spec](https://github.com/FuelLabs/fuels-ts/pull/2856/files#diff-91cdfcdf704955525b9cd0988d8abef2b29fb92406e961e786d571b0779786f5R20), this should be a string (hash). It would help reduce complexity if this were consistent.

packages/abi-typegen/src/types/interfaces/JsonAbiNew.ts
```diff
@@ -0,0 +1,101 @@
...
+  readonly metadataTypes: readonly MetadataType[];
+  readonly functions: readonly AbiFunction[];
+  readonly loggedTypes: readonly LoggedType[];
+  readonly messagesTypes: readonly MessageType[];
+  readonly configurables: readonly Configurable[];
+}
+
+export interface ConcreteType {
+  readonly type: string;
+  readonly concreteTypeId: string;
```
sdankel, line 19: Is `type` just the type name, or is it the full type string? i.e. `MyStruct` or `path::to::MyStruct<T>`

packages/abi-coder/src/types/JsonAbiNew.ts
sdankel: I need the name of the type without the full path and without the generics to display input fields in sway playground, for the UI that lets users call contract functions.

sdankel: I need the name of the type without the full path and without the generics to display input fields in sway playground, for the UI that lets users call contract functions. I can write a helper function that does this for myself, but it seems like a useful this for the SDK to provide.

packages/abi-typegen/src/types/interfaces/JsonAbiNew.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export interface TypeArgument {
+  readonly typeId: number | string; // the type metadata declaration ID or type concrete declaration hash based ID of the type of the component.
+  readonly typeArguments?: readonly TypeArgument[];
+}
+
+export interface AbiFunction {
+  readonly name: string;
+  readonly inputs: readonly AbiFunctionInput[];
+  readonly output: string;
```
sdankel, line 43: Is this string the concrete type hash or the decoded full type string?

packages/abi-coder/src/types/JsonAbiNew.ts
nedsalk: > Yes I have to do all of this in my application. See [here](https://github.com/FuelLabs/sway-playground/blob/6ae1b33ce1e4bcf589f2acfa25fb89c827b8e5ac/app/src/features/interact/components/ContractInterface.tsx#L27)

Okay, so you have to parse the abi to present the infinite amount of possible inputs the users can generate. The SDK doesn't have something to support this as it's a highly specific use case that we haven't encountered previously because people are creating apps for already known sway programs and they know their inputs. 

> I'm curious why this is the hash for concrete types rather than the array index.

Because of the need for deterministic log ids. Previously array indexes would change and somebody doing e.g. `decodeLog(56, value)` would get a failure if the array changed. Now they pass in a hash of the concrete type and it won't change when the abi changes if the abi change doesn't affect that type.

> This new interface in the SDK is much more complex. It would have been nice to preserve the old interface or at least provide helper functions for conversion.

It just follows the specs. We intentionally wanted to remove the old interface so that we can stop supporting it altogether as `forc 0.63.0` and upwards won't be producing the old abi format. 

You _could_ internalize the [transpiler](https://github.com/FuelLabs/fuels-ts/pull/2856/files#diff-84ad7158971cfbd3205763a6ca8cf5d047ee10833b29074ecc5e48cc3fbdc33dR87) from this PR in your application, as it's doing the job of converting the new abi into the old, as well as the [old `JsonAbi` interface](https://github.com/FuelLabs/fuels-ts/pull/2856/files#diff-57ca93f87291b2c564b725c691f0f4e100de69a659a9b0eda64786757bdbad00R5).

We will be removing the tranpiler as part of #2877 because it's just there to give us more room to think about refactoring the packages that interact with `JsonAbi`.

packages/abi-coder/src/types/JsonAbiNew.ts
nedsalk: The link you gave us points to this same field the comment is on...

Regardless, the [spec](https://github.com/FuelLabs/fuel-specs/blob/master/src/abi/json-abi-format.md) says this for `concreteType.metadataTypeId`:
> "metadataTypeId": the type metadata declaration ID of this type, if the type metadata has components or is generic, otherwise nonexistent.

And it defines `metadataType.metadataTypeId` as this:
> "metadataTypeId": a unique integer ID.

So this conforms with the specification.

packages/abi-typegen/src/types/interfaces/JsonAbiNew.ts
nedsalk: It's the full type string. You can find examples lower down in the [spec](https://github.com/FuelLabs/fuel-specs/blob/master/src/abi/json-abi-format.md).

packages/abi-typegen/src/types/interfaces/JsonAbiNew.ts
nedsalk: It's the concrete type hash.

packages/abi-typegen/package.json
```diff
@@ -43,12 +43,11 @@
     "dist"
   ],
   "scripts": {
-    "pretest": "run-s build:forc-callpaths build:forc",
+    "pretest": "run-s build build:forc",
```
danielbate, line 46: How come we are building in pretest now? Seems to me that this is probably the reason for the [flakiness](https://github.com/FuelLabs/fuels-ts/issues/2967#issuecomment-2298515385). If pretest executes concurrently packages dependent on typegen _could_ fail if the build has not finished in time.

arboleya: Thanks a lot! 🙏 
- https://github.com/FuelLabs/fuels-ts/pull/2990

packages/abi-coder/src/types/JsonAbiNew.ts
arboleya: That was changed in:
 - https://github.com/FuelLabs/sway/pull/6254

> Removes the flag `--json-abi-with-callpaths` and the behavior is as if it were true. We removed the flag because it is unsafe to produce JSON ABIs without callpaths, so we shouldn't allow it.

You'll have to do some string parsing to extract only the portion you want.

packages/abi-coder/src/types/JsonAbiNew.ts
nedsalk: It is an array of metadata type ids:
> "typeParameters": an array of type metadata declaration ID of the type parameters of the type, if the type is generic, otherwise nonexistent. Each type parameter is a type declaration and is represented as described in [Generic Type Parameter](https://github.com/FuelLabs/fuel-specs/blob/master/src/abi/json-abi-format.md#generic-type-parameter).

packages/abi-coder/src/types/JsonAbiNew.ts
nedsalk: The SDK just consumes the ABI as per the specification and doesn't modify it. Concrete types and their hashed ids are now the way to identify actual directly used types (input or output, log, configurable, etc.) in a sway program. You can decode the hashes via the `Interface.decodeType(concreteTypeId: string, data: Uint8Array)` method which we added in this PR to enable users to decode any concrete type.

packages/abi-coder/src/types/JsonAbiNew.ts
arboleya: Depending on your use case, this can lead to naming collisions since you can't differentiate types with identical names but different namespaces. I'm taking note of this, and we'll consider adding such parsing utility helpers in the future. For now, the quickest way is to parse it yourself, as you said.

packages/abi-coder/src/types/JsonAbiNew.ts
sdankel: Sorry, updated the link. It also says this:

```
- _type metadata declaration_: the declaration or definition of a type which can be generic. `struct Foo { .. }` and `struct Bar<T> { .. }` in the example above are both type declarations. The metadata declaration contains component details about the type. And a single _type metadata declaration_ is generated per type, even for generic types.
```

Which makes it sound like "type metadata declaration ID" is the string hash, not an array index.

packages/abi-coder/src/types/JsonAbiNew.ts
sdankel: Thanks for the pointer to the helper function. Does it also work for decoding metadata type hashes?

packages/abi-coder/src/types/JsonAbiNew.ts
sdankel: IDs as in the hashes or the array indexes? Everywhere else, it looks like the hashes are stored as strings. Are these hashes stored as numbers instead or are these array indexes?

The spec itself is unclear on this.

packages/abi-coder/src/types/JsonAbiNew.ts
sdankel: Will the parse functions like `parseComponents` and `parseConcreteTypes` also be removed? Or is this something I can use?

packages/abi-coder/src/types/JsonAbiNew.ts
nedsalk: Metadata type ids are _always_ numbers. On the topic of array indexes, I wouldn't use array indexes for anything; they might correlate with the `metadataTypeId` of an entry in the `metadataTypes` array, but the spec doesn't define that relationship.

packages/abi-coder/src/types/JsonAbiNew.ts
nedsalk: The string hashes are exclusively related to concrete types (hashes of their `type` property). The spec doesn't mention array indexes in relation to `metadataTypeId`s, it just defines them as a field of the object. I wouldn't rely on that correlation to build anything as it might change: a metadata type with id `12` might be put onto index `24` and it would still be spec-compliant.

packages/abi-coder/src/types/JsonAbiNew.ts
nedsalk: Hashes are reserved exclusively to concrete types. Metadata types are _never_ used by the users directly so the SDK doesn't provide methods to work with them. I believe that you also won't have a need for them in sway-playground in relation to encoding/decoding as function inputs/outputs are always concrete types.

packages/abi-coder/src/types/JsonAbiNew.ts
nedsalk: Both of those functions are internal to the SDK and you aren't able to access them as they aren't exported. Everything introduced in this PR in relation to transpiling will be removed - those functions included. You can only go to the source code and copy/paste the whole logic for yourself and it'll be something that you'll have to maintain in your apps.

Issue #2855, opened 2024-07-29
UTXOs Cache Mechanism Needs To Consider Message Nonces


Torres-ssf:
The current UTXOs [cache mechanism](https://github.com/FuelLabs/fuels-ts/pull/930) needs to be refactored to include caching of message nonces. This is essential because message coins, uniquely identified by their nonces, are resources that can be used to fund transactions.

The cache mechanism must account for message coins to ensure the SDK is aware of all resources to be spent. This refactor will prevent errors such as the SDK using the same message to fund multiple transactions submitted within a short period of time (to be processed and mined at the same node).

arboleya:
Considering mentioning why we enable cache by default in the docs:
 - https://github.com/FuelLabs/fuels-ts/issues/2858

PR #2872, opened 2024-08-01
feat!: consider message on resources cache

Torres-ssf:
- Close: https://github.com/FuelLabs/fuels-ts/issues/2875
- Close: https://github.com/FuelLabs/fuels-ts/issues/2855

# Release notes

In this release, we:

- Cache `Message` resources upon transaction submission

# Summary

This PR addresses two issues:

1 - Unsetting cached resources if the transaction fails during processing. Previously, UTXO(s) remained cached and unavailable for new transactions until the cache TTL expired, potentially leading to resource unavailability for subsequent transactions.
2- Ensure the resource cache considers Message coins upon transaction submission, not just UTXO(s).

# Breaking Changes

The provider option flag `cacheUtxo` was renamed to `resourceCacheTTL`

```ts
// before
const provider = await Provider.create(FUEL_NETWORK_URL, {
  cacheUtxo: 5000,
});


  providerOptions: {
    cacheUtxo: 5000,
  },
});
```

```ts
// after
const provider = await Provider.create(FUEL_NETWORK_URL, {
  resourceCacheTTL: 5000,
});

  providerOptions: {
    resourceCacheTTL: 5000,
  },
});
```

# Checklist


.changeset/lazy-plants-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/account": patch
+---
+
+fix: unset cached resource for failed TX
```
arboleya, line 5: Shouldn't this be a breaking due to `set`'s new signature?

Torres-ssf: @arboleya I did not specify this change because the `MemoryCache` ( renamed to `ResourceCache` ) was used only internally. Also, it is not exported by `fuels`.

apps/docs-snippets/src/guide/provider/provider.test.ts
```diff
@@ -107,10 +107,10 @@ describe('Provider', () => {
     // #endregion options-fetch
   });
 
-  it('options: cacheUtxo', async () => {
+  it('options: resourceCacheTTL', async () => {
     // #region options-cache-utxo
     const provider = await Provider.create(FUEL_NETWORK_URL, {
-      cacheUtxo: 5000, // cache UTXO for 5 seconds
+      resourceCacheTTL: 5000, // cache UTXO for 5 seconds
```
petertonysmith94, line 113:
```suggestion
      resourceCacheTTL: 5000, // cache resources (Coin's and Message's) for 5 seconds
```

apps/docs/src/guide/provider/provider-options.md
petertonysmith94, line 1: Would it be worth adding that messages are also caches with this feature?

petertonysmith94 (commented):
Would there be a possibility that we'd want to cache the coins, but not the messages, and vice versa?

packages/account/src/providers/provider.test.ts
```diff
@@ -375,51 +377,57 @@ describe('Provider', () => {
...
 
     expect(provider.cache).toBeUndefined();
   });
 
-  it('should cache UTXOs only when TX is successfully submitted', async () => {
+  it('should cache resources only when TX is successfully submitted', async () => {
+    const resourceAmount = 50_000;
+    const utxosAmount = 2;
+
+    const testMessage = new TestMessage({ amount: resourceAmount, data: '' });
```
nedsalk, line 421: Given that `data` needs to be an empty string for this to be treated as a `MessageCoin`, you can change the default value to `''` in `TestMessage` right away. You don't need to do this in this PR if you don't want to.

packages/account/src/providers/provider.ts
```diff
@@ -683,16 +683,24 @@ Supported fuel-core version: ${supportedVersion}.`
...
+          acc.utxos.push(input.id);
+        } else if (input.type === InputType.Message) {
+          acc.messages.push(input.nonce);
+        }
+        return acc;
+      },
+      { utxos: [], messages: [] } as Required<ExcludeResourcesOption>
+    );
+
+    this.cache.set(transactionId, ExcludeResources);
```
nedsalk, line 703:
```suggestion
    const inputsToCache = inputs.reduce(
      (acc, input) => {
        if (input.type === InputType.Coin) {
          acc.utxos.push(input.id);
        } else if (input.type === InputType.Message) {
          acc.messages.push(input.nonce);
        }
        return acc;
      },
      { utxos: [], messages: [] } as Required<ExcludeResourcesOption>
    );

    this.cache.set(transactionId, inputsToCache);
```

apps/docs/src/guide/provider/provider-options.md
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2872/commits/1a8ab1a10f379f06a349108b131ed2e7a33bcf96

apps/docs-snippets/src/guide/provider/provider.test.ts
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/2872/commits/8aca9c2697b96ba7f8249abee0580979ef72a287

packages/account/src/test-utils/launchNodeAndGetWallets.test.ts
```diff
@@ -57,7 +57,7 @@ describe('launchNode', () => {
     const { stop, wallets } = await launchNodeAndGetWallets({
       walletCount: 5,
       providerOptions: {
-        cacheUtxo: 1,
+        resourceCacheTTL: 1,
       },
```
nedsalk, line 61:
```suggestion
```

It seems to me that there's no point in specifying this in the test.

nedsalk: Ohh, this is the obsolete `launchNodeAndGetWallets` utility. Ignore the comment.

Torres-ssf:
> Would there be a possibility that we'd want to cache the coins, but not the messages, and vice versa?

@petertonysmith94 I can't imagine a scenario where someone would want to do that.

maschad (changes requested):
Great work on this @Torres-ssf 🚀 

Given the nature of this cache being a shared resource, I've decided to make some changes after going down a bit of a rabbithole 😅  

here's the rationale:

- I notice we only remove the expired cached UTXOs when `getActiveData` is called, but that is only called when `getResourcesToSpend` is. Correct me if I am wrong but a consumer could submit a transaction without calling `getResourcesToSpend` and thus expired UTXOs wouldn't have been cleared and the transaction submitted. So I wrote a cleanup process that runs in the background to clear the cache.

- Because that process could be trying to delete a cached UTXO that could be potentially be trying to be accessed or the cache itself could be accessed during a delete, I made the operations atomic through a semaphore. 


All the changes are on https://github.com/FuelLabs/fuels-ts/pull/2886 and I've added the rationale there as well for reviewing consistency

Torres-ssf:
> Correct me if I am wrong but a consumer could submit a transaction without calling getResourcesToSpend

@maschad How can one submit a transaction without calling `getResourcesToSpend` first to fetch resources to fund the transaction?

>  and thus expired UTXOs wouldn't have been cleared and the transaction submitted

If `getResourcesToSpend` was never called and the transaction was never funded, how would the transaction be submitted with expired UTXOs?

maschad:
> @maschad How can one submit a transaction without calling getResourcesToSpend first to fetch resources to fund the transaction?

Calling [`sendTransaction`](https://github.com/FuelLabs/fuels-ts/blob/st/fix/unset-cached-resource-for-failed-tx/packages/account/src/providers/provider.ts#L717) directly doesn't  call `getResourcesToSpend` in the function call chain

Torres-ssf:
> Calling [`sendTransaction`](https://github.com/FuelLabs/fuels-ts/blob/st/fix/unset-cached-resource-for-failed-tx/packages/account/src/providers/provider.ts#L717) directly doesn't call `getResourcesToSpend` in the function call chain

Why someone will call `sendTransaction` without funding the transaction first? 

The call to `getResourcesToSpend` will always need to happen to fetch resources to fund the transaction. A transaction without funds will not be processed.

packages/account/src/providers/provider.test.ts
Torres-ssf: @nedsalk https://github.com/FuelLabs/fuels-ts/pull/2891

maschad:
> The call to getResourcesToSpend will always need to happen to fetch resources to fund the transaction. A transaction without funds will not be processed.

That's correct the [submit call](https://github.com/FuelLabs/fuels-ts/blob/st/fix/unset-cached-resource-for-failed-tx/packages/account/src/providers/provider.ts#L737) will fail, and given we now cache inputs _after_ submission then that scenario can't occur. 

I suppose the only benefit of the changes I proposed would be that we clean up expired cached UTXOs periodically as opposed to only when we are going to submit another transaction. Feel free to disregard them though.

maschad (dismissed):
Nice work!

petertonysmith94 (dismissed):

packages/account/src/providers/provider.test.ts
nedsalk, line 421:
```suggestion
    const testMessage = new TestMessage({ amount: resourceAmount });
```

packages/account/src/providers/provider.test.ts
```diff
@@ -477,21 +493,80 @@ describe('Provider', () => {
...
+    [...utxos, ...messages].forEach((key) => {
+      expect(provider.cache?.isCached(key)).toBeFalsy();
+    });
   });
 
-  it('should ensure cached UTXOs are not being queried', async () => {
-    // Fund the wallet with 2 UTXOs
-    const totalUtxos = 2;
+  it('should unset cached resources when TX execution fails', async () => {
+    const message = new TestMessage({ amount: 100_000, data: '' });
```
nedsalk, line 503:
```suggestion
    const message = new TestMessage({ amount: 100_000 });
```

packages/account/src/providers/provider.test.ts
```diff
@@ -477,21 +493,80 @@ describe('Provider', () => {
...
+
+    // Ensure user's resouces were unset from the cache
+    [...utxos, ...messages].forEach((key) => {
+      expect(provider.cache?.isCached(key)).toBeFalsy();
+    });
+  });
+
+  it('should ensure cached resources are not being queried', async () => {
+    // Fund the wallet with 2 resources
+    const testMessage = new TestMessage({ amount: 100_000_000_000, data: '' });
```
nedsalk, line 561:
```suggestion
    const testMessage = new TestMessage({ amount: 100_000_000_000 });
```

packages/account/src/providers/resource-cache.ts
```diff
@@ -0,0 +1,79 @@
...
+interface CachedResource {
+  utxos: Set<string>;
+  messages: Set<string>;
+  timestamp: number;
+}
+
+const cache = new Map<string, CachedResource>();
+
+export class ResourceCache {
+  readonly ttl: number;
```
nedsalk, line 15:
```suggestion
export class ResourceCache {
  readonly ttl: number;
  private cache = new Map<string, CachedResource>();
```

If somebody were to instantiate two providers, the cache would be shared between them because this `Map` is shared between all `ResourceCache` instances. The test below fails as a result. I suggest you add it to `provider.test.ts`.
```ts
  it('cache is not shared between Predicate instances', async () => {
    using launched1 = await setupTestProviderAndWallets();
    using launched2 = await setupTestProviderAndWallets();

    const {
      wallets: [sender, receiver],
    } = launched1;

    await sender.transfer(receiver.address, 1000);

    expect(launched1.provider.cache?.getActiveData()).not.toMatchObject(
      launched2.provider.cache?.getActiveData() ?? {}
    );
  });
```

This shouldn't happen because `cache` [isn't](https://github.com/FuelLabs/fuels-ts/blob/9c3eb66e133489bf7eee2aa59de8b5c5745f9db5/packages/account/src/providers/provider.ts#L376) a static property on `Provider`.

To fix this, `cache` should be internalized into `ResourceCache`.

nedsalk (changes requested):

packages/account/src/providers/resource-cache.ts
Torres-ssf: @nedsalk This is intentional. A provider can be instantiated in different parts of an dApp application.

nedsalk: I discussed this privately with @Torres-ssf and he'll create a new issue/discussion on this topic, listing the implications of making the cache fully static or making it fully instance-specific. For now we agreed that it's fine as it is.

petertonysmith94 (approved):

petertonysmith94 (approved):

maschad (approved):
Nice work 🚀

PR #2874, opened 2024-08-01
chore!: reorder `Interface.decodeLog` arguments

nedsalk:
# Summary

Having this order makes more sense.

# Breaking Changes

The `Interface.decodeLog` arguments got reordered so that it now first accepts the `logId` and then the data.
```ts
// before
const abiInterface = new Interface(yourAbi);
const decodedLog = abiInterface.decodeLog(logData, logId);
```
```ts
// after
const decodedLog = abiInterface.decodeLog(logId, logData);
```

# Checklist

### DIFF
--- a/.changeset/curly-buttons-invent.md
+++ b/.changeset/curly-buttons-invent.md
@@ -1,0 +1,6 @@
+---
+"@fuel-ts/abi-coder": minor
+"@fuel-ts/account": patch
+---
+
+chore!: reorder `Interface.decodeLog` arguments
--- a/.changeset/small-lamps-study.md
+++ b/.changeset/small-lamps-study.md
@@ -1,0 +1,4 @@
+---
+---
+
+chore: stop skipping decoding for vectors in `Interface.test.ts`
--- a/packages/abi-coder/src/Interface.ts
+++ b/packages/abi-coder/src/Interface.ts
@@ -8,7 +8,7 @@ import { FunctionFragment } from './FunctionFragment';
 import type { InputValue } from './encoding/coders/AbstractCoder';
 import type { JsonAbi, JsonAbiConfigurable } from './types/JsonAbi';
 import { type EncodingVersion } from './utils/constants';
-import { findTypeById, getEncodingVersion } from './utils/json-abi';
+import { getEncodingVersion } from './utils/json-abi';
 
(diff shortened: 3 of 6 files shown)


petertonysmith94 (approved):

Torres-ssf (approved):

danielbate (commented):
I'm bikeshedding here so this isn't a hill I'm willing to die on as you've got 2 approvals, but I don't think we should be making breaking changes just because something seems more correct. I don't think that's enough value to justify breaking consumer apps and them having to search or deduce a resolution. And we can rewrite the API in #1795. Thoughts?

danielbate:
> I'm bikeshedding here so this isn't a hill I'm willing to die on as you've got 2 approvals, but I don't think we should be making breaking changes just because something seems more correct. I don't think that's enough value to justify breaking consumer apps and them having to search or deduce a resolution. And we can rewrite the API in #1795.

arboleya:
@danielbate You're probably right. My guess is that the `typegen`/`coder` API will inevitably break in the next big push anyway. Focusing on the new ABI spec and generated files should do the job.

.changeset/curly-buttons-invent.md
```diff
@@ -0,0 +1,6 @@
+---
+"@fuel-ts/abi-coder": minor
+"@fuel-ts/account": patch
+---
+
+chore!: reorder `Interface.decodeLog` arguments
```
arboleya, line 6: Approving with a comment to hold auto-merge from kicking before conversations are settled.

arboleya (approved):

danielbate:
@nedsalk #1795 is a `v1` change. I agree we can deliver value via the points you've mentioned and not introduce breaking changes. However, the current design has been proved and battle tested, and the real value is in rethinking and rewriting both the implementation and API.

I'm pro closing this PR. I don't think the breakage justifies the value it delivers.

petertonysmith94:
I'd also favour closing this PR until v1 after reflection.

Thoughts @nedsalk?

petertonysmith94:
There appears to consensus that we'll hold off on this change till:
- https://github.com/FuelLabs/fuels-ts/issues/1795

Closing until then.

PR #2886, opened 2024-08-05
chore: update resource cache to be atomic + cleanup process

maschad:
# Summary

I've moved these suggestions onto a separate branch given they are quite substantial. They are recommendations for https://github.com/FuelLabs/fuels-ts/pull/2872

- I noticed we only remove the expired cached UTXOs when `getActiveData` is called, but that is only called when `getResourcesToSpend` is. Correct me if I am wrong but a consumer could submit a transaction without calling `getResourcesToSpend` and thus expired UTXOs wouldn't have been cleared and the transaction submitted. So I wrote a cleanup process that runs in the background to clear the cache.

- Because that process could be trying to delete a cached UTXO that could be potentially be trying to be accessed or the cache itself could be accessed during a delete, I made the operations atomic through a semaphore. 

# Checklist


maschad:
Closing as these changes are no longer relevant (see https://github.com/FuelLabs/fuels-ts/pull/2872#issuecomment-2271748130)

PR #2888, opened 2024-08-06
fix: typegen reporting the correct versions

### LINKED ISSUE #2823: Typegend templates are misreporting the versions they are using
Our typegen templates report the versions they are using in the source code:

https://github.com/FuelLabs/fuels-ts/blob/e00db5f2276670edec2e7a625dbd2bc08fda9d98/packages/abi-typegen/src/templates/common/_header.hbs#L8-L9

These versions are taken from the currently supported version by the SDK `internal/forc` / `internal/fuel-core`. Wheras `fuels build` will use the users internal `forc` version i.e. a user could build with `forc@0.62.0` but the template will report `0.61.2` or whatever the supported version is at that time.

petertonysmith94:
- Closes #2823

# Release notes

In this release, we:

- Fixed `forc` and `fuel-core` version reporting on typegen-generated files

# Summary

- Currently, we add the "built in" versions of the SDK to the typegen'd files, which doesn't reflect that of the user.
- This change does two things:
  - `fuels build` - we add the versions of the `forc` and `fuel-core` that has been configured (via config).
  - `fuels typegen` - these versions will no longer be present, as we don't know which versions these were built with.

# Checklist


Torres-ssf (dismissed):
🚀

packages/abi-typegen/src/templates/common/_header.hbs
```diff
@@ -7,6 +7,10 @@
 
 /*
   Fuels version: {{FUELS}}
+{{#if FORC}}
   Forc version: {{FORC}}
+{{/if}}
+{{#if FUEL_CORE}}
   Fuel-Core version: {{FUEL_CORE}}
+{{/if}}
```
arboleya, line 15: Initially, these were the SDK-supported versions of each tool, but I can see how this PR makes more sense. 👍

After all, we already have the `fuels` version anyway.

arboleya (commented):
Conflicts to fix.

Dhaiwat10:
There is a typo in the release notes:

`Versions for FORC and FUEL_CORE in typegen files, better reflect that actual versions.`

should be

`Versions for FORC and FUEL_CORE in typegen files better reflect their actual versions.`

arboleya (approved):
Another take on the release notes:
 - Fixed `forc` and `fuel-core` version reporting on typegen-generated files

Torres-ssf (approved):

danielbate (approved):

Issue #2467, opened 2024-06-07
Map VM errors to `FuelError` error codes


nedsalk:
- Create `ErrorCode` entry for every VM error
- Every `ErrorCode` entry that maps to a VM error must have a test associated with it:
    - e.g. bring the VM to throw `not enough coins to fit the target` and:
        - Verify that a `FuelError` with the `not-enough-coins` code is thrown
        - Verify that `FuelError.rawError` corresponds to the serialized VM error - this would be done with a `fetch` spy

cc @LuizAsFight 

> [!NOTE]
> Should be done after:
> - https://github.com/FuelLabs/fuels-ts/issues/2521

LuizAsFight:
hey @nedsalk thanks for the heads up

just for context check [this file](https://github.com/FuelLabs/fuels-wallet/blob/4a20b3d8684a07e40bd2f7559032941b5adbe022/packages/app/src/systems/Transaction/services/transaction.tsx#L176) from wallet code

graphql api can return multiple errors, that's why we initially had only `e.response?.errors` , and after the `FuelError` was included we needed to add this conditional, forcing the creation of an array containing the error

do you think it makes sense to create 1 FuelError for each occurrence of errors that came from graphql api? that way returning always an array

nedsalk:
I asked @FuelLabs/client the following question:
> The graphql API returns an errors array when an error happens on the VM. In practice, are there cases where the VM will return more than one error? If there are none, could we make an assumption that there won't be any in the future? Having this assumption would improve ergonomics for users on our side - a thrown error would always be the error, not a wrapper for multiple errors in some cases and the error in others.

And @Dentosal responded:
>  My understanding is that the errors array is part of how GraphQL itself works. If you only request one VM operation, then there can only be one VM error.

@LuizAsFight based on this, we can extract the single error from the graphql `errors` array and throw it. We don't need to handle the _multiple errors_ case because it never happens in reality. The [wallet code](https://github.com/FuelLabs/fuels-wallet/blob/4a20b3d8684a07e40bd2f7559032941b5adbe022/packages/app/src/systems/Transaction/services/transaction.tsx#L176) that you linked to could then maybe be simplified because you wouldn't be working with arrays at all.

arboleya:
I reduced the scope of this issue and moved a portion of it to the below issue, as it will be easier and provide [immediate value](https://github.com/FuelLabs/fuels-ts/issues/2522) even before we can map "all" errors.

[...]

PR #2902, opened 2024-08-09
feat: map 'not enough coins' error

### LINKED ISSUE #2522: Map `not enough coins to fit the target` error
Errors like `not enough coins to fit the target` don't seem obvious to users:

- https://forum.fuel.network/t/fuelerror-not-enough-coins-to-fit-the-target/5818/2
- https://forum.fuel.network/t/creating-a-fuel-dapp-using-provided-example-in-docs/5796/1

We should consider mapping this one even before we get to https://github.com/FuelLabs/fuels-ts/issues/2467.

> [!NOTE]
> Should be done after:
> - https://github.com/FuelLabs/fuels-ts/issues/2521

Dhaiwat10:
- Closes #2522 

# Release notes

In this release, we:

- Mapped the 'not enough coins' error to a readable error message

# Summary

This PR maps the `not enough coins to fit the target error` to a proper readable error message. It also lays down the foundation for adding the mapping of more VM errors like this one (#2467).

# Checklist

### DIFF
--- a/packages/account/src/providers/fuel-graphql-subscriber.ts
+++ b/packages/account/src/providers/fuel-graphql-subscriber.ts
@@ -2,6 +2,8 @@ import { ErrorCode, FuelError } from '@fuel-ts/errors';
 import type { DocumentNode } from 'graphql';
 import { print } from 'graphql';
 
+import { handleGqlErrorMessage } from './utils/handle-gql-error-message';
+
 type FuelGraphQLSubscriberOptions = {
   url: string;
   query: DocumentNode;
@@ -48,10 +50,8 @@ export class FuelGraphqlSubscriber implements AsyncIterator<unknown> {
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         const { data, errors } = this.events.shift()!;
         if (Array.isArray(errors)) {
-          throw new FuelError(
-            FuelError.CODES.INVALID_REQUEST,
-            errors.map((err) => err.message).join('\n\n')
-          );
+          const errorMessage = errors.map((err) => err.message).join('\n\n');
+          handleGqlErrorMessage(errorMessage);
         }
         return { value: data, done: false };
       }
--- a/packages/account/src/providers/utils/handle-gql-error-message.ts
+++ b/packages/account/src/providers/utils/handle-gql-error-message.ts
@@ -1,0 +1,17 @@
+import { ErrorCode, FuelError } from '@fuel-ts/errors';
+
+export enum GqlErrorMessage {
+  NOT_ENOUGH_COINS = 'not enough coins to fit the target',
+}
+
+export const handleGqlErrorMessage = (errorMessage: string) => {
+  switch (errorMessage) {
+    case GqlErrorMessage.NOT_ENOUGH_COINS:
+      throw new FuelError(
+        ErrorCode.NOT_ENOUGH_FUNDS,
+        'This account does not have enough funds to cover this transaction.'
+      );
+    default:
+      throw new FuelError(ErrorCode.INVALID_REQUEST, `Unknown error: ${errorMessage}`);
+  }
+};
--- a/.changeset/rare-shrimps-clap.md
+++ b/.changeset/rare-shrimps-clap.md
@@ -1,0 +1,6 @@
+---
+"@fuel-ts/account": patch
+"@fuel-ts/errors": patch
+---
+
+feat: map 'not enough coins' error
--- a/packages/account/src/providers/provider.ts
+++ b/packages/account/src/providers/provider.ts
@@ -60,6 +60,7 @@ import {
 } from './utils';
 import type { RetryOptions } from './utils/auto-retry-fetch';
 import { autoRetryFetch } from './utils/auto-retry-fetch';
+import { handleGqlErrorMessage } from './utils/handle-gql-error-message';
 
 const MAX_RETRIES = 10;
 
@@ -567,11 +568,12 @@ Supported fuel-core version: ${supportedVersion}.`
       responseMiddleware: (response: GraphQLResponse<unknown> | Error) => {
         if ('response' in response) {
           const graphQlResponse = response.response as GraphQLResponse;
+
           if (Array.isArray(graphQlResponse?.errors)) {
-            throw new FuelError(
-              FuelError.CODES.INVALID_REQUEST,
-              graphQlResponse.errors.map((err: Error) => err.message).join('\n\n')
-            );
+            const errorMessage = graphQlResponse.errors
+              .map((err: Error) => err.message)
+              .join('\n\n');
+            handleGqlErrorMessage(errorMessage);
           }
         }
       },
--- a/packages/errors/src/error-codes.ts
+++ b/packages/errors/src/error-codes.ts
@@ -56,6 +56,7 @@ export enum ErrorCode {
   MISSING_REQUIRED_PARAMETER = 'missing-required-parameter',
   INVALID_REQUEST = 'invalid-request',
   INVALID_TRANSFER_AMOUNT = 'invalid-transfer-amount',
+  NOT_ENOUGH_FUNDS = 'not-enough-funds',
 
   // crypto
   INVALID_CREDENTIALS = 'invalid-credentials',
--- a/packages/fuel-gauge/src/not-enough-coins.test.ts
+++ b/packages/fuel-gauge/src/not-enough-coins.test.ts
@@ -1,0 +1,19 @@
+import { Contract, ErrorCode, Wallet } from 'fuels';
+import { expectToThrowFuelError } from 'fuels/test-utils';
+
+import { CallTestContractFactory } from '../test/typegen/contracts';
+
+import { launchTestContract } from './utils';
+
+test('not enough coins error', async () => {
+  using contract = await launchTestContract({ factory: CallTestContractFactory });
+
+  const emptyWallet = Wallet.generate({ provider: contract.provider });
+
+  const emptyWalletContract = new Contract(contract.id, contract.interface.jsonAbi, emptyWallet);
+
+  await expectToThrowFuelError(() => emptyWalletContract.functions.return_void().call(), {
+    code: ErrorCode.NOT_ENOUGH_FUNDS,
+    message: 'This account does not have enough funds to cover this transaction.',
+  });
+});


packages/account/src/providers/fuel-graphql-subscriber.ts
```diff
@@ -48,10 +50,8 @@ export class FuelGraphqlSubscriber implements AsyncIterator<unknown> {
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         const { data, errors } = this.events.shift()!;
         if (Array.isArray(errors)) {
-          throw new FuelError(
-            FuelError.CODES.INVALID_REQUEST,
-            errors.map((err) => err.message).join('\n\n')
-          );
+          const errorMessage = errors.map((err) => err.message).join('\n\n');
+          handleGqlErrorMessage(errorMessage);
         }
```
petertonysmith94, line 55: I'd of personally, of create a `assertNoGqlErrors` which takes the whole errors array. Then parse the individual errors from there and throw if required.

Currently, as it stands, if there were two errors then the detection in `handleGqlErrorMessage` would fall over.

packages/fuel-gauge/src/funding-transaction.test.ts
```diff
@@ -266,7 +266,21 @@ describe('Funding Transactions', () => {
      */
     await expectToThrowFuelError(
       () => sender.fund(request, txCost),
-      new FuelError(FuelError.CODES.INVALID_REQUEST, 'not enough coins to fit the target')
+      new FuelError(
+        FuelError.CODES.NOT_ENOUGH_FUNDS,
+        'This account does not have enough funds to cover this transaction.',
```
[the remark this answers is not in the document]
arboleya: Agreed. A slightly more general message seems like a better fit.

apps/docs/src/guide/errors/index.md
```diff
@@ -252,6 +252,10 @@ When the Fuel Node info cache is empty; This is usually caused by not being conn
 
 Ensure that the provider has connected to a Fuel Node successfully.
 
+### `NOT_ENOUGH_FUNDS`
+
+When the account sending the transaction does not have enough funds to cover the transaction fee.
+
```
petertonysmith94, line 258:
```suggestion
When the account sending the transaction does not have enough funds to cover the transaction fee.

Ensure that the account creating the transaction has been funded appropriately.

```

Suggest adding a description of resolution.

petertonysmith94 (commented):
Might be worthwhile to refactor some of the expect cases to use `expectToThrowFuelError` - then we can test the error codes as well.

packages/account/src/providers/utils/handle-gql-error-message.ts
```diff
@@ -0,0 +1,20 @@
+import { ErrorCode, FuelError } from '@fuel-ts/errors';
+import type { GraphQLError } from 'graphql';
+
+export enum GqlErrorMessage {
+  NOT_ENOUGH_COINS = 'not enough coins to fit the target',
+}
+
+export const handleGqlErrorMessage = (errorMessage: string, rawError?: GraphQLError) => {
```
nedsalk, line 8: Should the `rawError` be mandatory instead of optional? Every gql error must have an accompanying `GraphQLError` error.

Dhaiwat10: I kept it optional because I'm only getting the error _message_ from the GraphQL subscription implementation that we have :/

nedsalk: This error will never happen in `FuelGraphqlSubscriber` because we're not using the `submitAndAwait` endpoint anymore; we only use `statusChange`, which can't error with `NOT_ENOUGH_FUNDS` because it works with already-submitted transactions. I'm for completely removing the work done in `FuelGraphqlSubscriber` because of that and thinking about the wider problem in #2467, given that the scope of this PR/issue is only related to `NOT_ENOUGH_FUNDS`.

nedsalk: If you still want to keep it in `FuelGraphqlSubscriber` - which I'm also fine with - you can change the `rawError` parameter's type to be `FuelError['rawError']` which is `unknown` and then you can pass in the whole error object in `FuelGraphqlSubscriber` as well:
```ts
for (const error of errors) {
  handleGqlErrorMessage(error.message, error);
}
```

packages/account/src/providers/utils/handle-gql-error-message.ts
```diff
@@ -0,0 +1,20 @@
...
+export enum GqlErrorMessage {
+  NOT_ENOUGH_COINS = 'not enough coins to fit the target',
+}
+
+export const handleGqlErrorMessage = (errorMessage: string, rawError: GraphQLError) => {
+  switch (errorMessage) {
+    case GqlErrorMessage.NOT_ENOUGH_COINS:
+      throw new FuelError(
+        ErrorCode.NOT_ENOUGH_FUNDS,
+        'The transaction does not have enough funds to cover its execution.',
```
Torres-ssf, line 13: @Dhaiwat10 what do you think about: 

`The account does not have enough funds to cover the transaction execution.`?

Since `not enough coins to fit the target`error happens when we call `getResourcesToSpend` informing an amount higher than the one owned by the account.

Dhaiwat10:
@petertonysmith94 I'm in favour of asserting the error code as well but as you can see in the failing CI, looks like I'll have to assert the `rawError` object as well which is not ideal :/

packages/account/src/providers/utils/handle-gql-error-message.ts
Dhaiwat10: @Torres-ssf that was the message I used initially but the team asked for a change here: https://github.com/FuelLabs/fuels-ts/pull/2902#discussion_r1714965751

packages/fuel-gauge/src/funding-transaction.test.ts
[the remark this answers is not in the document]
Torres-ssf: The client throws the error `not enough coins to fit the target` when calling the query `getResourcesToSpend` specifying a value higher than the one owned by the account.

This happens before submitting the TX to the node.

The error `The transaction does not have enough funds to cover its execution.` feels a little off to me as the TX was not even submitted. It gives the idea that the transaction was submitted without being properly funded.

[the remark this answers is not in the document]
maschad: Agree here @Torres-ssf

petertonysmith94:
> @petertonysmith94 I'm in favour of asserting the error code as well but as you can see in the failing CI, looks like I'll have to assert the `rawError` object as well which is not ideal :/

I believe if we set `rawError` to be `null` by default, this will resolve the issue you're experiencing.

https://github.com/FuelLabs/fuels-ts/blob/3084cca14120f81309131f05c6b2fae5a2fd3bb3/packages/errors/src/fuel-error.ts#L40

As the following block will work as expected:

https://github.com/FuelLabs/fuels-ts/blob/3084cca14120f81309131f05c6b2fae5a2fd3bb3/packages/errors/src/test-utils/expect-to-throw-fuel-error.ts#L54-L56

Or check that the `rawError` is not an empty object above, however, I favour `null` by default :)

packages/fuel-gauge/src/funding-transaction.test.ts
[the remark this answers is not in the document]
petertonysmith94: I'd personally say leave this for #2467 as we can map based on the query path to generate more descriptive error message (or that's the plan). 

Does this satisfy both camps?

> The account or transaction does not have enough funds to cover this request.

packages/account/src/providers/utils/handle-gql-error-message.ts
petertonysmith94: > I'm for completely removing the work done in FuelGraphqlSubscriber because of that and thinking about the wider problem in https://github.com/FuelLabs/fuels-ts/issues/2467

@nedsalk could you just clarify what you mean by this? Maybe on the #2467 would be helpful :)

petertonysmith94: @nedsalk Thanks for the insight. Surely having the logic contained within `handleGqlErrorMessage` is keeping it centralized?

nedsalk: Indeed it is, but both subscriptions and non-subscriptions are going through the same fetch function so there _might_ be room for restructuring the code and using `handleGqlErrorMessage` only in that fetch function as opposed to the current usage within both `FuelGraphqlSubscriber` and `GraphqlRequest`.

packages/fuel-gauge/src/funding-transaction.test.ts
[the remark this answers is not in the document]
Dhaiwat10: @petertonysmith94 to be honest that sounds a little more confusing 😅 

How about
`The account(s) sending the transaction don't have enough funds to cover the transaction.`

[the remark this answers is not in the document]
nedsalk: @Torres-ssf is right here and the original message or what he proposed in https://github.com/FuelLabs/fuels-ts/pull/2902#discussion_r1715792581 is better.

arboleya:
@Dhaiwat10 Marking as a draft until conversations are resolved and CI is all green.

packages/account/src/account.test.ts
```diff
@@ -523,11 +523,13 @@ describe('Account', () => {
     }
 
     // Test excludes the UTXO where the assetIdA gets added to the senders wallet
-    await expect(
-      user.getResourcesToSpend([[1, ASSET_A, 500_000]], {
-        utxos: [assetAUTXO.id],
-      })
-    ).rejects.toThrow(/not enough coins to fit the target/);
+    await expectToThrowFuelError(
+      () => user.getResourcesToSpend([[1, ASSET_A, 500_000]], { utxos: [coins[0].id] }),
```
Torres-ssf, line 527:
```suggestion
      () => user.getResourcesToSpend([[1, ASSET_A, 500_000]], { utxos: [assetAUTXO.id] }),
```

packages/errors/src/fuel-error.ts
```diff
@@ -37,7 +37,7 @@ export class FuelError extends Error {
     code: ErrorCode,
     message: string,
     metadata: Record<string, unknown> = {},
-    rawError: unknown = {}
+    rawError: unknown = null
```
Torres-ssf, line 40: @Dhaiwat10 Why we had to change this to null?

Torres-ssf (changes requested):
Great job 👍

packages/account/src/account.test.ts
Torres-ssf: This should fix the test errors

packages/errors/src/fuel-error.ts
Dhaiwat10: Here is the reasoning: https://github.com/FuelLabs/fuels-ts/pull/2902#issuecomment-2288013444

maschad (dismissed):
Nice work 🚀

packages/errors/src/fuel-error.ts
arboleya: I think I'd prefer `undefined`.

```ts
 rawError?: unknown
```

petertonysmith94: I usually tend towards a `null` value when representing the absence of an object value.

I don't have a strong opinion on this.

Torres-ssf (dismissed):

packages/fuel-gauge/src/funding-transaction.test.ts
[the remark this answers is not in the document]
arboleya: @Dhaiwat10 So, are we going with singular or plural?

This conversation needs resolution.

apps/docs/src/guide/errors/index.md
```diff
@@ -264,6 +264,12 @@ When the Fuel Node info cache is empty; This is usually caused by not being conn
 
 Ensure that the provider has connected to a Fuel Node successfully.
 
+### `NOT_ENOUGH_FUNDS`
+
+When the account sending the transaction does not have enough funds to cover the transaction fee.
```
arboleya, line 269: Nit.

```suggestion
When the account sending the transaction does not have enough funds to cover the fee.
```

apps/docs/src/guide/wallets/wallet-transferring.md
```diff
@@ -53,9 +52,8 @@ To transfer assets to multiple wallets, use the `Account.batchTransfer` method:
 
 This section demonstrates additional examples of transferring assets between wallets and to contracts.
 
-
 ## Checking Balances
 
-Before transferring assets, ensure your wallet has sufficient funds. Attempting a transfer without enough funds will result in an error: `not enough coins to fit the target`.
+Before transferring assets, ensure your wallet has sufficient funds. Attempting a transfer without enough funds will result in an error: `The transaction does not have enough funds to cover its execution.`.
```
arboleya, line 57: Nit.

```suggestion
Before you transfer assets, please make sure your wallet has enough funds. Attempting a transfer without enough funds will result in the error: `The transaction does not have enough funds to cover its execution.`
```

arboleya (approved):

danielbate (dismissed):

petertonysmith94 (approved):

Torres-ssf (approved):

danielbate (approved):

packages/fuel-gauge/src/funding-transaction.test.ts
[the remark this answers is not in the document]
Dhaiwat10: @arboleya just going ahead with singular for simplicity right now.

Issue #2917, opened 2024-08-12
Blob upload optimisations


maschad:
Follow up optimisations for #2827

- ~~Improve sync flow. Since #2692 contract deployment has been non blocking. The blob deployment process first deploys blocking blob txs then a non blocking create tx. We should make the whole flow non blocking. Full context [here](https://github.com/FuelLabs/fuels-ts/pull/2827#discussion_r1714942020).~~
- Improve chunk size estimation. Based on [discussions](https://github.com/FuelLabs/fuels-ts/pull/2827#discussion_r1711251634) with @danielbate we determined chunk estimation can be improved. The rationale is that if we know the account's balance, the transaction limit, contract limit and maximum configurable request body size, we can determine an accurate cost before deployment, and thus reduce the need for the user provided value for chunk size tolerance. 
- Blob tx retry conditions - [comment](https://github.com/FuelLabs/fuels-ts/pull/2827/files/fd95a20d87038b30aa1d03fc356648027606ff57#r1715065398)

petertonysmith94:
Wouldn't this be a breaking change?

If yes, would it be worth basing off #2827 and getting it into the release?

danielbate:
IMO the above optimisations that @maschad mentions will definitely reduce the importance of the tolerance level. However unknowns at compile time such as the configurable graph QL body size limit means we should still keep this parameter, and this not introduce a breaking change.

danielbate:
@maschad I'm going to transition this whole issue to blob follow ups.

maschad:
https://github.com/FuelLabs/fuels-ts/pull/3434#issuecomment-2512170624

PR #2929, opened 2024-08-13
feat!: non-blocking blob deployment

danielbate:
- Relates to #2917 

# Release notes

In this release, we:

- Made `ContractFactory.deployAsBlobTx` a non-blocking call

# Summary

In #2827 `ContractFactory.deployAsBlobTx` was introduced as a blocking call. This is because we need to await the blob transactions before submitting a non-blocking create transaction.

This PR makes the blob txs non blocking, however we now need to make the transaction ID returned from contract deployment into a promise. This is because we do not have access to the transaction ID until we have funded the create call, due to potentially changing UTXOs.

# Breaking Changes

The transaction ID from a contract deployment is now returned as a promise.
```ts
// before
import { ContractFactory } from 'fuels'; 

const factory = new ContractFactory(bytecode, abi, wallet);
const { waitForResult, contractId, transactionId } = await factory.deploy();
console.log(transactionId); // 0x123....
```

```ts
// after
import { ContractFactory } from 'fuels'; 

const factory = new ContractFactory(bytecode, abi, wallet);
const { waitForResult, contractId, waitForTransactionId } = await factory.deploy();
const transactionId = await waitForTransactionId();
console.log(transactionId); // 0x123....
```

# Checklist


packages/contract/src/contract-factory.ts
```diff
@@ -5,6 +5,7 @@ import type {
   CreateTransactionRequestLike,
   Provider,
   TransactionRequest,
+  TransactionResponse,
```
arboleya, line 8:
```suggestion
```

packages/contract/src/contract-factory.ts
```diff
@@ -41,8 +41,8 @@ export type DeployContractOptions = {
 } & CreateTransactionRequestLike;
 
 export type DeployContractResult<TContract extends Contract = Contract> = {
-  transactionId: string;
   contractId: string;
+  getTransactionId: () => Promise<string>;
```
danielbate, line 45: Need some thoughts on this breaking change @FuelLabs/sdk-ts.

The problem with blob transactions is we don't have the transaction ID for the create TX until all blob TXs have been completed (because of spending / changing UTXOs that _can_ change when we come to funding the create TX.

To combat this transaction ID could become a promise. For `deployAsCreateTx` this can be returned as an immediately resolved promoise.

However for `deployAsBlobTx` this will be resolved once the blob transactions have completed and we have funded the create TX.

danielbate: Additional context [here](https://github.com/FuelLabs/fuels-ts/pull/2827/files#r1716619089).

danielbate: Or also open to other/fresh approaches to this problem.

arboleya: @danielbate I think your proposal to use promises is good enough.

However, if we add breaking changes, we must hurry up.

danielbate: I'll get it ready for review.

danielbate: I'm not entirely convinced this is a better approach than #2678 but let's see what people think.

arboleya: In case we don't have time for #2678 now, do you think we should still get this PR in and tackle that later?

danielbate: We definitely don't have time for #2678. The benefit of the type change is at least `string` -> `Promise<string>` gives us a lot more room to play with, regardless of the solution. I think it's worth getting this in.

packages/contract/src/contract-factory.ts
```diff
@@ -235,7 +235,11 @@ export default class ContractFactory {
       return { contract, transactionResult };
     };
 
-    return { waitForResult, contractId, transactionId: transactionResponse.id };
+    return {
+      contractId,
+      getTransactionId: () => Promise.resolve(transactionResponse.id),
```
arboleya, line 240: Did you consider following the `waitForResult` convention?

```ts
const { waitForResult, waitForTxId } = await factory.deploy();

const transactionId = await waitForTxId();
```

packages/contract/src/contract-factory.ts
Torres-ssf: So is this problem a side effect of treating normal and big contract deployment as the same instead of letting the developers handle this on their own?

I do agree that `getTransactionId` breaking change seems to be the best approach for now.

danielbate: > So is this problem a side effect of treating normal and big contract deployment as the same instead of letting the developers handle this on their own?

Yes correct, in theory someone could manage the transactions themselves and all transaction IDs would be available in succession. But as the returned tx ID reflects the create tx ID, we don't have that till further in the flow for big deployments.

packages/contract/src/contract-factory.ts
danielbate: I did not but this is better 👍🏻

packages/contract/src/contract-factory.ts
petertonysmith94: > So is this problem a side effect of treating normal and big contract deployment as the same instead of letting the developers handle this on their own?

It does appear that way.

> I do agree that getTransactionId breaking change seems to be the best approach for now.

I agree with @Torres-ssf, it does seem the way forward now.

`waitForTxId` would be my preference of method name :)

packages/contract/src/contract-factory.ts
danielbate: https://github.com/FuelLabs/fuels-ts/pull/2929/commits/b6814ce7bf26b16ba44712356f4027c7655a78c8

apps/docs/src/guide/contracts/deploying-contracts.md
```diff
@@ -39,11 +39,11 @@ Once you have the contract artifacts, it can be passed to the `ContractFactory`
...
+This call resolves as soon as the transaction to deploy the contract is submitted and returns three items: the `contractId`, a `waitForTransactionId` function and a `waitForResult` function.
 
 <<< @/../../docs-snippets/src/guide/contracts/deploying-contracts.test.ts#deploy{ts:line-numbers}
 
-The `contract` instance will be returned only after calling `waitForResult` and waiting for it to resolve. To avoid blocking the rest of your code, you can attach this promise to a hook or listener that will use the contract only after it is fully deployed.
+The `contract` instance will be returned only after calling `waitForResult` and waiting for it to resolve. To avoid blocking the rest of your code, you can attach this promise to a hook or listener that will use the contract only after it is fully deployed. Similarly, the transaction ID is only available once the underlying transaction has been funded. To avoid blocking the code until the ID is ready, you can use the `waitForTransactionId` function to await it's retrieval.
```
Torres-ssf, line 46: @danielbate, do you think it's worth mentioning the differences between normal and large contracts? 

Deploying large contracts will take more time. Each blob transaction has to wait for a block to be produced before it gets mined. After that, there's still the wait for the normal create transaction to be deployed.

danielbate: There is a note at the bottom of the doc:

Screenshot 2024-08-14 at 16 27 28

But on second thoughts Im going to flesh it some more.

danielbate: https://github.com/FuelLabs/fuels-ts/pull/2929/commits/9bb81024e19c89b6dabcdff23c2b123d258ff1ef

Torres-ssf (dismissed):
Great job 🚀

petertonysmith94 (approved):

packages/contract/src/contract-factory.ts
```diff
@@ -281,73 +294,90 @@ export default class ContractFactory {
...
+      const transactionResponse = await account.sendTransaction(createRequest);
       const transactionResult = await transactionResponse.waitForResult<TransactionType.Create>();
       const contract = new Contract(contractId, this.interface, account) as TContract;
 
       return { contract, transactionResult };
     };
 
-    return { waitForResult, contractId, transactionId: transactionResponse.id };
+    const waitForTransactionId = async () => {
+      const maxPollingTime = 15_000_000; // 15 Minutes
```
nedsalk, line 365: I believe that instead of doing it with polling you can to this:
```ts
let txIdResolver: (value: string | PromiseLike<string>) => void;
const txIdPromise = new Promise<string>((resolve) => {
  txIdResolver = resolve;
});
// ...

txIdResolver(createRequest.getTransactionId(account.provider.getChainId()))

// ...

return { waitForResult, contractId, waitForTransactionId: async () => txIdPromise }

```

danielbate: Way cleaner, thanks.

https://github.com/FuelLabs/fuels-ts/pull/2929/commits/980ec6675533ea5b32cfeabd72544e82ed01487c

Torres-ssf (approved):

petertonysmith94 (approved):

nedsalk (approved):

Issue #2892, opened 2024-08-07
Connector may arbitrary change on component re-render if several wallets are installed


svvald:
### fuels-ts SDK Version

0.92.1

### Toolchain Versions

forc: 0.62.0, fuel-core: 0.31.0, fuel-core-keygen: 0.31.0

### Node.js Version

v20.15.1

### Browser

Google Chrome 127.0.6533.89 (Official Build) (x86_64)

### Operating System

macOS Sonoma 14.5 (23F79)

### Describe the Problem

Pre-requisites:
React web application – `react v18.3.1`, `next v14.2.5`, `@fuels/react v0.20.0`
Wallets installed – `Fuel v0.23.0`, `Fuelet v1.0.29`

Steps to reproduce:
– Establish connection with Fuelet wallet
– Re-open/refresh the page
– Interact with some component which uses several hooks from `@fuels/react`, specifically `useIsConnected`. Interaction leads to component re-render upon which connection status from the hook may change to false

While investigating I discovered that on component re-render the active connector may arbitrary change to Fuel connector thus reporting the disconnected status. On debugging I discovered the following:
– `fuel` object seems to be re-instantiated on component re-render (it is used by `@fuels/react` internally via `useFuel` hook and, apparently, has to be memoized, I'm not sure why it doesn't happen, I do not re-render the whole application tree, so the context provider value doesn't have to be changed)

– While re-instantiating `fuel`, the following methods are invoked in its constructor:
```
this.setDefaultConnector();
this._targetUnsubscribe = this.setupConnectorListener();
```
Where the first one tries to setup the default provider via storage value if it's present (it is in my case, it is Fuelet wallet) or via config otherwise. And the second one setups listeners for connectors to be added to `fuel` object once they're initialized.
`addConnector` method on its turn tries to connect to the added connector, if no current connector is present:
```
if (!this._currentConnector) {
  await this.selectConnector(connector.name, {
    emitEvents: false,
  });
}
```
So the problem here is that `setDefaultConnector` is async, but its promise is ignored in constructor, so both methods are called synchronously, which may lead to the case when previously used connector is not yet obtained from the browser storage, but listeners for connectors are already set up, their events are invoked with Fuel being the first in the order and it ends up being set up as an active connector which results in connection loss for Fuelet wallet.

[...]

PR #2930, opened 2024-08-13
fix: avoid potential race-condition in fuel connectors init

maschad:
- Closes https://github.com/FuelLabs/fuels-ts/issues/2892

# Release notes

In this release, we:
- Fixed a potential race condition where the fuel connector may disconnect upon re-render
- Introduced an `init` method for synchronous instantiation of a fuel-connectors object

# Summary

- Ensure the promise when setting up the default connector is always awaited
- Builds on @svvald 's work in https://github.com/FuelLabs/fuels-ts/pull/2943

# Checklist


packages/account/src/connectors/fuel.ts
```diff
@@ -101,11 +102,21 @@ export class Fuel extends FuelConnector implements FuelSdk {
...
+  private async initialize(): Promise<void> {
+    try {
+      await this.setDefaultConnector();
+      this._targetUnsubscribe = this.setupConnectorListener();
+    } catch (error) {
+      throw new FuelError(ErrorCode.INVALID_PROVIDER, 'Error initializing Fuel Connector');
+    }
+  }
+
+  public async init(): Promise<Fuel> {
```
nedsalk, line 117: From https://github.com/FuelLabs/fuels-ts/issues/2892#issuecomment-2294451450:
> To avoid introducing breaking the constructor for now I've added a new [init](https://github.com/FuelLabs/fuels-ts/pull/2930/files#diff-f2f37286dc75888286dbb2da3164a73b4a4a1ea8a90d87fd1ec9bcee1666097cR117-R119) method which will allow a consumer to await all initialization before continuing.

One could say that this is a breaking change because fully instantiating a connector now requires an additional step. If nothing, we should at least update the documentation accordingly as this is unintuitive.

maschad: > One could say that this is a breaking change because fully instantiating a connector now requires an additional step.

I disagree, this change does not cause any failures for consumers nor does it **require** changes, although it is recommended. Consumers who are instantiating the constructor as is and are not experiencing race conditions will remain unaffected. 

> If nothing, we should at least update the documentation accordingly as this is unintuitive.

I agree here, I've updated the docs to recommend this method of instantiating the class.

dac5906

packages/account/test/fuel-wallet-connector.test.ts
```diff
@@ -557,49 +553,16 @@ describe('Fuel Connector', () => {
...
+    hasWallet = await (
+      await new Fuel({
+        connectors: [connector],
+      }).init()
+    ).hasWallet();
 
     expect(hasWallet).toBeTruthy();
   });
-
-  it('should ensure getProvider works just fine', async () => {
```
maschad, line 571: `getProvider` has been deprecated for some time now and so we no longer need to test this method.

apps/docs/src/guide/wallets/connectors.md
```diff
@@ -319,6 +319,9 @@ It can be instantiated as follows:
 
 <<< @/../../docs-snippets/src/guide/wallets/connectors.test.ts#fuel-instantiation-1{ts:line-numbers}
 
+> [!NOTE] Note
+> We recommend initializing the Fuel class with the `init` method to avoid the any potential race conditions that may arise from the async nature of instantiating a connector.
```
danielbate, line 323:
```suggestion
> We recommend initializing the Fuel class with the `init` method to avoid any potential race conditions that may arise from the async nature of instantiating a connector.
```

danielbate, line 323:
```suggestion
> We recommend initializing the Fuel class with the `init` method to avoid any potential race conditions that may arise from the async nature of instantiating a connector.
```

danielbate (approved):

nedsalk (approved):

apps/docs-snippets/src/guide/wallets/connectors.test.ts
```diff
@@ -140,17 +140,17 @@ class WalletConnector extends FuelConnector {
  */
 describe('connectors', () => {
   describe('instantiation', () => {
-    test('should be able to instantiate a Fuel SDK', () => {
+    test('should be able to instantiate a Fuel SDK', async () => {
       // #region fuel-instantiation-1
       // #import { Fuel };
 
-      const sdk = new Fuel();
+      const sdk = await new Fuel().init();
```
arboleya, line 147: Nit - just trying to make the `init` method more evident in the docs.

```suggestion
      const sdk = await new Fuel();

      /*
        Awaits for initialization to mitigate potential race conditions
        derived from the async nature of instantiating a connector.
      */
      await sdk.init(); 
```

arboleya (dismissed):

apps/docs-snippets/src/guide/wallets/connectors.test.ts
arboleya: Edited/fixed the code suggestion.

maschad: Good call

packages/account/test/fuel-wallet-connector.test.ts
arboleya: Has the method been removed or just deprecated?

In short, I think the test should be removed only when the method is removed.

Otherwise, people may still be using it, in which case the tests are still valid.

maschad: My perspective is that if it's deprecated we no longer support or maintain it (including running tests against it), so if some experiences an issue with it with should deter them from using it. But I'm happy to leave the tests here and we can address that on another issue if needs be.

arboleya: Deprecated methods are still supported, which makes the test deletion potentially dangerous.

If we remove the test, we can introduce some changes that break it silently, and then if we publish a `patch`, which should be backward compatible, and users update to it, it may break their apps.

danielbate (approved):

arboleya (approved):

Torres-ssf:
@LuizAsFight @helciofranco do the changes made here represent any breaking change into the `useFuel` React hook?

Torres-ssf (approved):

Issue #2936, opened 2024-08-14
Transactions squeezed out in-between `submit` and `statusChange` calls hang indefinitely


nedsalk:
@Torres-ssf did tests on devnet and found that sending a transaction would sometimes hang indefinitely. After some investigation, the problem was found to be transactions being squeezed out after submitting them but _before_ a `statusChange` subscription is established. This `statusChange` subscription would then hang indefinitely because the squeezed out transaction is unavailable on the node.

### Solution

To solve this, we'd have to use the `submitAndAwait` endpoint instead of `submit` in `Provider.sendTransaction`. This way the `SqueezedOutStatus` status wouldn't be lost as it'd be sent as the second event on the subscription after `SubmittedStatus`.

There are other considerations here. For example, what happens if we submit the transaction and the connection breaks? We'd have to use the `statusChange` subscription then, but we'd be back on square one because the transaction can be squeezed out in the meantime. That would require some internal "squeezed out timer"  on `statusChange` which'd throw an error if there's no response from the node after e.g. 20 seconds, and this error would be something like "The transaction was _probably_ squeezed out".

nedsalk:
Blocked by:
- https://github.com/FuelLabs/fuel-core/issues/2108

maschad:
https://github.com/FuelLabs/fuel-core/pull/2120 has been merged so this is no longer blocked

PR #2962, opened 2024-08-16
feat!: read malleable fields from transaction status on subscription

### LINKED ISSUE #1644: Use receipts from `TransactionStatus`

nedsalk:
- Closes #1644

# Release notes

In this release, we:
- Reduced required roundtrips & bandwidth for submitting TXs

# Summary

The `TransactionResponse` class can now work with transaction ids as well as the `TransactionRequest` class.

For transaction ids the use case is as previous - you use it to fetch the _whole_ transaction as-is on the node.

 The newly added `TransactionRequest` integration serves the use case when a transaction was just submitted in `Provider.sendTransaction`. Given that `TransactionRequest` already contains _most_ of the transaction data except the malleable fields that are only set upon transaction execution (e.g. receipts), we don't need to fetch the _whole_ transaction from the node; rather, `submitAndAwait` and `statusChange` subscriptions now return that malleable data, so we only take that data and integrate it with `TransactionRequest.toTransaction` to give a full `TransactionSummary`.

I have removed `TransactionResult.gqlTransaction` because it assumed that the whole `gqlTransaction` would always be fetched inside `TransactionResponse`. Given that this is no longer the case, we cannot support this field. There's no _real_ regression here, however, because `TransactionResult.transaction` contains the same data.

# Breaking Changes

Removed TransactionResult.gqlTransaction. You can use the `TransactionResult.transaction` field and other fields on the response instead, which have all the data that `TransactionResult.gqlTransaction` has.
```ts
// before
const { gqlTransaction } = await new TransactionResponse('your-tx-id').waitForResult();
```
```ts
// after
const { transaction, id, fee, ...rest } = await new TransactionResponse('your-tx-id').waitForResult();
```

# Checklist


packages/account/src/providers/operations.graphql
```diff
@@ -61,10 +61,64 @@ fragment transactionStatusFragment on TransactionStatus {
...
+      to
+      amount
+      assetId
+    }
+    ... on ContractCreated {
+      contract
+      stateRoot
+    }
+  }
+}
```
Torres-ssf, line 102: @nedsalk Why not just get the `rawPayload` property for both `SuccessStatus` and `FailureStatus`?

Wouldn't that avoid assembling a processed TX manually later on?

nedsalk: The `rawPayload` property only exists on `Transaction`, there is none on the statuses. If you mean to fetch that property, then there wouldn't be any bandwidth benefit as I'd be fetching the whole transaction.

Torres-ssf: @nedsalk Maybe I'm missing something, but wasn't the transaction added to both the success and failure statuses?

https://github.com/FuelLabs/fuels-ts/blob/5727b17913ad7a9f5342d0d7ad057735c1d442ed/packages/account/src/providers/fuel-core-schema.graphql#L1247-L1257

https://github.com/FuelLabs/fuels-ts/blob/5727b17913ad7a9f5342d0d7ad057735c1d442ed/packages/account/src/providers/fuel-core-schema.graphql#L1300

Torres-ssf: My idea was to have the `rawPayload` property available within a `submitAndAwait`/`statusUpdate` response to avoid manually assembling the TX

nedsalk: Yes, but we don't want to fetch the whole transaction via `rawPayload` because we already have most of it locally via `TransactionRequest.toTransaction`. We just want to fetch the _malleable_ fields who's values are set by the node.

packages/account/src/providers/transaction-response/transaction-response.ts
```diff
@@ -146,20 +255,14 @@ export class TransactionResponse {
...
+          this.status = statusChange;
           break;
         }
       }
 
       return this.fetch();
     }
 
-    // Refetch if the expected status is not the same as the response status
-    if (this.expectedStatus && response.transaction.status?.type !== this.expectedStatus) {
```
nedsalk, line 158: I removed this polling code added in #2896. The problem was happening because `statusChange` sent a `SuccessStatus` but fetching the whole transaction with receipts would return a transaction in `SubmittedStatus`. Because we're now reading all the malleable fields from the subscription directly and not doing that additional fetch, there's no need for this solution.

packages/account/src/providers/fuel-graphql-subscriber.ts
```diff
@@ -10,14 +10,12 @@ type FuelGraphQLSubscriberOptions = {
...
-  private stream!: ReadableStreamDefaultReader<Uint8Array>;
   private static textDecoder = new TextDecoder();
 
-  public constructor(private options: FuelGraphQLSubscriberOptions) {}
-
-  private async setStream() {
-    const { url, query, variables, fetchFn } = this.options;
+  private constructor(private stream: ReadableStreamDefaultReader<Uint8Array>) {}
 
+  public static async create(options: FuelGraphQLSubscriberOptions) {
```
nedsalk, line 17: I moved `FuelGraphqlSubcriber` instantiation to be async as a preparatory step for #2936. I tried introducing the solution for that issue here in commit https://github.com/FuelLabs/fuels-ts/pull/2962/commits/80511131b68754288f7e686fc040413e8ea14141 but I reverted it because the CI disagreed that it's a solution :sweat_smile:. 

This change is necessary for the usage of `submitAndAwait` endpoint because the transaction can fail when submitted, but the failure isn't manifested until a call to the `next` method is done. I will add this failure handling in the subsequent PR that will tackle #2936, after I sync with the `fuel-core` team. We currently aren't using `submitAndAwait` anywhere so it's not a problem to leave it as-is.

.changeset/famous-pans-press.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/account": minor
+---
+
+feat!: read malleable fields from transaction status on subscription
```
arboleya, line 5: Nit - in this release, we:
 - Reduced required roundtrips & bandwidth for submitting TXs

petertonysmith94 (commented):
@nedsalk should the transaction response created via an ID, be the same as that created via a request?

packages/account/src/providers/operations.graphql
```diff
@@ -61,10 +61,64 @@ fragment transactionStatusFragment on TransactionStatus {
   }
 }
 
+fragment malleableFieldsFragment on Transaction {
```
petertonysmith94, line 64: Nit: `malleableTransactionFieldsFragment` would be more descriptive.

petertonysmith94 (approved):
Small nitty comments :)

packages/account/src/providers/transaction-response/transaction-response.ts
```diff
@@ -78,10 +83,40 @@ export type TransactionResultReceipt =
...
+      case 'VariableOutput':
+        return { ...obj, type: OutputType.Variable } satisfies OutputVariable;
+      case 'ContractCreated':
+        return {
+          ...obj,
+          type: OutputType.ContractCreated,
+          contractId: obj.contract,
+        } satisfies OutputContractCreated;
+      default:
+        return assertUnreachable(obj);
```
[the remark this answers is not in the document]
nedsalk: The `assertUnreachable` will throw a compile-time error if the switch doesn't exhaust all types in the union, and thus won't allow a build at all. I think it's a _very_ useful thing to have for all switches. I could put the error you mentioned in `assertUnreachable`, though I'm not sure if we should do it because, in my mind, `FuelError` should be related to domain-specific errors, not stuff like unreachable code.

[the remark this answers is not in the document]
petertonysmith94: This is magic - I approve!

petertonysmith94 (approved):
Nice!

packages/account/src/providers/transaction-response/transaction-response.ts
```diff
@@ -93,22 +130,23 @@ export class TransactionResponse {
...
   /**
    * Constructor for `TransactionResponse`.
    *
-   * @param id - The transaction ID.
+   * @param tx - The transaction ID or TransactionRequest.
    * @param provider - The provider.
    */
-  constructor(id: string, provider: Provider, abis?: JsonAbisFromAllCalls) {
-    this.id = id;
+  constructor(tx: string | TransactionRequest, provider: Provider, abis?: JsonAbisFromAllCalls) {
```
danielbate, line 144: Not a fan of the type overload here, dealing with just the ID was fine IMO.

danielbate (approved):

packages/account/src/providers/transaction-response/transaction-response.ts
danielbate: Nevermind @petertonysmith94 taught me the error of my ways, we reusing the request as we have the fields, otherwise we fetch, of course 🤦🏻

nedsalk: The alternative would be:
```ts
constructor(
  id: string, 
  provider: Provider, 
  abis?: JsonAbiFromAllCalls, 
  private request?: TransactionRequest) { ... }

// Provider.sendTransaction
return new TransactionResponse(
  transactionRequest.getTransactionId(this.getChainId(),
  this,
  abis,
  transactionRequest)
```

The type union makes the two use cases of `TransactionResponse` explicit; you either specify the id and we'll fetch everything else, or you give us your `TransactionRequest` of the transaction you just sent and we'll fetch only the necessary data.

danielbate: Makes sense, LGTM 🫡

Torres-ssf (approved):

PR #2964, opened 2024-08-16
chore!: wrap subscriptions in promise

nedsalk:
# Summary

I moved `FuelGraphqlSubcriber` instantiation to be async as a preparatory step for #2936. I tried introducing the solution for that issue in #2962, commit https://github.com/FuelLabs/fuels-ts/pull/2962/commits/80511131b68754288f7e686fc040413e8ea14141, but I wasn't happy with the solution so I reverted it and created this to give myself room to figure out a solution.

This change is necessary for the usage of `submitAndAwait` endpoint because the transaction can fail when submitted, but the failure isn't manifested until a call to the `FuelGraphqlSubscriber.next` method is done. With this async instantiation, the subscription can be made to fail _before_ the actual subscription iterator is returned. I will add this failure handling in a subsequent PR that will tackle #2936, after I sync with the `fuel-core` team. We currently aren't using `submitAndAwait` anywhere so it's not a problem to leave it unhandled as-is.

# Breaking Changes

```ts
// before
const subscription = provider.operations.statusChange({ transactionId });
for await (const response of subscription) { ... }

const submittedTxSubscription = this.operations.submitAndAwait({ encodedTransaction: '0x' });
```

```ts
// after
const subscription = await provider.operations.statusChange({ transactionId });
for await (const response of subscription) { ... }

const submittedTxSubscription = await this.operations.submitAndAwait({ encodedTransaction: '0x' });
```

# Checklist

### DIFF
--- a/.changeset/smart-seals-care.md
+++ b/.changeset/smart-seals-care.md
@@ -1,0 +1,5 @@
+---
+"@fuel-ts/account": minor
+---
+
+chore!: wrap subscriptions in promise
--- a/packages/account/src/providers/fuel-graphql-subscriber.ts
+++ b/packages/account/src/providers/fuel-graphql-subscriber.ts
@@ -10,14 +10,12 @@ type FuelGraphQLSubscriberOptions = {
 };
 
 export class FuelGraphqlSubscriber implements AsyncIterator<unknown> {
-  private stream!: ReadableStreamDefaultReader<Uint8Array>;
   private static textDecoder = new TextDecoder();
 
-  public constructor(private options: FuelGraphQLSubscriberOptions) {}
+  private constructor(private stream: ReadableStreamDefaultReader<Uint8Array>) {}
 
-  private async setStream() {
-    const { url, query, variables, fetchFn } = this.options;
-
+  public static async create(options: FuelGraphQLSubscriberOptions) {
+    const { url, query, variables, fetchFn } = options;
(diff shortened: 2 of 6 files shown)


petertonysmith94 (approved):
May be worth adding `submitAndAwait` to the breaking changes (to be thorough):

```ts
// Before
this.operations.submitAndAwait({ encodedTransaction: '0x' })
```

```ts
// After
await this.operations.submitAndAwait({ encodedTransaction: '0x' })
```

LGTM though, nice one!

maschad (approved):

arboleya (commented):
> I tried introducing the solution for that issue in #2962, commit https://github.com/FuelLabs/fuels-ts/pull/2962/commits/80511131b68754288f7e686fc040413e8ea14141, but I reverted it afterwards because it caused issues.

**How this reads**: _We can't merge #2962 and #2963 together, or we'll have issues_.

nedsalk:
@arboleya it's not that we'll have issues by implementing both #2936 and #2962 together but that my solution for #2936 was naive so I reverted it. I actually made it work locally and all the tests were passing but when it came to the CI it started failing because the CI machine is slower and I was doing some `sleep(x)` tricks which just failed in the CI because the chosen `x` was not enough. It was generally a bad take so I just reverted it.

danielbate (approved):

Issue #2939, opened 2024-08-15
Transaction Summary showing incorrect operations for faucet transaction


LuizAsFight:
### fuels-ts SDK Version

0.93.0 (next)

### Describe the Problem
The operations being shown for a faucet transaction are like following:
<img width="348" alt="image" src="https://github.com/user-attachments/assets/53185345-0bb4-48af-bf74-616c5ecca8b0">

### What's wrong
- The first operation is a transfer from the user account to the user account
- The second operation is a transfer from the user account to the faucet account

### How it should be
- One transfer operation from the faucet account to the user account with the faucet amount ( 0.002 ETH )
- One transfer operation from the faucet account to the faucet account ( remaining ETH amount from faucet account )

Torres-ssf:
@LuizAsFight, I do not completely understand the problem here.

> How it should be
> - One transfer operation from the faucet account to the user account with the faucet amount ( 0.002 ETH )

This one is clear. The first shown operation cleared is wrong.

> One transfer operation from the faucet account to the faucet account ( remaining ETH amount from faucet account )

Why does the second operation needs to show a transfer operation from and to the faucet account with the remaining ETH amount owned by the faucet itself? 

Seems to me that this operation shouldn't exist in the first place.

LuizAsFight:
you're correct it shouldn't exist, but if you look at the faucet transactions, instead of putting the "ChangeOutput" as the faucet account, it does another transfer to the faucet account

Torres-ssf:
@LuizAsFight You’re right. The Faucet transfer indeed produces an unconventional use of outputs.

```ts
// Faucet Transfer TX outputs 
[
  {
    type: 0, // CoinOut to my address -> Everything is fine 👍
    to: '0xc2714e9248bc19510c412df5c6a6242abe36be85856e8be419baa78334ac7070', // My address
    amount: <BN: 0x1e8480>, // Normal value transferred by the faucet
    assetId: '0xf8f8b6283d7fa5b672b530cbb84fcccb4ff8dc40f8176ef4544ddb1f1952ad07'
  },
  {
    type: 2, // ChangeOutput to my address -> This was supposed to be the faucet address 👎
    to: '0xc2714e9248bc19510c412df5c6a6242abe36be85856e8be419baa78334ac7070', // My address
    amount: <BN: 0xc088a>, // It uses only a small amount.
[...]

PR #2969, opened 2024-08-17
testing: account transfer operation

petertonysmith94:
# Summary

- Writing tests around the transfer operation logic
  - Test for a use case of unlocking a predicate
  - Test for a use case of two accounts transferring the same amount of the same asset to another account

# Checklist

### DIFF
--- a/packages/fuel-gauge/src/transaction-summary.test.ts
+++ b/packages/fuel-gauge/src/transaction-summary.test.ts
@@ -381,6 +381,64 @@ describe('TransactionSummary', () => {
       });
     });
 
+    it('should ensure transfer operation is assembled (TWO ACCOUNTS TO ANOTHER ACCOUNT)', async () => {
+      const amountFromA = 10;
+      const amountFromB = 10;
+
+      using launched = await launchTestNode();
+
+      const {
+        provider,
+        wallets: [genesis, feePayee],
+      } = launched;
+
+      const baseAssetId = provider.getBaseAssetId();
+
+      // Sender A
+      const senderA = Wallet.generate({ provider });
+      const txA = await genesis.transfer(senderA.address, amountFromA, baseAssetId);
+      await txA.waitForResult();
+
+      // Sender B
+      const senderB = Wallet.generate({ provider });
+      const txBase = await genesis.transfer(senderB.address, amountFromB, baseAssetId);
+      await txBase.waitForResult();
+
+      const recipient = Wallet.generate({ provider });
+
+      let request = new ScriptTransactionRequest();
+
+      const resourcesA = await senderA.getResourcesToSpend([[amountFromA, baseAssetId]]);
+      const resourcesB = await senderB.getResourcesToSpend([[amountFromB, baseAssetId]]);
+      request.addResources([...resourcesA, ...resourcesB]);
+      request.addCoinOutput(recipient.address, amountFromA, baseAssetId);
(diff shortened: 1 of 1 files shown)


packages/fuel-gauge/src/transaction-summary.test.ts
```diff
@@ -314,6 +314,131 @@ describe('TransactionSummary', () => {
       });
     });
 
+    it('should ensure transfer operation is assembled (UNLOCKING PREDICATE)', async () => {
```
petertonysmith94, line 317: This test I was just playing around with an additional use case and exploring what should be display.

I'm happy with this case.

packages/fuel-gauge/src/transaction-summary.test.ts
```diff
@@ -314,6 +314,131 @@ describe('TransactionSummary', () => {
...
+        recipients: [
+          {
+            address: recipient.address,
+            quantities: [{ amount: amountToUnlockFromPredicate, assetId: baseAssetId }],
+          },
+        ],
+      });
+    });
+
+    it('should ensure transfer operation is assembled (TWO ACCOUNTS TO ANOTHER ACCOUNT)', async () => {
```
petertonysmith94, line 384: This case - when transfering the same quantities from two different accounts, then the resultant sender will be one of the accounts. Once I change the amount so they're not equal, we get two distinct operations (one with sender from A, and one from B) - which is what I expected in this case.

Torres-ssf: @petertonysmith94 Which one was not the expected case? Having just one sender?

Why this was not expected?

petertonysmith94: I would expect there to be two senders as there are two senders...?

When I updated `amountFromB` to 11, I obtained what I expected - two distinct transfer options for both senders (A and B).

Torres-ssf: @petertonysmith94 But how do you know that there were 2 senders?

In your example, you funded the TX with UTXOs from 2 users, and have added `CoinOutputs` with predefined values, but how do you know that all fetched UTXOs were spent? 

The first funding user might have a UTXO with the amount of 1 ETH, and the VM could have used just this one to pay for the whole TX, and the other UTXO(s), including the ones from the second user, were not spent at all.

There is only so much where this validation can go through. Trying to cover all the edge cases at some point will become just assumptions. 

We could use the amount on the `ChangeOutput` and the fee value to better determine who transfers to who. However, the VM only accepts one `ChangeOutput` for each `assetId`. So it will be available only for just one sender.

It means that in your example case, if the VM chooses to spend 1 ETH UTXO to pay the whole TX, from the user that does not have its address specified on the `ChangeOutput` (second user since it was the last one to add resources to the TX), the user will lose all its change.

It is very likely that what is happening [here](https://github.com/FuelLabs/fuels-ts/issues/2939#issuecomment-2291491338) is because of a similar reason. Whoever implemented this transfer decided to set the users to receive the change of the spent UTXOs. And since the `ChangeOuptut` is already set for someone, they had added another `CoinOutput` transferring to self, to not lose UTXOs.

petertonysmith94: Thank you @Torres-ssf for the through explanation, it's helped me better understand what is happening and my assumptions have been wrong in this case.

PR #2976, opened 2024-08-19
chore: group all dependabot upgrades

petertonysmith94:
# Summary

- Further groups dependabot upgrades by patch/minor and major. 
- This _should_ alleviate the issue we're experiencing when once we merge a single dependabot PR, we _usually_ have conflicts in all subsequential PR's

# Checklist


Torres-ssf (approved):

.github/dependabot.yml
```diff
@@ -10,20 +10,12 @@ updates:
       - "chore"
     commit-message:
       prefix: "build(deps):"
-      prefix-development: "build(deps-dev):"
     versioning-strategy: "increase"
     groups:
-      dev-deps:
-        dependency-type: "development"
+      all:
```
arboleya, line 15: Should we at least group `patch` updates from the rest? The rationale is that patches should always be backward-compatible and one click away, whereas minors/majors may need manual intervention, which could block the whole PR.

danielbate: https://github.com/FuelLabs/fuels-ts/pull/2976/commits/35debaa057b7903279d6620f9ae78b8e4b7b7bb1

danielbate (approved):

.github/dependabot.yml
arboleya: This way, we'll end up with four groups, right?

Could we achieve having only two?

1. deps + deps dev (patches)
1. deps + deps dev (major + minor)

danielbate: Correct, the docs sounds like it's going to split them so we lose the prefixing, but that's not necessarily a problem. And I might be wrong. [Let's see](https://github.com/FuelLabs/fuels-ts/pull/2976/commits/fbc2f587f74a499b2358066c91a5ae18d1d7eac9).

danielbate (approved):

arboleya (approved):

Torres-ssf (approved):

PR #2985, opened 2024-08-20
fix: made bytecode compression browser compatible

### LINKED ISSUE #2984: Re-evaluate usage of `Buffer` in bytecode compression
Incoming from the [forum](https://forum.fuel.network/t/new-ts-sdk-requires-buffer-which-hinders-frontend-dev/6487)

> The latest version of the TS SDK (0.94.0) uses decompressBytecode function in the factory, which requires the node builtin Buffer, which isn’t available in most Frontend Envs and causes issues especially with Vite.

This causes the following issue on a frontend application:

> ReferenceError: Buffer is not defined

petertonysmith94:
- Closes #2984 

# Release notes

In this release, we:

- Enabled typegen'd bytecode compression, to work in the browser.

# Checklist


.github/workflows/pr-release.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
     name: "Release PR to npm"
     runs-on: ubuntu-latest
     # comment out if:false to enable release PR to npm
-    if: false
+    # if: false
```
```suggestion
    if: false
```

Confirmed working in browser

packages/utils/src/utils/bytecode.ts
```diff
@@ -3,22 +3,28 @@ import { gzipSync, gunzipSync } from 'fflate';
...
+  if (!bytecodeAsBinary) {
     return '';
   }
 
-  const bytecodeBytes = arrayify(bytecode);
-  const bytecodeGzipped = gzipSync(bytecodeBytes);
-  const bytecodeEncoded = Buffer.from(bytecodeGzipped).toString('base64');
+  const bytecodeCompressBytes = arrayify(bytecodeAsBinary);
+  const bytecodeCompressGzipped = gzipSync(bytecodeCompressBytes);
+  const bytecodeCompressBinary = String.fromCharCode.apply(
```
maschad, line 13: Just a suggestion to reduce the type casting

```suggestion
const bytecodeCompressBinary = Uint8Array.from(bytecodeCompressGzipped).reduce(
    (str, byte) => str + String.fromCharCode(byte),
    ''
  );
```

maschad (commented):
Nice work @petertonysmith94 , we should probably disable the `Buffer` polyfill in the browser [test config](https://github.com/FuelLabs/fuels-ts/blob/ps/fix/compression-buffer-not-defined/vitest.browser.config.mts#L12-L17) so that our tests catch instances where we may expect browser compatibility but we have false negatives due to node polyfills. I noted that quite a few tests failed when I did that.

Torres-ssf (approved):

arboleya (approved):

packages/utils/src/utils/bytecode.ts
danielbate: `apply` is significantly more performant than `reduce` here so will resolve.

nedsalk (approved):

danielbate (approved):
@maschad I've added #3007 for us to look into reducing the number of polyfills we need.

Issue #2924, opened 2024-08-13
Handle unknown tx variants gracefully


arboleya:
Much like the Rust counterpart in https://github.com/FuelLabs/fuels-rs/issues/1477.

PR #3003, opened 2024-08-22
feat: added support for unknown txn types

maschad:
- Closes https://github.com/FuelLabs/fuels-ts/issues/2924
- Related to https://github.com/FuelLabs/fuels-rs/issues/1477

# Release notes

In this release, we:

- Added future-proofing to fail gracefully for different transaction types which may exist later on.

# Summary

This PR introduces a `console.warn` for transaction types which the SDK may not support yet. I've made an assumption about what a [`BaseTransactionType`](https://github.com/FuelLabs/fuels-ts/pull/3003/files#diff-09abe134d06fc6102e5aa90ea9bf92c090119ee9aca9f9a9000dcae7827d94c4R35-R60) would contain as to distinguish it from a completely unsupported transaction. 

# Checklist


packages/transactions/src/coders/transaction.ts
```diff
@@ -669,13 +580,79 @@ export class TransactionBlobCoder extends Coder<TransactionBlob, TransactionBlob
...
 }
 
+export type TransactionUnknown = BaseTransactionType & {
+  type: TransactionType.Unknown;
+
+  /** Data of the transaction */
+  data?: BytesLike;
+};
+
+export class TransactionUnknownCoder extends Coder<TransactionUnknown, TransactionUnknown> {
```
danielbate, line 590: Given we don't know the spec of an unknown transaction and won't be transmitting it, why would we want to encode/decode it?

packages/transactions/src/coders/transaction.ts
```diff
@@ -669,13 +580,79 @@ export class TransactionBlobCoder extends Coder<TransactionBlob, TransactionBlob
...
+    parts.push(new NumberCoder('u16', { padToWordSize: true }).encode(value.witnessesCount));
+    parts.push(new PoliciesCoder().encode(value.policies));
+    parts.push(new ArrayCoder(new InputCoder(), value.inputsCount).encode(value.inputs));
+    parts.push(new ArrayCoder(new OutputCoder(), value.outputsCount).encode(value.outputs));
+    parts.push(new ArrayCoder(new WitnessCoder(), value.witnessesCount).encode(value.witnesses));
+
+    return concat(parts);
+  }
+
+  decode(data: Uint8Array, offset: number): [TransactionUnknown, number] {
```
danielbate, line 610: Decoding here seems dangerous given we don't know the structure of the bytes.

packages/account/src/account.ts
```diff
@@ -632,6 +633,9 @@ export class Account extends AbstractAccount {
       );
     }
     const transactionRequest = transactionRequestify(transactionRequestLike);
+    if (isTransactionTypeUnknown(transactionRequest)) {
```
danielbate, line 636: Should this be warning before returning the transaction response?

packages/account/src/predicate/predicate.ts
```diff
@@ -117,6 +118,9 @@ export class Predicate<
    */
   sendTransaction(transactionRequestLike: TransactionRequestLike): Promise<TransactionResponse> {
     const transactionRequest = transactionRequestify(transactionRequestLike);
+    if (isTransactionTypeUnknown(transactionRequest)) {
```
danielbate, line 121: As above, should this be warning?

packages/account/src/providers/transaction-request/unknown-transaction-request.ts
```diff
@@ -0,0 +1,57 @@
...
+import { BaseTransactionRequest, TransactionType } from './transaction-request';
+
+/**
+ * @hidden
+ */
+export interface UnknownTransactionRequestLike extends BaseTransactionRequestLike {
+  data?: BytesLike;
+}
+
+export class UnknownTransactionRequest extends BaseTransactionRequest {
```
danielbate, line 15: Why would someone want to create an instance of an unknown transaction?

danielbate (commented):
Potentially I'm missing something but I'm not sure we need to treat this in the same way as other transaction types, with a request class and encoding. 🤔

packages/account/src/providers/transaction-request/unknown-transaction-request.ts
Torres-ssf: @maschad, I might be missing some context, but I also believe we don't need an Unknown transaction request. If there's something I'm overlooking, please let me know.

arboleya: I second others' thoughts. Ideally, this PR would be [minimal](https://github.com/FuelLabs/fuels-rs/issues/1477#issuecomment-2285907223).
 1. Try/catch unsupported TXs
 1. Ignoring them
 1. Print warning
 1. Move on

packages/account/src/account.ts
maschad: The [warning](https://github.com/FuelLabs/fuels-ts/pull/3003/files#diff-a7b58f093b62c1f41f0e8a5ea0700c9a0ba62d9f16a62c90e7e451cf30dba091R47-R49) is done in the `transactionRequestify` function which is what [this test](https://github.com/FuelLabs/fuels-ts/pull/3003/files#diff-3ccf1abbc4dc63f3169e41bf35f6a65f268903ebf6185701aff6ec308e975ce7R25-R54) verifies.

packages/account/src/predicate/predicate.ts
maschad: see [this comment](https://github.com/FuelLabs/fuels-ts/pull/3003#discussion_r1729473790)

packages/account/src/providers/transaction-request/unknown-transaction-request.ts
maschad: The `UnknownTransactionRequest` is needed because ultimately because the `sendTransaction` API must return a `TransactionResponse` otherwise it would be an error, in order to construct a `TransactionResponse` you must have a `TransactionRequest` . I've added the `hidden` tag as this is just purely internal mechanics. 

See 6b88425ac

packages/transactions/src/coders/transaction.ts
maschad: Good point @danielbate , I've removed the encoding/decoding 

6b88425ac

packages/transactions/src/coders/transaction.ts
maschad: Good point 6b88425ac

maschad:
> Potentially I'm missing something but I'm not sure we need to treat this in the same way as other transaction types, with a request class and encoding. 🤔

You make a good point re: enconding / decoding @danielbate  so I agree with you there, but I disagree re: the request class. 

Here's my rationale (which I tried to allude to in the PR description)

But the impetus for this PR is for consumers sending transactions to have _graceful_ failures, so therefore an `UnknownTransaction` is a _type_ of transaction which is not yet supported, and thus it should have at least the base qualities of a transaction with some new / unsupported properties. For that reason, and to avoid breaking types in the responses that consumers work with, it needs to be treated as a `TransactionRequestLike` class instance so that it can be treated as a `TransactionResponse` otherwise the behaviour would be the same as throwing an error.

Perhaps there may be another way to achieve the same result but I believe this most minimal way to do so, open to hearing peoples thoughts.

Torres-ssf:
> Perhaps there may be another way to achieve the same result but I believe this most minimal way to do so, open to hearing your and other peoples thoughts. cc @arboleya @Torres-ssf

@maschad, I believe a minimal solution for this issue would involve addressing three key methods: `Provider.getTransaction`, `Provider.getTransactions`, and `TransactionResponse.create`.

1 - `TransactionResponse.create`

This is the most complex part. A user could potentially call it like this:

```ts
const response = await TransactionReponse.create(idFromUnsupportedTx)
```

We could add a try/catch block during the [decoding](https://github.com/FuelLabs/fuels-ts/blob/682d62eb7c45cfb04816378b17453a7e5e25fd2b/packages/account/src/providers/transaction-response/transaction-response.ts#L272) to handle any errors. However, things become tricky afterward, as we can't simply throw the error—we still need to return a return a [transaction result](https://github.com/FuelLabs/fuels-ts/blob/682d62eb7c45cfb04816378b17453a7e5e25fd2b/packages/account/src/providers/transaction-response/transaction-response.ts#L85) with [TX summary](https://github.com/FuelLabs/fuels-ts/blob/682d62eb7c45cfb04816378b17453a7e5e25fd2b/packages/account/src/providers/transaction-summary/assemble-transaction-summary.ts#L106), just like any other regular transaction.

This might be where we need to introduce a `TransactionUnsupported`:

```ts
export type TransactionUnsupported = {
  type: TransactionType.Unsupported;

  // TX Bytes
  bytes: UInt8Array
};
```

This type might be necessary because the transaction property is required  [here](https://github.com/FuelLabs/fuels-ts/blob/682d62eb7c45cfb04816378b17453a7e5e25fd2b/packages/account/src/providers/transaction-summary/assemble-transaction-summary.ts#L129). Most other properties can be either booleans (defaulting to false), empty arrays, or values extracted from the receipts.

2 - `getTransactions`:

Since this method returns an array of transactions, we could simply log a warning and skip unsupported types, or alternatively, return them as `TransactionUnsupported`.

3 - `getTransaction`

This method returns a single transaction object. We could return a `TransactionUnsupported` here. If not, it's unclear to me what else we could return in this case.

packages/transactions/src/coders/transaction.ts
```diff
@@ -27,22 +28,27 @@ export enum TransactionType /* u8 */ {
   Upgrade = 3,
   Upload = 4,
   Blob = 5,
+  Unknown = 6,
```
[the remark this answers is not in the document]
maschad: That's a good point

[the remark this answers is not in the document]
maschad: I've undone this change, for two reasons: 

1) It's a reasonable assumption that there will never be a Transaction encoding that will ever begin with `-1` , certainly not intentionally.

2) that collusion is actually desirable, because the presumption is that a new transaction type would actually begin with the hex encoded `6` as that's the next available integer.

606bae5c6

maschad:
As discussed on 26-08-2024 and following @Torres-ssf suggestions, I've added capabilities to handle the scenarios for `getTransactions` and `getTransaction` 606bae5c6

We agreed we willl not have to handle the decoding in any nuanced way as any transaction hash, which, once decoded begins with a number greater than or equal to `6` will be treated as an _unknown_ and a console warning will be logged. 

> Does this PR address the following:
> 
> > New input or output types (i.e CoinInputV2)
> > New opcodes w/ brand new gas price in the consensus params
> 
> Or do we already handle these?

@petertonysmith94 we will not have to handle these as we are not decoding the transaction in it's entirety once it begins with the `UNKNOWN` type

packages/account/src/providers/transaction-request/unknown-transaction-request.ts
maschad: As we had discussed the goal is not for consumers to create this type (as they would have created a type that is unknown to the SDK i.e. has a encoded hash >= 6)

packages/transactions/src/coders/transaction.ts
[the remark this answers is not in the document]
petertonysmith94: Are we making an assumption that there will be a new transaction variant of type `6` for _all_ unknown transaction types? Might it be worthwhile confirming this assumption as that's not how I read the issue.

What happens when we add a new transaction variant for `ScriptV2`? This could be type `7` - will this not throw an error?

petertonysmith94:
> @petertonysmith94 we will not have to handle these as we are not decoding the transaction in it's entirety once it begins with the `UNKNOWN` type

We call `getBaseTransaction` within the `UnknownTransactionRequest`.

https://github.com/FuelLabs/fuels-ts/blob/e7daa0515d54e3317ecbc9a9830b2f52875ed1a2/packages/account/src/providers/transaction-request/unknown-transaction-request.ts#L51-L57

This performs `inputlify` and `outputlify` on all the inputs and outputs (respectively) of that request.

https://github.com/FuelLabs/fuels-ts/blob/e7daa0515d54e3317ecbc9a9830b2f52875ed1a2/packages/account/src/providers/transaction-request/transaction-request.ts#L178-L179

We throw an error within these functions when the `InputType` and `OutputType` are not known. If there was to be a new input or output type (i.e CoinInputV2), would this not throw an error?

packages/transactions/src/coders/transaction.ts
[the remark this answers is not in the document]
maschad: The assumption is the variants would be of type 6 **or greater**. 

The type when deserialized would be either `6` or greater, in which both cases when decoded [would thrown an error](https://github.com/FuelLabs/fuels-ts/pull/3003/files#diff-09abe134d06fc6102e5aa90ea9bf92c090119ee9aca9f9a9000dcae7827d94c4R699-R700) but that error is suppressed but the methods the consumer interacts with i.e.

- `getTransaction`
- `getTransactions`
- `sendTransaction`

maschad:
> We throw an error within these functions when the `InputType` and `OutputType` are not known. If there was to be a new input or output type (i.e CoinInputV2), would this not throw an error?

My apologies I should have been more explicit, the [context](https://github.com/FuelLabs/fuels-ts/pull/3003#issuecomment-2310202479) of this PR is that should a consumer try to send or retrieve a transaction that is currently not supported that it will fail gracefully. Those methods handle this error that would be thrown as discussed [here](https://github.com/FuelLabs/fuels-ts/pull/3003#discussion_r1733465552)

The only scenario someone could call `toTransaction` on an `UnknownTransactionRequest` type without interacting with those methods would be if they instantiated the class explicitly, in which case they would correctly have those errors thrown.

packages/transactions/src/coders/transaction.ts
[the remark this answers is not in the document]
Torres-ssf: I was under the impression that the `UnknownTransaction` type is intended to serve as a placeholder when the TypeScript SDK cannot identify the type of a given transaction. 

Based on this, @petertonysmith94's suggestion makes sense—using the type as `-1` seems like the appropriate approach.

packages/account/src/providers/provider.ts
```diff
@@ -739,12 +740,15 @@ Supported fuel-core version: ${supportedVersion}.`
...
-  // #region Provider-sendTransaction
   async sendTransaction(
     transactionRequestLike: TransactionRequestLike,
     { estimateTxDependencies = true }: ProviderSendTxParams = {}
   ): Promise<TransactionResponse> {
     const transactionRequest = transactionRequestify(transactionRequestLike);
+    if (isTransactionTypeUnknown(transactionRequest)) {
+      return new TransactionResponse(transactionRequest, this);
+    }
+    // #region Provider-sendTransaction
```
Torres-ssf, line 751: @maschad My understanding is that if someone attempts to submit a transaction that the SDK does not support, we should throw an error rather than simply resolving the `sendTransaction` promise with a dummy response.

Torres-ssf (changes requested):

packages/transactions/src/coders/transaction.ts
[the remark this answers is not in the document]
maschad: The `TransactionType` [is decoded](https://github.com/FuelLabs/fuels-ts/blob/mc/chore/handle-unknown-txns/packages/transactions/src/coders/transaction.ts#L671-L672) from the serialized transaction, what reason do you have to believe that the Fuel specification would intentionally encode  `-1` as a transaction type in the future? `6` is far more likely.

packages/account/src/providers/provider.ts
maschad: @Torres-ssf If a consumer calls [`batchTransfer`](https://github.com/FuelLabs/fuels-ts/blob/mc/chore/handle-unknown-txns/packages/account/src/account.ts#L371) (which calls `sendTransaction`) for 10 transactions and 1 of them is currently unsupported by the SDK version should all transactions not be submitted and an error thrown?

packages/transactions/src/coders/transaction.ts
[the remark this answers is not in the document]
petertonysmith94: I'm under the same impression as @Torres-ssf - that unknown type was just a placeholder and not going to added to the specification with a type of 6.

My understand is that a new `ScriptV2` _could_ be added with type `6`, then another called `CreateV2` could be added with type `7` and so on... I feel we need to test these assumptions and clear up the disparity between our understandings.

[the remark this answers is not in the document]
petertonysmith94: > The transaction type when deserialized would be either `6` or greater, in which both cases when decoded [would thrown an error](https://github.com/FuelLabs/fuels-ts/pull/3003/files#diff-09abe134d06fc6102e5aa90ea9bf92c090119ee9aca9f9a9000dcae7827d94c4R699-R700) but that error is suppressed by the methods the consumer interacts with i.e.
> - `getTransaction`
> - `getTransactions`
> - `sendTransaction`

Are we not just delaying the surfacing of these errors?

For example, with the `sendTransaction`, we still return a `TransactionResponse`. If the consumer is calling `getTransactionSummary` then this would lead to an error being thrown would it not?

maschad:
After discussing with @Torres-ssf and @petertonysmith94 on a call, we decided it's best to remove the `UNKNOWN` txn type as it's major use case would be for sending transactions, particularly batch transactions. This would perhaps be unideal as those transactions may have interdependence, in which case, it would not be advisable to submit some of the transactions excluding the unsupported ones. 

The PR has now been updated to handle the scenario where we are unable to decode a transaction hash when calling `getTransaction` or `getTransactions`  , and thus it logs a warning and returns a null response.

packages/account/src/providers/provider.ts
```diff
@@ -739,12 +739,12 @@ Supported fuel-core version: ${supportedVersion}.`
    * @param sendTransactionParams - The provider send transaction parameters (optional).
    * @returns A promise that resolves to the transaction response object.
    */
-  // #region Provider-sendTransaction
   async sendTransaction(
     transactionRequestLike: TransactionRequestLike,
     { estimateTxDependencies = true }: ProviderSendTxParams = {}
   ): Promise<TransactionResponse> {
     const transactionRequest = transactionRequestify(transactionRequestLike);
+    // #region Provider-sendTransaction
```
Torres-ssf, line 747: @maschad Why this code snippet region was placed in a different line?

maschad: In my opinion it better articulates the process of the dependency estimation,  for example the lines related to `transactionRequest` are irrelevant.

# Before

Screenshot 2024-08-29 at 1 35 30 PM


# After

Screenshot 2024-08-29 at 1 37 56 PM

Torres-ssf (approved):

petertonysmith94 (approved):

danielbate (approved):
Nice changes 👍🏻

arboleya (approved):
Clean! 🎯

PR #3038, opened 2024-08-27
fix: `fuels dev` cleanup not killing node

### LINKED ISSUE #2889: Exiting `fuels dev` with `CTRL+C` (`SIGINT`) doesn't kill `fuel-core` node
As the title says. Run the following: `echo | pnpm create fuels --pnpm && cd my-fuel-project && pnpm fuels dev`, wait for the node to spin up and deploy the contracts, then kill it with `CTRL+C`, run `pnpm fuels dev` again and you'll get an error:

```sh
Error: Address already in use (os error 98)

Error: Address already in use (os error 98)

Error: Address already in use (os error 98)
fuel-core node under pid xxxx does not exist. The node might have been killed before cleanup was called. Exiting cleanly.
```

nedsalk:
- Closes #2889

# Checklist


packages/account/src/test-utils/launchNode.ts
```diff
@@ -233,7 +233,6 @@ export const launchNode = async ({
       }
       childState.isDead = true;
 
-      removeSideffects();
```
nedsalk, line 236: Removing this line caused the issue to go away. I suspect that the root cause of the problem is somewhere in the `fuels` package because this line didn't cause problems to e.g. `launchTestNode`. It might be in the interplay between file watchers registered in the `dev` command and this `cleanup` function removing the files they're watching, but I couldn't pinpoint the *actual* root. The cleanup still happens because the `exit` and `error` event listeners registered on the child [above](https://github.com/FuelLabs/fuels-ts/pull/3038/files#diff-b8f07cac51692505e88e065262fd373293a373ed5f6196eb13ec811e37aee136L223) cause the cleanup to happen.

packages/fuels/test/features/dev.integration.test.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/**
+ * @group node
+ */
+describe('dev2', () => {
+  it(
+    'cleans up resources on graceful shutdown',
+    async () => {
+      using paths = runInit();
```
nedsalk, line 39: The `using` keyword is great because you don't have to think about where the test might fail so that you put the `cleanup` if it does fail there. With `using`, the cleanup will happen automatically both when the test finishes as well as when it throws an error.

packages/fuels/test/features/dev-2.test.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * @group node
+ */
+describe('dev', () => {
+  it(
+    'cleans up resources on graceful shutdown',
+    async () => {
+      using paths = runInit();
```
nedsalk, line 43: The `using` keyword is great because you don't have to think about where the test might fail so that you run the `cleanup` if it does fail there. With `using`, the cleanup will happen automatically both when the test finishes as well as when it throws an error.

nedsalk:
Superseded by #3508

PR #3047, opened 2024-08-28
feat: validate blob IDs against chain in chunk deploys

### LINKED ISSUE #3028: Check for Blobs' existence before uploading them
The main idea is to:
 1. Given a list of Blob IDs, use a new endpoint to check for their existence
 1. Iterate the Blobs list, upload only non-existent ones, and skip the rest

To achieve this, we can use this new GraphQL endpoint introduced in [v0.34.0](https://github.com/FuelLabs/fuel-core/releases/tag/v0.34.0):
 - https://github.com/FuelLabs/fuel-core/pull/2096

Important:
 1. Perform a single request with multiple GQL queries, one per Blob ID (avoids unnecessary roundtrips)
 1. Make sure not to ask for the Blob bytecode in the query (avoids bloating server response)

danielbate:
- Closes #3028

# Release notes

In this release, we:

- Reduced the cost of submitting blob contract deploys where a blob has already been uploaded.

# Summary

Uses the new `blob()` endpoint in `fuel-core@0.34.0` to validate a blob ID against before submitting it. Utilises GraqhQL to query for any number of blobs, as the `blob()` endpoint is constrained to a single ID.

# Checklist


packages/fuel-gauge/src/contract-factory.test.ts
```diff
@@ -518,4 +518,35 @@ describe('Contract Factory', () => {
...
+      salt: concat(['0x01', new Uint8Array(31)]),
+    });
+    const { contract: firstContract } = await firstDeploy.waitForResult();
+    const firstDeployCalls = sendTransactionSpy.mock.calls.length;
+    const secondDeploy = await factory.deployAsBlobTx<LargeContract>({
+      salt: concat(['0x02', new Uint8Array(31)]),
+    });
+    const { contract: secondContract } = await secondDeploy.waitForResult();
+    const secondDeployCalls = sendTransactionSpy.mock.calls.length;
+    expect(secondDeployCalls - firstDeployCalls).toBeLessThan(firstDeployCalls);
```
danielbate, line 542: I can check the exact number of calls if we prefer but potentially prone to flakiness over time. The important thing is that the second deploy submits less txs than the first as the blobs should already be uploaded.

arboleya: Should the fact we are using a node launcher in isolation be a guarantee against flakiness?

danielbate: Flakiness was probably the wrong word here. I mean that if we alter anything in the tx flow so we introduce more or less txs, the test will remain solid as the point is only that the second deploy should use less, rather than checking for a specific number of calls. Right now the number of calls is consistent.

packages/account/src/providers/provider.ts
```diff
@@ -1330,6 +1362,25 @@ Supported fuel-core version: ${supportedVersion}.`
     return coins;
   }
 
+  /**
+   * Returns an array of blobIds if they exist on chain, for a given array of blobIds.
+   *
+   * @param blobIds - blobIds to check.
+   * @returns - A promise that resolves to an array of blobIds that exist on chain.
+   */
+  async getBlobs(blobIds: string[]): Promise<(string | null)[]> {
```
nedsalk, line 1371:
```suggestion
  /**
   * Returns an array of blobIds that exist on chain, for a given array of blobIds.
   *
   * @param blobIds - blobIds to check.
   * @returns - A promise that resolves to an array of blobIds that exist on chain.
   */
  async getBlobs(blobIds: string[]): Promise<string[]> {
```

The `return blobs.filter((v) => v);` below filters out all `null`s so we can return a `string[]` type.

packages/account/src/providers/provider.ts
```diff
@@ -603,8 +606,37 @@ Supported fuel-core version: ${supportedVersion}.`
...
 
+    const customOperations = (requester: Requester) => ({
+      getBlobs(variables: { blobIds: string[] }) {
+        const queryParams = variables.blobIds.map((_, i) => `$blobId${i}: BlobId!`).join(', ');
+        const blobParams = variables.blobIds
+          .map(
+            (_, i) => `blob${i}: blob(id: $blobId${i}) {
+                        id
+                      }`
+          )
```
arboleya, line 617: Nit - maybe we don't need to worry about line breaks here.

```suggestion
          .map((_, i) => `blob${i}: blob(id: $blobId${i}) { id }`)
```

arboleya (dismissed):
Left an optional and trivial nit.

Looking good otherwise. 🚀

arboleya (dismissed):

Torres-ssf (approved):

packages/account/src/providers/operations.graphql
```diff
@@ -897,6 +897,12 @@ query getMessageByNonce($nonce: Nonce!) {
   }
 }
 
+query getBlob($blobId: BlobId!) {
+  blob(id: $blobId) {
+    id
+  }
+}
+
```
petertonysmith94, line 905:
```suggestion
```

Seems redundant

Torres-ssf (approved):

arboleya (approved):

petertonysmith94 (approved):

PR #3078, opened 2024-09-02
fix: TX estimation when `InputMessage` contains data

### LINKED ISSUE #3053: Error on contract call forwarding an assetId different from base assetId
### fuels-ts SDK Version

0.94.0

### Describe the Problem

when doing a contract call forwarding an assetId different from the baseAssetId of the provider, I am running into this error.

image

2 issues here:
1- estimated maxFee is not enough to pay for the transaction
2- even specifying a maxFee through `.txParams`, is not working, the estimation overrides my params

the issue 2 has already happened before and it was fixed, if it doesn't exist yet I suggest creating a test to make sure it doesn't happen anymore, something like:
- create a contract
- do a contract call infoming `txParams`
- in the last moment before sending the transaction, the `txParams` such as `gasLimit` , `maxFee`, `tip` should remain the same as set before in .txParams

### Reproduce
See it happening in this PR preview link https://github.com/FuelLabs/fuel-explorer/pull/455
you can use the [Fuel Dev Wallet](https://dev-wallet.fuel.network/docs/install/), or run `@FuelLabs/fuels-wallet` repo locally from master branch

Steps:
- open the preview link of bridge app
- click on asset "ETH" in input amount
- in the "TKN" card, click in the faucet button
- deposit TKN to Fuel
- after deposit is completed
- use the tokens received on fuel to start a Withdraw of erc20 TKN

### LINKED ISSUE #3077: Make Transaction Request Consider Message With Data
The TransactionRequest API [does not easily allow](https://github.com/FuelLabs/fuels-ts/blob/d19736b72f85cce90d4333252175e30be46bdb36/packages/account/src/providers/transaction-request/transaction-request.ts#L401) adding a message with data (a non-Coin Message) to its inputs.

An [InputMessage](https://docs.fuel.network/docs/specs/tx-format/input/#inputmessage) can either contain data or not. Messages without data are considered Coin Messages and can be used as resources to pay for the transaction fee when they are spent.

However, messages with data are not considered resources. Although they can still be added to the transaction inputs and spent, even if their amount is greater than 0, they cannot be used to pay transaction fees.

Torres-ssf:
- Closes #3077
- Closes #3053

# Release notes

In this release, we:

- Fixed TX estimation when an `InputMessage` contains data

# Summary

This PR addresses 2 things:

1. Fixed the `Account.getTransactionCost` method to ensure that it does not modify the amount of an `InputMessage` containing data when adding/editing fake resources before the estimation dry-run.
2. Updated the `BaseTransactionRequest.addMessageInput` method to accept messages with data. Although these messages are not considered Coin Messages and cannot be used to pay transaction fees, they can still be spent when included in a transaction's inputs.

# Checklist


packages/account/src/account.ts
```diff
@@ -542,13 +543,15 @@ export class Account extends AbstractAccount {
 
     const findAssetInput = (assetId: string) =>
```
maschad, line 544: We also use a [similar function](https://github.com/FuelLabs/fuels-ts/blob/st/feat/consider-message-with-data-input/packages/account/src/providers/transaction-request/transaction-request.ts#L592) when funding with fake UTXOs, should we just export this function and import it there? Otherwise you would have to update both no?

packages/fuel-gauge/src/transaction.test.ts
```diff
@@ -21,4 +23,120 @@ describe('Transaction', () => {
...
+    } = launched;
+
+    const baseAssetId = provider.getBaseAssetId();
+
+    const {
+      messages: [message],
+    } = await fundedWallet.getMessages();
+
+    const request = await contract.functions.foo(10).getTransactionRequest();
+    request.addMessageInput(message);
```
maschad, line 65: Why do we need to add the message input here if the wallet is already funded with the [`testMessage`](https://github.com/FuelLabs/fuels-ts/pull/3078/files#diff-3ccf1abbc4dc63f3169e41bf35f6a65f268903ebf6185701aff6ec308e975ce7R31) ?

maschad (commented):
Awesome job this fix @Torres-ssf and @LuizAsFight 🚀 

Overall LGTM but left two comments

packages/fuel-gauge/src/transaction.test.ts
Torres-ssf: @maschad But it is not funded.

Also, messages with data are not considered resources since they cannot pay for the TX fee.

Because of this the `getResourcesToSpend` will not return any message with data.

packages/account/src/account.ts
```diff
@@ -542,13 +543,15 @@ export class Account extends AbstractAccount {
...
       txRequestClone.inputs.find((input) => {
-        if ('assetId' in input) {
+        if (input.type === InputType.Coin) {
           return input.assetId === assetId;
         }
-        if ('recipient' in input) {
+
+        // We only consider the message input if it has no data.
+        // Messages with `data` cannot fund the gas of a transaction.
+        if (input.type === InputType.Message && bn(input.data).isZero()) {
```
petertonysmith94, line 552: Should we use the helper `isRequestInputMessageWithoutData` here?

petertonysmith94 (commented):
Nitty comment - nice one 🚀

petertonysmith94 (dismissed):

packages/account/src/account.ts
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/3083

packages/account/src/account.ts
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/3078/commits/06a2fbc36d40324be0689f2af01fa0a1263ffdbe

petertonysmith94 (approved):

packages/fuel-gauge/src/transaction.test.ts
maschad: I realize now that a consumer has to explicitly state that the request should be funded with the message input. Thanks for clarifying.

maschad (approved):

packages/account/src/providers/message.ts
```diff
@@ -71,3 +73,6 @@ export type MessageProof = {
 export type MessageStatus = {
   state: GqlMessageState;
 };
+
+export const isMessageCoin = (message: Message | MessageCoin): message is MessageCoin =>
```
danielbate, line 77: Nit but would prefer something like `isSpendableMessage` here due to the type `MessageCoin`.

danielbate (approved):
Nice one, approving with a nit but will leave it to you.

arboleya (approved):

packages/account/src/providers/message.ts
Torres-ssf: @danielbate, this was my initial idea, but it isn't entirely accurate semantically, as messages with data can also be spent.

Issue #3043, opened 2024-08-28
Add basic infra for benchmarks


arboleya:
The idea is to start measuring user-perceived network roundtrip/latency from the SDK perspective.

---

Points of analysis:
 1. Get Transaction Cost
 1. Submit TX (is tx submitted?)
 1. Wait for Result (has the tx finalized?)

---

Points of interaction:
 1. Read call (`getCount`)
 1. Write call (`incrementCounter`)
 1. Single Transfer (1-1)
    1. ~Test with Predicate variant~
 1. Batch Transfer (1-N)
 1. Withdraw to base layer
 1. Mint 
    1. ~Test with Predicate variant~

---

Contexts:
 - Local
 - ~Devnet~
 - ~Testnet~

arboleya:
A few possibly related refs/occurrences:
- https://github.com/FuelLabs/fuels-ts/issues/2465
- https://github.com/FuelLabs/fuels-ts/issues/1875

arboleya:
Potentially nice tool to integrate:
 - https://codspeed.io/

Issue #1753, opened 2024-02-15
Introduce Turbo cache in CI


danielbate:
In #1649 we introduced a Turbo cache in CI, however it was prone to [flakiness](https://github.com/FuelLabs/fuels-ts/actions/runs/7904898719/job/21577360284?pr=1750). As the `docs` build is dependent on an autogenerated file from `providers`, if `providers` hits the cache but docs misses, the job will fail.

Adding a build cache would improve the time of the second job in our test CI, so we should look to introduce it. However something that is more robust than what was added in #1649, #1694 and #1721.

danielbate:
> FWIW, it seems that the error happens when re-running a failed workflow.

Ah okay this is good information. A similar implementation to work reverted in #1752 may suffice then, as long as it accounts for the failed stage. Unsure why at the moment though, I'd expect the cache entries to still be present and it would still run using the same SHA.

danielbate:
Closing this for now. This has been investigated by the above PRs and someone is welcome to have another go however the real value for our CI and better use of time will come from #2493 and #1755.

danielbate:
Closing this for now. This has been investigated by the above PRs and someone is welcome to have another go however the real value for our CI and better use of time will come from https://github.com/FuelLabs/fuels-ts/issues/2493 and https://github.com/FuelLabs/fuels-ts/issues/1755.

PR #3084, opened 2024-09-03
ci: add initial benchmark infrastructure

maschad:
> Closes TS-604

- Closes https://github.com/FuelLabs/fuels-ts/issues/3043

# Summary

This PR seeks to introduce some initial benchmarks that will be on master post-merge.  It adds some metrics around base functions of the SDK and is by no means a finished product. The overall intention is to eventually run these benchmarks against nodes in `devnet` and `testnet` to get more realistic assessments around consumer's experiences. 

# Checklist


.github/workflows/bench.yml
```diff
@@ -0,0 +1,38 @@
+name: Benchmarks
```
arboleya, line 1: I did not review it; I just glanced over it.

I wasn't expecting this to grow roots into existent packages.

Could we have an `internal/benchmarks` package where we isolate everything about it?

maschad: Good suggestion @arboleya

packages/account/src/test-utils/launchNode.ts
```diff
@@ -265,15 +267,19 @@ export const launchNode = async ({
...
+            resolve({
+              cleanup,
+              ip: realIp,
+              port: realPort,
+              url: `http://${realIp}:${realPort}/v1/graphql`,
+              snapshotDir: snapshotDirToUse as string,
+              pid: child.pid as number,
+            }),
+          delay ?? 0
+        );
```
Torres-ssf, line 282: Why is this delay needed?

packages/abi-coder/test/utils/constants.d.ts
```diff
@@ -0,0 +1,38 @@
+export declare const U8_MAX: number;
+export declare const U8_MAX_ENCODED: Uint8Array;
```
Torres-ssf, line 2: @maschad Just out of curiosity, why do we need these constants?

internal/benchmarks/src/transaction-results.bench.ts
```diff
@@ -0,0 +1,110 @@
...
+import type { TransferParams } from 'fuels';
+import { launchTestNode, TestAssetId } from 'fuels/test-utils';
+import { bench } from 'vitest';
+
+/**
+ * @group node
+ */
+describe('Transaction Submission Benchmarks', () => {
+  bench('should successfully transfer a single asset between wallets', async () => {
+    using launched = await launchTestNode();
```
Torres-ssf, line 13: @maschad For transaction submissions I believe it is better to set up the node with the following:

```ts
 args: ['--poa-instant', 'false', '--poa-interval-period', '1s'],
``` 

This will give us results closer to reality.

packages/abi-coder/test/utils/constants.d.ts
maschad: I think this was added from a very old commit, they aren't needed.

packages/account/src/test-utils/launchNode.ts
maschad: I had introduced it when running some previous tests to try and simulate a delayed connection to the node. My new approach will be to run these tests against a devnet node so this is no longer necessary

internal/benchmarks/src/transaction-results.bench.ts
maschad: There will be three environments which these tests will run in - local, devnet and testnet.  Running them in the devnet and testnet environments will give us results _closer_ to reality, so no need to introduce an artificial delay.

.github/workflows/bench.yml
```diff
@@ -0,0 +1,26 @@
...
+on:
+  pull_request:
```
maschad, line 3: > [!NOTE]
> Do we want to run these on master or are we comfortable with this being per PR?

package.json
```diff
@@ -30,6 +30,13 @@
     "test:validate": "./scripts/tests-validate.sh",
     "test:browser:filter": "vitest --run --coverage false --config vitest.browser.config.mts",
     "test:e2e": "vitest --run --config vitest.node.config.mts $(scripts/tests-find.sh --e2e)",
+    "bench:node": "vitest bench --run --config vitest.node.config.mts $(scripts/tests-find.sh)",
+    "bench:node:filter": "vitest bench --run --config vitest.node.config.mts",
+    "clinic:flame": "clinic flame --autocannon [ / ] -- node packages/${npm_config_package_name}/dist/profile.js",
```
maschad, line 35: I've decided to still include these performance analysis scripts for digging into issues but they do require setting up a package with profiling infrastructure. I am open to removing them if we determine this is not germane to the PR

.github/workflows/bench.yml
Dhaiwat10: What will be the cost of doing this per PR? Both time-wise and $$$ ?

internal/benchmarks/test/fixtures/forc-projects/call-test-contract/src/main.sw
petertonysmith94, line 1: We are only using the `sum` and `return_context_amount` functions being used. Might be worthwhile only defining the functions we require for benchmarking.

package.json
```diff
@@ -30,6 +30,13 @@
     "test:validate": "./scripts/tests-validate.sh",
     "test:browser:filter": "vitest --run --coverage false --config vitest.browser.config.mts",
     "test:e2e": "vitest --run --config vitest.node.config.mts $(scripts/tests-find.sh --e2e)",
+    "bench:node": "vitest bench --run --config vitest.node.config.mts $(scripts/tests-find.sh)",
+    "bench:node:filter": "vitest bench --run --config vitest.node.config.mts",
```
petertonysmith94, line 34: Is it possible to run benchmarking on browser as well?

package.json
petertonysmith94: Could we have these script within the `@internal/benchmarks` packages?

Some documentation on their usages might be beneficial as well.

petertonysmith94 (commented):
V Nice 🚀

internal/benchmarks/package.json
```diff
@@ -0,0 +1,19 @@
...
+  ],
+  "scripts": {
+    "type:check": "tsc --noEmit",
+    "pretest": "run-s build:forc type:check",
+    "build:forc": "pnpm fuels build"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@fuel-ts/crypto": "workspace:*",
+    "@fuel-ts/account": "workspace:*",
```
danielbate, line 15: Redundant imports? Maybe we should also just benchmark the umbrella package?

internal/benchmarks/src/crypto.bench.ts
```diff
@@ -0,0 +1,61 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import type { Keystore } from '@fuel-ts/crypto';
+import { bufferFromString, pbkdf2, computeHmac, encrypt, decrypt } from '@fuel-ts/crypto';
```
danielbate, line 3:
```suggestion
import type { Keystore } from 'fuels;
import { bufferFromString, pbkdf2, computeHmac, encrypt, decrypt } from 'fuels';
```

Pretty sure all these can come from `fuels`

internal/benchmarks/src/crypto.bench.ts
```diff
@@ -0,0 +1,61 @@
...
+import { bench } from 'vitest';
+
+/**
+ * @group node
+ * @group browser
```
danielbate, line 8: Does mixing browser and node benchmarks work well together?

internal/benchmarks/tsdoc.json
danielbate, line 1: Do we need this file if we aren't using typedoc here?

package.json
```diff
@@ -30,6 +30,13 @@
...
+    "clinic:flame": "clinic flame --autocannon [ / ] -- node packages/${npm_config_package_name}/dist/profile.js",
+    "clinic:bubble": "clinic bubbleprof --autocannon [ -c 5 -a 500 / ] -- node packages/${npm_config_package_name}/dist/profile.js",
+    "clinic:doctor": "clinic doctor --autocannon [ / ] -- node packages/${npm_config_package_name}/dist/profile.js",
+    "clinic:heap": "clinic heapprofiler --autocannon [ / ] -- node packages/${npm_config_package_name}/dist/profile.js",
+    "0x": "0x -o -P 'autocannon localhost:$PORT' -- node packages/${npm_config_package_name}/dist/profile.js",
```
danielbate, line 39: Is it worth having all of these packages under the `bench:` prefix for grouping, like we do with `test`, `build`, `lint` etc.

.github/workflows/bench.yml
danielbate: I'd be pro doing this on ever PR, if something is going to bomb the performance we probably want to know before merging.

internal/benchmarks/src/cost-estimation.bench.ts
```diff
@@ -0,0 +1,139 @@
...
+import { CallTestContractFactory } from '../test/typegen/contracts';
+
+/**
+ * @group node
+ */
+describe('Cost Estimation Benchmarks', () => {
+  bench(
+    'should successfully get transaction cost estimate for a single contract call',
+    async () => {
+      using launched = await launchTestNode({
```
nedsalk, line 17: Not sure if we want to have the test setup be part of the benchmarks because
- it's not the part of the code that we're benchmarking
- it might introduce flakiness due to variations in launch startup + contract deployment speed because of the unpredictability of the CI environment speed-wise, which could force us to increase the threshold of allowed variation for reasons irrelevant to the code path we're benchmarking

Perhaps we could make use of something like this:
```ts
  let contract: CallTestContract;
  let cleanup: () => void;
  beforeEach(async () => {
    const launched = await launchTestNode({
      contractsConfigs: [{ factory: CallTestContractFactory }],
    });

    cleanup = launched.cleanup;
    contract = launched.contracts[0];
  });

  afterEach(() => {
    cleanup();
  });


bench('should successfully get transaction cost estimate for a single contract call',
    async () => {
      const cost = await contract.functions
        .return_context_amount()
        .callParams({
          forward: [100, contract.provider.getBaseAssetId()],
        })
        .getTransactionCost();

      expect(cost.minFee).toBeDefined();
      expect(cost.maxFee).toBeDefined();
      expect(cost.gasPrice).toBeDefined();
      expect(cost.gasUsed).toBeDefined();
      expect(cost.gasPrice).toBeDefined();
    })
```

internal/benchmarks/test/fixtures/forc-projects/call-test-contract/src/main.sw
Torres-ssf: Yeah, we should simplify this contract for now

internal/benchmarks/test/fixtures/forc-projects/call-test-contract/src/main.sw
```diff
@@ -0,0 +1,193 @@
...
+struct SumStruct {
+    a: u64,
+    b: u64,
+}
+
+struct SingleParamStruct {
+    a: u64,
+}
+
+abi TestContract {
```
Torres-ssf, line 39: @maschad Should we add a contract fn to Mint coins?

.github/workflows/bench.yml
maschad: > What will be the cost of doing this per PR? Both time-wise and $$$ ?

@Dhaiwat10  the benchmarks took around [11 minutes](https://github.com/FuelLabs/fuels-ts/actions/runs/10754180931/job/29824507491) I can't speak to the $ cost but it looks like according to the [pricing plans](https://codspeed.io/pricing) we would get unlimited usage.  @arboleya do you have any idea about costs?

I second what you say @danielbate

package.json
maschad: @petertonysmith94 I can definitely add some documentation on their usage although they are more for our use cases than consumers. 

I wouldn't put the `bench` commands in the `@internal/benchmarks` though since they rely on vitest config which is at the root, I think it would be awkward for a package to rely on files outside of it.

Nor would I put profiling in the `@internal/benchmarks` package because the intention is to use them with a specific package in the monorepo e.g.

```
npm_config_package_name=account pnpm clinic:flame  // runs flame against the account package
```

so it doesn't really have a use case outside of the monorepo

package.json
maschad: They are profiling tools not _necessarily_ benchmarking tools i.e. you would only use them if you suspect that there is a performance regression related to our code.

internal/benchmarks/tsdoc.json
maschad: Good catch, I suppose not.

internal/benchmarks/src/cost-estimation.bench.ts
maschad: Agreed @nedsalk

internal/benchmarks/src/crypto.bench.ts
maschad: They are run separately by the `bench:node` and `bench:browser` commands as well as on the workflow a0597b9d7

internal/benchmarks/src/crypto.bench.ts
maschad: a0597b9d7

internal/benchmarks/package.json
maschad: Good catch a0597b9d7

internal/benchmarks/test/fixtures/forc-projects/call-test-contract/src/main.sw
maschad: For sure @Torres-ssf a0597b9d7

package.json
maschad: So setting up the browser environment for the browser is a [bit more involved](https://github.com/FuelLabs/fuels-ts/actions/runs/10780020394/job/29894818040?pr=3084#step:6:108) but should be solvable. My concern though is that if the node tests take 11/12 mins on avg then adding browser setup + browser tests we could easily add another 15 mins, meaning benchmark tests could add on another 25+ mins to workflows, which is unreasonable. 

I've created [this](https://github.com/FuelLabs/fuels-ts/issues/3136) follow-up issue

petertonysmith94: Could we not run them as separate jobs?

IMO browser performance is more important than NodeJS.

package.json
petertonysmith94: @maschad Understood!

~Documentation would be good (doesn't have to be part of this PR) for our internal use would be wonderful.~

CONTRIBUTING.md
```diff
@@ -136,6 +136,28 @@ pnpm test:filter packages/my-desired-package/src/my.test.ts
...
+pnpm bench:node packages/my-desired-package
+```
+
+# Profiling
+
+We currently use [`clinic`](https://clinicjs.org/) to profile and debug our tooling. For instance you can run clinic's flame command to create a flamegraph for a specific package:
+
+```sh
+# creates a flamegraph for a specific package
+pm_config_package_name=account pnpm clinic:flame  // runs flame against the account package
```
petertonysmith94, line 158:
```suggestion
npm_config_package_name=account pnpm clinic:flame  // runs flame against the account package
```

Could we add a note on `0x` as well?

internal/benchmarks/src/cost-estimation.bench.ts
danielbate, line 1: Do we need to continue the conversation on what we are benchmarking?

I note the issues references contract reads, writes, mint. Any reason these were left out?

danielbate (approved):
Approving as this is great work, with one comment on what we are benchmarking.

Torres-ssf:
@maschad I've noticed that we are not benchmarking a deploy contract transaction. I am aware that we forgot to add it to https://github.com/FuelLabs/fuels-ts/issues/3043 but I believe this is a valid case to benchmark.

Do you think it is worth to add this case now? Or maybe in a future PR?

internal/benchmarks/src/cost-estimation.bench.ts
maschad: Good point @danielbate I think I may have forgotten the actual contract read/write calls themselves but rather just the cost estimation 

c4bbf94bc

Torres-ssf (approved):

danielbate (approved):
:shipit:

.github/workflows/bench.yml
```diff
@@ -0,0 +1,25 @@
...
+  push:
+    branches:
+      - master
+
+jobs:
+  benchmarks:
```
arboleya, line 9: Why is this step taking 50% more time than our current huge test suite?

This seems off.

diff

Torres-ssf: @maschad I share @arboleya concerns here. It is really strange why it is taking so long.

Could it be because of the codspeed setup process? Which involves downloading and installing it? https://github.com/FuelLabs/fuels-ts/actions/runs/10795148772/job/29941117888?pr=3084#step:5:37

maschad: @Torres-ssf @arboleya that's certainly apart of the reason (probably in tandem with uploading that performance data as well) but I don't think there's any easy way around that looking at the [GH action](https://github.com/CodSpeedHQ/action)  

That's why I suggested that we try to optimize the CI setup and `pretest` elements since that's within our control, even if it's only for this job.

arboleya: The `bench:node` takes 5-6 secs locally for me.

Torres-ssf: @maschad Is the Codspeed usage tied to the Github action `CodSpeedHQ/action@v3`? 

Is there no other way to use it to avoid the setup process?

maschad: From the looks of it the [action essentially just installs](https://github.com/CodSpeedHQ/action/blob/main/action.yml#L43-L81)  the [runner](https://github.com/CodSpeedHQ/runner). Of course we need to use the runner to upload the results to Codspeed so I don't think avoiding the action would necessarily yield better results.

Looking at other runs across [other repos](https://github.com/prisma/prisma/actions/runs/10802495620/job/29964692421?pr=25209#step:7:42) it takes on average 9 - 12 minutes depending on the amount of benchmarks, so our 6 mins is less than others

You can look at more [here](https://codspeed.io/explore) 

We can optimize both our build times and the CodSpeed Github action, and even the runner if we determine it's worth the engineering effort.

danielbate: May be time to dig up #1753.

maschad: @arboleya the issue is not with the installation script, that takes roughly ~20 seconds or less, as I had mentioned the bottleneck is [the runner](https://github.com/CodSpeedHQ/runner) which executes the vitest benchmarks in it's  [own environment](https://github.com/CodSpeedHQ/runner/blob/main/src/run/runner/run.rs#L18) and [measures the results](https://github.com/CodSpeedHQ/runner/blob/main/src/run/runner/valgrind.rs). This takes up 90%+ of the job time. 

I was looking to see if there was a way I could optimize it but to my dismay

image

arboleya: My bad; I misinterpreted what you said at first.

This is annoying, but I also don't want to hold progress.

If we exhaust all options, we can merge it as is and see how it goes. 🙂

maschad: No worries @arboleya , I think your concern about the length of time this takes is very valid. For a performance tool, this is poor performance 😆 

We should definitely raise an issue on the runner to see if there is a way to optimize their execution flow

Issue #3076, opened 2024-09-02
Make `ScriptTransactionRequest.from` clone a given transaction


arboleya:
Currently, this utility method creates a new instance that references received object.

Quick example:

```ts
const base_tx: ScriptTransactionRequest = { .. }

const estimation_tx = ScriptTransactionRequest.from(base_transaction);

estimation_tx.inputs.push(...spendableInputs);

const final_tx = ScriptTransactionRequest.from(base_transaction)

final_tx.inputs.push(...otherSpendableInputs);

// Result: all base_tx, estimation_tx, and final_tx will contain `spendableInputs` and `otherSpendableInputs`
```

It should ideally clone the received transaction and return a new one, that doesn't reference anything.

PR #3097, opened 2024-09-03
feat: return deep clone on transaction request `from` method

maschad:
- Closes https://github.com/FuelLabs/fuels-ts/issues/3076

# Release notes

In this release, we:

- We now return a deep clone of the transaction request in the static `from` method

# Summary

I've applied this feat to the other `from` methods on the other `BaseTransactionRequest` subclasses

# Checklist

### DIFF
--- a/packages/account/src/providers/transaction-request/blob-transaction-request.ts
+++ b/packages/account/src/providers/transaction-request/blob-transaction-request.ts
@@ -18,9 +18,9 @@ export interface BlobTransactionRequestLike extends BaseTransactionRequestLike {
 export class BlobTransactionRequest extends BaseTransactionRequest {
   static from(obj: BlobTransactionRequestLike) {
     if (obj instanceof this) {
-      return obj;
+      return structuredClone(obj);
     }
-    return new this(obj);
+    return structuredClone(obj);
   }
 
   /** Type of the transaction */
--- a/packages/account/src/providers/transaction-request/blob-transaction-request.ts
+++ b/packages/account/src/providers/transaction-request/blob-transaction-request.ts
@@ -17,10 +17,7 @@ export interface BlobTransactionRequestLike extends BaseTransactionRequestLike {
 
 export class BlobTransactionRequest extends BaseTransactionRequest {
   static from(obj: BlobTransactionRequestLike) {
-    if (obj instanceof this) {
-      return structuredClone(obj);
-    }
-    return structuredClone(obj);
+    return new this(structuredClone(obj));
   }
 
   /** Type of the transaction */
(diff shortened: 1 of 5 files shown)


packages/account/src/providers/transaction-request/blob-transaction-request.ts
```diff
@@ -17,10 +18,7 @@ export interface BlobTransactionRequestLike extends BaseTransactionRequestLike {
 
 export class BlobTransactionRequest extends BaseTransactionRequest {
   static from(obj: BlobTransactionRequestLike) {
-    if (obj instanceof this) {
-      return obj;
-    }
-    return new this(obj);
+    return new this(clone(obj));
```
maschad, line 21: My preference would be to use the native [`structuredClone` algorithm](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) but this was not returning an [exact copy](https://github.com/FuelLabs/fuels-ts/actions/runs/10689518036/job/29631794565?pr=3097) , I'm not too sure why though

packages/account/src/providers/transaction-request/script-transaction-request.ts
```diff
@@ -39,10 +40,7 @@ export interface ScriptTransactionRequestLike extends BaseTransactionRequestLike
  */
 export class ScriptTransactionRequest extends BaseTransactionRequest {
   static from(obj: ScriptTransactionRequestLike) {
-    if (obj instanceof this) {
-      return obj;
-    }
-    return new this(obj);
+    return new this(clone(obj));
```
arboleya, line 43: Should we have some simple tests on all these `from` methods?

 1. Clone source object/instance
 1. Modify source object/instance
 1. Validate if the new instance remained pristine

I'd assume we have no tests like these, or we'd have caught the issue earlier.

maschad: Good point ee922c837

Torres-ssf (approved):

danielbate (approved):

petertonysmith94 (approved):

packages/account/src/providers/transaction-request/utils.test.ts
```diff
@@ -42,3 +45,88 @@ describe('isTransactionTypeBlob', () => {
     expect(isTransactionTypeCreate(request)).toBe(false);
   });
 });
+
+describe('transactionRequestify', () => {
```
petertonysmith94, line 49: May be worth cloning with some inputs on the `baseRequest`:

```ts
const baseRequest = new ScriptTransactionRequest();
newRequest.inputs.push(baseInput);
expect(baseRequest.inputs).toHaveLength(1);

const newRequest = ScriptTransactionRequest.from(baseRequest);
expect(newRequest).to.not.equal(baseRequest);
expect(newRequest.inputs).toHaveLength(1);

newRequest.inputs.push(newInput);
expect(newRequest.inputs).to.not.equal(baseRequest.inputs);
```

maschad: I don't think we need extensive tests on the [clone](https://ramdajs.com/docs/#clone) method since it's from an external library, I only added these tests so that should someone remove the _clone_ functionality, we would catch it.

PR #3100, opened 2024-09-04
feat: add sway tests to `create-fuels`

### LINKED ISSUE #3025: Implement contract testing in `create fuels`
`create fuels` aims to provide a full stack and feature complete solution for working with and deploying dApps on fuel.

We should provide a contract testing solution so that consumers test their contracts in isolation. We should utilise `forc test` to keep the tool within the ecosystem.

Dhaiwat10:
- Closes #3025 

# Release notes

In this release, we:

- Added Sway tests to `create-fuels`

# Summary

This PR adds native Sway tests for all three programs in the `create-fuels` template.

# Checklist


danielbate (changes requested):
I think we should also extend the [`create fuels` testing documentation](https://docs.fuel.network/docs/fuels-ts/creating-a-fuel-dapp/#3-extending-the-test-suite-optional).

templates/nextjs/package.json
```diff
@@ -9,7 +9,9 @@
     "build": "pnpm run xprebuild && next build",
     "start": "next start",
     "lint": "next lint",
-    "test": "vitest"
+    "test": "run-s test:*",
+    "test:e2e": "vitest",
```
arboleya, line 13: This is just a terminology thing, but are these end-to-end tests?

Dhaiwat10: No, that specific script is for the e2e tests that we already have. For the tests added in this PR, the command is `test:unit`

arboleya: Yes, I was referring to the existent tets (`test` -> `test:e2e`).

Are these actual end-to-end tests?

Dhaiwat10: Well, it depends on what you classify as end-to-end. 

These tests do cover starting a node, deploying a contract instance, calling that contract and assserting the result. 

But they don't cover the UI part.

arboleya: Given how we currently use these terms, I guess it'd be an integration test.

@danielbate Thoughts?

templates/nextjs/package.json
```diff
@@ -9,7 +9,9 @@
     "build": "pnpm run xprebuild && next build",
     "start": "next start",
     "lint": "next lint",
-    "test": "vitest"
+    "test": "run-s test:*",
```
petertonysmith94, line 12: I believe we removed all instances of wildcard `*` running for a reason.

```suggestion
    "test": "run-s test:unit test:e2e",
```

templates/vite/package.json
```diff
@@ -9,7 +9,9 @@
     "build": "pnpm run xprebuild && tsc -b && vite build",
     "lint": "eslint .",
     "fuels:dev": "fuels dev",
-    "test": "vitest"
+    "test": "run-s test:*",
```
petertonysmith94, line 12:
```suggestion
    "test": "run-s test:unit test:e2e",
```

petertonysmith94 (commented):
Nice 🚀

templates/nextjs/package.json
danielbate: Yah agreed let's stick with our internal classification.

```jsonc
{
  "test": "run-s test:*",
  "test:unit": "forc test",
  "test:integration": "vitest ..", // launchTestNode
  "test:ui": "vitest ..", // playwright
}
```

`e2e` would be against a live node.

danielbate: I'd consider removing the serial test and letting `vitest` run off `test` as I think that will be more common practice, like we do internally.

Dhaiwat10: Renamed scripts here and updated folders/scripts in https://github.com/FuelLabs/fuels-ts/pull/3104 too

templates/vite/package.json
Dhaiwat10: Agreed, implemented Dan's suggestion:

https://github.com/FuelLabs/fuels-ts/pull/3100#discussion_r1746752310

templates/nextjs/package.json
arboleya: @Dhaiwat10 @danielbate Another [slightly shorter] term could be `feat`_ures_. ([example](https://github.com/FuelLabs/fuels-ts/tree/master/packages/fuels/test/features))

Regarding the commands, what if we suggest a default and provide a **_parallel_** `run all` option?

In this case, more options may also provide workflow inspiration, even for CI.

```json
{
  "test": "run-p test:feat",
  "test:all": "run-p test:feat test:unit test:ui",
  "test:feat": "vitest",
  "test:unit": "forc test",
  "test:ui": "sh ./scripts/test-ui.sh.",
}
```

maschad (dismissed):

templates/nextjs/package.json
Dhaiwat10: `feat` sounds good to me.

Like the idea of suggesting a default too. Makes the most sense.

danielbate (dismissed):

nedsalk (dismissed):

petertonysmith94 (dismissed):

templates/vite/package.json
```diff
@@ -9,7 +9,8 @@
     "build": "pnpm run xprebuild && tsc -b && vite build",
     "lint": "eslint .",
     "fuels:dev": "fuels dev",
-    "test": "vitest"
+    "test": "vitest",
+    "test:unit": "cd sway-programs && forc test"
```
petertonysmith94, line 13: We should probably run these tests in the CI to ensure they pass. Same for all packages.

Torres-ssf (dismissed):

templates/nextjs/package.json
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/3100/commits/3dedab6ac535909a6812cfb8291d17fb893b3691

templates/vite/package.json
Dhaiwat10: I [tried](https://github.com/FuelLabs/fuels-ts/pull/3100/commits/a69b34972ceb37a6c423103340c90f7f39b6c3da) adding them to CI but this would require us to install `forc` in CI. One thing we could do is use our internal binaries to run these tests only for CI instead of `forc`. Mind if I tackle this in a separate issue?

Dhaiwat10:
Marking this as blocked and closing for now since I've been unable to run these tests in CI after multiple attempts

PR #3107, opened 2024-09-05
chore: deprecate network URLs, add asset ID and chain ID for mainnet

### LINKED ISSUE #2509: Add ChainID and AssetID for `mainnet`
Before mainnet release, we must to double-check these items:

- ~[ ] Populate [`NETWORK_URL`](https://github.com/FuelLabs/fuels-ts/blob/54e16f03d93ae320cd66da6c50804e507246896c/packages/account/src/configs.ts#L5) with the mainnet URL.~
  - [ ] Instead of adding URL for mainnet, we should deprecate the other constants and later remove them.
- [ ] Add and populate the chain ID for `mainnet` in the static [`CHAIN_IDS` export](https://github.com/FuelLabs/fuels-ts/blob/master/packages/account/src/providers/chains.ts#L7-L8).
- [ ] Add and populate the asset for `mainnet` for our [`assets` export](https://github.com/FuelLabs/fuels-ts/blob/master/packages/account/src/providers/assets/assets.ts#L33).

Dhaiwat10:
- Closes #2509 

# Release notes

In this release, we:

- Deprecated the two network URLs, added asset ID and chain ID for `mainnet`

# Summary

This PR deprecates the two network URLs currently exported by the SDK, and adds the asset ID and chain ID for `mainnet`. 

The constants will eventually be deleted, so we have moved them inside of `@internal/utils` so that we can still use them in our tests.

# Checklist


packages/account/src/providers/chains.ts
```diff
@@ -6,5 +6,6 @@ export const CHAIN_IDS = {
   fuel: {
     devnet: 0,
     testnet: 0,
+    mainnet: 9889,
```
danielbate, line 9: Where did this come from?

danielbate (dismissed):
Approving with comment 👍🏻

petertonysmith94:
> I haven't deprecated the old URLs like the original issue suggests because I think people will still need them, at least the testnet one for sure. Thoughts? cc @petertonysmith94 @arboleya

I personally believe deprecation would be the best way forward.

Users managing their own endpoints seems like a reasonable ask to me. Also removes coupling our SDK to specific endpoints.

packages/account/src/providers/chains.ts
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/issues/2509#issuecomment-2329371444

petertonysmith94: https://github.com/FuelLabs/chain-configuration/tree/master/ignition#the-configuration-of-the-ignition-network

packages/account/src/configs.ts
```diff
@@ -1,4 +1,4 @@
 export const DEVNET_NETWORK_URL = 'https://devnet.fuel.network/v1/graphql';
 export const TESTNET_NETWORK_URL = 'https://testnet.fuel.network/v1/graphql';
-// TODO: replace placeholder with mainnet network url
-// export const NETWORK_URL = '';
+
+export const MAINNET_NETWORK_URL = 'https://mainnet.fuel.network/v1/graphql';
```
arboleya, line 4: We don't want to provide constants with URLs for users anymore because we may soon have several other URLs from different providers, and we don't want to concentrate everything on a single point of failure.

Yes, users may still be using the existing constants for Devnet/Testnet, but _deprecating_ is safe and will give them time to adapt before we remove them altogether—which we should do when we release the next breaking changes.

That's why I modified the issue to:
 1. Do not add another constant
 2. Deprecate the existing ones

arboleya (changes requested):

packages/account/src/configs.ts
Dhaiwat10: Got it. I interpreted the issue as we still wanted the new constant due to the todo comment in our code. I made the corrections

packages/account/src/providers/assets/assets.ts
```diff
@@ -30,6 +30,12 @@ export const rawAssets: Assets = [
         chainId: CHAIN_IDS.fuel.testnet,
         decimals: 9,
         assetId: '0xf8f8b6283d7fa5b672b530cbb84fcccb4ff8dc40f8176ef4544ddb1f1952ad07'
+      },
+      {
+        type: 'fuel',
+        chainId: CHAIN_IDS.fuel.mainnet,
+        decimals: 9,
+        assetId: '0xf8f8b6283d7fa5b672b530cbb84fcccb4ff8dc40f8176ef4544ddb1f1952ad07',
```
arboleya, line 38: @xgreenx Can you please confirm this?

arboleya (approved):

packages/account/src/configs.ts
```diff
@@ -1,4 +1,9 @@
+/**
+ * @deprecated This export is deprecated and will be removed in the future.
+ */
 export const DEVNET_NETWORK_URL = 'https://devnet.fuel.network/v1/graphql';
+
+/**
+ * @deprecated This export is deprecated and will be removed in the future.
+ */
```
arboleya, line 8: @Dhaiwat10 Could we replace these constants on the code snippets in the docs with hardcoded strings for each net? If we're deprecating them, we should stop suggesting their use.

Off the top of my head, I remember these two, but there can be more:
- https://docs.fuel.network/docs/fuels-ts/getting-started/connecting-to-a-local-node/
- https://docs.fuel.network/docs/fuels-ts/getting-started/connecting-to-testnet/

arboleya (dismissed):

packages/account/src/configs.ts
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/3107/commits/d548512f5dfe2376f1aec0cf90ec123349297008

xgreenx (dismissed):

apps/docs-snippets/src/guide/predicates/predicate-custom-transactions.test.ts
petertonysmith94, line 1: We should probably replace the `TESTNET_NETWORK_URL` handed to the provider as the hardcoded string.

```ts
const provider = await Provider.create('https://testnet.fuel.network/v1/graphql'));
```

petertonysmith94: https://github.com/FuelLabs/fuels-ts/pull/3107/files#diff-20a9255d49cad714208c0b31531b4152e5ed129089318d657befd213ac447f9cR33

apps/docs-snippets/src/guide/contracts/deploying-contracts.test.ts
petertonysmith94, line 1: We should probably replace the `TESTNET_NETWORK_URL` handed to the provider as the hardcoded string.

```ts
const provider = await Provider.create('https://testnet.fuel.network/v1/graphql'));
```

apps/docs-snippets/src/guide/predicates/predicate-custom-transactions.test.ts
Dhaiwat10: Same as here: https://github.com/FuelLabs/fuels-ts/pull/3107#discussion_r1747485296

apps/docs-snippets/src/guide/contracts/deploying-contracts.test.ts
Torres-ssf: @Dhaiwat10 perhaps here we should rename the variable to something simpler, like `url`, to avoid using the name of deprecated variable

Dhaiwat10: Renamed: https://github.com/FuelLabs/fuels-ts/pull/3107/commits/3f2c719f430096dbb64bd903e5ba2f5638702dd4

Torres-ssf (approved):

maschad (changes requested):
I should have commented this on the issue itself so please forgive me, but I wonder if we should just internalize these constants as opposed to removing them all together? If we still need the constants for certain cases (such as in the [e2e tests](https://github.com/FuelLabs/fuels-ts/pull/3107/files#diff-cbaea013d66ca16ee1630d21b77f3d7f972bb4f9eacc4f3be8dac3f6a2032744R24) for example) then a constant is better than a string literal.

arboleya:
@maschad In this case, they must be defined under the `internal` namespace and can only be used by tests.

Is that what you had in mind?

maschad:
@arboleya yes I think so, my main point is essentially we no longer need to export these constants, **but** they are still useful for our tests.

Dhaiwat10:
Great suggestion @maschad. Done

apps/docs-snippets/src/guide/contracts/deploying-contracts.test.ts
petertonysmith94: It's be nice if we we're defining the `providerUrl` within the context of the snippet.

internal/utils/tsup.config.ts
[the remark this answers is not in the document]
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/3107/commits/2a2bb4fe2155faa1a2e3f073899f58b11ff489f7

templates/vite/vite.config.ts.timestamp-1725856816588-3e7d000dce87b.mjs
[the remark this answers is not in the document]
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/3107/commits/e0916e6f14817c351f6614de9baf672b33960ced

petertonysmith94 (approved):

.changeset/silent-bees-give.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/account": patch
```
petertonysmith94, line 2:
```suggestion
"@fuel-ts/account": patch
"create-fuels": patch
```

nedsalk (dismissed):

danielbate (approved):

maschad (commented):
LGTM but we should update the changeset as @petertonysmith94 suggested @Dhaiwat10

petertonysmith94 (approved):

maschad (approved):
Great job @Dhaiwat10

internal/utils/src/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const DEVNET_NETWORK_URL = 'https://devnet.fuel.network/v1/graphql';
+export const TESTNET_NETWORK_URL = 'https://testnet.fuel.network/v1/graphql';
```
arboleya, line 2: @maschad @Dhaiwat10  Not gonna lie; a new package for this felt a bit over in the end. 🫠

maschad: I agree @arboleya if it remains only for these two constants but if we eventually migrate any other utilities that we want to internalize here then it may be more worthwhile. I don't have a list of those current utilities atm though 😅

danielbate: Yah +1, we could've left the constants internally in `account` but not exported them from `fuels` (once deprecated) 🤔

danielbate (approved):

Torres-ssf (approved):

internal/utils/src/constants.ts
Dhaiwat10: Gonna merge this PR with this new package, I think we'll probably be adding other constants and utilities to this package in the future like @maschad mentioned

PR #3108, opened 2024-09-05
chore: inform users if their `fuels` version is outdated

### LINKED ISSUE #2968: Inform users if their `fuels` version is outdated
We could provide a banner, such as Turbo does, to let users know if they're outdated:

```console
❯ pnpm build

> fuels-ts@0.0.0 build /fuel/fuels-ts
> turbo run build --cache-dir=.turbo

╭───────────────────────────────────────────────────────────────────────╮
│                                                                       │
│                  Update available v2.0.12 ≫ v2.0.14                   │
│    Changelog: https://github.com/vercel/turbo/releases/tag/v2.0.14    │
│           Run "npx @turbo/codemod@latest update" to update            │
│                                                                       │
│        Follow @turborepo for updates: https://x.com/turborepo         │
╰───────────────────────────────────────────────────────────────────────╯
turbo 2.0.12

[...]
```

Dhaiwat10:
Closes #2968 

# Release notes

In this release, we:

- Added a banner for users running an outdated `fuels` version

# Summary

This PR adds a feature in the `fuels` CLI regarding the user's `fuels` version. If the user's installed `fuels` version is outdated, the CLI will print out the latest version for the user.

# Breaking Changes

`runVersions` is now `async`.

# Checklist


packages/versions/src/lib/getFuelsVersion.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  return null;
+};
+
+export const getLatestFuelsVersion = async () => {
+  const response = await fetch('https://registry.npmjs.org/fuels/latest');
+  const data = await response.json();
+  return data.version;
+};
```
Torres-ssf, line 26: @Dhaiwat10 Why not use [getBuiltinVersions](https://github.com/FuelLabs/fuels-ts/blob/master/packages/versions/src/lib/getBuiltinVersions.ts) instead of a new helper?

arboleya: We need to compare both, actually:
 - last published version on NPM
 - output of `getBuiltInVersions`

This will tell us whether the user has the last version.

AFAICT, we don't need to read `package.json` files.

packages/versions/src/bin.ts
```diff
@@ -1,3 +1,7 @@
 import { runVersions } from './cli';
 
-runVersions();
+runVersions()
+  .then(() => process.exit(0))
+  .catch(() => {
+    process.exit(1);
+  });
```
arboleya, line 7: Technically, this could be considered a breaking change.

But only for someone extending our CLIs.

packages/versions/src/lib/getFuelsVersion.ts
Torres-ssf: @arboleya Ok, so there is no way to avoid fetching data from NPM, and therefore, dealing with a promise.

Torres-ssf: @arboleya I was under the impression that the `FUELS` constant within the response of `getBuiltInVersions` would always be the latest release.

Dhaiwat10: @Torres-ssf that's a good point, I can use that helper to get the version of `fuels` installed by the user. But I cannot use it to get the latest version published on npm. We still need to fetch from npm for that

Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/3108/commits/ee603f2632fcf7e90e92cd3b70836884f18d54ca

packages/versions/src/bin.ts
danielbate: Annoyingly I think we should exercise caution and just be hard on every breaking change, but I like the use case, this comes up a lot.

I'd opt for a `fuels version:check` command that implements this, and maybe we add to the deprecation issue that `fuels version:check` becomes `fuels:version`.

petertonysmith94:
I believe this will require changes to the [`fuels` CLI](https://github.com/FuelLabs/fuels-ts/blob/master/packages/fuels/src/cli.ts#L99-L102) as well :)

packages/versions/src/lib/getFuelsVersion.ts
Torres-ssf: @Dhaiwat10 @arboleya Oh I understand now. I'm sorry for the confusion.

packages/versions/src/bin.ts
arboleya: Left a suggestion for an alternative: https://github.com/FuelLabs/fuels-ts/pull/3108#issuecomment-2346355732

Dhaiwat10:
@arboleya noted, will make the needed changes

packages/fuels/src/cli/utils/checkForAndDisplayUpdates.ts
```diff
@@ -0,0 +1,42 @@
...
+        `\n⚠️ There is a newer version of fuels available: ${latestFuelsVersion}. Your version is: ${userFuelsVersion}\n`
+      );
+      return;
+    }
+
+    if (isFuelsVersionUpToDate) {
+      log(`\n✅ Your fuels version is up to date: ${userFuelsVersion}\n`);
+    }
+  } catch {
+    /* Fail gracefully */
```
maschad, line 40: I agree that we should fail gracefully, but I do wonder how we would ever know if this is failing for consumers if we don't log something. Perhaps we should at least acknowledge that we were unable to fetch the latest fuel version.

```suggestion
log(`\n currently unable to fetch latest fuels version. Skipping`)
```

maschad (commented):
Great job @Dhaiwat10 LGTM but left one comment

packages/fuels/src/cli/utils/checkForAndDisplayUpdates.ts
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/3108/commits/5df5b7868361f4e8d5b3f6f44d7f5052650988ec

packages/fuels/src/cli/utils/checkForAndDisplayUpdates.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    const latestFuelsVersion = await Promise.race<string | undefined>([
+      new Promise((resolve) => {
+        setTimeout(resolve, 3000);
+      }),
+      getLatestFuelsVersion(),
+    ]);
+
+    if (!latestFuelsVersion) {
+      return;
```
petertonysmith94, line 23:
```suggestion
      log(`\n Unable to fetch latest fuels version. Skipping...\n`);
      return;
```

packages/fuels/src/cli/utils/checkForAndDisplayUpdates.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  test('should handle fetch timing out', async () => {
+    vi.spyOn(global, 'fetch').mockImplementation(
+      () =>
+        new Promise((resolve) => {
+          setTimeout(resolve, 5000);
+        })
+    );
+    await expect(checkForAndDisplayUpdatesMod.checkForAndDisplayUpdates()).resolves.not.toThrow();
```
petertonysmith94, line 64:
```suggestion
    await expect(checkForAndDisplayUpdatesMod.checkForAndDisplayUpdates()).resolves.not.toThrow();
    expect(log).toHaveBeenCalledWith('\n Unable to fetch latest fuels version. Skipping...\n');
```

Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/3108/commits/cd5434ec1971678648792f1e44fb4b99c5ff5e35

Good catch!

packages/fuels/src/cli/utils/checkForAndDisplayUpdates.ts
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/3108/commits/cd5434ec1971678648792f1e44fb4b99c5ff5e35

petertonysmith94 (approved):

danielbate (approved):

packages/fuels/src/cli/utils/checkForAndDisplayUpdates.ts
```diff
@@ -0,0 +1,43 @@
...
+export const checkForAndDisplayUpdates = async () => {
+  try {
+    const { FUELS: userFuelsVersion } = versions;
+
+    const latestFuelsVersion = await Promise.race<string | undefined>([
+      new Promise((resolve) => {
+        setTimeout(resolve, 3000);
+      }),
+      getLatestFuelsVersion(),
+    ]);
```
arboleya, line 20: @Dhaiwat10 I couldn't review this one final time before the merge, but I will leave notes here.

As I mentioned in my previous [comment](https://github.com/FuelLabs/fuels-ts/pull/3108#issuecomment-2346355732), I really think we should use a [local cache](https://github.com/node-cache/node-cache) here.

Requesting npm versions for every CLI command we run seems unnecessary.

A TTL of 6 hours should do wonders already. Thoughts?

packages/fuels/src/run.ts
```diff
@@ -1,6 +1,8 @@
 import { configureCli } from './cli';
+import { checkForAndDisplayUpdates } from './cli/utils/checkForAndDisplayUpdates';
+import { error } from './cli/utils/logger';
 
 export const run = async (argv: string[]) => {
   const program = configureCli();
-  return program.parseAsync(argv);
+  return Promise.all([await checkForAndDisplayUpdates().catch(error), program.parseAsync(argv)]);
```
arboleya, line 7: @Dhaiwat10 This could potentially be simplified to:

```ts
  await checkForAndDisplayUpdates().catch(error)
  const program = configureCli();
  return program.parseAsync(argv);
```

packages/fuels/src/cli/utils/checkForAndDisplayUpdates.ts
Dhaiwat10: @arboleya my bad, I missed that point of yours. I have filed an issue and will do a follow-up PR soon: https://github.com/FuelLabs/fuels-ts/issues/3177

arboleya: Thanks.

Issue #3119, opened 2024-09-06
Deprecations pending removal


arboleya:
List of deprecations to go out in the next wave of breaking changes:
 - https://github.com/FuelLabs/fuels-ts/issues/3105
 - https://github.com/FuelLabs/fuels-ts/issues/2509

nedsalk:
- Remove `submitAndAwait` operation [deprecated](https://github.com/FuelLabs/fuels-ts/pull/3101/files#diff-69a6670d0426c74935ed2830ce552e7df2379c2d112a582306e57569fcd17029R395) in #3101.

arboleya:
@petertonysmith94 Can we close this since you created individual issues?

petertonysmith94:
Closed in favour of removal issues:
- https://github.com/FuelLabs/fuels-ts/issues/3516
- https://github.com/FuelLabs/fuels-ts/issues/3517
- https://github.com/FuelLabs/fuels-ts/issues/3518
- https://github.com/FuelLabs/fuels-ts/issues/2348
- https://github.com/FuelLabs/fuels-ts/issues/3519
- https://github.com/FuelLabs/fuels-ts/issues/3520

Issue #3032, opened 2024-08-25
Abi - Refactor / Gen


arboleya:
# Summary
 - [X] Refactor and move `abi-typegen` into it
 - [X] Plug existent strategies with `abi/parser` public interface
 - [X] Prepare the foundation for multiple output format (Javascript)

# Final items
  - [ ] Perform a clean-up on the PR to get ready for review
  - [ ] A diagram to show the overarching flow and key concepts

petertonysmith94:
- We will move the static `storageSlots` variable from the Contract (e.g. `MyTypegenContract.storageSlots`) to the Factory (e.g. `MyTypegenContractFactory.storageSlots`). As it is only relevant for deployment purposes.
- This is a breaking change, so we should make note on the relevant PR of this.

petertonysmith94:
Closed in favour of:
- https://github.com/FuelLabs/fuels-ts/issues/3650

PR #3134, opened 2024-09-09
fix: storage slots auto-load from typegen

### LINKED ISSUE #3130: `StorageSlots` not auto-loaded by Typegen
My understanding is that https://github.com/FuelLabs/fuels-ts/pull/1346 introduced the autoloading of storage slots, as stated in [the documentation](https://docs.fuel.network/docs/fuels-ts/contracts/storage-slots/#auto-load-of-storage-slots) , code generated using [Typegen](https://docs.fuel.network/docs/fuels-ts/fuels-cli/generating-types/) should automatically [load](https://docs.fuel.network/docs/fuels-ts/fuels-cli/using-generated-types/#autoloading-of-storage-slots) Storage Slots for consumers.

But as referenced in both [this issue](https://github.com/FuelLabs/fuels-ts/issues/2484#issuecomment-2176832668) and this [forum post](https://forum.fuel.network/t/mint-native-tokens-with-sdk/6568/13?u=maschad) this doesn't seem to be the case.

petertonysmith94:
- Closes #3130

# Release notes

In this release, we:

- Fixed issue with storage slots not being auto-loaded when deploying a contract. 

# Checklist

### DIFF
--- a/packages/abi-typegen/src/templates/contract/factory.hbs
+++ b/packages/abi-typegen/src/templates/contract/factory.hbs
@@ -1,6 +1,6 @@
 {{header}}
 
-import { ContractFactory, decompressBytecode } from "fuels";
+import { Contract, ContractFactory, decompressBytecode } from "fuels";
 import type { Provider, Account, DeployContractOptions, DeployContractResult } from "fuels";
 
 import { {{capitalizedName}} } from "./{{capitalizedName}}";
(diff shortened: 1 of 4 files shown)


packages/abi-typegen/src/templates/contract/factory.hbs
```diff
@@ -15,15 +15,20 @@ export class {{capitalizedName}}Factory extends ContractFactory {
     super(bytecode, {{capitalizedName}}.abi, accountOrProvider);
   }
 
+  deploy<TContract extends Contract = Contract>(
+    deployOptions?: DeployContractOptions
+  ): Promise<DeployContractResult<TContract>> {
+    return super.deploy({
+      storageSlots: {{capitalizedName}}.storageSlots,
```
nedsalk, line 22: We add storage slots on the contract while they're relevant only for deployment purposes, so it seems to me that we should move them to the factory. That would technically constitute a breaking change because we'd remove the static `MyTypegenContract.storageSlots` field, so we should add this as a TODO to
- #3119

We could add a deprecation note on the `MyTypegenContract.storageSlots` field.

This move of storage slots to from contract to contract factory can also be done as part of #3032.

Dhaiwat10 (approved):

danielbate (approved):
👏🏻

packages/abi-typegen/src/templates/contract/factory.hbs
petertonysmith94: I agree that moving this variable makes sense. I've added the following [comment](https://github.com/FuelLabs/fuels-ts/issues/3032#issuecomment-2337740796), on the issue.

I'd personally shy away from the deprecation notice (for now).

Issue #3029, opened 2024-08-25
Abi - Refactor / Kickoff


arboleya:
Instead of merely combining both packages in a single folder, the aim is to refactor them using the same structural approach. After recently modified Spec and Encoding strategies, and with future-proofing in mind, we can look ahead and strategize a proper foundation that should be resilient to future needs.

- [X] Draft new directory structure
- [X] Draft high-level package diagram/flowchart
- [X] Define basic concepts for types' classes (oop, functions, etc.)
  - We will follow the general principles that internally we will strive for functional based, with external entrypoint generally exposed a class.
- [X] Define essential layers: AbiParser, AbiCoder, AbiGen (typegen)
- [X] Add foundation for the new `abi` package, basic infra for tests, etc.
- [X] Consolidate and share existent fixtures between all layers
- [X] Favor feature-complete contracts over small minimalistic ones
  - Consolidate + refactor `abi-coder/exhaustive-examples` test and to `fuel-gauge/abi-contract`
  - Consolidate `abi-typegen/*` tests and merge into the related `fuel-gauge/abi-*` Sway program.
  - Consolidate all `abi-coder/encoding/**Coder` related tests into Sway

arboleya:
![overview](https://github.com/user-attachments/assets/6c3bb1c7-bbc3-4673-8102-19c388ff52a3)

arboleya:
Basic sketch of what could be a starting point for the file structure.


```console
abi
├── coder
│   ├── AbiCoder.ts
│   ├── encodings
│   │   ├── v1
│   │   │   ├── array.ts
│   │   │   └── enum.ts
│   │   └── v2
│   │       ├── array.ts
│   │       └── enum.ts
│   └── types
├── gen
│   ├── AbiGen.ts
│   ├── formatters
│   │   ├── configurables.ts
│   │   ├── enums.ts
│   │   ├── functions.ts
│   │   └── ...
│   ├── renders
│   │   ├── renderContract.ts
│   │   └── ...
│   ├── templates
│   │   ├── typescript
│   │   │   ├── contract-factory.hbs
│   │   │   ├── contract.hbs
│   │   │   └── ...
│   │   └── javascript
│   └── types
├── parser
│   ├── AbiInterface.ts
│   ├── AbiParser.ts
│   ├── specs
│   │   ├── v1
│   │   │   ├── interface.ts
│   │   │   └── transpiler.ts
│   │   └── v2
│   └── types
└── test
[...]

PR #3137, opened 2024-09-10
feat: sway type matchers

nedsalk:
> Part of TS-597

Relates to:
- #3029 

# Summary

This PR gives a proof-of-concept to how we'd go from a sway type to a corresponding coder or typer.

A coder example would be:
```ts
const matcher = createMatcher<Coder>({
  string: StringCoder,
  empty: EmptyCoder,
  enum: EnumCoder,
  result: EnumCoder,
  ...
});

const resultCoder = matcher({ swayType: 'enum std::result::Result' });
```

The same approach would be for typers:
```ts
const matcher = createMatcher<Typer>({
  string: StringTyper,
  empty: EmptyTyper,
  enum: EnumTyper,
  result: ResultTyper,
  ...
});

const resultTyper = matcher({ swayType: 'enum std::result::Result' });
```

# Checklist


packages/abi/src/matchers/sway-type-matchers.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  evmAddress: 'struct std::vm::evm::evm_address::EvmAddress',
+  stdString: 'struct std::string::String',
+  vector: 'struct std::vec::Vec',
+
+  enum: 'enum MyEnum',
+  option: 'enum std::option::Option',
+  result: 'enum std::result::Result',
+
+  rawUntypedPtr: 'raw untyped ptr',
+  rawUntypedSlice: 'raw untyped slice',
```
nedsalk, line 31: I question the need for these types and we'll review if we need them later on when we start implementing more.

packages/abi/src/matchers/sway-type-matchers.ts
```diff
@@ -0,0 +1,89 @@
...
+  | 'struct'
+  | 'b512'
+  | 'bytes'
+  | 'vector'
+  | 'tuple'
+  | 'array'
+  | 'assetId'
+  | 'evmAddress'
+  | 'rawUntypedPtr' // might not need it
+  | 'rawUntypedSlice'; // might not need it
```
nedsalk, line 25: I question the need for these last two types and we'll review if we need them later on when we start implementing more.

packages/abi/src/coder/abi-coder.ts
```diff
@@ -1,2 +1,11 @@
+import { getCoder } from './encoding/v1/coder-matcher';
+import type { Coder } from './encoding/v1/coders/types';
+
 // Placeholder
-export class AbiCoder {}
+export class AbiCoder {
+  public encode(type: string, value: unknown) {
+    const coder = getCoder(type) as Coder & { encode: (value: unknown) => unknown };
```
arboleya, line 7: Please do not construct types on the fly.

Have types clearly defined in a way they can be imported and used.

packages/abi/src/matchers/sway-type-matchers.test.ts
```diff
@@ -0,0 +1,48 @@
...
+};
+
+describe('sway type matchers', () => {
+  test.each(Object.entries(testCases))('%s - %s', (type, value) => {
```
arboleya, line 35: I have given this feedback many times before, but I will repeat:
- Please do not use this sort of `test.each(...)`. As the set grows, isolating things becomes impossible, and understanding anything becomes a pain.

packages/abi/src/matchers/sway-type-matchers.ts
Torres-ssf: @nedsalk Why they might not be needed?

nedsalk: We don't use them for encoding and decoding, nor for generating types in typegen.

packages/abi/src/coder/abi-coder.ts
nedsalk: I agree and I wouldn't do it if we had decided on what the `Coder` interface would look like. This is all a work in progress to show how it'd look like to get the coder and call a fictitious `encode` method "in action". The point of this PR is to showcase how we'd go from a sway type to a coder or typer, whatever they look like in the end. Subsequent PRs will introduce coders and typers properly.

packages/abi/src/matchers/sway-type-matchers.ts
Torres-ssf: @nedsalk Thanks for the clarification

packages/abi/src/matchers/sway-type-matchers.test.ts
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/3137/commits/19f793df60c5e9eecfe44d13b8c2a6130da13dc9

packages/abi/src/matchers/sway-type-matchers.ts
petertonysmith94, line 1: For me, a helper function around this matching functionality would be ideal. This would enable us to change the strategy of matching to be more performant down the line (without considerable changes).

```ts
export const createMatcher =
  <T>(mappings: Record<SwayType, T>) =>
  (opts: { swayType: string }): T => {
    const { swayType } = opts;

    for (const [key, matcher] of Object.entries(swayTypeMatchers)) {
      if (matcher(swayType)) {
        return mappings[key as SwayType];
      }
    }

    throw new Error(`Matcher not found for ${swayType}`);
  };
```

Also enables partial matching:
```ts
const match = createMatcher<string | undefined>({
  array: 'array',
});
```

Also makes the test more readable IMO:
```ts
const matcherKey = 'array';
const swayType = '[_; 3]';
const expectedResult = 'array';

const result = match({ swayType });

expect(result).toEqual(expectedResult);
verifyOtherMatchersDontMatch(matcherKey, swayType);
```

packages/abi/src/matchers/sway-type-matchers.ts
```diff
@@ -0,0 +1,98 @@
...
+const generic: Matcher = (type) => genericRegEx.test(type);
+
+export const stringRegEx = /str\[(?<length>[0-9]+)\]/;
+const string: Matcher = (type) => stringRegEx.test(type);
+
+export const tupleRegEx = /^\((?<items>.+)\)$/m;
+const tuple: Matcher = (type) => tupleRegEx.test(type);
+
+export const arrayRegEx = /\[(?<item>[\w\s\\[\]]+);\s*(?<length>[0-9]+)\]/;
+const array: Matcher = (type) => arrayRegEx.test(type);
```
petertonysmith94, line 48: IMO, all constants should be `ALL_CAPS` to enhance readability.

```suggestion
export const GENERIC_REGEX = /^generic ([^\s]+)$/m;
const generic: Matcher = (type) => GENERIC_REGEX.test(type);

export const STRING_REGEX = /str\[(?<length>[0-9]+)\]/;
const string: Matcher = (type) => STRING_REGEX.test(type);

export const TUPLE_REGEX = /^\((?<items>.+)\)$/m;
const tuple: Matcher = (type) => TUPLE_REGEX.test(type);

export const ARRAY_REGEX = /\[(?<item>[\w\s\\[\]]+);\s*(?<length>[0-9]+)\]/;
const array: Matcher = (type) => ARRAY_REGEX.test(type);
```

petertonysmith94 (commented):
Our tests should illustrate how the functionality works, and therefore the additions to `coder` and `gen` should be removed IMO.

packages/abi/src/matchers/sway-type-matchers.ts
nedsalk: I implemented it in the last few commits.

packages/abi/src/coder/abi-coder.ts
nedsalk: After a discussion with @petertonysmith94, I removed the coder and typer examples out of the code and into the PR description.

packages/abi/src/matchers/sway-type-matchers.ts
```diff
@@ -0,0 +1,116 @@
...
+
+const option: Matcher = (type) => type === 'enum std::option::Option';
+const result: Matcher = (type) => type === 'enum std::result::Result';
+const enumMatcher: Matcher = (type) => !option(type) && !result(type) && /^enum .*$/m.test(type);
+
+const rawUntypedPtr: Matcher = (type) => type === 'raw untyped ptr';
+const rawUntypedSlice: Matcher = (type) => type === 'raw untyped slice';
+
+export const swayTypeMatchers: Record<SwayType, Matcher> = {
+  empty,
```
petertonysmith94, line 68:
```suggestion
  void,
```

Sway-end calls these a `void` type rather than `empty`, I'd favour syncing our terminology.

packages/abi/src/matchers/sway-type-matchers.test.ts
```diff
@@ -0,0 +1,247 @@
+import type { SwayType, swayTypeMatchers } from './sway-type-matchers';
+import { createMatcher } from './sway-type-matchers';
+
+const testMappings: Record<keyof typeof swayTypeMatchers, `${string}-matched`> = {
+  string: 'string-matched',
+  empty: 'empty-matched',
```
petertonysmith94, line 6:
```suggestion
  void: 'void-matched',
```

packages/abi/src/matchers/sway-type-matchers.ts
```diff
@@ -0,0 +1,116 @@
...
+export const STRING_REGEX = /str\[(?<length>[0-9]+)\]/;
+const string: Matcher = (type) => STRING_REGEX.test(type);
+
+export const TUPLE_REGEX = /^\((?<items>.+)\)$/m;
+const tuple: Matcher = (type) => TUPLE_REGEX.test(type);
+
+export const ARRAY_REGEX = /\[(?<item>[\w\s\\[\]]+);\s*(?<length>[0-9]+)\]/;
+const array: Matcher = (type) => ARRAY_REGEX.test(type);
+
+const struct: Matcher = (type) =>
```
arboleya, line 50: Adhering to the REGEX/const pattern.

```suggestion
const STRUCT_REGEX = /^struct .+$/m;
const STRUCT_STD_REGEX = /^struct std::.*(AssetId|B512|Vec|RawVec|EvmAddress|Bytes|String|RawBytes)$/m;
const struct: Matcher = (type) => STRUCT_REGEX.test(type) && !STRUCT_STD_REGEX.test(type);
```

Torres-ssf (approved):

Dhaiwat10 (approved):
Neat work

petertonysmith94 (approved):

PR #3125, opened 2024-09-06
feat: `create fuels` template revamp

danielbate:
> Closes TS-620, TS-580, TS-622

- Closes #3086 
- Closes #2998 
- Closes #3088

# Release notes

In this release, we:

- Updated the style and UX of the `create fuels` templates

# Summary

Refresh of the `create fuels` template to simplify DX and add some simple style. Working preview against testnet can be found [here](https://create-fuels-template-git-db-featcreate-fuels-styled-fuel-labs.vercel.app/).

Screenshot 2024-09-27 at 12 37 47

**_Special thanks_** to @rishabhkeshan for his instrumental feedback and work to improve the templates that was superseded, but did inform this work.

**Features:**
- Removed wallet abstractions and uses only connectors
- Introduces a _splash screen_
- Added styles with tailwind
- UX/DX improvements (improved flows, transaction states, notifications, simplified/lighter routing etc)

# Checklist

- [x] All **changes** are **covered** by **tests** (or not applicable)
- [x] All **changes** are **documented** (or not applicable)
- [x] I **reviewed** the **entire PR** myself (preferably, on GH UI)
- [x] I **described** all **Breaking Changes** (or there's none)

PR #3143, opened 2024-09-11
feat: `create fuels` template optimisations

### LINKED ISSUE #3088: `create fuels` Template Optimisations
Various feedback from @SilentCicero that we should include in the templates, these should be reflected in both the `nextjs` and `vite` templates.

If any of the below introduces too much specific code then we should not introduce it to ensure the template remains barebones.

- [x] Replace the `Top-up Wallet` button with a single faucet button (remove the faucet button from the left hand nav).
- [x] Reduce the transfer amounts for the predicate example, the amount is not suitable for testnet
- [x] Predicate text not displaying correctly on mobile, may be resolved by generally reducing mobile body padding
- [x] Ensure all externally routing links have `target="_blank"` and opens in a new tab. The `Learn More about Scripts` links do not but also check other links.
- [x] Ensure faucet is embedded. I believed this was done in #3036 but doesn't seem to be happening in the [deployed env](https://create-fuels-template.vercel.app/), we should verify this.
- [x] Balance should update on tx submission
- [x] Mobile formatting fixes - reduce body padding significantly and make mobile nav links cover entire width
- [x] Move `Fuel Docs` link beside the `Home` button as another nav link
- [x] Improve error handling when _unconnected_, currently it has the default tx fail error messages, we should provide something more specific, we could grey out the action buttons when no wallet is connected
- [x] Remove the `Back to home` links, we already have a home button in the nav
- [x] Transaction pending state, when a transaction is pending the active button should reflect that
- [x] Burner wallet issues? _We may need to do some additional debugging / integration tests here to debug_

rishabhkeshan:
> Fixes TS-622

- Closes #3088 

# Summary

Includes the following optimizations on the Next and Vite templates as per @SilentCicero's feedback:

- [x] Replace the `Top-up Wallet` button with a single faucet button (remove the faucet button from the left hand nav).
- [x] Reduce the transfer amounts for the predicate example, the amount is not suitable for testnet
- [x] Predicate text not displaying correctly on mobile, may be resolved by generally reducing mobile body padding
- [x] Ensure all externally routing links have `target="_blank"` and opens in a new tab.
- [x] Ensure faucet is embedded and for Safari/Mobile there is a redirect to the faucet url and back.
- [x] Balance should update on tx submission
- [x] Move `Fuel Docs` link beside the `Home` button as another nav link
- [x] Improve error handling when _unconnected_, currently it has the default tx fail error messages, we should provide something more specific, we could grey out the action buttons when no wallet is connected
- [x] Remove the `Back to home` links, we already have a home button in the nav
- [x] Transaction pending state, when a transaction is pending the active button should reflect that
- [x] Mobile formatting fixes - reduce body padding significantly and make mobile nav links cover entire width
- [x] Burner wallet issues? _We may need to do some additional debugging / integration tests here to debug_

# Checklist


Dhaiwat10 (changes requested):
Currently, if the user has not connected anything, the app does not even _read_ the value of the counter. We don't need a wallet to read a value, a provider is enough. Can we facilitate that? The provider always exists, so there is no reason for us to not be reading this value at all times.

image

.changeset/many-bananas-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-fuels": patch
+---
+
+feat: `create fuels` template optimisations
```
arboleya, line 5: @danielbate How does this PR relate to the template revamp, and will the changes in this PR be overridden by those from 👇 ?
 - https://github.com/FuelLabs/fuels-ts/pull/3125

danielbate: This PR is improvements to the current templates based on feedback from @SilentCicero / @rishabhkeshan .

#3125 completely changes and overwrites the current templates, but either implements or supersedes the above feedback.

Unfortunately this PR took a little longer than hoped. Hopefully we can get this in for tomorrows release with #3125 making next week.

Dhaiwat10 (approved):

templates/nextjs/src/app/faucet/page.tsx
```diff
@@ -24,6 +29,25 @@ export default function Faucet() {
     }
   }, [wallet]);
 
+  useEffect(() => {
+    const interval = setInterval(async () => {
+      await refreshWalletBalance?.();
+    }, 500);
```
danielbate, line 35: Couple you up this to 5s to be safe? I kept getting rate limited when working against testnet with a 1s interval.

templates/nextjs/src/app/page.tsx
```diff
@@ -59,18 +75,36 @@ export default function Home() {
...
+
+      // Wait for the transaction to be mined, and then read the value returned
+      const { value, transactionId } = await waitForResult();
+      toast.success(() => (
+        <span>
+          Counter Incremented! View it on the
+          <a
+            target="_blank"
+            className="pl-1 underline"
+            href={`https://app.fuel.network/tx/${transactionId}`}
```
danielbate, line 93: Should this be conditional so it's only when the app is running against testnet?

templates/vite/src/components/Navbar.tsx
```diff
@@ -116,28 +106,46 @@ export const Navbar: FC = () => {
...
 
-            <Link href="/faucet">Faucet</Link>
-
-            {isBrowserWalletConnected && (
-              <Button onClick={disconnect}>Disconnect Wallet</Button>
+            <Link href={DOCS_URL} target="_blank">
+              Docs
+            </Link>
+
+            {isConnected && wallet && (
```
danielbate, line 113: Does this need a local check?

danielbate (commented):
Nice work @rishabhkeshan, my comments should be applied to all instances across both templates.

@Dhaiwat10 Is this likely to impact docs or integration tests?

templates/vite/src/routes/faucet.lazy.tsx
```diff
@@ -27,6 +31,25 @@ function Index() {
     }
   }, [wallet]);
 
+  useEffect(() => {
+    const interval = setInterval(async () => {
+      await refreshWalletBalance?.();
+    }, 500);
+    return () => clearInterval(interval);
+  }, []);
```
Torres-ssf, line 39: Should we add the `wallet` to the  `refreshWalletBalance` `useEffect` ?

```ts
  useEffect(() => {
    const interval = setInterval(async () => {
      await refreshWalletBalance?.();
    }, 500);
    return () => clearInterval(interval);
  }, [wallet]);
```

Dhaiwat10:
@danielbate it _might_ if the UI has changed in a certain way. The only way to really know is to run them I think

danielbate:
Yeah unfortunately the UI tests are failing for me. Docs will also need updating 🤔 


Screenshot 2024-09-18 at 10 43 50

danielbate:
Closing this _for now_, we've run into issues with the integration tests when removing the wallet abstractions. These are problems we'll also need to solve for #3125 so we'll be syncing up with @rishabhkeshan to get these over the line, but it's likely that this PR will be superceded.

PR #3147, opened 2024-09-11
feat: `provider.url` now returns auth url

### LINKED ISSUE #3142: `this.provider.url` is not considering auth url passed
when the provider is created, if a basic auth is sent, it gets lost as the sdk implementation will put basic auth in headers function, and create the provider with an URL without basic auth

this becomes a problem when it's needed to create a new provider from the `provider.url` we have, for instance in string communication protocols, like JSON, state saving

I think `this.provider.url` should be as it was inputted, then when using it internally in fetch request, basic auth part can be taken out

petertonysmith94:
- Closes #3142
- Closes TS-643

# Release notes

In this release, we:

- The `provider.url` now returns an authenticated URL
- The `Provider` now accepts a `headers` field

# Summary

- `Provider.connect` was not working with the authentication URL, so that is also working now.

# Checklist

### DIFF
--- a/.changeset/afraid-starfishes-joke.md
+++ b/.changeset/afraid-starfishes-joke.md
@@ -1,0 +1,6 @@
+---
+"@fuel-ts/transactions": patch
+"@fuel-ts/account": patch
+---
+
+feat: introduce upload and upgrade transaction request
--- a/.changeset/long-ravens-explain.md
+++ b/.changeset/long-ravens-explain.md
@@ -1,5 +1,0 @@
----
-"@fuel-ts/account": patch
----
-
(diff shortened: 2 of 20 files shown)


.changeset/orange-hornets-help.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/account": patch
```
petertonysmith94, line 2:
```suggestion
"@fuel-ts/account": patch
"@fuel-ts/errors": patch
```

arboleya (approved):
Headers, of course! ⚡ ✨ 🎯

maschad (approved):

Torres-ssf (approved):

PR #3166, opened 2024-09-13
chore: run benchmarking utility in devnet environment

### LINKED ISSUE #3138: Run Benchmarking utility in devnet environment
#3043 seeks to introduce the basic benchmarking infrastructure which has now given us benchmarks against a local `fuel-core` node in a Node.js environment (for browser support see https://github.com/FuelLabs/fuels-ts/issues/3136)

We should also run these benchmarks in devnet, also introducing blob transactions.

maschad:
> Fixes TS-639
- Closes #3138 

# Summary

This PR introduces devnet benchmarking on master builds, as well as blob transaction deployment testing.

Also previous benchmarking tests are now looped 10 times to have more consistent results.

# Checklist


.github/workflows/bench-testnet.yaml
```diff
@@ -0,0 +1,32 @@
+name: "Bench Devnet"
+
+on:
+  # Remove this before merging, we only want devnet tests on master
```
maschad, line 4: > [!NOTE]
> Remove this before merging so that we only run the devnet workflow on master builds

arboleya: @maschad I'd say it's best to run only for the [release PRs](https://github.com/FuelLabs/fuels-ts/pull/3174).

We'll still have the chance to review it before merging and identify which PR caused a regression — which, by the way, I think we should disable for now and switch to _observation_ mode until we are comfortable with our numbers and thresholds; then, we can re-enable it. 

With this, we should not have any blocked PRs and releases that can't be merged because of regressions, like what is happening in:
 - https://github.com/FuelLabs/fuels-ts/pull/3174

<br/>

regression

internal/benchmarks/src/transaction-results.bench.ts
```diff
@@ -55,45 +69,55 @@ describe('Transaction Submission Benchmarks', () => {
...
+          destination: receiver1.address,
+          amount: amountToTransfer1,
+          assetId: provider.getBaseAssetId(),
+        },
+        { destination: receiver2.address, amount: amountToTransfer2, assetId: TestAssetId.A.value },
+        { destination: receiver3.address, amount: amountToTransfer3, assetId: TestAssetId.B.value },
+      ];
 
-    const tx = await wallet.batchTransfer(transferParams);
+      const tx = await wallet.batchTransfer(transferParams);
```
arboleya, line 102: @maschad It seems one of the [failing tests](https://github.com/FuelLabs/fuels-ts/actions/runs/10894541904/job/30231421374#step:5:143) is triggered by this `batchTransfer` call.

I just double-checked the wallet pvt key and updated the GH secret to be sure, then re-ran the workflow, and it failed like before.

@Torres-ssf Thoughts?

maschad: @arboleya that's strange because based on the [CI logs](https://github.com/FuelLabs/fuels-ts/actions/runs/10894541904/job/30257363622?pr=3166#step:5:113) it looks like it's failing at the [contract read call](https://github.com/FuelLabs/fuels-ts/pull/3166/files#diff-31f4ad5f8775c72e8c13a0bc07d99b7dba56c4bf9914d6af75eab3a146689f3fR63)

.github/workflows/bench-testnet.yaml
maschad: @arboleya I agree with you re: release PRs 

I assume you mean by _observation mode_ is that the workflow still runs but we can still go ahead and merge to prevent PRs from being blocked. As we have been doing already

Screenshot 2024-09-17 at 8 56 13 AM

internal/benchmarks/src/transaction-results.bench.ts
arboleya: Oh, right, the first error changed between the two runs. I linked the first, and you linked the second.

I was referring to the second error on the bottom rectangles.
runs-diff

Torres-ssf: @arboleya @maschad, I am unsure about what is happening here. I will take a closer look

maschad: I've resolved it. @Torres-ssf did some digging and he discovered that Codspeed actually runs our benchmarks twice in it's own environment (independent of any loop) as a result looping this 10 times in quick succession seems to be problematic on devnet - perhaps UTXO caching? Not sure.

.github/workflows/bench-devnet.yaml
```diff
@@ -0,0 +1,32 @@
+name: "Bench Devnet"
+
+on:
+  # Remove this before merging, we only want devnet tests on master
```
maschad, line 4: As discussed [here](https://github.com/FuelLabs/fuels-ts/pull/3166#discussion_r1763085556)

> [!NOTE]
> Remove this before merging so that we only run the devnet workflow on master builds

We will only run this workflow on release

.github/workflows/bench.yml
```diff
@@ -1,9 +1,11 @@
 name: Benchmarks
 on:
   pull_request:
-  push:
```
maschad, line 4: I suppose we still want to run the local benchmark workflow on master builds

internal/benchmarks/test/fixtures/forc-projects/pyth-contract/src/main.sw
```diff
@@ -0,0 +1,934 @@
+contract;
```
maschad, line 1: Based on a conversation at sync we probably don't want to maintain this large and complicated a dependency

@petertonysmith94 had recommended this [large contract](https://github.com/FuelLabs/fuels-ts/pull/3171/files#diff-1a0ad96ab237032f55549b8e2f5bab4e04d5235a3b134a794d7d17f5090d2f7b) do we agree?

petertonysmith94: What was the rational behind not using the [large contract](https://github.com/FuelLabs/fuels-ts/blob/master/packages/fuel-gauge/test/fixtures/forc-projects/large-contract) with an ASM blob?

petertonysmith94 (commented):
I'm not a massive fan of the for-each iterations within every bench marks test and the conditional based deployments for every file. It seems there is a pattern and we could use some helper functions remove the complexity around this.



For the iterations, maybe we could extend the vitest `bench` function:

```ts
// Psuedocodeish
import { bench as vitestBench } from 'vitest';

// TODO: add conditionals for devnet
const bench = (name: string, fn: () => void) => {
  const iterateFunction = async (iterations: number = 10) => {
    for (let i = 0; i < iterations; i++) {
      await fn();
    }
  };
  return vitestBench(`${name} [iterations = 10]`, iterateFunction);
};
```

For the local/devnet deployments, could we wrap the devnet deployment functionality behind the `launchTestNode` interface?

```ts
// Psuedocodeish
const deployLiveNode = ({
  environmentConfig: typeof DEVNET_CONFIG,
  contractConfigs: DeployContractConfig[]
}) => {
  const { networkUrl } = environmentConfig;
  const provider = await Provider.create(networkUrl);
  const wallet = new WalletUnlocked(process.env.DEVNET_WALLET_PVT_KEY as string, provider);

  const deployedContracts = [];

  for (const contractConfig of contractsConfigs) {
    const { waitForResult } = await contractConfig.factory.deploy(wallet, contractConfig.options);
    const { contract } = await waitForResult();
    deployedContracts.push(contract);
  }

  return {
    provider,
    wallets: [wallet],
    contracts: deployedContracts,
    cleanup,
  };
}
```

I think this would greatly reduce the complexity and the maintenance burden of these benchmarks.

.github/workflows/bench.yml
Dhaiwat10: I think we can run the local ones on every PR too, shouldn't be a big issue?

internal/benchmarks/test/fixtures/forc-projects/pyth-contract/src/main.sw
maschad: I had wanted something that more closely resembles a realistic contract

.github/workflows/bench.yml
maschad: of course, I was saying as a well as the master branch

maschad:
@petertonysmith94 good recommendation, I've taken some inspiration from it and done a [refactor](535145d04671fd2480c6df804bbaf114e8c2311f),  but I also wanted to keep it relatively readable without too much syntactic sugar

internal/benchmarks/test/fixtures/forc-projects/pyth-contract/src/main.sw
petertonysmith94: As we're only benchmarking the deployment of blobs, I would favour using the [large contract](https://github.com/FuelLabs/fuels-ts/blob/master/packages/fuel-gauge/test/fixtures/forc-projects/large-contract) with an ASM blob.

arboleya: ASM blob seems enough.

internal/benchmarks/src/contract-interaction.bench.ts
```diff
@@ -16,55 +22,93 @@ describe('Contract Interaction Benchmarks', () => {
...
 
-  bench('should successfully execute a contract read function', async () => {
-    const tx = await contract.functions.get_count().call();
+  const runBenchmark = (name: string, benchmarkFn: () => Promise<void>) => {
+    bench(isDevnet ? name : `${name} (x${iterations} times)`, async () => {
+      for (let i = 0; i < iterations; i++) {
+        await benchmarkFn();
+      }
+    });
+  };
```
arboleya, line 69: @maschad Why not simply use `vitest` features?

https://vitest.dev/api/#bench
 - `iteration`
 - `time`
 - `warmupIteration`
 - `warmupTime`
 - `setup`
 - `teardown`

> cc @petertonysmith94

arboleya (approved):
This PR has grown out of control.

My main concern with the bench initiative was to keep it short and sweet.

Can you please try to get it back to a minimal state and manageable size?

internal/benchmarks/src/config.ts
```diff
@@ -0,0 +1,11 @@
+import { DEVNET_NETWORK_URL } from '@internal/utils';
+import { TransactionType } from 'fuels';
+
+export const DEVNET_CONFIG = {
```
danielbate, line 4: Could we merge the `e2e` configs here and import to `fuel-gauge`?

danielbate (changes requested):
I like the initiative here but I really think we should keep the Pyth contract out of the repo. I've worked this contract and it requires an understanding of the flow to viably use a lot of the functions, that doesn't really make it suitable for a quick setup / tear down test. We'd be ignoring most of the functions and have a lot of redundancy in the repo. There are probably better / simpler real life use cases out there.

internal/benchmarks/src/contract-interaction.bench.ts
petertonysmith94: I assumed that these options had been explored and that was the reason behind the iterations.

I would favour this.

internal/benchmarks/src/config.ts
maschad: I've removed this all together aea62fb1f

internal/benchmarks/src/contract-interaction.bench.ts
maschad: So I've done another set of refactors aea62fb1f 

But @arboleya @petertonysmith94 apart from the `iterations` option, where would I use the other options?

The diff now is 

```ts
const runBenchmark = (name: string, benchmarkFn: () => Promise<void>) => {
    bench(isDevnet ? name : `${name} (x${iterations} times)`, async () => {
        await benchmarkFn();
    }, {iterations});
  };
```

Which is not substantial, but I agree preferable. Recall the test name is contingent on whether it's devnet or a local node.

internal/benchmarks/test/fixtures/forc-projects/pyth-contract/src/main.sw
maschad: aea62fb1f

maschad:
Agreed @danielbate  I've removed it aea62fb1f

I've also done a series of other refactors just to clean this up a bit @petertonysmith94 @arboleya

internal/benchmarks/src/config.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { bench } from 'vitest';
+
+export const isDevnet = process.env.DEVNET_WALLET_PVT_KEY !== undefined;
+
+const iterations = isDevnet ? 1 : 10;
+
+export const runBenchmark = (name: string, benchmarkFn: () => Promise<void>) => {
+  bench(
+    isDevnet ? name : `${name} (x${iterations} times)`,
```
danielbate, line 11: Is this a big enough value add to justify having a function that we need to wrap every test?

We could just import the iterations value from a config file, that is used against the `describe`?

internal/benchmarks/src/contract-interaction.bench.ts
```diff
@@ -16,55 +23,82 @@ describe('Contract Interaction Benchmarks', () => {
...
 
-  bench('should successfully execute a contract deploy', async () => {
+  runBenchmark('should successfully execute a contract deploy', async () => {
     const factory = new CounterContractFactory(wallet);
     const { waitForResult } = await factory.deploy();
     const { contract: deployedContract } = await waitForResult();
+    expect(deployedContract).toBeDefined();
+  });
 
+  bench('should successfully execute a contract deploy as blobs', async () => {
```
danielbate, line 96: How come this isn't a `runBenchmark`?

internal/benchmarks/src/cost-estimation.bench.ts
```diff
@@ -1,36 +1,75 @@
...
+      provider = contract.provider;
+    }
+
+    request = new ScriptTransactionRequest({ gasLimit: 1000000 });
+    recipient = Wallet.generate({ provider });
+    receiver1 = Wallet.generate({ provider });
+    receiver2 = Wallet.generate({ provider });
+    receiver3 = Wallet.generate({ provider });
+    sender = Wallet.fromPrivateKey(
+      '0x30bb0bc68f5d2ec3b523cee5a65503031b40679d9c72280cd8088c2cfbc34e38',
```
danielbate, line 51: Whats the reason for this private key specifically?

internal/benchmarks/src/contract-interaction.bench.ts
maschad: Given that this is a longer operation, there is no need to do 10 iterations of it, that would not only be a lengthy benchmark but an expensive one on `devent`

internal/benchmarks/src/config.ts
maschad: This arose out of a refactor where we had the `isDevnet` conditional on each test + adding the iterations per test, **but** not all tests use the 10 iterations though. So this refactor is the sweet spot I feel.

internal/benchmarks/src/cost-estimation.bench.ts
maschad: No reason, I've removed it be9151dc3

internal/benchmarks/src/config.ts
danielbate: Just floating ideas, what about:

```ts
// config.ts
export const isDevnet = process.env.DEVNET_WALLET_PVT_KEY !== undefined;
export const iterations = isDevnet ? 1 : 10;
```
```ts
// bench.test.ts
import { iterations } from './config.ts';

describe('benchmarks', () => {
  bench('tests a thing', () => {});
  bench('tests another thing thing', () => {});
}, { iterations });
```

Then we don't need to augment `bench` and can use the other config options if we need them.

petertonysmith94: I personally like having the iterations in the name, as it's explicit to why one bench takes for example: 100ms and another takes 1000ms.

petertonysmith94 (dismissed):
🔥

internal/benchmarks/src/config.ts
maschad: Based on our sync today I clarified this refactor with @danielbate 

Also to corroborate @petertonysmith94 it's actually **required** to have these distinctions in the test names so that we get accurate reporting on Codspeed, otherwise the statistics will be convoluted (single iterations mixed with multiple iterations)

Torres-ssf (dismissed):

internal/benchmarks/src/config.ts
danielbate: Resolving due to codspeed requirement, thankyou for the clarification.

danielbate (approved):

petertonysmith94:
@maschad just a conflict with the `pnpm-lock.yaml` file here.

internal/benchmarks/src/config.ts
petertonysmith94: @danielbate can we resolve this?

petertonysmith94 (approved):
🔥

danielbate (approved):

PR #3175, opened 2024-09-16
fix: generate types for `create fuels` users when extracted

Dhaiwat10:
> Closes TS-651

# Summary

 Currently, the artifacts generated by typegen are ignored by git. But in order for someone to deploy their dapp's frontend to the cloud, they will need these artifacts to be included in their git repo - if they follow the instructions listed on our Deploying to Testnet guide. 

There is no other way to deploy the frontend since we deprecated the built-in binaries for users. This is why we need to remove these artifacts from the gitignore of the template, and include them in the source.

This PR generates these types for users when the template is extracted, and removes these files from gitignore, hence 'fixes' our deploying to testnet guide.

# Checklist


templates/nextjs/gitignore
```diff
@@ -35,11 +35,6 @@ next-env.d.ts
 
 .fuels
 
-src/sway-api/contracts
-src/sway-api/predicates
-src/sway-api/scripts
-src/sway-api/index.ts
-
```
arboleya, line 42: Is there any value in having these as a comment?

```bash
# Uncomment the lines below if you don't wish
# to commit the generated files to git.

# src/sway-api/contracts
# src/sway-api/predicates
# src/sway-api/scripts
# src/sway-api/index.ts
```

arboleya (dismissed):

Torres-ssf (dismissed):

maschad (commented):
LGTM but I wonder we should add a `pretest` equivalent build step that will generate these at least every build (perhaps at the root of the templates folder) if not each time we make a change to typegen or we when we release a new `fuels` version, otherwise we could run the risk of these becoming outdated

Dhaiwat10:
Good catch @maschad but this will be slightly more tricky than our usual `pretest` build steps because those artifacts aren't supposed to be included in the source, unlike these artifacts. We'll probably need an additonal CI step like the one that checks forc formatting, and fails if all files aren't properly formatted. Maybe we could have a similar one that fails if the geneated types are out of date? Thoughts?

Dhaiwat10:
@arboleya two questions:

1. If we don't ship the generated files in our repo, then aren't we still left with the same issue that PR is trying to solve?
2. We already have a `prebuild` script that generates these types but they are not _added_/committed to the source by the CI workflow. Does that make sense? Sorry if I'm missing something

templates/nextjs/gitignore
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/3175/commits/34bb78f9b62c05f1717d41c4d5e0f04d331e6665

arboleya:
@Dhaiwat10 
1. Isn't the issue happening when our users try to deploy? They must build the project to generate the files in advance before deploying, correct? Or does this apply to us as well?
2. Yes, if the above is correct, I'm saying that we, `fuels-ts`, should continue **_ignoring_** these files. Then, for the app we generate for users, we can override the `.gitignore` (we use in the `fuels-ts` repo) file with another one that doesn't ignore the generated files.

maschad:
> We'll probably need an additonal CI step like the one that checks forc formatting, and fails if all files aren't properly formatted. Maybe we could have a similar one that fails if the geneated types are out of date? Thoughts?

@Dhaiwat10 essentially yes, what I am suggesting is that workflow will:

1) Generate the types
2) check the diff between the currently committed ones and the generated ones.
3) If there is a diff, fail the job. Otherwise pass it.

Dhaiwat10:
Thanks @arboleya, applied your suggestion.

@maschad with the approach that Anderson suggested, we don't need the extra workflow. The files are only generated freshly when the template is extracted.

maschad (approved):
Nice work @Dhaiwat10 🚀

packages/create-fuels/src/cli.ts
```diff
@@ -139,18 +139,31 @@ export const runScaffoldCli = async ({
...
   fileCopySpinner.succeed('Copied template files!');
 
+  // Remove typegen files from gitignore
+  const gitignorePath = join(projectPath, '.gitignore');
+  const gitignoreContents = readFileSync(gitignorePath, 'utf-8');
+  const newGitIgnoreContents = gitignoreContents
+    .replace(/src\/sway-api\/contracts\n/g, '')
+    .replace(/src\/sway-api\/predicates\n/g, '')
+    .replace(/src\/sway-api\/scripts\n/g, '')
+    .replace(/src\/sway-api\/index.ts/g, '');
```
arboleya, line 149: Is there any reason you haven't kept the paths [as comments](https://github.com/FuelLabs/fuels-ts/pull/3175#discussion_r1761369499)?

packages/create-fuels/src/cli.ts
```diff
@@ -139,18 +139,31 @@ export const runScaffoldCli = async ({
...
     }).start();
-
     process.chdir(projectPath);
     execSync(packageManager.install, { stdio: verboseEnabled ? 'inherit' : 'pipe' });
-
     installDepsSpinner.succeed('Installed dependencies!');
   }
 
+  // Generate typegen files
+  if (!process.env.VITEST) {
```
arboleya, line 163: I have often given this feedback in past PRs - we should _not_ have test-related constants, conditionals, or env vars in the shipped source code. Can we achieve the same without this workaround?

Dhaiwat10: @arboleya My bad, I was just using this temporarily locally to quickly debug something and forgot to remove it. I've removed it now

packages/create-fuels/src/cli.ts
Dhaiwat10: Forgot to include this change in the new approach. Added it back. Thanks for pointing it out

packages/create-fuels/src/cli.ts
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/3175/commits/5fa7fc879060829ec73f82390a9c40524db544ee

packages/create-fuels/src/cli.ts
Dhaiwat10: https://github.com/FuelLabs/fuels-ts/pull/3175/commits/d7147517acb035863184f5b68d9f10825338c2fe

maschad (dismissed):

templates/nextjs/.gitignore
```diff
@@ -34,13 +34,13 @@ yarn-error.log*
...
-src/sway-api/index.ts
-
 sway-programs/**/out
 .turbo
 
 test-results
-playwright-report
\ No newline at end of file
+playwright-report
+
```
arboleya, line 42:
```suggestion
# Uncomment the lines below if you don't wish
# to commit the generated files to git

```

packages/create-fuels/src/cli.ts
```diff
@@ -139,18 +139,32 @@ export const runScaffoldCli = async ({
...
+  const gitignorePath = join(projectPath, '.gitignore');
+  const gitignoreContents = readFileSync(gitignorePath, 'utf-8');
+  const newGitIgnoreContents = gitignoreContents
+    .replace(
+      /src\/sway-api\/contracts\n/g,
+      "# Uncomment the lines below if you don't wish\n# to commit the generated files to git.\n\n# src/sway-api/contracts\n"
+    )
+    .replace(/src\/sway-api\/predicates\n/g, '# src/sway-api/predicates\n')
+    .replace(/src\/sway-api\/scripts\n/g, '# src/sway-api/scripts\n')
+    .replace(/src\/sway-api\/index.ts/g, '# src/sway-api/index.ts');
```
arboleya, line 152:
```suggestion
    .replace(/^(src\/sway-api\.+)$/mg, '# $1');
```

templates/nextjs/gitignore
```diff
@@ -35,13 +35,13 @@ next-env.d.ts
...
 sway-programs/**/out
 .turbo
 
 test-results
-playwright-report
\ No newline at end of file
+playwright-report
+
```
arboleya, line 43:
```suggestion
# Uncomment the lines below if you don't wish
# to commit the generated files to git

```

templates/vite/.gitignore
```diff
@@ -25,13 +25,13 @@ dist-ssr
...
-src/sway-api/index.ts
-
 sway-programs/**/out
 .turbo
 
 test-results
-playwright-report
\ No newline at end of file
+playwright-report
+
```
arboleya, line 33:
```suggestion
# Uncomment the lines below if you don't wish
# to commit the generated files to git

```

templates/vite/gitignore
```diff
@@ -25,13 +25,13 @@ dist-ssr
...
-src/sway-api/index.ts
-
 sway-programs/**/out
 .turbo
 
 test-results
-playwright-report
\ No newline at end of file
+playwright-report
+
```
arboleya, line 33:
```suggestion
# Uncomment the lines below if you don't wish
# to commit the generated files to git

```

arboleya (dismissed):

Torres-ssf (approved):

danielbate (approved):

packages/create-fuels/src/cli.ts
```diff
@@ -139,18 +139,25 @@ export const runScaffoldCli = async ({
...
     }).start();
-
     process.chdir(projectPath);
     execSync(packageManager.install, { stdio: verboseEnabled ? 'inherit' : 'pipe' });
-
     installDepsSpinner.succeed('Installed dependencies!');
   }
 
+  // Generate typegen files
+  execSync(packageManager.run('prebuild'), { stdio: verboseEnabled ? 'inherit' : 'pipe' });
```
petertonysmith94, line 159: Is it worth extract these options `{ stdio: verboseEnabled ? 'inherit' : 'pipe' }` elsewhere?

petertonysmith94 (commented):
@Dhaiwat10 might be worthwhile releasing to NPM so we can test.

packages/create-fuels/src/cli.ts
Dhaiwat10: It's only used in two places tbf, feels like overkill

Dhaiwat10:
https://github.com/user-attachments/assets/207a6c5e-1044-459c-ab95-216e793d1034

@petertonysmith94 this video should demonstrate the purpose of this PR. As you can see, the `sway-api` folder is already populated with the types and the types have been removed from the gitignore, enabling users to deploy dapps to the cloud even if the cloud environment doesnt have forc and fuel-core available at build-time.

arboleya (approved):

maschad (approved):

Torres-ssf (approved):

danielbate (approved):

packages/create-fuels/src/cli.ts
danielbate: Resolving as outdated and fixed in above commit.

packages/create-fuels/src/cli.ts
danielbate: Resolving as fixed in above commit.

petertonysmith94:
> types-pr-screen.mp4 
> @petertonysmith94 this video should demonstrate the purpose of this PR. As you can see, the `sway-api` folder is already populated with the types and the types have been removed from the gitignore, enabling users to deploy dapps to the cloud even if the cloud environment doesnt have forc and fuel-core available at build-time.

Beautiful video @Dhaiwat10, thanks for that. What screen grabber do you use to record this?

PR #3106, opened 2024-09-04
chore: deprecate all receipt coders

Torres-ssf:
- Closes #3105

# Release notes

In this release, we:

- Deprecate all receipt coders

# Checklist

- [x] All **changes** are **covered** by **tests** (or not applicable)
- [x] All **changes** are **documented** (or not applicable)
- [x] I **reviewed** the **entire PR** myself (preferably, on GH UI)
- [x] I **described** all **Breaking Changes** (or there's none)

PR #3020, opened 2024-08-23
chore: made `chainConfig` files match testnet configuration

Torres-ssf:
# Summary

This PR updates the `chainConfig` files in our tests to align with the latest testnet upgrade, ensuring consistency across the testing environment. The `gasCosts` new values resulted in expressively reduced transaction fees, requiring some fine-tuning of a few tests.

# Checklist

- [x] All **changes** are **covered** by **tests** (or not applicable)
- [x] All **changes** are **documented** (or not applicable)
- [x] I **reviewed** the **entire PR** myself (preferably, on GH UI)
- [x] I **described** all **Breaking Changes** (or there's none)

PR #3180, opened 2024-09-18
chore: deprecate unused operation names

### LINKED ISSUE #3008: Unused `OperationName` Enum Entries in the SDK
Not all entries of the `OperationName` enum seem to be in use. Right now the SDK does not assemble operations for the following enum values:

1. `OperationName.mint`
_Intended to be used when the transaction minted some coins_
2. `OperationName.predicatecall`
_When a predicate code was executed within the TX_
3. `OperationName.script`
_Presumably for script calls, though its exact usage is unclear._
4. `OperationName.sent`
_The purpose of this entry is not well defined._ 

These entries might need to be reviewed and either implemented where necessary or removed if they are irrelevant/redundant.

Torres-ssf:
- Closes #3008

# Checklist


maschad (changes requested):
Although the `OperationName` enum is _hidden_ it's still exported by the `account` package and therefore it's a breaking change no? Perhaps we should add a deprecation notice instead.

petertonysmith94:
> Although the `OperationName` enum is _hidden_ it's still exported by the `account` package and therefore it's a breaking change no? Perhaps we should add a deprecation notice instead.

I would concur.

Torres-ssf:
But @maschad this is only internally used.

This enum doesn't have any purpose for users whatsoever. Especially the removed entries as the SDK is not even using them.

We don't even have operations that use these enums.

Dhaiwat10:
@maschad Is it _practically_ a breaking change though? Like Sergio mentioned, there is no use of this enum for anyone really

maschad:
The issue here is that we can't guarantee that a consumer downstream didn't use this enum in their code since it was exported and as such if we remove properties from it then it would break their code.

arboleya (approved):

arboleya:
@maschad This seems like junk code that doesn't fit and can't be used anywhere - or can it?

@Torres-ssf I believe we had another PR like this a few weeks ago, removing some broken/obsolete code. I couldn't find it, maybe you remember something.

maschad:
@arboleya I believe you may be referring to [this PR](https://github.com/FuelLabs/fuels-ts/pull/3106)

My understanding from [that discussion](https://github.com/FuelLabs/fuels-ts/pull/3106#discussion_r1744994813) as well as [this one](https://github.com/FuelLabs/fuels-ts/pull/2963/files#r1720170592)  (where I had previously advocated for it to be non-breaking) is that even in scenarios where consumers _more than likely_ would **not** have used those exports, we were erring on the side of caution, and thus taking the more conservative approach to treat the removal of any exported code as breaking. I am open to discuss but this is my current understanding of the consensus around these types of changes.

Torres-ssf:
I believe @arboleya is talking about https://github.com/FuelLabs/fuels-ts/pull/3020, specifically about this [change](https://github.com/FuelLabs/fuels-ts/pull/3020/files#diff-e832863a29008135e9565254c0c198c218d2f69be4cc80b0550b578554d7f04b) here. 

These operations were never used and implemented on the TS SDK, therefore aside from their names, they do not exist.

petertonysmith94 (dismissed):

petertonysmith94 (commented):
The `OperationName` being removed here, are being [used](https://github.com/FuelLabs/fuels-wallet/blob/9f4334cd2c8d7f56bc2f101ddd3c052e0d1a4303/packages/app/src/systems/Transaction/utils/icon.ts#L6-L13) in the `fuels-wallet`.

I believe we should just depreciate these.

danielbate (commented):
+1 with the [sentiment here](https://github.com/FuelLabs/fuels-ts/pull/3180#issuecomment-2359462420) and exercising caution. Lets add to #3119.

Torres-ssf:
> The `OperationName` being removed here, are being [used](https://github.com/FuelLabs/fuels-wallet/blob/9f4334cd2c8d7f56bc2f101ddd3c052e0d1a4303/packages/app/src/systems/Transaction/utils/icon.ts#L6-L13) in the `fuels-wallet`.
> 
> I believe we should just depreciate these.

Good catch @petertonysmith94 

@LuizAsFight I was under the [impression](https://github.com/FuelLabs/fuels-ts/issues/3008#issuecomment-2347931044) that this was not being used.

petertonysmith94:
@Torres-ssf Do you think we should just add deprecation notices to these enum values for now, and add to #3119 issue?

petertonysmith94:
@Torres-ssf Not sure why this one is [failing](https://vercel.com/fuel-labs/ts-api-docs/5RAh9uRvUqw2K7vzgiR4GxktBJu3?filter=errors#L5)

Torres-ssf:
> The `OperationName` being removed here, are being [used](https://github.com/FuelLabs/fuels-wallet/blob/9f4334cd2c8d7f56bc2f101ddd3c052e0d1a4303/packages/app/src/systems/Transaction/utils/icon.ts#L6-L13) in the `fuels-wallet`.
> 
> I believe we should just depreciate these.

@petertonysmith94 The enums are only being used within an image map, where the Wallet SDK assigns different images to operations based on their names. The idea is to display different images within the Wallet APP, based on the operation.

However, these operation enum entries were never actually used in assembling any operation. Operations are assembled entirely within the TS SDK side, and we never had code that used these enum entries.

Therefore, the Wallet SDK would simply need to remove the unused entries from the image map, when updating to the new `fuels` version, as this has no side effects whatsoever.

To address everyone's concerns that some users might be relying on these enums (even though it's highly unlikely, like a bald man using a comb 😆 ), I've marked each unused enum entry with the `@deprecated` tag.

> @Torres-ssf Not sure why this one is [failing](https://vercel.com/fuel-labs/ts-api-docs/5RAh9uRvUqw2K7vzgiR4GxktBJu3?filter=errors#L5)


I am unsure why this one is falling. It seems to be failing on other [PRs](https://github.com/FuelLabs/fuels-ts/pull/3238) as well

arboleya (approved):

maschad (approved):

danielbate (approved):

PR #3184, opened 2024-09-19
feat: support proxied contract calls

### LINKED ISSUE #3123: Add support for proxy contracts in `fuels deploy`
# Intro

It goes like this:

```toml
[proxy]
enabled = true
```

Docs:
 - https://docs.fuel.network/docs/forc/plugins/forc_client/#proxy-contracts

# Tips

We currently have this helper:
https://github.com/FuelLabs/fuels-ts/blob/751d638fd3324e41906f9170adb3d8caa781414a/packages/fuels/src/cli/config/forcUtils.ts#L35

It is used here, but the `toml` contents get discarded when the method finishes:
https://github.com/FuelLabs/fuels-ts/blob/56018e3720898571074bd2d3994ab48ca44d7185/packages/fuels/src/cli/config/loadConfig.ts#L92

The final touch will probably have to be around here:
https://github.com/FuelLabs/fuels-ts/blob/8faeaa401723f98b5f573700934b786d2ae4dbca/packages/fuels/src/cli/commands/deploy/deployContract.ts#L28

Ultimately, we should probably persist the TOML contents with the other artifacts.

# Notes

I wonder if it'd be a good moment to address this issue together with:
 - https://github.com/FuelLabs/fuels-ts/issues/989

### LINKED ISSUE #3145: Provide better ergonomics for proxied contract calls
We should provide a type-safe API that can work behind proxied contracts.

This should go in tandem with:
 - https://github.com/FuelLabs/fuels-ts/issues/3123

Original request:
 - So, the proxy contract works on dynamically dispatching calls based on the function selector. Previously, to call a contract method, they could use the TS SDK to generate the types for the contract and call the contract by `contractA.functions.someRandomFunc().call()`. However, they're saying for the proxy contract, there isn't any way to call a particular function (say, "someRandomFunc") that is defined in the implementation because the base proxy contract has essentially just a "fallback" method. They've found [this](https://github.com/FuelLabs/proxy-chunks-minimal-example/tree/main), but it mentions the usage of forc that isn't ideal for their use case.

Torres-ssf:
> Closes TS-644

 - Closes #3145

# Checklist

### DIFF
--- a/packages/fuel-gauge/src/contract.test.ts
+++ b/packages/fuel-gauge/src/contract.test.ts
@@ -13,14 +13,19 @@ import {
   Predicate,
   PolicyType,
   buildFunctionResult,
+  ReceiptType,
 } from 'fuels';
-import type { ScriptTransactionRequest, TransferParams } from 'fuels';
+import type { ScriptTransactionRequest, TransactionResultCallReceipt, TransferParams } from 'fuels';
 import { expectToThrowFuelError, ASSET_A, ASSET_B, launchTestNode } from 'fuels/test-utils';
 import type { DeployContractConfig } from 'fuels/test-utils';
 
 import {
   CallTestContract,
   CallTestContractFactory,
+  CoverageContract,
+  CoverageContractFactory,
+  ProxySrc14Contract,
+  ProxySrc14ContractFactory,
   StorageTestContract,
   StorageTestContractFactory,
 } from '../test/typegen/contracts';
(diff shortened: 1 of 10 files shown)


packages/fuel-gauge/test/fixtures/forc-projects/proxy-src14-contract/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "proxy-src14-contract"
+
```
danielbate, line 6: Should this have the `proxy` flag?

packages/fuel-gauge/src/contract.test.ts
```diff
@@ -1111,4 +1116,61 @@ describe('Contract', () => {
...
+    await call.waitForResult();
+
+    const proxyId = proxyContract.id;
+    const echoValuesId = coverageContract.id;
+
+    const proxiedConverageContract = new Contract(
+      echoValuesId,
+      CoverageContract.abi,
+      wallet,
+      proxyId
```
danielbate, line 1150: Do you think we could just read this value from the SRC14 storage? Then we could just pass some `{ proxy: true }` config and the user only needs to be aware of one contract ID that doesn't change. 

Only issue is in the contract call, we'd first need to read the value, with a function call to `proxy_target` using the existing contract ID. Then, we'd perform the user dictated function call against the received contract ID.

danielbate: We could then possibly use the `proxy` flag to modify the `Interface` and merge the passed ABI with an `SRC14` ABI?

packages/fuel-gauge/test/fixtures/forc-projects/proxy-src14-contract/Forc.toml
Torres-ssf: @danielbate Good catch.

packages/fuel-gauge/src/contract.test.ts
Torres-ssf: @danielbate I was planning to solve this with Typegen. Since the flag `[proxy]` is available in the `Forc` file, then we could use it. So when connecting with the contract or deploying it, if the flag is present we can already handle things on the fly.

With that, I believe we solve the problem of having the user manually inform the proxy contract ID and avoid executing the call to read the target contract at the proxy contract, which would represent an additional `dryRun` since it is a read call.

About modifying the proxy contract `Interface`, I made a couple of tries but didn't succeed. Maybe I am missing something.

danielbate: Typegen is a good idea, will we still need a way for users to connect that haven't used a typegen factory though? Just a standard `new Contract()`?

Even with typegen though, how do we know the ID of the target contract without the `dryRun`, if it hasn't been deployed by us?

> About modifying the proxy contract Interface, I made a couple of tries but didn't succeed. Maybe I am missing something.

Lets see if we can solve this later, not a massive deal as the user will ideally just want to call the target contract functions anyways.

Torres-ssf: > Even with typegen though, how do we know the ID of the target contract without the dryRun, if it hasn't been deployed by us?

@danielbate It seems the solution you idealize for this is different from the one that I am approaching. In the context described here, the user would have only in its hands the contract Proxy ID and wants to execute a contract call for the target proxy, and they do not have the ID of the deployed target contract, right?

This is why you are suggesting a way to modify the proxy contract interface using the target contract interface. Therefore the user would use just the proxy contract ID to execute the call, then, behind the scenes, we would:

1 - Instantiate the Proxy contract, using also the target contract JSON ABI to have all its existent functions.

2 - Execute a read only call to the Proxy contract to get the target contract ID

3 - Build the script for the contract with the Proxy contract ID and execute the actual contract call specified by the user.

Is that your proposal?

danielbate: Yes exactly, well summarised.

packages/fuel-gauge/test/fixtures/forc-projects/proxy-src14-contract/Forc.toml
danielbate: From my investigation, I actually think the flag goes on the target contract. Not the SRC14 compliant contract.

Torres-ssf:
> Closes TS-644

 - Closes #3145

PR #3282, opened 2024-10-09
feat: source request header

### LINKED ISSUE #3280: Consider adding especial header to GQL requests
The idea is that we could add a `source: ts-sdk-<version>` to know which transactions originated from the TS SDK and what version submitted them.

danielbate:
- Closes #3280

# Release notes

In this release, we:

- Added a new header for GraphQL requests with the `fuels` version used

# Checklist

### DIFF
--- a/packages/account/src/providers/provider.ts
+++ b/packages/account/src/providers/provider.ts
@@ -5,7 +5,7 @@ import { BN, bn } from '@fuel-ts/math';
 import type { Transaction } from '@fuel-ts/transactions';
 import { InputType, InputMessageCoder, TransactionCoder } from '@fuel-ts/transactions';
 import { arrayify, hexlify, DateTime, isDefined } from '@fuel-ts/utils';
-import { checkFuelCoreVersionCompatibility } from '@fuel-ts/versions';
+import { checkFuelCoreVersionCompatibility, versions } from '@fuel-ts/versions';
 import { equalBytes } from '@noble/curves/abstract/utils';
 import type { DocumentNode } from 'graphql';
 import { GraphQLClient } from 'graphql-request';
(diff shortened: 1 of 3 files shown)


nedsalk (approved):

petertonysmith94 (approved):
🚀

packages/account/src/providers/provider.ts
```diff
@@ -471,16 +471,20 @@ export default class Provider {
...
     this.url = rawUrl;
     this.urlWithoutAuth = urlWithoutAuth;
-    this.options = { ...this.options, ...options };
     this.url = url;
 
-    if (headers) {
-      this.options = { ...this.options, headers: { ...this.options.headers, ...headers } };
-    }
+    const { FUELS } = versions;
+    const headers = { ...authHeaders, ...options.headers, Source: `ts-sdk-${FUELS}` };
```
petertonysmith94, line 481:
```suggestion
    const headers = { ...authHeaders, ...options.headers, source: `ts-sdk-${FUELS}` };
```

petertonysmith94 (commented):
What happens when this is a tmp release?

petertonysmith94 (approved):

packages/account/src/providers/provider.ts
petertonysmith94, line 481:
```suggestion
    const headers = { ...authHeaders, Source: `ts-sdk-${FUELS}`, ...options.headers };
```

Maybe allowing for overrides here would be beneficial, then we could also determine if say it's a connector/app?

nedsalk: They can always add their own header here, e.g. `source-connector: 'best-connector-ever'`.

nedsalk:
> What happens when this is a tmp release?

I'd add this support in a follow-up PR as it'd require more work to implement given that the version of a temporary release isn't stored in the sdk. We'd either have to read the version installed on the user's disk or add some automation to update the `getBuiltinVersions` return for temporary releases.

petertonysmith94:
> > What happens when this is a tmp release?
> 
> I'd add this support in a follow-up PR as it'd require more work to implement given that the version of a temporary release isn't stored in the sdk. We'd either have to read the version installed on the user's disk or add some automation to update the `getBuiltinVersions` return for temporary releases.

Already on it

Dhaiwat10 (approved):

Issue #3281, opened 2024-10-08
Review and optimize GQL operations


arboleya:
During some load tests, we realized a couple of issues ([#3279](https://github.com/FuelLabs/fuels-ts/issues/3279), [#3277](https://github.com/FuelLabs/fuels-ts/issues/3277)) that we're fixing and optimizing.

In keeping with this theme, we should continue the effort and review our predefined GQL queries and fragments.

Every field removal _is_ a breaking change and _can_ break things for apps, so we need to be extra careful here.

Taking from the [`providers/operations.graphql`](https://github.com/FuelLabs/fuels-ts/blob/d0e9b1a110721fa721aa40f87d7b5515be16b9b3/packages/account/src/providers/operations.graphql):

- [ ] `getVersion`
- [ ] `getNodeInfo`
- [ ] `getChain`
- [ ] `getTransaction`
- [ ] `getTransactionWithReceipts`
- [ ] `getTransactions`
- [ ] `getTransactionsByOwner`
- [ ] `estimatePredicates`
- [ ] `getBlock`
- [ ] `getBlockWithTransactions`
- [ ] `getBlocks`
- [ ] `getCoin`
- [ ] `getCoins`
- [ ] `getCoinsToSpend`
- [ ] `getContract`
- [ ] `getContractBalance`
- [ ] `getBalance`
- [ ] `getLatestGasPrice`
- [ ] `estimateGasPrice`
- [ ] `getBalances`
- [ ] `getMessages`
- [ ] `getMessageProof`
- [ ] `getMessageStatus`
- [ ] `getRelayedTransactionStatus`
- [ ] `getMessageByNonce`

petertonysmith94:
Currently what @Green is doing to capture the GQL is setup a proxy to intercept the requests and dump them into the file `traffic.dump` via the following command:

```
mitmdump --mode reverse:https://mtls-testnet.fuel.network -w traffic.dump
```

petertonysmith94:
List of all operations used **internally** in `provider.ts`

```markdown
dryRun
estimateGasPrice
estimatePredicates
getBalance
getBalances
getBlobs
getBlock
getBlockWithTransactions
getBlocks
getChain
getCoins
getCoinsToSpend
getContract
getContractBalance
getLatestGasPrice
getMessageByNonce
getMessageProof
getMessageStatus
getMessages
getNodeInfo
getRelayedTransactionStatus
getTransaction
getTransactions
getVersion
produceBlocks
[...]

PR #2925, opened 2024-08-13
chore: improve node incompatibility warning

Torres-ssf:
- Closes https://github.com/FuelLabs/fuels-ts/issues/2283

# Checklist

- [x] All **changes** are **covered** by **tests** (or not applicable)
- [x] All **changes** are **documented** (or not applicable)
- [x] I **reviewed** the **entire PR** myself (preferably, on GH UI)
- [x] I **described** all **Breaking Changes** (or there's none)

PR #3284, opened 2024-10-09
chore: merge `chain` and `nodeInfo` queries into one

nedsalk:
- Part of #3281 

# Summary

This PR adds a new `getChainAndNodeInfo` query which returns the `chain` and `nodeInfo` data in one request, as opposed to the way we've been doing up until now where we've been sending them separately.

# Checklist

### DIFF
--- a/packages/account/src/providers/provider.test.ts
+++ b/packages/account/src/providers/provider.test.ts
@@ -1051,36 +1051,31 @@ describe('Provider', () => {
     const { provider } = launched;
 
     expect(spyFetchChainAndNodeInfo).toHaveBeenCalledTimes(1);
-    expect(spyFetchChain).toHaveBeenCalledTimes(1);
-    expect(spyFetchNode).toHaveBeenCalledTimes(1);
 
     provider.getChain();
     provider.getNode();
 
     expect(spyFetchChainAndNodeInfo).toHaveBeenCalledTimes(1);
-    expect(spyFetchChain).toHaveBeenCalledTimes(1);
-    expect(spyFetchNode).toHaveBeenCalledTimes(1);
+    expect(spyFetchChain).toHaveBeenCalledTimes(0);
+    expect(spyFetchNode).toHaveBeenCalledTimes(0);
   });
 
   it('should ensure fetchChainAndNodeInfo always fetch new data', async () => {
     Provider.clearChainAndNodeCaches();
 
     const spyFetchChainAndNodeInfo = vi.spyOn(Provider.prototype, 'fetchChainAndNodeInfo');
-    const spyFetchChain = vi.spyOn(Provider.prototype, 'fetchChain');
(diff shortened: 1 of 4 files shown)


packages/account/src/providers/provider.test.ts
```diff
@@ -1188,50 +1183,6 @@ Supported fuel-core version: ${mock.supportedVersion}.`
     );
   });
 
-  it('should ensure fuel node version warning is shown before chain incompatibility error', async () => {
```
nedsalk, line 1191: This was added in #2925. My argument is that these chain incompatibility errors won't be happening anymore due to great care taken by the `fuel-core` team in ensuring backwards compatibility, so we might as well favor optimizing the query over showing the warning.

Dhaiwat10: Do we need to make a network request to make this check?

nedsalk: The flow was:
1. fetch node info
2. if fuels doesn't support the node, write a warning to the user
3. fetch the chain info
4. if the chain info query fails, the user will at least have the knowledge that the sdk doesn't support the `fuel-core` version they're querying

The feature was motivated by the frequent breaking changes happening to the graphql api, but it's stable now and these errors shouldn't occur, so IMO we should favor performance.

packages/account/src/providers/provider.ts
```diff
@@ -612,13 +612,26 @@ export default class Provider {
...
+
+    const processedNodeInfo: NodeInfo = {
+      maxDepth: bn(nodeInfo.maxDepth),
+      maxTx: bn(nodeInfo.maxTx),
+      nodeVersion: nodeInfo.nodeVersion,
+      utxoValidation: nodeInfo.utxoValidation,
+      vmBacktrace: nodeInfo.vmBacktrace,
+    };
+
+    Provider.ensureClientVersionIsSupported(processedNodeInfo);
```
petertonysmith94, line 625: @nedsalk We still print the warning tho?

https://github.com/FuelLabs/fuels-ts/pull/3284/files#r1793066249

nedsalk: Yes.

packages/account/src/providers/provider.test.ts
nedsalk: > I'll have a think on this one - although I agree with the sentiment and there has been a lot of confusion around this warning - maybe removal at this time is not right.

The warning won't be there only if this specific query `getChainAndNodeInfo` fails because it's logged right after the query is successful. If this query is successful, the warning will be logged and can be of help if some subsequent queries fail.

Dhaiwat10: Fair. I think the benefits outweigh the cons of removing this feature here.

danielbate (approved):

arboleya (approved):

maschad (approved):

PR #3301, opened 2024-10-10
chore!: optimize coin gql queries

### LINKED ISSUE #3291: Optimize coin's graphql queries
During some load tests, we realized a couple of issues ([#3279](https://github.com/FuelLabs/fuels-ts/issues/3279), [#3277](https://github.com/FuelLabs/fuels-ts/issues/3277)) that we're fixing and optimizing.

In keeping with this theme, we should continue the effort and review our predefined GQL queries and fragments.

Every field removal _is_ a breaking change and _can_ break things for apps, so we need to be extra careful here.

Taking from the [`providers/operations.graphql`](https://github.com/FuelLabs/fuels-ts/blob/d0e9b1a110721fa721aa40f87d7b5515be16b9b3/packages/account/src/providers/operations.graphql):

- [x] `getCoin`
- [x] `getCoins`
- [x] `getCoinsToSpend`

maschad:
- Closes #3291

# Release notes

In this release, we:

- Optimized the `getCoins` provider endpoint

# Summary

- Removed `getCoin` gql query from the provider operations as it is unused in our SDK.
- Removed the `owner` from the `coinFragment` as it can be populated in the calling functions for both:
  - `getCoins`
  - `getResourcesToSpend`

# Breaking Changes

- The `Provider.operations.getCoins()` and  `Provider.operations.getCoinsToSpend` function no longer return the owner. These methods shouldn't be called directly but are used internally to formulate responses from the SDK.

- Removed the property `owner` from the `Provider.operations.getCoinsToSpend()` function. Suggest to use the owner from the input parameters.

# Checklist


.changeset/young-steaks-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/account": patch
+---
+
+fix: optimize coin gql queries
```
danielbate, line 5:
```suggestion
---
"@fuel-ts/account": minor
---

fix!: optimize coin gql queries
```

This will be a breaking change because we are removing the `getCoin` query, which you'll know longer be able to call on `Provider.operations.getCoin()`. This is okay with me, we can just recommend `Provider.operations.getCoins()` in breaking change notes.

danielbate (changes requested):

.changeset/young-steaks-brake.md
danielbate: For speed, I'm pushing this change and will add the breaking change notes.

danielbate (approved):

nedsalk (dismissed):

Dhaiwat10 (dismissed):
Can you also mention the removal of the `owner` field in the breaking changes notice?

.changeset/young-steaks-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/account": minor
+---
+
+fix!: optimize coin gql queries
```
petertonysmith94, line 5:
```suggestion
chore!: optimize coin gql queries
```

packages/account/src/providers/provider.ts
```diff
@@ -1360,7 +1360,7 @@ Supported fuel-core version: ${supportedVersion}.`
       id: node.utxoId,
       assetId: node.assetId,
       amount: bn(node.amount),
-      owner: Address.fromAddressOrString(node.owner),
+      owner: Address.fromAddressOrString(owner),
```
petertonysmith94, line 1363:
```suggestion
      owner: ownerAddress,
```

Dhaiwat10:
I was wondering if this is actually an optimization since this just reduces the size of our source code. If someone was making a `getCoin` query, they will still need to make that query - just via a different function now. The network payloads are not being reduced apart from the `owner` field being removed 👀

nedsalk:
> @Dhaiwat10 [...] The network payloads are not being reduced apart from the owner field being removed 👀

That's the optimization, and he removed the other query just because he saw that it's not necessary, which I'm okay with.

Dhaiwat10:
@nedsalk but one of our users could be using it, right? This feels like a rather unnecessary breaking change

maschad:
I've decided to restore the `getCoins` method but this is still breaking as the owner is no longer returned in the `...coinFragment` gql fragment

nedsalk (approved):

danielbate (approved):

Torres-ssf (approved):

PR #3307, opened 2024-10-10
feat: add feature to check if hex is an account

maschad:
# Release notes

In this release, we:

- Added Provider methods `isUserAccount` and `getAddressType` which indicate the type of the hex passed

# Summary

We now have two methods on the `Provider` to check the account type

# Checklist

### DIFF
--- a/.github/workflows/pr-release.yaml
+++ b/.github/workflows/pr-release.yaml
@@ -8,7 +8,7 @@ jobs:
     name: "Release PR to npm"
     runs-on: ubuntu-latest
     # comment out if:false to enable release PR to npm
-    # if: false
+    if: false
     permissions: write-all
     steps:
       - name: Checkout
--- a/.github/workflows/pr-release.yaml
+++ b/.github/workflows/pr-release.yaml
@@ -8,7 +8,7 @@ jobs:
     name: "Release PR to npm"
     runs-on: ubuntu-latest
     # comment out if:false to enable release PR to npm
-    if: false
+    # if: false
     permissions: write-all
     steps:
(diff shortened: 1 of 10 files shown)


packages/account/src/providers/provider.ts
```diff
@@ -1950,6 +1950,53 @@ Supported fuel-core version: ${supportedVersion}.`
...
+        transactionId: id,
+      });
+
+      if (contract || blob || transaction) {
+        return false;
+      }
+      return true;
+    } catch (error) {
+      return true;
+    }
```
Torres-ssf, line 1973: @Why do we need this validation here? I wouldn't make it possible to false positives?

maschad: Good catch 8bb3be33f

Torres-ssf (approved):

.github/workflows/pr-release.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
     name: "Release PR to npm"
     runs-on: ubuntu-latest
     # comment out if:false to enable release PR to npm
-    if: false
+    # if: false
```
Torres-ssf: We need to re-comment this line

```suggestion
    if: false
```

Torres-ssf (approved):

.github/workflows/pr-release.yaml
Torres-ssf: My bad it was removed already

arboleya (approved):

petertonysmith94 (approved):

PR #3319, opened 2024-10-13
feat: using gas modifier at `Provider.getTransactionCost`

### LINKED ISSUE #3317: Add gas padding when estimating transactions
The SDK should implement a gas padding strategy. If a transaction is estimated to require `10_000` gas, the SDK can automatically add a 20% padding, increasing the gas limit to `12_000`.

This is necessary because scripts and contracts are stateful, and the actual gas consumption can change due to another transaction modifying the state. Adding this padding would help avoid unexpected out-of-gas errors and improve transaction reliability.

Torres-ssf:
- Closes #3317

# Release notes

In this release, we:

- Use a modifier of `20%` for estimated gas 

# Checklist


packages/account/src/providers/provider.ts
```diff
@@ -69,6 +69,7 @@ export const RESOURCES_PAGE_SIZE_LIMIT = 512;
 export const TRANSACTIONS_PAGE_SIZE_LIMIT = 60;
 export const BLOCKS_PAGE_SIZE_LIMIT = 5;
 export const DEFAULT_RESOURCE_CACHE_TTL = 20_000; // 20 seconds
+export const GAS_USED_MODIFIER = 1.3;
```
arboleya, line 72: Shouldn't it be 20%?

Torres-ssf: @arboleya Yes, @luizstacio suggested using 30%, but after running some tests, I believe it’s better to stick with the initial value of 20%. The issue with using a higher value is that it might cause us to exceed the maximum allowed gas.

arboleya: That can happen with any multiplier, right?

Perhaps a `Math.max(gas, MAX_GAS)` could have a place here.

Torres-ssf: @arboleya I was on that already. I decided to spend some time running some validations https://github.com/FuelLabs/fuels-ts/pull/3319/commits/66b43b129481dc5b092133f6b259451f50137e0e

arboleya: I'm sure you were. 😄

maschad (approved):

arboleya (approved):

PR #3324, opened 2024-10-13
feat: ensure that we fetch node info per estimation

### LINKED ISSUE #3315: Improve `Provider` caching strategy
### Summary 

Yesterday, we upgraded the gas price factor to be bigger, and due to caching, the SDK used old values for consensus parameters.

It would be nice if SDK periodically, on each request, or in the case of an error, requested `consensusParametersVersion` and compared it with the local version. If it is different, fetch new consensus parameters. 

Currently, if the gas prices are wrong when submitting a transaction, we will receive an error , and the transaction will fail

> Since it is launched, we can adjust them maybe each week to make it cheaper/expensive. In the case of an attack, we can change them immediately.

```graphql
query {
  chain {
    latestBlock {
      header {
        consensusParametersVersion
        stateTransitionBytecodeVersion
      }
    }
  }
}
```

Incoming from @xgreenx

maschad:
Closes #3315

# Release notes

In this release, we:

-  Improved `Provider` cache to self-clean on TX dry-run/estimation

# Summary

Given that the consensus parameters _may_ be updated on a weekly basis, we should try and fetch new consensus parameters before submitting a transaction. Kudos to @nedsalk and @arboleya for the assistance.

# Checklist


packages/account/src/providers/provider.ts
```diff
@@ -1150,6 +1152,8 @@ Supported fuel-core version: ${supportedVersion}.`
     const { transactionRequest } = params;
     let { gasPrice } = params;
 
+    await this.fetchChainAndNodeInfo(true);
```
arboleya, line 1155: Would it be easier to keep everything as it was and have this here instead?

```suggestion
    await Provider.clearChainAndNodeCaches();
    await this.fetchChainAndNodeInfo();
```

.changeset/silent-otters-accept.md
```diff
@@ -0,0 +1,6 @@
+---
+"@fuel-ts/account": patch
+"fuels": patch
+---
+
+fix: ensure that we fetch node info per estimation
```
arboleya, line 6: Can we go with a more concise release note in the PR desription?

 - Improved `Provider` cache to self-clean on TX dry-run/estimation

packages/account/src/providers/provider.ts
```diff
@@ -598,39 +598,41 @@ export default class Provider {
...
+      vmBacktrace: data.nodeInfo.vmBacktrace,
+    };
+
+    Provider.ensureClientVersionIsSupported(nodeInfo);
+    const chain = processGqlChain(data.chain);
+    Provider.chainInfoCache[this.urlWithoutAuth] = chain;
+    Provider.nodeInfoCache[this.urlWithoutAuth] = nodeInfo;
+
+    return { chain, nodeInfo };
+  }
```
Torres-ssf, line 618: @maschad It seems this is just a code refactoring. What am I missing?

packages/account/src/providers/provider.ts
arboleya: On a second look, invalidating the cache on every tx estimation doesn't seem as efficient as only when the consensus version changes.

packages/account/src/providers/provider.ts
Torres-ssf: Ok, you are using the `refreshCache` flag. I understood it now

arboleya: Maybe [this](https://github.com/FuelLabs/fuels-ts/pull/3324/files/00fd2ee1f9b6397bb698251774663d1338221fa3#diff-69a6670d0426c74935ed2830ce552e7df2379c2d112a582306e57569fcd17029R1155).

packages/account/src/providers/provider.ts
```diff
@@ -598,39 +598,41 @@ export default class Provider {
     await this.fetchChainAndNodeInfo();
   }
 
+  async refreshCacheFunction(): Promise<{ chain: ChainInfo; nodeInfo: NodeInfo }> {
```
nedsalk, line 601:
```suggestion
  private async refreshCacheFunction(): Promise<{ chain: ChainInfo; nodeInfo: NodeInfo }> {
```

But I'm thinking we could just do what's mentioned in @arboleya's [comment](https://github.com/FuelLabs/fuels-ts/pull/3324/files#r1798469752) and call it a day.

packages/account/src/providers/provider.ts
nedsalk: Yes, this is just a refactoring. The main change related to the cache is the addition of the `refreshCache` parameter in `fetchChainAndNodeInfo` below.

.changeset/silent-otters-accept.md
maschad: Good idea

packages/account/src/providers/provider.ts
maschad: You're right, and it's definitely introduced some performance regressions, better to monitor the consensus version.

arboleya: On a third look, this blocks every tx estimation and degrades the experience. This check must be non-blocking.


```ts
// no await / non-blocking
Provider.autoRefreshCache().catch(..);
```

Then:

```ts
// pseudo-code
function autoRefreshCache () {
  const previous = Provider.consensusParamsVersion;
  const current = Provider.getConsensusVersion();

  if (previous !== current) {
    Provider.clearChainAndNodeCaches();
    await Provider.fetchChainAndNodeInfo();
  }
}
```

arboleya: Btw, this seems more like a feature/enhancement than a bug/fix.

maschad: Good idea @arboleya c3eadcea4

packages/account/src/providers/provider.ts
maschad: c3eadcea4

packages/account/src/providers/provider.ts
```diff
@@ -485,6 +485,13 @@ export default class Provider {
     } else {
       this.cache = new ResourceCache(DEFAULT_RESOURCE_CACHE_TTL);
     }
+
+    // Call auto refresh cache on 1 minute intervals continuously
+    setInterval(() => {
+      this.autoRefreshCache().catch((error) => {
+        console.error('Error during auto refresh cache:', error);
+      });
+    }, 60000); // 60000 milliseconds = 1 minute
```
arboleya, line 494: You must clear the previous interval before setting another one, otherwise it'll pile up.

maschad: That's done in the [fetchChainAndNodeInfo](https://github.com/FuelLabs/fuels-ts/blob/mc/feat/fetch-updated-consensus-params/packages/account/src/providers/provider.ts#L631-L634) method

arboleya: I'm referring to the `setInterval` — if you create 10x `Provider` instances, you'll have 10x `setInterval` in parallel doing the same thing, which will pile up, bloat network requests, eat memory, etc. The easiest way to prevent this is to ensure you call `setInterval` only once for the lifetime of the `Provider` class.

luizstacio: Also, for the FE application, it should have a way to specify if we want to have the internal, as it may be a short-lived instance of the Provider.

Also ensure we add a destroy method that allows to destroy the provider, that in this case would be mainly clean timeout.

maschad: My apologies I misread @arboleya adjusted in d78d37bac

Good suggestion @luizstacio added in d78d37bac

packages/account/src/providers/provider.ts
```diff
@@ -746,6 +758,13 @@ Supported fuel-core version: ${supportedVersion}.`
     return bn(height);
   }
 
+  /**
+   * Destroys the provider instance.
+   */
+  destroy() {
+    clearInterval(this.autoRefreshIntervalId);
+  }
```
Torres-ssf, line 766: When destroying should we also unset the `autoRefreshIntervalId`?

```ts
this.autoRefreshIntervalId = undefined
```

maschad: 6f1bf7e69d05a48e34715551a22ed9ed58b503ce

arboleya (dismissed):

Torres-ssf (approved):

arboleya (approved):

Issue #3392, opened 2024-11-13
Enable creating a new predicate from an existing one


nedsalk:
We should perhaps have a way of creating a new predicate from an existing one:
```ts
const loaderPredicate = await waitForDeploy();

const newLoaderPredicate = loaderPredicate.toNewInstance({ configurableConstants, data });
```
Because we currently don't have a nice way of duplicating a predicate but with new configurable constants and data.

_Originally posted by @nedsalk in https://github.com/FuelLabs/fuels-ts/pull/3389#discussion_r1838694868_

YaTut1901:
@petertonysmith94 may I take this one? I am interested in smth more domain-related

petertonysmith94:
@YaTut1901 yes you may take this one. It'll be a good one to get a better understanding of the domain, pop a message if you need anything :)

PR #3389, opened 2024-11-12
chore!: refactor predicate and script deployment

nedsalk:
- Closes #3361 

# Summary

Moves predicate and script deployment from `ContractFactory` to their respective classes as a `deploy` instance method.

# Breaking Changes

`ContractFactory.deployAsBlobTxForScript` has been removed in favor of `Predicate.deploy` and `Script.deploy`:

```ts
// before
const factory = new ContractFactory(scriptBytecode, scriptAbi, wallet);
const { waitForResult } = await factory.deployAsBlobTxForScript();
const { loaderBytecode, configurableOffsetDiff } = await waitForResult();

// after
const script = new Script(scriptBytecode, scriptAbi, wallet);
const { blobId, waitForResult } = await script.deploy(deployerWallet);
const loaderScript = await waitForResult();

const predicate = new Predicate({ bytecode, abi, provider });
const { blobId, waitForResult } = await predicate.deploy(deployerWallet);
const loaderPredicate = await waitForResult();
```

# Checklist

- [x] All **changes** are **covered** by **tests** (or not applicable)
- [x] All **changes** are **documented** (or not applicable)
- [x] I **reviewed** the **entire PR** myself (preferably, on GH UI)
- [x] I **described** all **Breaking Changes** (or there's none)

PR #3395, opened 2024-11-14
chore(): added method to duplicate predicate

YaTut1901:
Closes https://github.com/FuelLabs/fuels-ts/issues/3392

# Summary
Added method `toNewInstance` which returns new instance of predicate with same abi, loaderBytecode, provider and different configurableConstants and data to be passed:
```typescript
toNewInstance(params: {
  configurableConstants?: TConfigurables;
  data?: TData;
}): Predicate<TData, TConfigurables> { }
```

# Checklist

### DIFF
--- a/packages/account/src/predicate/predicate.ts
+++ b/packages/account/src/predicate/predicate.ts
@@ -96,6 +96,26 @@ export class Predicate<
   }
 
   /**
+   * Creates a new Predicate instance with updated parameters while maintaining the original bytecode and ABI
+   *
+   * @param params - Object containing optional new configurableConstants and data
+   * @returns A new Predicate instance with the updated parameters
+   */
+  toNewInstance(params: {
+    configurableConstants?: TConfigurables;
+    data?: TData;
+  }): Predicate<TData, TConfigurables> {
+    return new Predicate({
+      bytecode: this.bytes,
+      abi: this.interface?.jsonAbi,
+      provider: this.provider,
+      data: params.data ?? this.predicateData,
+      configurableConstants: params.configurableConstants,
+      loaderBytecode: this.loaderBytecode,
+    });
+  }
+
+  /**
    * Populates the transaction data with predicate data.
    *
    * @param transactionRequestLike - The transaction request-like object.
--- a/packages/account/test/predicate-functions.test.ts
+++ b/packages/account/test/predicate-functions.test.ts
@@ -64,5 +64,28 @@ describe('Predicate', () => {
         });
       }).toThrow('Cannot use ABI without "main" function');
     });
+
+    it('creates a new instance with updated parameters', async () => {
+      using launched = await setupTestProviderAndWallets();
+      const { provider } = launched;
+
+      const predicate = new Predicate({
+        bytecode: predicateBytecode,
+        abi: predicateAbi,
+        provider,
+        configurableConstants: { value: false },
+      });
+
+      const newPredicate = predicate.toNewInstance({
+        configurableConstants: { value: true },
+        data: ['NADA'],
+      });
+
+      expect(newPredicate.predicateData).toEqual(['NADA']);
+      expect(newPredicate.bytes.slice(1)).toEqual(predicate.bytes.slice(1));
+      expect(newPredicate.bytes[0]).not.toEqual(predicate.bytes[0]);
+      expect(newPredicate.interface?.jsonAbi).toEqual(predicate.interface?.jsonAbi);
+      expect(newPredicate.provider).toEqual(predicate.provider);
+    });
   });
 });
--- a/packages/account/test/fixtures/predicate-abi.ts
+++ b/packages/account/test/fixtures/predicate-abi.ts
@@ -30,5 +30,11 @@ export const predicateAbi: JsonAbi = {
   ],
   loggedTypes: [],
   messagesTypes: [],
-  configurables: [],
+  configurables: [
+    {
+      name: 'value',
+      concreteTypeId: 'b760f44fa5965c2474a3b471467a22c43185152129295af588b022ae50b50903',
+      offset: 0,
+    },
+  ],
 };


Torres-ssf:
Hi @YaTut1901,

Great work on this! We really appreciate your help.

Could you edit the PR description, specifically this part:

```
Closes https://github.com/FuelLabs/fuels-ts/issues/3392
```

and change it to:

```
- Closes https://github.com/FuelLabs/fuels-ts/issues/3392
```

Thank you!

packages/account/src/predicate/predicate.ts
```diff
@@ -80,6 +80,26 @@ export class Predicate<
...
+    configurableConstants?: TConfigurables;
+    data?: TData;
+  }): Predicate<TData, TConfigurables> {
+    return new Predicate<TData, TConfigurables>({
+      bytecode: this.bytes,
+      abi: this.interface?.jsonAbi,
+      provider: this.provider,
+      data: params.data ?? this.predicateData,
+      configurableConstants: params.configurableConstants,
+      loaderBytecode: this.loaderBytecode,
```
petertonysmith94, line 99:
```suggestion
```

The `loaderBytecode` is no longer a field on the `Predicate` class (after https://github.com/FuelLabs/fuels-ts/pull/3389).

packages/account/test/predicate-functions.test.ts
petertonysmith94, line 1: Could we check the `data` + `configurableConstants` have been set initially (on the first `Predicate` instantiation) and then set again after using `toNewInstance`?

packages/account/test/predicate-functions.test.ts
```diff
@@ -64,5 +64,32 @@ describe('Predicate', () => {
...
+      const predicate = new Predicate({
+        bytecode: predicateBytecode,
+        abi: predicateAbi,
+        provider,
+        configurableConstants: { value: false },
+        data: ['DADA'],
+      });
+
+      expect(predicate.predicateData).toEqual(['DADA']);
+      expect(predicate.bytes[0]).toEqual(0);
```
danielbate, line 81: This check seems redundant, can we check the entire bytecode?

danielbate (changes requested):
Would be good to have an e2e test inside `fuel-gauge` that uses a predicate via `toNewInstance`

Torres-ssf (changes requested):
@YaTut1901 Thanks for your help 🙏 

However, I am skeptical that the current solution will fully resolve the issue.

When a new predicate is instantiated, its bytes property does not always reflect the original raw predicate bytes with default configurable values.

Configurable values are embedded directly into the predicate bytes during processing. Consequently, the stored bytes reflect the instance's configurable values rather than the pristine, default state. This can be observed on [processPredicateData](https://github.com/FuelLabs/fuels-ts/blob/537fdc287d53aac6b8bf29d665b3929e324e8814/packages/account/src/predicate/predicate.ts#L154) and [setConfigurableConstants](https://github.com/FuelLabs/fuels-ts/blob/537fdc287d53aac6b8bf29d665b3929e324e8814/packages/account/src/predicate/predicate.ts#L228).

When using `toNewInstance` to create a new predicate, the bytes property inherits the values from the previous instance, which might include non-default configurable values. This makes it impossible to rely on `toNewInstance` to produce a new instance with the pristine, default configurable values.

It’s possible that someone might want to use `toNewInstance` without setting new configurable values, relying instead on the defaults.

I believe we need to store the predicate's pristine bytes separately, those that remain untouched by configurable values.

YaTut1901:
@Torres-ssf I pushed a commit where refactored predicate class to keep configurable constants in separate field and set them to bytecode at the moment of operation execution. I see this as the option to keep initial bytecode untouched. And in this case there is no need to have a method to create new predicate object - existing one can be modified. What do you think about this approach?

YaTut1901:
Hey, @Torres-ssf @danielbate @petertonysmith94 @Dhaiwat10, could anybody comment on my refactoring?

Torres-ssf:
> Hey, @Torres-ssf @danielbate @petertonysmith94 @Dhaiwat10, could anybody comment on my refactoring?

@YaTut1901 We’ll need to add some e2e tests to validate the changes introduced here.

These tests should be added to `packages/fuel-gauge/src/predicate`. A new test suite can be created for this purpose.

The test suite should ensure the following:

1 - **Configurable Values**: When creating a new Predicate instance from an existing one, ensure that any newly provided configurable values are correctly applied to the new instance.

2 - **Default Values**: If no configurable values are provided, confirm that the new instance uses the hardcoded default values from the Predicate code, not those from the previous instance.

Additionally, remember that configurable values are embedded within the predicate bytes. Any changes to these values change the predicate address, resulting in a new instance with a different address. Consequently, the latest instance will be a new account, without any of the previous funds added.

We also need to add some documentation for this. We need to tell our users about this new feature and how they can use it.

Torres-ssf:
> I can do this, but overall does this change worth it? Or there can be other easier way?

@YaTut1901 Implementing a method to re-instantiate a predicate from an already instantiated one?

Yes, I believe this change is worthwhile as it enhances the flexibility and clarity of how Predicate instances handle configurable values. That said, I understand this implementation, along with the required tests and documentation, is not trivial and requires a deeper understanding of the stack.

If you feel this might be too much to take on right now, that's completely fine. You're more than welcome to step away from it, and we can reassign or revisit it later. Just let us know how you'd like to proceed.

YaTut1901:
> > I can do this, but overall does this change worth it? Or there can be other easier way?
> 
> @YaTut1901 Implementing a method to re-instantiate a predicate from an already instantiated one?
> 
> Yes, I believe this change is worthwhile as it enhances the flexibility and clarity of how Predicate instances handle configurable values. That said, I understand this implementation, along with the required tests and documentation, is not trivial and requires a deeper understanding of the stack.
> 
> If you feel this might be too much to take on right now, that's completely fine. You're more than welcome to step away from it, and we can reassign or revisit it later. Just let us know how you'd like to proceed.

I meant the way I proposed, like not having method to re-instantiate predicate, but have an option to update values on existing one. It is possible in case when all changes to initial bytecode are applied only when transaction is sent

danielbate:
> I meant the way I proposed, like not having method to re-instantiate predicate, but have an option to update values on existing one. It is possible in case when all changes to initial bytecode are applied only when transaction is sent

The reason though is that if the values are changed on a predicate and the underlying bytecode has changed, then you have a new predicate address which inherently is a new predicate instance. 

Therefore it makes sense for us to replicate this behaviour in the SDK so it is clear to the user that when these values change, they are dealing with a new predicate instance. Hence it is important to document and test this behaviour due to the risk of loss of funds, as we are dealing with a new account. With that in mind, we can not do it at TX submission.

YaTut1901:
> > I meant the way I proposed, like not having method to re-instantiate predicate, but have an option to update values on existing one. It is possible in case when all changes to initial bytecode are applied only when transaction is sent
> 
> The reason though is that if the values are changed on a predicate and the underlying bytecode has changed, then you have a new predicate address which inherently is a new predicate instance, as @Torres-ssf has mentioned.
> 
> Therefore it makes sense for us to replicate this behaviour in the SDK so it is clear to the user that when these values change, they are dealing with a new predicate instance. Hence it is important to document and test this behaviour due to the risk of loss of funds, as we are dealing with a new account. With that in mind, we can not do it at TX submission.

Oh, okay, got the point. So I will take it

YaTut1901:
created new PR with implementation from scratch: https://github.com/FuelLabs/fuels-ts/pull/3432

PR #3396, opened 2024-11-14
fix: typegen and storage slots integration

### LINKED ISSUE #3355: Typegen'd contract factories don't properly integrate storage slots for deployment methods
Given a typegen'd contract factory `MyContractFactory` that has some storage slots, the following works and includes the storage slots:
```ts
const factory = new MyContractFactory(wallet);
await factory.deploy();
```

However, the two examples below don't integrate the storage slots:
```ts
await factory.deployAsCreateTx();
await factory.deployAsBlobTx();
```

This is because the storage slots in the typegen outputs aren't properly integrated for these methods.

There are two possible solutions:
1. Override the two methods in the typegen'd contract factory the same way we do it with the `deploy` method,
2. Add a property on `ContractFactory` that can be overriden in the typegen'd factory and which would be read in all the deployment methods and merged with the `deployOptions` passed via arguments.

The second option looks better because it encapsulates the logic in `ContractFactory`, thereby having the typegen'd factory do as little runtime logic as possible. A possible implementation showcasing the core components is shown below.

```ts
// contract-factory.ts
export default class ContractFactory {
  // should it be protected or public?
  protected _deployOptions: DeployContractOptions = { };
  private getDeployOptions(deployOptions) { return mergeDeepRight(this._deployOptions, deployOptions); }

  // same pattern for all other public methods that have deployOptions as a parameter
  async deploy(deployOptions) {
    const options = this.getDeployOptions(deployOptions);
  }
}

// typegen/contracts/MyContractFactory.ts
export class MyContractFactory extends ContractFactory {
  protected _deployOptions: DeployContractOptions = {
    storageSlots: MyContract.storageSlots
  }
}
```

nedsalk:
- Closes #3355 

# Summary

This PR fixes integration of typegen'd storage slots with the `ContractFactory.deployAsCreateTx` and `ContractFactory.deployAsBlobTx` methods by making it possible to specify `deployOptions` in the `ContractFactory` constructor, passing the storage slots via the constructor in typegen'd factories, and using them in those methods.

This approach also enables us to remove the override of the `deploy` method in typegen.

# Checklist


packages/fuel-gauge/src/storage-test-contract.test.ts
petertonysmith94, line 1: Is it worthwhile testing the merge functionality, I assume the `mergeRight` should enable individual storage slots to be overridden?

packages/abi-typegen/src/templates/contract/factory.hbs
```diff
@@ -7,27 +7,23 @@ import { {{capitalizedName}} } from "./{{capitalizedName}}";
...
-      ...deployOptions,
-    });
-  }
-
-  static async deploy (
+  static deploy (
     wallet: Account,
     options: DeployContractOptions = {}
-  ): Promise<DeployContractResult<{{capitalizedName}}>> {
+  ): ReturnType<{{capitalizedName}}Factory['deploy']> {
```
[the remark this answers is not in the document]
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/3396/commits/73504472403534b62dc43f18cf26c4c9da1c9d0a

packages/fuel-gauge/src/storage-test-contract.test.ts
nedsalk: Great question, it led me to finding a bug. https://github.com/FuelLabs/fuels-ts/pull/3396/commits/7cb0db61cfafd5a07e546295923f7661e8ea0537

petertonysmith94: Nice, should there be a test for this in the commit?

nedsalk: Ugh, I did some commit reverting to tidy it up and didn't do it properly. The linked commit above now contains the tests.

Torres-ssf (commented):
@nedsalk Maybe I am missing something, but not why simplifying override the two methods in the typegen'd contract factory the same way we do it with the deploy method?

nedsalk:
@Torres-ssf I prefer this approach over the overrides because it just passes the storage slots data via the constructor and doesn't add any logic to typegen outputs; the logic lives in the base `ContractFactory`. If/when we decide for typegen to also generate `.js` outputs, the generated factories for that mode would also just pass it via the constructor. By using overrides, we'd have to override all public methods in every typegen output mode, which I feel is a bigger maintenance burden.

packages/contract/src/contract-factory.ts
```diff
@@ -52,11 +53,12 @@ export type DeployContractResult<TContract extends Contract = Contract> = {
 /**
  * `ContractFactory` provides utilities for deploying and configuring contracts.
  */
-export default class ContractFactory {
+export default class ContractFactory<TContract extends Contract = Contract> {
   bytecode: BytesLike;
   interface: Interface;
   provider!: Provider | null;
   account!: Account | null;
+  deployOptions: DeployContractOptions;
```
Torres-ssf, line 61: @nedsalk Adding the whole `deployOptions` to the constructor seems to be adding a lot of complexity later on. 

The `mergeDeployOptions` implementation appears to be a workaround for this complexity, as it consolidates the options into a unified structure for later use.

Why not simply add the `storageSlots` to the constructor instead of the whole deploy options?

```ts
export default class ContractFactory {
  bytecode: BytesLike;
  interface: Interface;
  provider!: Provider | null;
  account!: Account | null;
  storageSlots!: StorageSlot[];

  constructor(
    bytecode: BytesLike,
    abi: JsonAbi | Interface,
    accountOrProvider: Account | Provider | null = null,
    storageSlots: StorageSlot[] = []
  ) {
    // Force the bytecode to be a byte array
    this.bytecode = arrayify(bytecode);
    this.storageSlots = storageSlots;

    if (abi instanceof Interface) {
      this.interface = abi;
    } else {
      this.interface = new Interface(abi);
    }
```

The `createTransactionRequest` method can then use the `storageSlots` directly from the instance or override them using the `deployOptions`:


```ts
createTransactionRequest(deployOptions?: DeployContractOptions & { bytecode?: BytesLike }) {
    const storageSlots = (deployOptions?.storageSlots || this.storageSlots)
      .map(({ key, value }) => ({
        key: hexlifyWithPrefix(key),
        value: hexlifyWithPrefix(value),
      }))
      .sort(({ key: keyA }, { key: keyB }) => keyA.localeCompare(keyB));

    const options = {
      salt: randomBytes(32),
      ...deployOptions,
      storageSlots,
    };
```

Since both `deployAsCreateTx` and `deployAsBlobTx` invoke `createTransactionRequest`, this change ensures a consistent and simplified flow, right?

When generating types, `storageSlots` can simply be passed to the `ContractFactory` constructor.

Am I missing something?

nedsalk: > The `mergeDeployOptions` implementation appears to be a workaround for this complexity, as it consolidates the options into a unified structure for later use.

The `mergeDeployOptions` also integrates new storage slots with the original ones such that you can do partial overriding of storage slots, which I verified [here](https://github.com/FuelLabs/fuels-ts/pull/3396/files#diff-b763698f1c0a38daedd295807af1d6e2737d99e78838dab76ac9dca56bd6fef4R129). The linked test would fail on current `master`.

> Why not simply add the `storageSlots` to the constructor instead of the whole deploy options?
The createTransactionRequest method can then use the storageSlots directly from the instance or override them using the deployOptions:
Since both deployAsCreateTx and deployAsBlobTx invoke createTransactionRequest, this change ensures a consistent and simplified flow, right?
When generating types, storageSlots can simply be passed to the ContractFactory constructor.

I like this and will go with it. It doesn't seem to me that we'll need anything else from `DeployContractOptions` in a similar manner which would warrant us changing the constructor further down the line, and the solution you proposed is simpler. I'll have to keep the storage slot merging logic I mentioned above.

Torres-ssf (approved):

maschad (dismissed):
Great work on this @nedsalk  🚀

.changeset/cyan-panthers-carry.md
```diff
@@ -0,0 +1,6 @@
+---
+"@fuel-ts/abi-typegen": patch
+"@fuel-ts/contract": patch
```
petertonysmith94, line 3:
```suggestion
"@fuel-ts/contract": patch
"fuels": patch
```

petertonysmith94 (changes requested):
I meant to request changes with the [following comment](https://github.com/FuelLabs/fuels-ts/pull/3396/files/1fcb22ae121ccfe43d8b51c2af4346457704e331#r1851646460).

packages/contract/src/contract-factory.ts
```diff
@@ -192,7 +198,7 @@ export default class ContractFactory {
    * @param deployOptions - Options for deploying the contract.
    * @returns A promise that resolves to the deployed contract instance.
    */
-  async deploy<TContract extends Contract = Contract>(
+  async deploy(
```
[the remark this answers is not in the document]
nedsalk: Great catch! https://github.com/FuelLabs/fuels-ts/pull/3396/commits/6836f2b8c2a42632d83b382e7f43ed2b86ca534f

petertonysmith94 (approved):

maschad (approved):

PR #3400, opened 2024-11-15
chore!: limit TX pagination number for `getTransactionsSummaries`

Torres-ssf:
# Release notes

In this release, we:

- Limit pagination number for `getTransactionsSummaries` to `60`

# Breaking Changes

The pagination number for `getTransactionsSummaries` is limited to `60` now

```ts
// before
const { transactions } = await getTransactionsSummaries({
  provider,
  filters: {
    owner: account.address.toB256(),
    first: 200,
  },
});
```

```ts
// after
const { transactions } = await getTransactionsSummaries({
  provider,
  filters: {
    owner: account.address.toB256(),
    first: 60, // Limit is 60 now. A higher value will result in an error
  },
});
```

# Checklist


packages/account/src/providers/transaction-summary/get-transaction-summary.ts
```diff
@@ -142,7 +142,17 @@ export async function getTransactionsSummaries(
...
+  const { owner, ...inputArgs } = filters;
+
+  const validPaginationParams = provider.validatePaginationArgs({
+    inputArgs,
+    paginationLimit: TRANSACTIONS_PAGE_SIZE_LIMIT,
+  });
+
+  const { transactionsByOwner } = await provider.operations.getTransactionsByOwner({
+    owner,
+    ...validPaginationParams,
```
danielbate, line 154:
```suggestion
    ...validPaginationParams,
    owner,
```

Just to make sure it's not overwritten

packages/account/src/providers/provider.ts
```diff
@@ -2106,7 +2106,7 @@ Supported fuel-core version: ${supportedVersion}.`
   /**
    * @hidden
    */
-  private validatePaginationArgs(params: {
+  validatePaginationArgs(params: {
```
nedsalk, line 2109: You can make this static, but it's okay like this as well.

nedsalk (approved):

packages/account/src/providers/provider.ts
Torres-ssf: I plan to externalize this as a helper. I did not do that right because it involved moving some tests around. Do you think it is better to make it static instead?

danielbate (dismissed):

packages/account/src/providers/provider.ts
nedsalk: A helper function is even better. The problem is that removing this method later down the line would constitute a breaking change now that you made it public. I think you can move it to a helper function right away. You can write the tests for it later if you want. The most important thing is that you added the test for `getTransactionsSummaries` in this PR.

nedsalk (approved):

arboleya (approved):

packages/account/src/providers/provider.ts
Torres-ssf: @nedsalk https://github.com/FuelLabs/fuels-ts/pull/3400/commits/0ad5fd65f0d1e9627db648160505f01d1d09dec7

nedsalk (approved):

arboleya (approved):

Issue #3203, opened 2024-09-24
Deprecate `bech32` address format


arboleya:
We should first deprecate this (non-breaking change), and then move on to:
 - https://github.com/FuelLabs/fuels-ts/issues/2348

arboleya:
@maschad Because we implemented the checksum utilities around addresses during the launch, I believe this issue is unblocked, and we can move on with the deprecation.

PR #3204, opened 2024-09-24
chore: deprecate bech32 addresses

maschad:
> Closes TS-660

- Closes https://github.com/FuelLabs/fuels-ts/issues/3203

# Release notes

In this release, we:

- Deprecated the `bech32Address` type in favour of hexadecimal address format going forward.

# Summary

This PR uses adds the `deprecated` JS-doc tag as well as makes note in the relevant documentation of the deprecation notice.

# Checklist

- [x] All **changes** are **covered** by **tests** (or not applicable)
- [x] All **changes** are **documented** (or not applicable)
- [x] I **reviewed** the **entire PR** myself (preferably, on GH UI)
- [x] I **described** all **Breaking Changes** (or there's none)

PR #3411, opened 2024-11-20
chore: deprecate bech32 addresses

maschad:
- Closes https://github.com/FuelLabs/fuels-ts/issues/3203

# Release notes

In this release, we:

- Deprecated the `bech32Address` type in favour of hexadecimal address format going forward.

# Summary

This PR uses adds the `deprecated` JS-doc tag as well as makes note in the relevant documentation of the deprecation notice.

# Checklist


packages/address/src/address.ts
petertonysmith94, line 1: May be worthwhile adding a deprecation to the constructor suggesting to use `Address.fromB256`?

maschad: We had [deliberated on this](https://github.com/FuelLabs/fuels-ts/pull/3204#discussion_r1774958640) on the previous PR in the discussions, my understanding this is the final resolution. I am open to either to be fair.

packages/address/src/address.ts
```diff
@@ -137,8 +136,9 @@ export default class Address extends AbstractAddress {
...
 
   /**
-   * returns the address `checksum` as a string
-   *
+   * Returns the value of the `bech32Address` property
+   * @deprecated
+   * Type `Bech32Address` is now deprecated, as is this method. Use `B256` addresses instead. ([help](https://docs.fuel.network/docs/specs/abi/argument-encoding/#b256))
    * @returns The value of `bech32Address` property
    */
   override valueOf(): string {
```
Torres-ssf, line 144: I don't think we will deprecate `valueOf`.

Torres-ssf (changes requested):

packages/address/src/address.ts
maschad: Why would one use the `valueOf` method over the `toHexString` method for B256 addresses?

danielbate: Once we deprecate the `bech32Address` property won't `toHexString` also be redundant?

Torres-ssf: @maschad I’m not sure why this was implemented initially, as we already had `toString` with the same functionality.

However, since this is no longer related to `bech32`, I believe it’s outside the scope of this PR and doesn’t need to be addressed here.

packages/address/src/address.ts
petertonysmith94: IMO offering users an option to migrate early before deprecation would be beneficial. I believe the following would be a non-breaking change.

```ts
/**
 * @param address - A Bech32 address
 *
 * @deprecated
 * We will be deprecating the `Bech32Address` type. Use a `B256` address instead.
 */
constructor(address: B256Address | Bech32Address) {
  super();

  if (isBech32(address)) {
    this.bech32Address = normalizeBech32(address as Bech32Address);
  } else if (isB256(address)) {
    this.bech32Address = toBech32(address);
  } else {
    throw new FuelError(
      FuelError.CODES.INVALID_BECH32_ADDRESS,
      `Invalid Bech32 Address: ${address}.`
    );
  }
}
```

.changeset/late-pugs-carry.md
```diff
@@ -0,0 +1,7 @@
+---
+"@fuel-ts/interfaces": patch
+"@fuel-ts/address": patch
+"@fuel-ts/errors": patch
+---
+
+chore: deprecate bech32 addresses
```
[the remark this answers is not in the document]
maschad: Added

packages/address/src/address.ts
maschad: 7b97f008ccf9132f3cfaec6c4bd63fda45981f7b

maschad:
> Should this PR introduce a new `b256Address` property on the `Address` class that we can then phase over to?

Agreed see https://github.com/FuelLabs/fuels-ts/commit/7b97f008ccf9132f3cfaec6c4bd63fda45981f7b @danielbate

danielbate (approved):

packages/address/src/address.ts
petertonysmith94: Looks like the tests are failing with this change. It is probably worthwhile adding tests for the new constructor via `B256` as well.

packages/address/src/address.ts
```diff
@@ -30,17 +30,24 @@ import {
...
   // #region address-2
+  /**
+   * @deprecated
+   * Type `Bech32Address` is now deprecated, as is this property. Use `B256` addresses instead. ([help](https://docs.fuel.network/docs/specs/abi/argument-encoding/#b256))
+   */
   readonly bech32Address: Bech32Address;
   // #endregion address-2
 
   /**
    * @param address - A Bech32 address
```
nedsalk, line 41:
```suggestion
   * @param address - A B256 address
```

packages/address/src/address.ts
```diff
@@ -137,8 +139,7 @@ export default class Address extends AbstractAddress {
   }
 
   /**
-   * returns the address `checksum` as a string
-   *
+   * Returns the value of the `bech32Address` property
```
nedsalk, line 142: `this.toChecksum()` doesn't return a `bech32Address`. Was this change a mistake?

Torres-ssf: @nedsalk Good catch.

I think this change came by accident

maschad: Technically it says that it returns the value of the property, not that it is a `bech32address`  , but I've updated the description to be more explicit 0568c2955

packages/address/src/address.ts
maschad: This constructor now accepts both types, so I've updated the language as such 0568c2955

petertonysmith94 (approved):

Torres-ssf (approved):

Issue #3429, opened 2024-11-28
Add docs around UTXO dismantling/consolidation


Torres-ssf:
Typical use cases are:
 1. Split one UTXO into several
 1. Consolidating several UTXOs into one

Examples:
- 1x UTXO with a value of `1000` could be split into 10x UTXOs, each valued at `100`
- 100x UTXOs with a value of `10` could be consolidated into 1x UTXO valued at `1000`

Notes:
- Fee must be subtracted before the split/consolidation
- Chain constraints limit the maximum number of outputs allowed per transaction

maschad:
@Torres-ssf @arboleya perhaps I may be misinterpreting but after reading through the [combining UTXO docs](https://docs.fuel.network/docs/fuels-ts/cookbook/combining-utxos/) I think the only documentation that is missing is regarding the `maxOutputs` which are determined by the chain config. I've added that to the docs in [this PR](https://github.com/FuelLabs/fuels-ts/pull/3435)

PR #3435, opened 2024-11-30
docs: add max outputs documentation to combining UTXO docs

maschad:
- Closes https://github.com/FuelLabs/fuels-ts/issues/3429

# Summary

Merely added a section about `maxOutputs` to the [Combining UTXOs docs](https://docs.fuel.network/docs/fuels-ts/cookbook/combining-utxos/)

# Checklist

### DIFF
--- a/.changeset/rotten-chefs-shake.md
+++ b/.changeset/rotten-chefs-shake.md
@@ -1,0 +1,4 @@
+---
+---
+
+docs: add max outputs documentation to combining UTXO docs
--- a/apps/docs/src/guide/cookbook/combining-utxos.md
+++ b/apps/docs/src/guide/cookbook/combining-utxos.md
@@ -9,3 +9,7 @@ One way to avoid these errors is to combine your UTXOs. This can be done by perf
 > **Note:** You will not be able to have a single UTXO for the base asset after combining, as one output will be for the transfer, and you will have another for the fees.
 
 <<< @./snippets/combining-utxos.ts#combining-utxos{ts:line-numbers}
+
+## Max Outputs
+
+It's also important to note that depending on the chain configuration, you may be limited on the number of outputs you can have in a transaction. This amount can be queried via the [TxParameters](https://docs.fuel.network/docs/graphql/reference/objects/#txparameters) GQL query.


.changeset/rotten-chefs-shake.md
```diff
@@ -0,0 +1,4 @@
+---
+---
+
+docs: add max outputs documentation to combining UTXO docs
```
danielbate, line 4: Should this be closing #3429, looks like it should just be related to?

apps/docs/src/guide/cookbook/combining-utxos.md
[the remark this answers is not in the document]
maschad: Personally I don't see the need for that, open to hearing others thoughts

.changeset/rotten-chefs-shake.md
maschad: I put it as closing given we already have docs around combining UTXOs and this just seeks to add the last part of the issue, I clarified [here](https://github.com/FuelLabs/fuels-ts/issues/3429#issuecomment-2508797600)

apps/docs/src/guide/cookbook/combining-utxos.md
```diff
@@ -9,3 +9,7 @@ One way to avoid these errors is to combine your UTXOs. This can be done by perf
 > **Note:** You will not be able to have a single UTXO for the base asset after combining, as one output will be for the transfer, and you will have another for the fees.
 
 <<< @./snippets/combining-utxos.ts#combining-utxos{ts:line-numbers}
+
+## Max Outputs
+
+It's also important to note that depending on the chain configuration, you may be limited on the number of outputs you can have in a transaction. This amount can be queried via the [TxParameters](https://docs.fuel.network/docs/graphql/reference/objects/#txparameters) GraphQL query
```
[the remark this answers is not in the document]
maschad: bf6a2daa4

.changeset/rotten-chefs-shake.md
Torres-ssf: @maschad Since we already have docs related to combining UTXOs, perhaps we can now add docs related to dismantling UTXOs, which is splitting one into several.

danielbate: Yeah let's be as explicit as possible given this came up very recently.

apps/docs/src/guide/cookbook/combining-utxos.md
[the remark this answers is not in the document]
petertonysmith94: Related to https://github.com/FuelLabs/fuels-ts/pull/3435#discussion_r1866828355 so resolving.

.changeset/rotten-chefs-shake.md
maschad: I've decided to add a function inspired by @luizstacio 's that splits the UTXOs for the consumers and documented it [here](https://github.com/FuelLabs/fuels-ts/pull/3435/files#diff-27051d152c0be799da72bb993355ee169e2b7a9457d003fdba0c1573e16f9951)

apps/docs/src/guide/cookbook/combining-utxos.md
```diff
@@ -9,3 +9,9 @@ One way to avoid these errors is to combine your UTXOs. This can be done by perf
 > **Note:** You will not be able to have a single UTXO for the base asset after combining, as one output will be for the transfer, and you will have another for the fees.
 
 <<< @./snippets/combining-utxos.ts#combining-utxos{ts:line-numbers}
+
+## Max Outputs
```
danielbate, line 13:
```suggestion
## Max Inputs and Outputs
```

apps/docs/src/guide/cookbook/snippets/splitting-utxos.ts
```diff
@@ -0,0 +1,55 @@
...
+
+// let's fund the wallet with 1000 of the base asset
+const fundingTx = await fundingWallet.transfer(
+  wallet.address,
+  1000,
+  provider.getBaseAssetId()
+);
+await fundingTx.waitForResult();
+
+// We can fetch the coins to see how many UTXOs we have, and confirm it is 1.
```
danielbate, line 19:
```suggestion
// We can fetch the coins to see how many UTXOs we have and confirm it is 1
```

apps/docs/src/guide/cookbook/snippets/splitting-utxos.ts
```diff
@@ -0,0 +1,55 @@
...
+import { BN, Provider, Wallet, splitUTXOs } from 'fuels';
+
+import { LOCAL_NETWORK_URL, WALLET_PVT_KEY } from '../../../env';
+
+const provider = await Provider.create(LOCAL_NETWORK_URL);
+const fundingWallet = Wallet.fromPrivateKey(WALLET_PVT_KEY, provider);
+
+const wallet = Wallet.generate({ provider });
+
+// let's fund the wallet with 1000 of the base asset
```
danielbate, line 11:
```suggestion
// Let's fund the wallet with 1000 of the base asset
```

apps/docs/src/guide/cookbook/snippets/splitting-utxos.ts
```diff
@@ -0,0 +1,55 @@
...
+await fundingTx.waitForResult();
+
+// We can fetch the coins to see how many UTXOs we have, and confirm it is 1.
+const { coins: initialCoins } = await wallet.getCoins(
+  provider.getBaseAssetId()
+);
+console.log('Initial Coins Length', initialCoins.length);
+// 1
+
+// Now we can split the UTXO into 5 UTXOs of 200 each.
```
danielbate, line 26:
```suggestion
// Now we can split the UTXO into 5 UTXOs of 200 each
```

packages/utils/src/utils/split-utxos.ts
```diff
@@ -0,0 +1,23 @@
+import { ErrorCode, FuelError } from '@fuel-ts/errors';
+import type { AbstractAddress } from '@fuel-ts/interfaces';
+import type { BN } from '@fuel-ts/math';
+
+export function splitUTXOs(
+  balance: BN,
+  amount: BN,
+  assetId: string,
+  destination: AbstractAddress,
```
danielbate, line 9: Maybe `destination` could just take a string as it is returning a string? And then we don't need the unknown cast in the above test?

packages/utils/src/utils/split-utxos.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    const destination = '0x...' as unknown as AbstractAddress;
+
+    expect(splitUTXOs(new BN(100), new BN(10), '0x0', destination, 3)).toEqual([
+      { amount: new BN(10), assetId: '0x0', destination },
+      { amount: new BN(10), assetId: '0x0', destination },
+      { amount: new BN(10), assetId: '0x0', destination },
+    ]);
+  });
+
+  it('should generate the maximum number of UTXOs possible', () => {
```
danielbate, line 42: Hmm should this throw if we can't fill the expected number of UTXOs? jw if that would be frustrating for a consumer that would unknowingly submit this tx for it to be invalid.

apps/docs/src/guide/cookbook/splitting-utxos.md
```diff
@@ -0,0 +1,7 @@
+# Splitting UTXOs
+
+There may be times when you want to split a UTXO into multiple UTXOs. This can be useful if you want to reduce the size of your transaction or you are receiving the `MAX_INPUTS_EXCEEDED` error.
```
nedsalk, line 3: > This can be useful if you want to reduce the size of your transaction

Adding more outputs to a transaction would only increase its size, not reduce it, or I'm missing something here.

> or you are receiving the `MAX_INPUTS_EXCEEDED` error

How would one receive this error when splitting UTXOs only works in the context of outputs? Also, the way it's phrased, substituting the error code with `MAX_OUTPUTS_EXCEEDED` is still confusing as you'd want to _merge_ outputs to avoid this error, not _split_ them up further.

Torres-ssf, line 3: Using multiple UTXOs instead of a single one increases the transaction size rather than reducing it. 

One advantage that we can point out is that having multiple UTXOs allows the user to submit concurrent transactions without needing to wait for each transaction to be processed sequentially.

apps/docs/src/guide/cookbook/combining-utxos.md
```diff
@@ -9,3 +9,9 @@ One way to avoid these errors is to combine your UTXOs. This can be done by perf
...
+
+It's also important to note that depending on the chain configuration, you may be limited on the number of inputs and/or outputs you can have in a transaction. These amounts can be queried via the [TxParameters](https://docs.fuel.network/docs/graphql/reference/objects/#txparameters) GraphQL query
+
+<< @./snippets/max-outputs.ts#max-outputs{ts:line-numbers}
```
Torres-ssf, line 17:
```suggestion
<<< @./snippets/max-outputs.ts#max-outputs{ts:line-numbers}
```

The missing `<` results in the snippet not rendering properly:

Screenshot 2024-12-04 at 7 57 24 AM

apps/docs/src/guide/cookbook/splitting-utxos.md
```diff
@@ -0,0 +1,7 @@
...
+
+To split a UTXO, you can use the `splitUTXOs` function. This will return an array of the number of UTXOs each with the specified amount, as long as the balance is greater than the amount.
+
+<< @./snippets/splitting-utxos.ts#splitting-utxos{ts:line-numbers}
```
Torres-ssf, line 7:
```suggestion
<<< @./snippets/splitting-utxos.ts#splitting-utxos{ts:line-numbers}
```

packages/utils/src/utils/split-utxos.test.ts
```diff
@@ -0,0 +1,52 @@
...
+import { splitUTXOs } from './split-utxos';
+
+/**
+ * @group node
+ * @group browser
+ */
+describe('splitUTXOs', () => {
+  it('should generate two UTXOs by default', () => {
+    // using unknown as AbstractAddress to avoid importing @fuel-ts/address as that's a circular dependency
+    const destination = '0x...' as unknown as AbstractAddress;
```
Torres-ssf, line 15: This indicates that we should define this helper somewhere else. We can put it within the `account` package. The result will be the same since users will import it from `fuels`.

apps/docs/src/guide/cookbook/snippets/splitting-utxos.ts
```diff
@@ -0,0 +1,55 @@
...
+  5
+);
+
+// We will also add some funds to the wallet to cover the fee
+const fundTx = await fundingWallet.transfer(
+  wallet.address,
+  500,
+  provider.getBaseAssetId()
+);
+await fundTx.waitForResult();
```
Torres-ssf, line 41: I don't understand this part of the documentation. Why is the wallet being funded a second time?

packages/utils/src/utils/split-utxos.ts
Torres-ssf, line 9: I believe this helper would benefit from some TSDoc

Torres-ssf (changes requested):

packages/utils/src/utils/split-utxos.ts
maschad: In my opinion the benefits of enforcing the address type outweigh the cast in the test. The primary goal here is to validate user input not to have "clean" tests.

packages/utils/src/utils/split-utxos.test.ts
maschad: I disagree, this is a utility specific to UTXOs and had I chosen to accept the `destination` param as a string this wouldn't be necessary. 
I could move it to the account to avoid this cast but I don't think structurally this is a function of the account.

apps/docs/src/guide/cookbook/snippets/splitting-utxos.ts
maschad: Because the wallet has a balance of 1000 comprised of 5 UTXOs each worth 200. Therefore there is an insufficient amount to cover the fee.

packages/utils/src/utils/split-utxos.ts
maschad: 0bad71912

packages/utils/src/utils/split-utxos.test.ts
maschad: Agreed 0568b1197

apps/docs/src/guide/cookbook/splitting-utxos.md
maschad: Agreed 0568b1197

maschad: Well said 0568b1197

apps/docs/src/guide/cookbook/combining-utxos.md
```diff
@@ -9,3 +9,9 @@ One way to avoid these errors is to combine your UTXOs. This can be done by perf
 > **Note:** You will not be able to have a single UTXO for the base asset after combining, as one output will be for the transfer, and you will have another for the fees.
 
 <<< @./snippets/combining-utxos.ts#combining-utxos{ts:line-numbers}
+
+## Max Inputs and Outputs
+
+It's also important to note that depending on the chain configuration, you may be limited on the number of inputs and/or outputs that you can have in a transaction. These amounts can be queried via the [TxParameters](https://docs.fuel.network/docs/graphql/reference/objects/#txparameters) GraphQL query
```
Dhaiwat10, line 15: Nit

```suggestion
It's also important to note that depending on the chain configuration, you may be limited on the number of inputs and/or outputs that you can have in a transaction. These amounts can be queried via the [TxParameters](https://docs.fuel.network/docs/graphql/reference/objects/#txparameters) GraphQL query.
```

apps/docs/src/guide/cookbook/snippets/splitting-utxos.ts
Dhaiwat10: Can we better clarify this in the comment where we fund the wallet the second time?

apps/docs/src/guide/cookbook/splitting-utxos.md
```diff
@@ -0,0 +1,7 @@
+# Splitting UTXOs
+
+There may be times when you want to split one large UTXO into multiple smaller UTXOs. This can be useful if you want to send multiple concurrent transactions without having to wait for them to be processed sequentially.
+
+To split a UTXO, you can use the `splitUTXOs` function. This will return an array of the number of UTXOs each with the specified amount, as long as the balance is greater than the amount.
```
Dhaiwat10, line 5:
```suggestion
To split a UTXO, you can use the `splitUTXOs` function. This will return an array of the number of UTXOs each with the specified amount, as long as the balance is greater than the combined amount of the split UTXOs.
```

Maybe more clear?

packages/utils/src/utils/split-utxos.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * This function allows a consumer to split a UTXO into multiple smaller UTXOs
+ * @param balance - The total balance that the user wants to split, this is equivalent of the amount you would like to send to the recipient
+ * @param amount - The amount of each UTXO
+ * @param assetId - The asset ID of the UTXOs
+ * @param destination - The destination address for the UTXOs
+ * @param number_of_coins - The number of UTXOs to split into, it defaults to 2
+ * @returns An array of the desired UTXOs with the amount, assetId, and destination address
+ */
+export function splitUTXOs(
```
nedsalk, line 14: The naming `splitUTXOs` implies that this function would be accepting actual UTXOs and creating transaction outputs that would be added to a transaction's outputs in a fashion like this:
```ts
const txRequest = new ScriptTransactionRequest();

txRequest.addCoinInput(coin1);
txRequest.addCoinInput(coin2);

const outputs = splitUTXOs({
  utxos: txRequest.getCoinInputs(),
  destination: '0x...',
  amount: bn(10),
});

txRequest.addCoinOutputs(outputs);
```

With the current implementation of this function, a more appropriate name would be `generateTransactionOutputs`. One could still use it in a similar fashion as in the code block above by summing the `amount`s of the coins, but because it's not taking in any UTXO its name is misaligned.

packages/utils/src/utils/split-utxos.ts
```diff
@@ -0,0 +1,36 @@
+import { ErrorCode, FuelError } from '@fuel-ts/errors';
+import type { AbstractAddress } from '@fuel-ts/interfaces';
+import type { BN } from '@fuel-ts/math';
+
+/**
+ * This function allows a consumer to split a UTXO into multiple smaller UTXOs
```
nedsalk, line 6: UTXOs only become such after a transaction has been processed. This utility produces transaction outputs that will become UTXOs if the user puts them in a transaction they submit.

>  consumer to split a UTXO

With regards to this, I wrote a comment [here](https://github.com/FuelLabs/fuels-ts/pull/3435/files#r1871389903).

```suggestion
 * This function allows a consumer to split a UTXO into multiple smaller transaction outputs.
```

packages/utils/src/utils/split-utxos.ts
arboleya: @Torres-ssf @maschad I thought we had agreed about this being a docs-only thing, no?

> [!NOTE]
> Special methods in SDKs are helpful, but add a lot of complexity for documentation and future proofing

The PR isn't even merged, and we're already watching these ☝️ complexities taking shape.

Torres-ssf: @maschad I completely agree with @arboleya here. Taking the approach of simply adding docs would make things much simpler.

packages/utils/src/utils/split-utxos.ts
maschad: This is no longer relevant 7165a1a85

apps/docs/src/guide/cookbook/splitting-utxos.md
maschad: No longer relevant 7165a1a85

packages/utils/src/utils/split-utxos.ts
maschad: No longer relevant 7165a1a85

packages/utils/src/utils/split-utxos.test.ts
maschad: No longer relevant 7165a1a85

Dhaiwat10 (approved):

petertonysmith94 (approved):

danielbate (approved):

Torres-ssf (approved):

PR #3438, opened 2024-12-01
fix: enable `fuel-core` node cleanup to work in Bun

### LINKED ISSUE #3359: `fuel-core` node cleanup doesn't work in bun
When using `bun v1.1.33`, trying to call `cleanup` on the return of `launchTestNode` results in the following error:
```
282 |       }
283 |       childState.isDead = true;
284 |       removeSideffects();
285 |       if (child.pid !== void 0) {
286 |         try {
287 |           process.kill(-child.pid);
                ^
RangeError: The value of "pid" is out of range. It must be a positive integer.
```
There is the issue https://github.com/oven-sh/bun/issues/8787 which documents an additional occurrence of this problem, so it might not be our fault.

Run the following to reproduce:
```bash
mkdir bun-test && cd bun-test && bun init -y && bun add fuels@0.96.1 && echo 'import { test } from "bun:test";
import { launchTestNode } from "fuels/test-utils";

test("test", async () => {
});' > failing.test.ts && bun test
```

petertonysmith94:
- Closes #3359

# Release notes

In this release, we:

- Fixed issue with `fuel-core` node cleanup operation failing in Bun.

# Summary

- Followed a similar approach to a related [workaround](https://github.com/antongolub/zx/blob/e69d477b505bb42dbeea9cf037299a8480a89800/src/core.ts#L583C1-L597C2) from the [issue](https://github.com/oven-sh/bun/issues/8787).

# Checklist


packages/account/src/test-utils/launchNode.test.ts
```diff
@@ -29,6 +29,10 @@ vi.mock('fs', async () => {
  * @group node
  */
 describe('launchNode', () => {
+  beforeEach(() => {
+    vi.spyOn(console, 'log').mockImplementation(() => {});
+  });
```
danielbate, line 34: How come your adding this in `beforeEach` if it's only needed in one test? Seems like unnecessary addition of logic for most tests.

petertonysmith94: This was my lazy attempt at suppressing our noisy CI.

A better solution: edd959d1227c6e65f4dd4ec6520f9827a8e07a34

danielbate (approved):

arboleya (approved):
❤️

Torres-ssf (approved):

PR #3449, opened 2024-12-05
chore: dependency pinning and auditing

### LINKED ISSUE #3448: Dependency Pinning & Auditing
- Implement dependency pinning to defend against malicious software updates.
- Implement `pnpm audit` on a release

danielbate:
- Closes #3448 

# Summary

- Pins dependency version
- A workflow to audit releases

# Checklist


maschad (changes requested):
Dependabot already alerts us of existing vulnerabilities in our dependencies, so I don't believe this CI check is necessary. 

Also I'm not sure if dependency pinning serves as a defence against upgrading to malicious software:

1. We already commit our lockfile, and so `pnpm` will install the same versions of the dependency. So if a vulnerability is discovered in an audit, we should actually upgrade to the patched version as opposed to keeping the dependency pinned at a particular version.

2. Even pinned versions can become compromised retroactively - an attacker could gain access to the package registry account and replace the package content while keeping the same version number  - we could explore using [provenance logs](https://docs.npmjs.com/generating-provenance-statements) for npm packages in such a case, but not my deps provide them unfortunately. That being said It still may be worthwhile to run `pnpm audit signatures` in the future.

danielbate:
@maschad neither of these changes were intended as full-proof solutions, but merely mitigation. Agreed deps can be retroactively compromised and we commit our lock file.

Pinning dependencies does mitigate risk of us upgrading to a compromised version, which in my experience is a more common risk. 

Good point on the `audit`, with pinning in place we would only be upgrading from dependabot, so it would be redundant.

.changeset/fluffy-cooks-worry.md
```diff
@@ -0,0 +1,20 @@
...
+"@fuel-ts/crypto": patch
+"@fuel-ts/hasher": patch
+"@fuel-ts/logger": patch
+"@fuel-ts/merkle": patch
+"fuels": patch
+"@fuel-ts/utils": patch
+"@fuel-ts/math": patch
+---
+
+chore: dependency pinning and audit
```
petertonysmith94, line 20:
```suggestion
chore: dependency pinning
```

petertonysmith94 (commented):
Nice 🪨 🌟

Dhaiwat10 (dismissed):

petertonysmith94 (dismissed):

Torres-ssf (dismissed):

maschad:
> @maschad neither of these changes were intended as full-proof solutions, but merely mitigation. Agreed deps can be retroactively compromised and we commit our lock file.
> 
> Pinning dependencies does mitigate risk of us upgrading to a compromised version, which I believe is a more common risk.
> 
> Good point on the `audit`, with pinning in place we would only be upgrading from dependabot, so it would be redundant.


Thanks for the clarification @danielbate 

Upon further consideration, I believe we could strengthen our security approach by adding `pnpm audit` specifically to our release workflow. This would ensure we never release with any _known_ vulnerabilities. So I would welcome that change. Of course we can also simultaneously still benefit from the dependabot security alerts as well, perhaps we need to revisit the dependabot upgrade workflows to ensure that process is more streamlined.

While version pinning seems like it would increase security through strict control, it can actually work against us by making it harder to receive automatic security fixes, bug fixes, and improvements. Since we already commit our _lockfile_, we get the best of both worlds: reproducible builds through the _lockfile_, plus the ability to receive ranged upgrades with important fixes.

What are your thoughts on this approach?

Dhaiwat10:
I agree with adding `pnpm audit` as a part of our CI. I believe the frontend team has this as well

danielbate:
Thankyou for the investigation @maschad, agreed with you and @Dhaiwat10 I'll implement audit here. Although that will introduce some upgrade problems which I'll need to fix, so I'll convert this to draft.

Re dep pinning, I did take a look at something @petertonysmith94 mentioned in sync yesterday, that although we do commit the lock file, it isn't included in our [published distribution](https://www.npmjs.com/package/fuels?activeTab=code) so only benefits us in development and not consumers installing `fuels`. So they would still be at risk of installing a compromised package installed by a range dependency?

.github/workflows/audit.yaml
```diff
@@ -0,0 +1,20 @@
+name: "Audit"
```
danielbate, line 1: We may want to hold this till after the release this week as there audit issues to resolve.

.changeset/fluffy-cooks-worry.md
```diff
@@ -0,0 +1,20 @@
...
+"@fuel-ts/crypto": patch
+"@fuel-ts/hasher": patch
+"@fuel-ts/logger": patch
+"@fuel-ts/merkle": patch
+"fuels": patch
+"@fuel-ts/utils": patch
+"@fuel-ts/math": patch
+---
+
+chore: dependency pinning
```
danielbate, line 20:
```suggestion
chore: dependency pinning and auditing
```

danielbate:
I caught up with @maschad earlier and we agreed that implementing audit on release is a good idea to ensure we aren't shipping with compromised versions. We also agreed that committing the lock file doesn't protect consumers of `fuels`. So we are left with a file predicament that is do we:
- Pin versions in an attempt to protect consumers from malicious upgrades, but they also miss out on security updates, and we're still at risk of a pinned version being compromised
- Keep the range versions to protect consumers by ensuring they can receive security updates by having the latest version

This feels like a bit of play off and would appreciate another opinion @FuelLabs/sdk-ts.

A potential solution that @maschad did propose would be to pin the consumer versions (prod deps inside `fuels`) but allow ranges for our dev deps so we can still benefit from security upgrades internally.

petertonysmith94 (dismissed):

Torres-ssf (approved):

Dhaiwat10 (dismissed):

arboleya (approved):

danielbate:
Okay I'm taking the approvals as agreement. Will hold off till after this weeks release window and then will merge, as I'll resolve the audit issues over the holiday period.

arboleya (dismissed):

petertonysmith94 (dismissed):

.github/workflows/audit.yaml
petertonysmith94: Have we resolved the audit issues?

arboleya (approved):

.github/workflows/audit.yaml
petertonysmith94: I have skipped the audit and added this following issue:
- https://github.com/FuelLabs/fuels-ts/issues/3524

petertonysmith94 (approved):

Torres-ssf (approved):

nedsalk (approved):

petertonysmith94 (approved):

nedsalk (approved):

arboleya (approved):

PR #3454, opened 2024-12-08
feat: allow for instance specific cache on `Provider`

### LINKED ISSUE #2893: Evaluate Global Resources Cache Within Provider Instances
The current implementation of the resources cache, managed by the [Provider instance](https://github.com/FuelLabs/fuels-ts/blob/4e82ad42b84e520c3133907aabeea2aad1c1a199/packages/account/src/providers/provider.ts#L376), utilizes a global cache. 

This design ensures that the cache is shared among all `Provider` instances within the same dApp. This approach was chosen due to the possibility that multiple providers may be instantiated from different parts of the dApp, and transactions can be submitted from these various instances using resources owned by the same address.

While the global cache design has its advantages, it also introduces some unintentional issues:

1. Each `Provider` instance can set its cache Time-To-Live (TTL), leading to confusion since the actual cache is shared among all instances.
2. Developers might expect each `Provider` instance to manage its own cache independently, leading to potential misunderstandings.

Given the potential unintentional problems, this issue should serve as a reminder to continuously evaluate if another approach is more suitable.

maschad:
- Closes #2893 

# Release notes

In this release, we:

- introduced the `resourceCacheStrategy` parameter on the `Provider` class to allow for instance-specific cache

# Summary

If one is using multiple Providers and would like to have different TTLs for each of them, one may want to use an _instance-specific_ cache strategy, which means that each `Provider` instance will have its own cache.

This introduces that parameter to allow for an alternative to a global cache shared by all provider instances.

# Checklist


packages/account/src/providers/resource-cache.ts
petertonysmith94, line 1: For me, the `ResourceCache` is difficult to understand and not entirely extensible. Could we not implement an interface for each caching strategy and have the `ResourceCache` just a facade over these?

```ts
interface CacheStrategy<T> {
  all(): Map<string, T>;
  ttl(): number;
  get(key: string): T;
  set(key: string, value: T): void;
  clear(): void;
}

class InstanceCache implements CacheStrategy {}
class GlobalCache implements CacheStrategy {}
```

maschad: What would be the use case for extending it? My understanding is that the consumer uses exclusively either an instance specific strategy or a global one.

The implementation details of those methods would be pretty much the same in the `Instance` and `GlobalCache` classes.

maschad: What specific extensibility needs did you have in mind? 

I can appreciate the suggestion for a cleaner abstraction but the core difference between global and instance strategies is really just about where we store the Map - all the implementation logic of the methods(TTL handling, resource management, etc.) remains the same, so I'm not sure it's necessary here.

petertonysmith94: It could be something we want to use for the node and chain information. The above approach would also mean we don't check the strategy every time we get the cache.

```ts 
getActiveCache(): Map<string, CachedResource> {
  return this.strategy === 'global' ? this.globalCache.getCache() : this.instanceCache;
}
```

We could just do:
```ts
getActiveCache(): Map<string, CachedResource> {
  return this.strategy.all();
}
```

maschad: Thanks for the clarification, but I am not sure I see the immediate use case(s) for node / chain info since at the moment this cache is only used for storing transactions. 

Also I don't think that the removing the simple boolean check on the _strategy_ field justifies duplicating all the caching logic across two separate classes. 

I definitely do think there is validity in storing recently queried node and chain info though :) so that's a good insight, perhaps we should flesh that out in a separate issue? I think it would be more fruitful in exploring if we need a refactor at that point.

packages/account/src/providers/global-cache.ts
```diff
@@ -0,0 +1,41 @@
...
+    GlobalCache.instance = new GlobalCache();
+    this.cache = new Map();
+    this.ttl = 0;
+  }
+
+  setTTL(ttl: number) {
+    // Only set TTL if not already set or if new TTL is smaller
+    if (this.ttl === 0 || ttl < this.ttl) {
+      this.ttl = ttl;
+    }
```
Torres-ssf, line 31: Why is the TTL only updated if the provided value is lower than the previously set one?

packages/account/src/providers/resource-cache.ts
```diff
@@ -1,33 +1,67 @@
...
+ */
 export class ResourceCache {
-  readonly ttl: number;
-
-  constructor(ttl: number) {
-    this.ttl = ttl; // TTL in milliseconds
+  private readonly ttl: number;
+  private readonly strategy: CacheStrategy;
+  private instanceCache: Map<string, CachedResource> = new Map();
+  private globalCache: GlobalCache;
```
Torres-ssf, line 25: Why does `ResourceCache` store an instance of `GlobalCache`, even though it might not be used when the `instance` strategy is selected?

packages/account/src/providers/resource-cache.ts
```diff
@@ -73,7 +111,16 @@ export class ResourceCache {
...
-  clear() {
-    cache.clear();
+  clear(): void {
+    this.getActiveCache().clear();
+  }
+
+  reset() {
+    this.clear();
+    this.globalCache.reset();
+  }
```
Torres-ssf, line 121: Will the `GlobalCache` be used exclusively by the `ResourceCache` class? I’m asking to ensure that resetting it won’t inadvertently clear data we still depend on.

Torres-ssf (commented):
Hey @maschad, good job! 👍

I've left some questions here to better understand your motivations.

packages/account/src/providers/resource-cache.ts
Torres-ssf: @maschad, I believe what @petertonysmith94 is suggesting is that by using a simple, non-opinionated cache interface, we can reuse it for any future caching needs. 

This would include the Provider cache for [chain and node info](https://github.com/FuelLabs/fuels-ts/blob/3a178b030b4aa00d0663825776d02a258832c5c9/packages/account/src/providers/provider.ts#L421-L423). Is this correct, @petertonysmith94?

petertonysmith94: @Torres-ssf, that is what I was trying to suggest.

arboleya: > I definitely do think there is validity in storing recently queried node and chain info though :) so that's a good insight, perhaps we should flesh that out in a separate issue?

We have this already:
- https://github.com/FuelLabs/fuels-ts/pull/3324

packages/account/src/providers/resource-cache.ts
arboleya: Yes, this seems wrong.

If one instantiates two providers, one `local` and one `global,` and then I call `local.reset()`, it will call the `global.reset()` and inadvertently clear the global cache.

Additionally, if one instantiates the `local` provider from a place/function that goes out of scope before the `global`, the latter will also be reset. This will happen because you chose to use `Symbol.dispose` and call reset from inside it.

https://github.com/FuelLabs/fuels-ts/blob/350dfd01107d7e9ff5cc919d37f3c3659a28f303/packages/account/src/providers/resource-cache.ts#L123-L125

This case is even worse since users will never manually call `.reset()` and will still have their cache wiped out.

packages/account/src/providers/global-cache.ts
maschad: This is a preventative measure by taking a just conservative approach for a shared cache service. By keeping the smallest TTL it respects stricter freshness requirements. 

Here's an example scenario with a global caching strategy:

- User instantiates Provider 1 and Provider 2 with TTL = 60
- User updates Provider 1 to use TTL = 300

If we allow the larger TTL to override the smaller one, then Provider 2 might may have stale data (up to 300 seconds old) which would violate Provider 2's original requirement for 60 second freshness. In this case Provider 2 may reject a 299 second old UTXO for instance, when that wasn't necessarily desired when a user set a 60 second TTL. 

I think by taking this approach we give the user less chance of erring.

arboleya: @maschad This reasoning sounds counterintuitive and exposes the [negative] nuances of going down this opinionated caching rabbit hole. The interplay between instance and singleton cannot happen like this; it is confusing and would require an in-depth DX reevaluation.

@Torres-ssf @nedsalk I know you guys discussed this in the past, and although I can't remember exactly the context of where it came from, I must repeat my [YAGNI-inspired] statement from the last sync: I don't see a strong use case to justify such a feature right now.

packages/account/src/providers/resource-cache.ts
maschad: When I had designed this the [original intention was for consumers](https://github.com/FuelLabs/fuels-ts/pull/3454/files#diff-4dc05f389c3f0a28d095f41e799a4c9c4e024cb6a39f0ce935fe8e03da453984R73) to use exclusively a global cache strategy or an instance specific cache strategy, not a combination of both, but you're right in that this isn't programmatically enforced so the above scenario would happen.

PR #3461, opened 2024-12-09
chore: avoid re-estimate `gasPrice` at `estimateTxDependencies`

Torres-ssf:
# Summary

This PR sets the `gasPrice` parameter to `0` when dry-running a transaction in `estimateTxDependencies`. This eliminates the need to re-estimate the `gasPrice` during subsequent dry-run attempts for the same method.

# Checklist


arboleya (dismissed):

.changeset/mean-sheep-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/account": minor
```
Torres-ssf, line 2:
```suggestion
"@fuel-ts/account": patch
```

arboleya (approved):
Good catch!

Do you know if the extra request was happening only in subsequent dry-run calls?

Just trying to understand where the performance improvement should be perceived.

petertonysmith94 (approved):
Might be worthwhile adding a benchmark around the `sendTransaction` function where `estimateTxDependencies` is true :)

Torres-ssf:
> Good catch!
> 
> Do you know if the extra request was happening only in subsequent dry-run calls?
> 
> Just trying to understand where the performance improvement should be perceived.

@arboleya I've also avoided calling the `estimateGasPrice` query when re-calculating the TX `maxFee` at the `Account.fund`method by informing the already estimated value [here](https://github.com/FuelLabs/fuels-ts/pull/3461/commits/4c564032ee548fe996f42e7b528a468239b5ce3f)

Torres-ssf:
> Might be worthwhile adding a benchmark around the `sendTransaction` function where `estimateTxDependencies` is true :)

@petertonysmith94 For local node or test-net?

petertonysmith94:
@Torres-ssf was thinking a benchmark within our [`@internal/benchmarks`](https://github.com/FuelLabs/fuels-ts/blob/master/internal/benchmarks) package - can't see anywhere using this flow.

Issue #3017, opened 2024-08-23
Investigate options for registry cleanup


arboleya:
> _This is a follow-up issue after https://github.com/FuelLabs/fuels-ts/issues/2377_

When someone runs `npm info fuels,` the output is far from ideal:

<details>
  <summary>Click for screenshot</summary>

![npm-info-fuels](https://github.com/user-attachments/assets/a84ac437-3d9b-42c5-b866-b2a78a85b0d1)

</details>

The solution to this would include two steps:
 - [x] Contact NPM support to remove all irrelevant tags
 - [X] Deprecate irrelevant tags
 - [x] Stop publishing temporary tags to NPM

The ability to publish temporary PRs is helpful, but it'd be great if we could do that separately from the NPM registry.

Could we use GitHub packages to resolve the second point?
 - https://docs.github.com/en/packages/learn-github-packages/introduction-to-github-packages

Ideally, we'd still be able to publish temp tags while keeping our packages clean and tidy.

petertonysmith94:
@danielbate I think the second point here is super valuable, as it'd enable us to run the UI integration tests by default, and not only on merge to master.

arboleya:
Also, here's the NPM policy:
https://docs.npmjs.com/policies/unpublish

danielbate:
I have contacted support to remove the [following versions](https://gist.github.com/danielbate/ed8839088c2f48b456f5ff1ba9446fdf), let's see how it goes. Worst case scenario we can alter the script and [deprecate](https://docs.npmjs.com/cli/v8/commands/npm-deprecate) these versions.

arboleya:
Easy, _just_ 21k+ versions. 😅

danielbate:
Okay so unfortunately support cannot remove our versions... Our weekly download count is too high and we have too many dependents. 

I'm going to make an adjustment to the script so that it deprecates these versions rather than attempting to un-publish. I'd also favour having it run on a cron whilst we go through this process.

I'll also look to move future published PR versions to a seperate registry.

danielbate:
Deprecation will clean up `info` alot, this is the result for nextjs who have more published versions than us, but deprecate.

<img width="809" alt="Screenshot 2024-12-11 at 11 03 02" src="https://github.com/user-attachments/assets/b0c82fb8-6df0-451b-b9b5-c8daf93cab88">

danielbate:
[...]

PR #3465, opened 2024-12-11
chore: release deprecation script

danielbate:
- Part of #3017 

# Summary

This PR adds a new release deprecation workflow, that will clean up our [pnpm info fuels](https://github.com/FuelLabs/fuels-ts/issues/3017#issuecomment-2535538553) result. Unfortunately, we are [unable to unpublish](https://github.com/FuelLabs/fuels-ts/issues/3017#issuecomment-2535478742) so this is the next best. In a follow up, I will provide a solution to mitigate future PR versions.

# Checklist


.changeset/lucky-flies-search.md
```diff
@@ -0,0 +1,4 @@
+---
+---
+
+chore: release deprecation script
```
danielbate, line 4: This has been tested as far as seeing the correct versions and failing the deprecate command, but needs to be run with the correct perms (in CLI) to do a full test.

.github/workflows/release-deprecate.yaml
```diff
@@ -0,0 +1,22 @@
+name: "Deprecate Old Versions"
+
+on:
+  workflow_dispatch:
```
petertonysmith94, line 4: A `dry_run` flag might be nice here, that is defaulted to true to check the output IMO.

scripts/release-deprecate.ts
```diff
@@ -0,0 +1,75 @@
...
+    } catch (err) {
+      return null;
+    }
+  });
+  return packagesNames.filter((p) => !!p);
+};
+
+const main = async () => {
+  const packages = getPublicPackages();
+  await Promise.all(
```
petertonysmith94, line 30:
```suggestion
  await Promise.allSettled(
```

Should this be all settled?

scripts/release-deprecate.ts
```diff
@@ -0,0 +1,75 @@
...
+
+      const versionsToDelete = Object.keys(packageVersions).filter(
+        (packageVersion) =>
+          packageVersion.search(UNPUBLISH_TAGS) > -1 &&
+          !compare(packageVersion, CURRENT_VERSION, '>=')
+      );
+      log('The following versions will be deprecated:');
+      log(versionsToDelete.map((v) => `   - ${v}`).join('\n'));
+
+      await Promise.all(
```
petertonysmith94, line 44:
```suggestion
      await Promise.allSettled(
```

Should this be all settled?

scripts/release-deprecate.ts
petertonysmith94, line 1: Might it be easier to read to have a `getVersionsToDeprecate`?

petertonysmith94 (approved):

Torres-ssf (approved):

scripts/release-deprecate.ts
```diff
@@ -0,0 +1,86 @@
...
+      return packageJson.private ? null : packageJson.name;
+    } catch (err) {
+      return null;
+    }
+  });
+  return packagesNames.filter((p) => !!p);
+};
+
+const getVersionsToDeprecate = async (packageName: string) => {
+  const { versions } = await fetch(`https://registry.npmjs.org/${packageName}`).then((resp) =>
```
maschad, line 30: I wonder if we should do a `try/catch` and throw an error here, given npm _may_ return a non-200 status for the job given that amount of packages we are checking for. Have you seen such situations in the past?

scripts/release-deprecate.ts
maschad: What kinds of failures were happening that lead you to choose this @petertonysmith94 @danielbate ?

maschad (commented):
Looks good but just have two questions

scripts/release-deprecate.ts
petertonysmith94: The underlying functionality can fail - such as the `fetch` requests. I haven't seen it with this script but it did occur on the unpublish script when I tried using it.

danielbate: Yeah the initial run contains around 20K versions and the deprecate CLI is rate limited (although it is pretty generous) so it was a good change.

scripts/release-deprecate.ts
danielbate: I haven't seen this endpoint fail, the only place I have seen it was on the deprecate CLI where we're doing thousands of versions rather than 10 packages, and even then it's still pretty resiliant.

.changeset/lucky-flies-search.md
arboleya: @danielbate Can't our CI user run this?

arboleya (approved):

.changeset/lucky-flies-search.md
danielbate: It can, but I was awaiting approvals before I tried a run in deprecation mode 👍🏻

PR #3508, opened 2024-12-30
fix: unable to kill `fuels dev` with `pnpm`

### LINKED ISSUE #3479: Can't kill `fuels:dev` process when using `pnpm`
After creating a new app with:
```
pnpm create fuels
```

And running it with:

```
pnpm fuels:dev
```

One cannot kill the script with `CTRL+C` — only with double `CTRL+C`.

Apparently the above doesn't happen with `bun` or `npm`.

petertonysmith94:
- Closes #3479 

# Release notes

In this release, we:

- Did this and that 

# Summary

# Breaking Changes

# Checklist

### DIFF
--- a/.changeset/thin-planes-cover.md
+++ b/.changeset/thin-planes-cover.md
@@ -1,0 +1,3 @@
+---
+"create-fuels": patch
+---
--- a/.github/workflows/pr-release.yaml
+++ b/.github/workflows/pr-release.yaml
@@ -8,7 +8,7 @@ jobs:
     name: "Release PR to npm"
     runs-on: ubuntu-latest
     # comment out if:false to enable release PR to npm
-    if: false
+    # if: false
     permissions: write-all
     steps:
       - name: Checkout


.changeset/thin-planes-cover.md
```diff
@@ -0,0 +1,7 @@
+---
+"@fuel-ts/account": patch
+"fuels": patch
+"create-fuels": patch
```
petertonysmith94, line 4:
```suggestion
```

.github/workflows/pr-release.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
     name: "Release PR to npm"
     runs-on: ubuntu-latest
     # comment out if:false to enable release PR to npm
-    if: false
+    # if: false
```
```suggestion
    if: false
```

arboleya (approved):

danielbate (approved):

nedsalk (changes requested):
We need tests for both issues the PR is closing (#2889, #3479).
You can find a test for #2889 in #3038. The same infrastructure in that test can be used for this PR. An approach could be to use a modified version of [`runInit`](https://github.com/FuelLabs/fuels-ts/pull/3038/files#diff-9f399cacc815c05a7c7ca9338f3e0e8738ae80c47a0b0ef11152d48e95c21916R9) where one would add a `"fuels:dev": "fuels dev"` script into the `package.json` file and then run a detached `spawn('pnpm fuels:dev')` command. This should simulate the same behavior that we do manually when testing with releases of PRs.

Torres-ssf (approved):

petertonysmith94:
> We need tests for both issues the PR is closing (#2889, #3479). You can find a test for #2889 in #3038. The same infrastructure in that test can be used for this PR. An approach could be to use a modified version of [`runInit`](https://github.com/FuelLabs/fuels-ts/pull/3038/files#diff-9f399cacc815c05a7c7ca9338f3e0e8738ae80c47a0b0ef11152d48e95c21916R9) where one would add a `"fuels:dev": "fuels dev"` script into the `package.json` file and then run a detached `spawn('pnpm fuels:dev')` command. This should simulate the same behavior that we do manually when testing with releases of PRs.

I found that the test within #3038 only working with Linux due to the `ps` command used.

I found that when I re-added the following ["fix" line](https://github.com/FuelLabs/fuels-ts/pull/3038/files#diff-b8f07cac51692505e88e065262fd373293a373ed5f6196eb13ec811e37aee136L264) to the source code (e2998d822fc4d7d5869de9661a1d0f23308bafe5), the tests can still [pass](https://github.com/FuelLabs/fuels-ts/actions/runs/12595438541/job/35104687935). I experienced a lot of timeouts with this test in CI, have you experienced similar?

For me, we could investigate this further but I don't think the time required outweighs the value that this test would provide. Processes and process events are inherently difficult to test and I believe manual testing is adequate for now. Happy to create an issue to revisit this and add automated tests for this scenario, but IMO it's not top priority.

nedsalk:
@petertonysmith94 sounds good. I have created #3547 and we can continue the discussion there. With regards to the test I wrote in #3038, it was failing at that time's `master` and succeeding with the fix. I'm not sure what happened in the meantime. All in all, this PR fixes it all as well so let's go with it :)

nedsalk (approved):

PR #3528, opened 2025-01-03
fix: usage of `providerUrl` in `fuels dev` command

### LINKED ISSUE #3527: Add support for `fuelCorePort: 0` in `fuels.config.ts`
We previously had [issues](https://github.com/FuelLabs/fuels-ts/issues/3263) due to `fuels deploy/dev` not working with `fuelCorePort: 0` as a config value and we had to write [this](https://github.com/FuelLabs/fuels-ts/blob/master/apps/docs/scripts/pretest.sh) pretest script. By adding support for the ephemeral port, we can remove this script.

nedsalk:
- Closes #3527 

# Summary

The `autoStartFuelCore` function wasn't using the provider url returned by `launchNode` but was rather constructing its own url, which led to the inability to use the ephemeral port `0` because it'd construct an invalid `http://127.0.0.1:0/v1/graphql` url.

# Checklist


package.json
```diff
@@ -124,7 +124,8 @@
     "vite-plugin-json5": "1.1.2",
     "vite-plugin-node-polyfills": "0.22.0",
     "vite-plugin-plain-text": "1.4.2",
-    "vitest": "2.0.5"
+    "vitest": "2.0.5",
+    "fuels": "workspace:*"
```
nedsalk, line 128: I had to add this so that I could run [this command in the test](https://github.com/FuelLabs/fuels-ts/pull/3528/files#diff-37ed97cd9ea1a119d0e728d36d7c5bd32d1983a202bd48688124c682515ae7c2R93). This is because the `spawn` is running in the root of the repo and there is no `fuels` dependency there that it can run. Upon adding it, the `spawn` runs as expected.

packages/fuels/src/cli/commands/dev/autoStartFuelCore.ts
```diff
@@ -25,9 +25,7 @@ export const autoStartFuelCore = async (config: FuelsConfig) => {
 
     const port = config.fuelCorePort ?? (await getPortPromise({ port: 4000 }));
 
-    const providerUrl = `http://${accessIp}:${port}/v1/graphql`;
-
-    const { cleanup, snapshotDir } = await launchNode({
+    const { cleanup, url, snapshotDir } = await launchNode({
```
arboleya, line 28: I think this PR could be reduced to this:

```ts
const port = config.fuelCorePort ?? 0;

const { cleanup, snapshotDir, url: providerUrl } = await launchNode({
```

And then changing the title to something like:

```console
fix: usage of `providerUrl` in `fuels dev` command
```

Then, remove the `portfinder` if it's not used elsewhere.

nedsalk: I did consider this but was hesitant to suggest because it would be a breaking change due to the default previously being port `4000` and now it'd become a random port unless specified by the user.

I don't see a way to mitigate this breaking change technically. For example, if we set a default value for `config.fuelCorePort` to `4000` and then, if `launchNode` fails to create a node on that port that we retry with port `0`, we wouldn't be able to differentiate between the default `4000` and an explicit `4000` that the user set and for which we _don't_ want to default to `0` and retry.

We could, however, just make the PR a breaking change one and tell the users that they have to specify the `fuelCorePort` value now if they want their `fuel-core` node to always run on the same port.

nedsalk: It would always start at `4000` and probably would always stay at `4000` for the majority of users and people might have taken it as a static default, hence why I saw it as a breaking change. But you're right, we didn't guarantee it being static. I made the changes, please review again.

arboleya: The tests are now failing, but you have a good point: we must consider these soft changes on behalf of users since `4000` is usually the default port used to start `fuel-core`, and users may be accustomed to it.

If we default to `0`, which port does the search start at?

nedsalk: It will be a random port assigned by the OS in the ephemeral port range. The below command gives the range in linux systems and is 32768 to 60999.
```sh
sysctl net.ipv4.ip_local_port_range
# net.ipv4.ip_local_port_range = 32768    60999
```

arboleya: Ok, it's way distant than the default `4000`.

Let me go back for a second here: Why can't we continue to use `getPortPromise()`?

```ts
const port = config.fuelCorePort ?? (await getPortPromise({ port: 4000 }));

const { cleanup, snapshotDir, url: providerUrl } = await launchNode({
```

nedsalk: We can continue to use it - that was the original solution in the PR. I'd just have to revert https://github.com/FuelLabs/fuels-ts/pull/3528/commits/2e2016fdf9157932add34012c07ae83c5f100a7e and https://github.com/FuelLabs/fuels-ts/pull/3528/commits/1fb0813534661a7b2003a4865e4cda426aa5fc1b and it would be like that again.

arboleya: Conflicts to fix.

package.json
```diff
@@ -118,7 +118,8 @@
...
```
petertonysmith94, line 122: I'm not a fan of this.

Could we not move the [test](https://github.com/FuelLabs/fuels-ts/pull/3528/files#diff-37ed97cd9ea1a119d0e728d36d7c5bd32d1983a202bd48688124c682515ae7c2R83) for these into `fuel-gauge` and avoid this import?

arboleya: The test seems redundant now, doesn't it?

At least, it seems disconnected from the current purpose of PR.

nedsalk: No, because the test itself is running in the root directory of the repo so the `spawn` is run there, regardless of where the test is located. I'd have to change the `spawn`'s `cwd` to be in `packages/fuel-gauge` and that just feels wrong because it'd be introducing infra-related quirks into the test itself.

arboleya: I also prefer a simple `{ cwd }` option passed to `spawn()` with a simple descriptive comment rather than adding the dep on the repo root since we'll soon forget why it's there. A simple CWD on the test makes it obvious—if the test is still necessary.

nedsalk: https://github.com/FuelLabs/fuels-ts/pull/3528/commits/b3280903e036cdfb3d484be9b970b4d82f2326c5

arboleya (approved):

petertonysmith94 (approved):

danielbate (approved):

PR #3540, opened 2025-01-03
feat: prevent illicit burn

### LINKED ISSUE #3489: Prevent implicit burn
Given how delicate and disastrous this can be, I think it's worth considering.

Inspired by:
 - https://github.com/FuelLabs/fuels-rs/pull/1562

petertonysmith94:
- Closes #3489 

# Release notes

In this release, we:

- Added a guard to transaction to ensure that we prevent implicit asset burns. 

# Summary

- When building a transaction, if inputs are added to a request without respective change outputs, then we will throw an error on sending a transaction.
- Unless you set the `enableBurn` flag on the transaction.  

# Checklist

### DIFF
--- a/.changeset/wild-avocados-carry.md
+++ b/.changeset/wild-avocados-carry.md
@@ -1,0 +1,6 @@
+---
+"@fuel-ts/account": patch
+"@fuel-ts/errors": patch
+---
+
+feat: prevent implicit asset burn
--- a/packages/account/src/providers/provider.ts
+++ b/packages/account/src/providers/provider.ts
@@ -866,6 +866,13 @@ Supported fuel-core version: ${supportedVersion}.`
         `The transaction exceeds the maximum allowed number of outputs. Tx outputs: ${tx.outputs.length}, max outputs: ${maxOutputs}`
       );
     }
+
+    if (tx.hasBurnableAssets()) {
+      throw new FuelError(
+        ErrorCode.ASSET_BURN_DETECTED,
+        'Asset burn detected.\nAdd relevant coin change outputs to the transaction, or enable asset burn in the transaction request (`request.enableBurn()`).'
+      );
+    }
   }
 
   /**
--- a/packages/account/src/providers/transaction-request/transaction-request.ts
+++ b/packages/account/src/providers/transaction-request/transaction-request.ts
@@ -118,6 +118,8 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
   outputs: TransactionRequestOutput[] = [];
   /** List of witnesses */
   witnesses: TransactionRequestWitness[] = [];
+  /** Whether the transaction request should enable asset burn */
+  burnEnabled: boolean = false;
 
   /**
    * Constructor for initializing a base transaction request.
@@ -708,5 +710,31 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
 
   byteLength(): number {
     return this.toTransactionBytes().byteLength;
+  }
+
+  /**
+   * Enables asset burn for the transaction request.
+   *
+   * @param burnEnabled - Whether the transaction request should enable asset burn.
+   * @returns This transaction request.
+   */
+  enableBurn(burnEnabled: boolean = true): this {
+    this.burnEnabled = burnEnabled;
+    return this;
+  }
+
+  hasBurnableAssets(): boolean {
+    if (this.burnEnabled) {
+      return false;
+    }
+
+    const coinInputs = new Set(this.getCoinInputs().map((input) => input.assetId));
+    const changeOutputs = new Set(
+      this.outputs
+        .filter((output) => output.type === OutputType.Change)
+        .map((output) => output.assetId)
+    );
+    const difference = new Set([...coinInputs].filter((x) => !changeOutputs.has(x)));
+    return difference.size > 0;
   }
 }
--- a/packages/fuel-gauge/src/transaction.test.ts
+++ b/packages/fuel-gauge/src/transaction.test.ts
@@ -1,12 +1,14 @@
+import type { CoinTransactionRequestInput } from 'fuels';
 import {
   FuelError,
   hexlify,
   InputMessageCoder,
+  InputType,
   ScriptTransactionRequest,
   sleep,
   TransactionType,
 } from 'fuels';
-import { expectToThrowFuelError, launchTestNode, TestMessage } from 'fuels/test-utils';
+import { ASSET_A, expectToThrowFuelError, launchTestNode, TestMessage } from 'fuels/test-utils';
 
 import { CallTestContractFactory } from '../test/typegen';
 
@@ -149,7 +151,6 @@ describe('Transaction', () => {
 
   it('should allow an asset burn when enabled', async () => {
     const {
-      provider,
       wallets: [sender],
     } = await launchTestNode();
 
@@ -158,9 +159,8 @@ describe('Transaction', () => {
     // Set the asset burn flag
     request.enableBurn(true);
 
-    // Add a coin input, without any output change
-    const baseAssetId = provider.getBaseAssetId();
-    const { coins } = await sender.getCoins(baseAssetId);
+    // Add a coin input, which adds the relevant coin change output
+    const { coins } = await sender.getCoins(ASSET_A);
     const [coin] = coins;
     request.addCoinInput(coin);
 
@@ -169,29 +169,34 @@ describe('Transaction', () => {
     request.maxFee = cost.maxFee;
     await sender.fund(request, cost);
 
-    const { waitForResult } = await sender.sendTransaction(request);
-    const { isStatusSuccess } = await waitForResult();
+    const tx = await sender.sendTransaction(request);
+    const { isStatusSuccess } = await tx.waitForResult();
     expect(isStatusSuccess).toEqual(true);
   });
 
-  it('should throw an error when an asset burn is detected', async () => {
+  it.only('should throw an error when an asset burn is detected', async () => {
     const {
-      provider,
       wallets: [sender],
     } = await launchTestNode();
 
     const request = new ScriptTransactionRequest();
 
     // Add a coin input, without any output change
-    const baseAssetId = provider.getBaseAssetId();
-    const { coins } = await sender.getCoins(baseAssetId);
+    const { coins } = await sender.getCoins(ASSET_A);
     const [coin] = coins;
-    request.addCoinInput(coin);
-
-    const cost = await sender.getTransactionCost(request);
-    request.gasLimit = cost.gasUsed;
-    request.maxFee = cost.maxFee;
-    await sender.fund(request, cost);
+    const { id, owner, amount, assetId, predicate, predicateData } = coin;
+    const coinInput: CoinTransactionRequestInput = {
+      id,
+      type: InputType.Coin,
+      owner: owner.toB256(),
+      amount,
+      assetId,
+      txPointer: '0x00000000000000000000000000000000',
+      witnessIndex: request.getCoinInputWitnessIndexByOwner(owner) ?? request.addEmptyWitness(),
+      predicate,
+      predicateData,
+    };
+    request.inputs.push(coinInput);
 
     await expectToThrowFuelError(
       () => sender.sendTransaction(request),
--- a/packages/fuel-gauge/src/transaction.test.ts
+++ b/packages/fuel-gauge/src/transaction.test.ts
@@ -1,5 +1,12 @@
-import { hexlify, InputMessageCoder, sleep, TransactionType } from 'fuels';
-import { launchTestNode, TestMessage } from 'fuels/test-utils';
+import {
+  FuelError,
+  hexlify,
+  InputMessageCoder,
+  ScriptTransactionRequest,
+  sleep,
+  TransactionType,
+} from 'fuels';
+import { expectToThrowFuelError, launchTestNode, TestMessage } from 'fuels/test-utils';
 
 import { CallTestContractFactory } from '../test/typegen';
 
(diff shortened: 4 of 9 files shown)


packages/fuel-gauge/src/transaction.test.ts
```diff
@@ -139,4 +148,62 @@ describe('Transaction', () => {
...
+    request.gasLimit = cost.gasUsed;
+    request.maxFee = cost.maxFee;
+    await sender.fund(request, cost);
+
+    const tx = await sender.sendTransaction(request);
+    const { isStatusSuccess } = await tx.waitForResult();
+    expect(isStatusSuccess).toEqual(true);
+  });
+
+  it.only('should throw an error when an asset burn is detected', async () => {
```
arboleya, line 177: Oops.

```suggestion
  it('should throw an error when an asset burn is detected', async () => {
```

arboleya (dismissed):

packages/fuel-gauge/src/transaction.test.ts
arboleya: Oh sorry, it's still a draft. I was going on with the flow.

.changeset/wild-avocados-carry.md
```diff
@@ -0,0 +1,6 @@
+---
+"@fuel-ts/account": patch
+"@fuel-ts/errors": patch
+---
+
+feat: prevent implicit asset burn
```
nedsalk, line 6: Preventing implicit asset burns would constitute a breaking change because those who have been burning their assets won't be able to do so anymore until they enable the flag.

danielbate (commented):
I think we probably want some docs [here](https://docs.fuel.network/docs/fuels-ts/transactions/transaction-request/) on burning assets for a transaction request

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -709,4 +711,39 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
...
+
+  /**
+   * @hidden
+   *
+   * Checks if the transaction request has burnable assets.
+   *
+   * @returns a boolean indicating whether the transaction request has burnable assets.
+   */
+  hasBurnableAssets(): boolean {
+    if (this.burnEnabled) {
```
nedsalk, line 736: The way this method is named, IMO it shouldn't be concerned with the `burnEnabled` flag as it's just a query to see if the tx request has burnable assets. I'd say remove this `if(this.burnEnabled)` check and update [the provider](https://github.com/FuelLabs/fuels-ts/pull/3540/files#diff-69a6670d0426c74935ed2830ce552e7df2379c2d112a582306e57569fcd17029R870) to check:
```ts
if(!tx.burnEnabled && tx.hasBurnableAssets) { throw new FuelError(...) }
```

nedsalk (commented):
Looks good, but I'd prefer putting all the `hasBurnableAssets` logic into a helper function which'd be invoked in the `Provider` and can be something like:
```ts
export function validateAssetBurnability(tx: TransactionRequest) {
  if (tx.burnEnabled) { return; }
  
    const coinInputs = new Set(tx.getCoinInputs().map((input) => input.assetId));
    const changeOutputs = new Set(
      tx.outputs
        .filter((output) => output.type === OutputType.Change)
        .map((output) => output.assetId)
    );
    const difference = new Set([...coinInputs].filter((x) => !changeOutputs.has(x)));

    if (difference.size > 0) { 
      throw new FuelError(
        ErrorCode.ASSET_BURN_DETECTED,
        'Asset burn detected.\nAdd the relevant change outputs to the transaction, or enable asset burn in the transaction request (`request.enableBurn()`).'
      ); }
}
```

We can then satisfy ourselves with just the `fuels-gauge/src/transaction.test.ts` you implemented and maybe a test to verify that `txRequest.enableBurn` does its thing. Perhaps we don't even need the `enableBurn` method but rather just allow the user to update the field directly:
```ts
txRequest.canBurnAssets = true;
```

.changeset/wild-avocados-carry.md
petertonysmith94: Made this a breaking change

apps/docs/src/guide/transactions/snippets/transaction-request/asset-burn.ts
```diff
@@ -0,0 +1,39 @@
...
+  assetId: coin.assetId,
+  txPointer: '0x00000000000000000000000000000000',
+  witnessIndex:
+    transactionRequest.getCoinInputWitnessIndexByOwner(coin.owner) ??
+    transactionRequest.addEmptyWitness(),
+});
+
+// Enable asset burn
+transactionRequest.enableBurn();
+
```
Torres-ssf, line 31: @petertonysmith94 Have you considered controlling this behavior with a flag in `Provider.sendTransaction`? 

One approach could be adding an `enableBurn` flag to the `ProviderSendTxParams` type. 

Then, the missing `ChangeOutput` validation could be handled directly within the `Provider.sendTransaction` method.

petertonysmith94: I did consider it to start with, but never revisited it. Thanks for the suggestion.

4004a0a5f8c1b188845bee2fca4ddda9d40f7da5

apps/docs/src/guide/transactions/transaction-request.md
```diff
@@ -97,3 +97,11 @@ The transaction ID is a SHA-256 hash of the entire transaction request. This can
 <<< @./snippets/transaction-request/add-witness.ts#transaction-request-11{ts:line-numbers}
 
 > **Note**: Any changes made to a transaction request will alter the transaction ID. Therefore, you should only get the transaction ID after all modifications have been made.
+
+### Burning assets
+
+Assets can be burnt as part of a transaction that has inputs without associated output change. The SDK validates against this behavior, so we need to implicitly enable this by sending the transaction with the `enableAssetBurn` option set to `true`.
```
nedsalk, line 103:
```suggestion
Assets can be burnt as part of a transaction that has inputs without associated output change. The SDK validates against this behavior, so we need to explicitly enable this by sending the transaction with the `enableAssetBurn` option set to `true`.
```

packages/account/src/providers/provider.ts
```diff
@@ -853,9 +859,11 @@ Supported fuel-core version: ${supportedVersion}.`
    */
   async sendTransaction(
     transactionRequestLike: TransactionRequestLike,
-    { estimateTxDependencies = true }: ProviderSendTxParams = {}
+    { estimateTxDependencies = true, enableAssetBurn = false }: ProviderSendTxParams = {}
   ): Promise<TransactionResponse> {
     const transactionRequest = transactionRequestify(transactionRequestLike);
+    validateTransactionForAssetBurn(transactionRequest, enableAssetBurn);
+
```
nedsalk, line 866:
```suggestion
    { estimateTxDependencies = true, enableAssetBurn }: ProviderSendTxParams = {}
  ): Promise<TransactionResponse> {
    const transactionRequest = transactionRequestify(transactionRequestLike);
    validateTransactionForAssetBurn(transactionRequest, enableAssetBurn);
```

packages/account/src/providers/transaction-request/helpers.ts
```diff
@@ -93,3 +96,38 @@ export const cacheRequestInputsResourcesFromOwner = (
...
+  const changeOutputs = new Set(
+    outputs.filter((output) => output.type === OutputType.Change).map((output) => output.assetId)
+  );
+  const difference = new Set([...coinInputs].filter((x) => !changeOutputs.has(x)));
+  return difference.size;
+};
+
+export const validateTransactionForAssetBurn = (
+  tx: TransactionRequest,
+  enableAssetBurn: boolean = false
```
nedsalk, line 115:
```suggestion
  enableAssetBurn: boolean | undefined
```

packages/account/src/wallet/base-wallet-unlocked.ts
```diff
@@ -112,15 +112,15 @@ export class BaseWalletUnlocked extends Account {
    */
   override async sendTransaction(
     transactionRequestLike: TransactionRequestLike,
-    { estimateTxDependencies = false }: ProviderSendTxParams = {}
+    { estimateTxDependencies = false, enableAssetBurn = false }: ProviderSendTxParams = {}
```
nedsalk, line 115:
```suggestion
    { estimateTxDependencies = false, enableAssetBurn }: ProviderSendTxParams = {}
```

packages/account/src/wallet/base-wallet-unlocked.ts
```diff
@@ -112,15 +112,15 @@ export class BaseWalletUnlocked extends Account {
...
   ): Promise<TransactionResponse> {
     const transactionRequest = transactionRequestify(transactionRequestLike);
     if (estimateTxDependencies) {
```
nedsalk, line 118: Can we do a validation before this `if(estimateTxDependencies)` here like we [do](https://github.com/FuelLabs/fuels-ts/pull/3540/files#diff-69a6670d0426c74935ed2830ce552e7df2379c2d112a582306e57569fcd17029R865) in `Provider.sendTransaction` to avoid network calls?

nedsalk (commented):
Looks good, just some nits and a question.

packages/account/src/providers/transaction-request/helpers.ts
Torres-ssf: @nedsalk Why not:

```suggestion
  enableAssetBurn?: boolean
```
🤣

packages/account/src/wallet/base-wallet-unlocked.ts
petertonysmith94: 82f472d680af20c24a61b458f7da334733a4aacd

packages/account/src/providers/transaction-request/helpers.ts
petertonysmith94: IMO setting an explicit default here is clearer to read.

packages/account/src/providers/provider.ts
petertonysmith94: 82f472d680af20c24a61b458f7da334733a4aacd

packages/account/src/providers/transaction-request/helpers.test.ts
```diff
@@ -196,5 +202,131 @@ describe('helpers', () => {
...
+          { type: OutputType.Change, assetId: ASSET_B, to: owner.toB256() },
+        ];
+        const enableAssetBurn = false;
+
+        const burnableAssets = validateTransactionForAssetBurn(
+          { inputs, outputs },
+          enableAssetBurn
+        );
+
+        expect(burnableAssets).toBeUndefined();
```
nedsalk, line 269:
```suggestion
        expect(() =>
          validateTransactionForAssetBurn({ inputs, outputs }, enableAssetBurn)
        ).not.toThrow();
```

packages/account/src/providers/transaction-request/helpers.test.ts
```diff
@@ -196,5 +202,131 @@ describe('helpers', () => {
...
+        ];
+        const outputs: TransactionRequestOutput[] = [];
+        const enableAssetBurn = true;
+
+        const burnableAssets = validateTransactionForAssetBurn(
+          { inputs, outputs },
+          enableAssetBurn
+        );
+
+        expect(burnableAssets).toBeUndefined();
```
nedsalk, line 308:
```suggestion
        expect(() =>
          validateTransactionForAssetBurn({ inputs, outputs }, enableAssetBurn)
        ).not.toThrow();
```

nedsalk (approved):
Approved with nits, you can ignore them if you want.

nedsalk (dismissed):

.changeset/wild-avocados-carry.md
```diff
@@ -0,0 +1,6 @@
+---
+"@fuel-ts/account": minor
+"@fuel-ts/errors": patch
+---
+
+feat!: prevent implicit asset burn
```
arboleya, line 6: Please, correct me if I'm wrong, but before, we'd add outputs automatically, right?

If that's the case, and I know the title came from the issue I created, but wouldn't this title be more accurate?

```
feat!: allow explicitly asset burn
```

petertonysmith94: We still do add corresponding outputs when adding inputs via the `transactionRequest.addCoinInput` method.

The only means to cause an asset burn at the moment would be to adding the inputs directly on the request `transactionRequest.input.push(...)`, without adding the associated output change.

danielbate (dismissed):

.changeset/wild-avocados-carry.md
arboleya: Can users use `addCoinInput()` and still choose not to set outputs, thus burning the assets?

arboleya (approved):

packages/fuel-gauge/src/transaction.test.ts
```diff
@@ -129,4 +138,60 @@ describe('Transaction', () => {
...
+    const {
+      wallets: [sender],
+    } = await launchTestNode();
+
+    const request = new ScriptTransactionRequest();
+
+    // Add a coin input, which adds the relevant coin change output
+    const { coins } = await sender.getCoins(ASSET_A);
+    const [coin] = coins;
+    request.addCoinInput(coin);
```
arboleya, line 152: Just as @segfault-magnet said it [here](https://github.com/FuelLabs/fuels-rs/pull/1562#discussion_r1905377027), I think we should also double-check and consider messages.

petertonysmith94: b321ec8f8c9bdeb5c5bf6caf1e14c9b131985927

packages/account/src/providers/transaction-request/helpers.ts
```diff
@@ -93,3 +95,61 @@ export const cacheRequestInputsResourcesFromOwner = (
...
+ * @returns The number of burnable assets in the transaction request.
+ */
+export const getBurnableAssetCount = (
+  baseAssetId: string,
+  opts: Pick<TransactionRequest, 'inputs' | 'outputs'>
+) => {
+  const { inputs, outputs } = opts;
+  const coinInputs = new Set(inputs.filter(isRequestInputCoin).map((input) => input.assetId));
+  // If there is a message input without data, we need to add the base asset to the set
+  if (inputs.some(isRequestInputMessageWithoutData)) {
```
Torres-ssf, line 114:
```suggestion
  if (inputs.some((i) => isRequestInputMessage(i) && bn(i.amount).gt(0))) {
```

Even though messages with data (non-coin messages ) cannot be used to pay for the transaction fee, they are spendable and they can have amounts, which will result in a UTXO being created by a `ChangeOutput` or/and `CoinOutput`

packages/account/src/providers/transaction-request/helpers.ts
```diff
@@ -93,3 +95,61 @@ export const cacheRequestInputsResourcesFromOwner = (
...
+ *
+ * Get the number of burnable assets in the transaction request.
+ *
+ * @param opts - The transaction request to get the burnable asset count from.
+ * @returns The number of burnable assets in the transaction request.
+ */
+export const getBurnableAssetCount = (
+  baseAssetId: string,
+  opts: Pick<TransactionRequest, 'inputs' | 'outputs'>
+) => {
```
Torres-ssf, line 110: Why not? 

```suggestion
export const getBurnableAssetCount = (
  baseAssetId: string,
  transactionRequest: TransactionRequest,
) => {
```

Torres-ssf (changes requested):
Nice job @petertonysmith94 👍

packages/account/src/providers/transaction-request/helpers.ts
petertonysmith94: I was not aware, thanks for the lowdown.

packages/account/src/providers/transaction-request/helpers.ts
petertonysmith94: Personal preference - generally I do it to document the minimal arguments required.

33e5ea9af0d6469b509683adcf07b5c57f21f854

nedsalk (approved):

arboleya (approved):

danielbate (approved):

packages/account/src/providers/transaction-request/helpers.ts
petertonysmith94: 62dc663ebc3c93dddfdd92030dad5e029947e05d

PR #3559, opened 2025-01-07
feat!: remove redundant gas price call for tx summary

### LINKED ISSUE #3536: Remove redundant gas price call in `TransactionResponse`
In instances where we are assembling a transaction summary for a transaction where we already have it's status, and therefore it's total fee, we are making a redundant network request to obtain the latest gas price. In these instances, we should instead build the response using the obtained fee, rather than recalculating it.

### LINKED ISSUE #3545: Reduce number of calls to send txs via connectors
> [!NOTE]  
> This may have crossover issues in the connectors repo, still TBD

For a basic transaction, submitted via the `FuelConnector`, the following network requests are made:
- Estimate gas price
- Dry run
- Get coins to spend
- Submission subscription
- Status subscription
- Get transaction with receipts
- Get latest gas price

Part of this can be mitigated through tx pre bundling (#3495, #3539) and offering less abstract approaches for consumers (#3541), but we should also investigate the source of the problems in our primary flows (`call()`->`waitForResult()` etc.) to improve performance for the majority of our consumers.

danielbate:
- Closes #3536 

# Release notes

In this release, we:

- Improved speed in assembling result for settled transactions

# Summary

Removes a redundant get latest gas price network call for transactions that have settled, where we can obtain the fee from the status check, rather than recalculating it.

```ts
const tx = await sender.transfer(receiver.address, 1_000, await provider.getBaseAssetId());
const result = await tx.waitForResult(); // no longer needs to call `await provider.getLatestGasPrice()`
console.log(result.fee); // bn('0x01')
```

# Breaking Changes

- `calculateTXFeeForSummary` and subsequently the `CalculateTXFeeForSummaryParams` no longer accept a `totalFee` property. If you have the `totalFee`, then there is no need to call the `calculateTxFeeForSummary()` function.

```ts
// Before
const totalFee = bn(..):
calculateTXFeeForSummary({ ..., totalFee } as CalculateTXFeeForSummaryParams);

// After
calculateTXFeeForSummary({ ... } as CalculateTXFeeForSummaryParams);
```

# Checklist


packages/account/src/providers/transaction-summary/assemble-transaction-summary.test.ts
```diff
@@ -104,7 +105,7 @@ describe('TransactionSummary', () => {
       type: expect.any(String),
     };
 
-    runTest(MOCK_SUCCESS_STATUS, expected, await provider.getBaseAssetId());
+    runTest(MOCK_SUCCESS_STATUS, expected, await provider.getBaseAssetId(), 0);
```
danielbate, line 108: On success we have the fee, no need to recalculate.

packages/account/src/providers/transaction-summary/assemble-transaction-summary.ts
```diff
@@ -55,6 +56,7 @@ export function assembleTransactionSummary<TTransactionType = void>(
     maxGasPerTx,
     gasPrice,
     baseAssetId,
+    totalFee,
```
danielbate, line 59: I tried to mitigate breaking changes here but open to ideas on the API.

.changeset/short-bears-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/account": patch
```
danielbate, line 2:
```suggestion
"@fuel-ts/account": minor
```

nedsalk (dismissed):

petertonysmith94 (approved):

packages/account/src/providers/transaction-summary/assemble-transaction-summary.ts
Torres-ssf: @danielbate Maybe I am missing something but I reckon we don't need the `totalFee` as a new parameter for the `assembleTransactionSummary` helper.

The `totalFee` is always extracted from the TX status and the `assembleTransactionSummary` helper code is already doing that when calls `processGraphqlStatus`.

packages/account/src/providers/transaction-summary/assemble-transaction-summary.ts
```diff
@@ -76,23 +78,32 @@ export function assembleTransactionSummary<TTransactionType = void>(
...
+      gasPrice,
+      rawPayload,
+      tip,
+      consensusParameters: {
+        gasCosts,
+        maxGasPerTx,
+        feeParams: {
+          gasPerByte,
+          gasPriceFactor,
+        },
```
Torres-ssf, line 104: Here `totalFee` ( `assembleTransactionSummary`'s parameter ) and `totalFeeFromStatus` ( returned by `processGraphqlStatus` ) are the same thing as they were extracted from the same TX status.

packages/account/src/providers/transaction-summary/calculate-tx-fee-for-summary.ts
```diff
@@ -37,18 +36,9 @@ export const calculateTXFeeForSummary = (params: CalculateTXFeeForSummaryParams)
...
   } = params;
 
-  /**
-   * If totalFee is provided it means that the TX was already processed and we could extract the fee
-   * from its status
-   */
-  if (totalFee) {
-    return totalFee;
-  }
-
```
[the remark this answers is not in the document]
Torres-ssf: Yes we don't need this anymore since this code was changed at `calculateTXFeeForSummary`

```ts
const fee =
    totalFee ??
    totalFeeFromStatus ??
    calculateTXFeeForSummary({...
```

Torres-ssf (changes requested):

packages/account/src/providers/transaction-summary/assemble-transaction-summary.ts
danielbate: I'm probably being overly cautious, but my concern was if we are using `totalFee` to dictate whether we are fetching the latest gas price, then we should pass down that reference to `totalFee` for the summary rather than reprocess it and create a new reference i.e. `totalFeeFromStatus`. And as it didn't introduce a breaking change, I went ahead for safety.

Give me a few to think about this.

packages/account/src/providers/transaction-summary/assemble-transaction-summary.ts
danielbate: They _should_ be the same, I was just being cautious and using the same reference. Continuing conversation [here](https://github.com/FuelLabs/fuels-ts/pull/3559/files#r1907114704) as it's for the same reason.

packages/account/src/providers/transaction-summary/assemble-transaction-summary.ts
danielbate: I've removed it, I was being overly cautious, it's very unlikely to change.

https://github.com/FuelLabs/fuels-ts/pull/3559/commits/b9c811366f3faa70cd08c214c00d7ac732514036

Torres-ssf: @danielbate I agree with you. I believe these values will always match, given that they’re derived from the transaction status.

After reviewing everything, I think we can remove the call to get the gas price entirely. When this logic was originally added, `totalFee` was not part of the transaction status, so fetching the gas price was the only way to calculate the fee. 

Now that `totalFee` is available, that extra call may no longer be necessary. Though some further investigation would be required.

I'm happy to approve these changes as is. If you’d like, I can open an issue to do an investigation later and handle it myself. What do you think?

Torres-ssf (dismissed):

packages/fuel-gauge/src/transaction-response.test.ts
```diff
@@ -306,4 +306,54 @@ describe('TransactionResponse', () => {
...
+    const result = await tx.waitForResult();
+
+    // fee is used from the success status, latest gas price not needed
+    expect(getLatestGasPriceSpy).toHaveBeenCalledTimes(0);
+    expect(result.fee.toNumber()).toBeGreaterThan(0);
+    expect(result.id).toBe(tx.id);
+  });
+
+  it('builds response and assembles result [fetches gas price then uses fee]', async () => {
+    using launched = await launchTestNode();
```
nedsalk, line 334:
```suggestion
    using launched = await launchTestNode({
      nodeOptions: {
        args: [
          '--poa-instant',
          'false',
          '--poa-interval-period',
          '2sec',
        ],
      },
    });
```

This test could be flaky because the transaction could be processed and successful before the `getLatestGasPrice` call is run, so better to mitigate that.

nedsalk (dismissed):

nedsalk (approved):

Torres-ssf (approved):

petertonysmith94 (approved):

PR #3562, opened 2025-01-08
chore!: removed `outputVariables` from the transaction cost estimation

### LINKED ISSUE #1763: Optimize `estimateTxDependencies` method
The [`estimateTxDependencies`](https://github.com/FuelLabs/fuels-ts/blob/a2b1a545b66adf9424bbd3be049025f65e4f7a8b/packages/account/src/providers/provider.ts#L682) automates the addition of missing [OutputVariables](https://specs.fuel.network/master/tx-format/output.html#outputvariable), essential for Sway's [transfer](https://fuellabs.github.io/sway/master/reference/documentation/operations/asset/transfer/index.html?highlight=transfer#transfer) or [mint](https://fuellabs.github.io/sway/master/reference/documentation/operations/asset/mint/mint.html?highlight=mint#mint) functions.

With this, users don't need to specify the `variableOutputs` parameter: 

```ts
const { transactionResult } = await contract
  .functions
  .mint_to_addresses(address, subId, mintAmount)
  .txParams({
    variableOutputs: 3, // <—————————— this
  })
  .call();
```

> [!WARNING]
> While providing a nice DX, this approach may lead to multiple dry-run calls, potentially slowing down the transaction execution and representing more use of resources, which translates into more costs.

How it works:
1. The method performs dry runs, checking output receipts for a `Revert` with `val` equal to `TransferToAddressFailed`. This error indicates a missing `OutputVariable`.
1. If such a `Revert` receipt is detected, an `OutputVariable` is added, and another dry run is executed. This loop continues until no related revert receipt is found.

Consider a contract call minting assets to 3 addresses:

```rs
    fn mint_to_addresses(addresses: [Address; 3], sub_id: b256, mint_amount: u64) {
        assert(sub_id == TOKEN_1 || sub_id == TOKEN_2 || sub_id == TOKEN_3);
        let mut counter = 0;
        while counter < 3 {
            mint_to_address(addresses[counter], sub_id, mint_amount);
            counter = counter + 1;
        }
    }
```
Outcome:
 1. Since this function mints a coin to **3x** addresses, it will need:
    - **3x** `OutputVariable`
 1. This means the dry-run call will fail:
    - **3x** times
 1. And succeed only in the:
    - **4th** execution
 1. This will result in:
    - **4x** total dry-runs

Proposed change:
- Discontinue automatic addition of `OutputVariable`
  - Shift this responsibility to users
  - Save up resources
- Docs
  - Document how to use `OutputVariable` in the SDK documentation.
  - Include information about the error returned when an `OutputVariable` is missing

----

Related:
 - https://github.com/FuelLabs/fuels-ts/issues/925

petertonysmith94:
- Closes #1763

# Release notes

In this release, we:

- TBD

# Summary

TBD

# Breaking Changes

TBD

# Checklist


Torres-ssf:
@FuelLabs/sdk-ts Have we all agreed on that? 

Stopping adding `outputVariables` on the fly can result in a major breaking change for some applications.

petertonysmith94:
> @FuelLabs/sdk-ts Have we all agreed on that?
> 
> Stopping adding `outputVariables` on the fly can result in a major breaking change for some applications.

So me and @danielbate had a chat around this, and maybe instead of this _highly breaking_ change we could flag this behaviour behind a flag for `optimizedTransaction` (or something similar).

We were in agreement that this would be beneficial for more experienced users, however, for new users this could be a barrier to entry.

nedsalk:
> So me and @danielbate had a chat around this, and maybe instead of this highly breaking change we could flag this behaviour behind a flag for `optimizedTransaction` (or something similar).

What would be the difference between the `optimizedTransaction` flag and setting `estimateTxDependencies` to `false`?

petertonysmith94:
> What would be the difference between the optimizedTransaction flag and setting estimateTxDependencies to false?

Setting `estimateTxDependencies` to `false`, would skip all estimation logic entirely.

Setting `estimateTxDependencies` to `true` and `optimizedTransaction` to `true` would result in an error if missing output variables are detected, therefore, skipping the brute force approach to finding the output variables.

Torres-ssf:
@petertonysmith94 I am missing out on how this would be optimized for more experienced users. What would be the default value for the `optimizedTransaction`?

petertonysmith94:
Closed with comment on the [issue](https://github.com/FuelLabs/fuels-ts/issues/1763#issuecomment-2581018876)

PR #3580, opened 2025-01-13
chore: prepend version mismatch to GraphQL errors

### LINKED ISSUE #3476: Reconsider version mismatch warning
Currently, someone using an outdated version of the TS SDK gets a warning like this:

```
The Fuel Node that you are trying to connect to is using fuel-core version 0.40.1,
which is not supported by the version of the TS SDK that you are using.
Things may not work as expected.
Supported fuel-core version: 0.39.0.
```

When things go wrong, people automatically assume this may be the cause, although it is rarely the case.

Removing it completely may not be the best approach, but there may be a middle ground to explore.

Changing the message is easy, but it requires careful consideration.

petertonysmith94:
- Closes #3476

# Release notes

In this release, we:

- Removed warning for incompatible node version in favour of prepending the warning to GQL errors.

# Summary

- Before, we warned users (via `console`) of incompatibilities with the `fuel-core` node version on Provider instantiation. This could often lead to misinterpretation of issues.
- Now, we prepend any error messages thrown from GraphQL, with the same incompatibility message.

# Checklist


Torres-ssf (commented):
@petertonysmith94 Should we also remove the warning here?
https://github.com/FuelLabs/fuels-ts/blob/71691a218aad8dcc3837c41598b8cbe48ef4f8f7/packages/versions/src/lib/checkFuelCoreVersionCompatibility.ts#L7-L11

petertonysmith94:
> @petertonysmith94 Should we also remove the warning here?
> 
> https://github.com/FuelLabs/fuels-ts/blob/71691a218aad8dcc3837c41598b8cbe48ef4f8f7/packages/versions/src/lib/checkFuelCoreVersionCompatibility.ts#L7-L11

@Torres-ssf I believe this is only a warning for when using unofficial versions of `fuel-core`, which I haven't seen any instance of. I'd say this one is alright to leave in.

arboleya (approved):

Torres-ssf (approved):

PR #3600, opened 2025-01-20
chore: `Address` constructor now accepts a range of inputs.

### LINKED ISSUE #3598: Improve `Address` DX
We should consider making the [`Address.constructor()`](https://github.com/FuelLabs/fuels-ts/blob/edefe5901768cf1eec08e04e3b5533e9fca1555a/packages/address/src/address.ts#L28) accept the same input as [`Address.fromDynamicInput()`](https://github.com/FuelLabs/fuels-ts/blob/edefe5901768cf1eec08e04e3b5533e9fca1555a/packages/address/src/address.ts#L205).

```ts
// before
constructor(address: B256Address) {}


// after
constructor(address: string | Address) {}
```

This could make DX a bit nicer:

```ts
// before
Address.fromDynamicInput(xyz)

// after
new Address(xyz)
```

petertonysmith94:
- Closes #3598

# Release notes

In this release, we:

- The `Address` constructor now accepts a range of different inputs

# Summary

- The `Address` constructor now accepts a range of inputs:
  - an `Address` instance
  - a `B256` string
  - an `EVM` address string
  - a public key

# Checklist


petertonysmith94:
Opened this PR early (need to update documentation, etc), just to ensure this is the direction we wish to pursue. Does anyone have any issues with this approach?

cc @FuelLabs/sdk-ts @LuizAsFight

packages/address/src/address.ts
```diff
@@ -240,8 +221,7 @@ export class Address {
       );
     }
 
-    const paddedAddress = padFirst12BytesOfEvmAddress(evmAddress);
-    return new Address(paddedAddress);
+    return new Address(evmAddress);
```
nedsalk, line 224: I don't understand why we'd want to have the `Address` class represent anything but a fuel-compatible address. It seems more appropriate to me that we differentiate between an `Address` and an `EvmAddress`, the same way Sway does it.

@LuizAsFight @luizstacio What are your thoughts here?

Torres-ssf: @nedsalk I believe this is done for simplicity since on Fuel an EVM address is represented the same way as Fuel addresses, using 32 bytes. The difference is that an EVM address is padded with zeros to make it 32 bytes.

Torres-ssf (approved):
Great job @petertonysmith94 🚀

packages/address/src/address.ts
```diff
@@ -140,13 +134,11 @@ export class Address {
...
+   *
+   * @deprecated Use `new Address` instead
    */
   static fromPublicKey(publicKey: string): Address {
-    if (!isPublicKey(publicKey)) {
-      throw new FuelError(FuelError.CODES.INVALID_PUBLIC_KEY, `Invalid Public Key: ${publicKey}.`);
-    }
-
-    const b256Address = hexlify(sha256(arrayify(publicKey)));
+    const b256Address = fromPublicKeyToB256(publicKey);
```
Torres-ssf, line 141: @petertonysmith94 Should we update the docs to conform with these new guidelines?

Torres-ssf (dismissed):

packages/address/src/address.ts
petertonysmith94: @Torres-ssf as stated in the [comment](https://github.com/FuelLabs/fuels-ts/pull/3600#issuecomment-2602761499), opened this early to get feedback/confirmation to proceed.

If we're happy with the approach, I will finalise the documentation + update references :)

petertonysmith94: f3010f91643ffbec91b7dc2f6cf189bfbcd98118

.changeset/chilly-buses-add.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fuel-ts/address": patch
```
Torres-ssf, line 2:
```suggestion
"@internal/check-imports": patch
"@fuel-ts/account": patch
"@fuel-ts/address": patch
"@fuel-ts/program": patch
```

Torres-ssf (changes requested):

arboleya (approved):

Torres-ssf (approved):

PR #3460, opened 2024-12-09
Create TestCoin feat: add TestCoin helper class (#3445)

crStiv:
## Release Notes

In this release, we:
- Added the `TestCoin` helper class to simplify the creation of test `Coin` objects for testing purposes.
- Introduced two static methods:
  - `TestCoin.create`: Generates a single test coin with customizable parameters.
  - `TestCoin.many`: Generates an array of test coins based on shared parameters.

---

## Summary

This pull request introduces the `TestCoin` helper class, designed to streamline the creation of test `Coin` objects. The `create` method allows for the generation of a single coin with default or customized properties, while the `many` method supports bulk creation of coins. These utilities will improve the efficiency and readability of tests that rely on `Coin` objects.

---

## Related Issues

- Closes #3445

---

## Breaking Changes

None.

---

## Checklist

- [x] All **changes** are **covered** by **tests**.
- [x] All **changes** are **documented**.
- [x] I **reviewed** the **entire PR** myself.
- [x] I **described** all **Breaking Changes** (or there's none).

---

_Allow edits by maintainers._

PR #3647, opened 2025-01-30
feat(account): Add TestCoin helper class for testing purposes

### LINKED ISSUE #3445: Add `TestCoin` helper class
Similar to the `TestMessage` helper class, it might be worth having one for coins as well. 

We could have a static many function which could generate multiple coins.

```ts
const coins: Coin[] = TestCoin.many({
  owner: '0x09c0b2d1a486c439a87bcba6b46a7a1a23f3897cc83a94521a96da5c23bc58db'
}, 4);
```

crStiv:
## Overview
This PR introduces the `TestCoin` helper class to simplify coin creation during testing. The class provides utility methods for generating test coins with random or specified parameters, making it easier to write and maintain tests that require coin instances.

## Features
- `TestCoin.create()`: Creates a single test coin with customizable parameters
- `TestCoin.many()`: Generates multiple test coins with shared base parameters
- Uses crypto-secure random generation via `@fuel-ts/crypto`
- Properly handles big numbers using `@fuel-ts/math`

## Related Issues
Closes #3445

## Related PR's
Fixed #3460


petertonysmith94 (changes requested):
Hey @crStiv 👋🏼 

Thanks for the review, could you add a test around this behaviour?

See [`test-message.test.ts`](https://github.com/FuelLabs/fuels-ts/blob/master/packages/account/src/test-utils/test-message.test.ts) for some inspiration.

packages/account/src/test-utils/test-coin.ts
danielbate, line 1: Please can we use the same instantiation pattern as `packages/account/src/test-utils/test-message.ts` whereby the instance is created via the constructor, rather than using a `create` method. So the flow is similar for all test utils.

danielbate (changes requested):

packages/account/src/test-utils/test-coin.ts
danielbate: Yes this is as expected, thankyou

danielbate (changes requested):
Please could you also include a test case that uses `TestCoin` in an end to end test using a transaction? `packages/account/src/providers/provider.test.ts` may be the best place for this, or inside `fuel-gauge`.

packages/account/src/test-utils/test-coin.test.ts
```diff
@@ -0,0 +1,67 @@
+import { describe, expect, test } from 'vitest';
+import { type Coin } from '@fuel-ts/account';
+import { bn, type BN } from '@fuel-ts/math';
+import { getRandomB256 } from '@fuel-ts/crypto';
```
Torres-ssf, line 4:
```suggestion
import { getRandomB256 } from '@fuel-ts/address';
```

packages/account/src/test-utils/test-coin.ts
```diff
@@ -0,0 +1,51 @@
+import { Coin } from '@fuel-ts/account';
+import { bn, type BN } from '@fuel-ts/math';
+import { getRandomB256 } from '@fuel-ts/crypto';
```
Torres-ssf, line 3:
```suggestion
import { getRandomB256 } from '@fuel-ts/address';

```

packages/account/src/test-utils/test-coin.ts
```diff
@@ -0,0 +1,51 @@
...
+  amount: BN;
+  type: number;
+}
+
+export class TestCoin {
+  public readonly id: string;
+  public readonly owner: string;
+  public readonly amount: BN;
+  public readonly type: number;
+
```
Torres-ssf, line 17: The `TestCoin` should follow the same approach as the existent `TestMessage`:

https://github.com/FuelLabs/fuels-ts/blob/01375ce29bf5f18b5d8f5e97fec5d82f4fb6ae5f/packages/account/src/test-utils/test-message.ts#L18-L24

It is intended to be used when setting up a local node, just like the `TestMessage` is used:

https://github.com/FuelLabs/fuels-ts/blob/01375ce29bf5f18b5d8f5e97fec5d82f4fb6ae5f/apps/docs/src/guide/provider/snippets/functionality/get-messages-by-nonce.ts#L4-L17

Therefore, its properties should be the same ones as the [Coin type for the stateConfig](https://github.com/FuelLabs/fuel-core/blob/05538a5de35f8b9ec5d46f9fc37b043b1c6a566e/version-compatibility/forkless-upgrade/chain-configurations/v36/state_config.json#L4-L10):

packages/account/src/test-utils/test-coin.ts
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Creates a chain-compatible coin object
+   */
+  toCoin(): Coin {
+    return {
+      id: this.id,
+      owner: this.owner,
+      amount: this.amount,
+      type: this.type,
```
Torres-ssf, line 41: The property `type` does not exist on the Coin type

packages/fuel-gauge/src/transaction.test.ts
```diff
@@ -317,4 +318,46 @@ describe('Transaction', () => {
...
+
+    const baseAssetId = await provider.getBaseAssetId();
+    const initialBalance = await wallet.getBalance(baseAssetId);
+
+    // Create test coins with specific parameters
+    const testCoin = new TestCoin({
+      amount: bn(1000),
+      owner: wallet.address.toB256(),
+      assetId: baseAssetId,
+    });
```
Torres-ssf, line 334: Is this test working? The transaction is being submitted with a coin not created on the local chain.

Torres-ssf (changes requested):

Torres-ssf:
Hey @crStiv, how are you? I noticed it's been a while since your last interaction on this, and I just wanted to check if you're still interested in working on it.

Torres-ssf:
@crStiv No worries. Take your time. I was just checking if you were still interested in working on this. 

Let us know if you have any questions, we are happy to help.

Issue #3683, opened 2025-02-10
Provide a way to submit TXs without using subscriptions


arboleya:
According to @LuizAsFight:
- The Fuel Wallet does the `sendTransaction` which calls `submitAndAwaitStatus` on ts sdk side. Then the wallet just gets the transaction id , returns do the dapp and abruptly  closes the wallet popup, not waiting for anything from this subscription anymore. could we have a way to submit a transaction that doesn't open a sub and instead just submit the tx? in the wallet use cases we dont need to wait for anything.

Follow-up issue:
- https://github.com/FuelLabs/fuels-ts/issues/3684

danielbate:
We used to use the `submit` operation, and `submitAndAwait` would be opt in via an `awaitExecution` flag on `sendTransaction`. 

The `submit` endpoint still exists on the client:

```gql
submit(
  tx: HexString!
): Transaction!
```

We could reimplement the reverse functionality, making the subscription opt out?

Voxelot:
If the default behavior is to have a subscription, it seems like it would be a breaking change that default to submit only.

danielbate:
@Voxelot agreed and we won't change the default behaviour, this will be optional. Ideally most consumers will still use the `submitAndAwaitStatus` route as that will use the smallest number of requests from submission -> summary.

I'd expect something _like_ the below snippet:

```ts
    // Default behaviour
    const { waitForResult } = await wallet.sendTransaction(txRequest); // 1 request via `submitAndAwaitStatus`
    const result = await waitForResult(); // 0 requests to assemble summary, previous subscription used
```

```ts
    // `awaitResult: true`
    const { waitForResult } = await wallet.sendTransaction(txRequest, { subscribe: false }); // 1 request made via `submit`
    const result = await waitForResult(); // 1 request made to `statusChange` to assemble summary
```

arboleya:
I'm closing for now, as subscription management should ideally not depend on the client since it will always be out of our control (connection drop, browser window being closed, etc.).

PR #3695, opened 2025-02-11
fix: ignored subscriptions not being read

### LINKED ISSUE #3684: Improve GraphQL subscriptions handling
It has been observed that sometimes subscriptions are left open.

We must ensure all subscriptions are closed to avoid bloating the nodes.

Quoting @Voxelot:
- My assumption was that the node creates a buffer for each subscription that hangs around either until 1) it is fuly consumed or 2) a timeout occurs and the buffer is dropped.
The subscription wouldn't automatically close immediately when the final state comes, because if it didn't wait for the sdk to consume the messages then the buffer would close before the user sees the messages. So if we have a lot of junk subscriptions hanging around, the problem is either with 1 or 2. Ie. there's something in the frontend opening subscriptions but ignoring the result and not consuming the messages or the client isn't properly closing the buffer when all the messages are read.

Follow-up after:
 - https://github.com/FuelLabs/fuels-ts/issues/3683

nedsalk:
- Closes #3684

# Summary

Fuel network logs showed some subscriptions weren't being closed and the current hypothesis is that the TS SDK is the culprit because it doesn't automatically read data as it comes from the node but rather waits for the user to start reading it. Even though the fuel-core nodes do close subscriptions once they stream all the data, they still have to wait for ACK messages from the clients and it might be the case that sometimes the clients' TCP receive buffer gets filled up to the point that they stop sending ACK messages to the server (because they don't read the stream), putting the server in a stalemate position until it kills the subscription via timeouts.

The next step would be to integrate the `submit` mutation ergonomically via
- #3683 

so that users can decide to not use the subscription if they don't need it.

# Checklist

### DIFF
--- a/packages/account/src/providers/fuel-graphql-subscriber.ts
+++ b/packages/account/src/providers/fuel-graphql-subscriber.ts
@@ -32,7 +32,12 @@ export class FuelGraphqlSubscriber implements AsyncIterator<unknown> {
     });
 
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    const [errorReader, resultReader] = response.body!.tee().map((stream) => stream.getReader());
+    const [backgroundStream, resultStream] = response.body!.tee();
+
+    // eslint-disable-next-line no-void
+    void this.readInBackground(backgroundStream.getReader());
+
+    const [errorReader, resultReader] = resultStream.tee().map((stream) => stream.getReader());
 
     /**
      * If the node threw an error, read it and throw it to the user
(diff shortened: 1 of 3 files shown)


packages/account/src/providers/fuel-graphql-subscriber.ts
petertonysmith94, line 1: Could we not delay the subscription until the point of a user reading the stream?

Moving [this](https://github.com/FuelLabs/fuels-ts/pull/3695/files#diff-9ede913edf9204f25a9b2444a220e8f638d39ac84e082d234b2c981a4f3dd792R22-R32) into the [`next()`](https://github.com/FuelLabs/fuels-ts/pull/3695/files#diff-9ede913edf9204f25a9b2444a220e8f638d39ac84e082d234b2c981a4f3dd792R74) iterator.

nedsalk: That would mean that doing this
```ts
const mySub = await this.operations.submitAndAwaitStatus(encodedTx);
```
wouldn't _actually_ submit the transaction, which would be unexpected.

petertonysmith94: Right you are - my mistake.

petertonysmith94 (approved):

Torres-ssf (approved):

danielbate (approved):

PR #3697, opened 2025-02-11
chore!: made `ResourceCache` consider resource owner

### LINKED ISSUE #3696: Make `ResourceCache` Owner-Aware
Because of [updates](https://github.com/FuelLabs/fuel-core/pull/2463) in `fuel-core 0.41`, the [coinsToSpend](https://github.com/FuelLabs/fuels-ts/blob/6eb113c64b25bc4b392bb64160f48d6e57f9b750/packages/account/src/providers/provider.ts#L1508) query now limits how many resources can be ignored when fetching resources. This limit is set by the `max_inputs` value in the chain configuration.

As a result, we need to update the resource cache logic on the `ResourceCache` class. Currently, the cache does not track which owner a resource belongs to; it combines UTXO IDs and Message nonces from different owners. 

However, since the `coinsToSpend` query is meant for a single owner, the cache needs to be changed to handle ownership. This way, when calling `coinsToSpend`, we will be able to exclude resources that belong to one specific user, making it less likely to exceed the `max_inputs` limit.

Torres-ssf:
- Closes #3696

# Summary

This PR addresses 3 things:

1 - Resources IDs are now cached considering their owner.

2 - A check has been added to prevent `Provider.getResourcesToSpend` from being called with more cached resource IDs than the `max_inputs` value.

3 - All resource cache-related tests from `provider.test.ts` have been moved to `resource-cache.test.ts` to reduce the number of test cases in the original file, which was already too large.

# Breaking Changes

```ts
//before
provider.cache?.getActiveData()
provider.cache?.isCached(key);
```

```ts
//after
const owner = wallet.address.toB256();

provider.cache?.getActiveData(owner)
provider.cache?.isCached(owner, key);
```

# Checklist


packages/account/src/providers/utils/helpers.ts
```diff
@@ -0,0 +1,29 @@
...
+  const { userInput, cached, maxInputs } = params;
+
+  const final = { ...userInput };
+
+  let total = final.utxos.length + final.messages.length;
+  if (total >= maxInputs) {
+    return final;
+  }
+
+  final.utxos = [...final.utxos, ...cached.utxos.slice(0, maxInputs - total)];
```
petertonysmith94, line 20:
```suggestion
  final.utxos = [...final.utxos, ...cached.utxos.reverse().slice(0, maxInputs - total)];
```

The active cache will return in order of the expiry.

| Tx ID | Number of UTXOs | Notes                       |
| ----- | --------------- | --------------------------- |
| 0x01  | 100             | (first transaction submitted)            |
| 0x02  | 100             |                             |
| 0x03  | 100             | <--- We remove these resources from the exclude first |

I am assuming that the earlier the transaction (`0x01`), the higher the likelihood of the transaction being processed by the time we fetch UTXOs/messages (less reason to exclude these resources). Would it be beneficial to reverse the order of excluding? 

| Tx ID | Number of UTXOs | Notes                       |
| ----- | --------------- | --------------------------- |
| 0x01  | 100             | <---  We want to remove these from the exclude first           |
| 0x02  | 100             |                             |
| 0x03  | 100             |  |

packages/account/src/providers/provider.ts
```diff
@@ -1511,15 +1500,25 @@ export default class Provider {
...
 
     if (this.cache) {
-      const cached = this.cache.getActiveData();
-      excludeInput.messages.push(...cached.messages);
-      excludeInput.utxos.push(...cached.utxos);
+      const cached = this.cache.getActiveData(ownerAddress.toB256());
+      if (cached.utxos.length || cached.messages.length) {
+        const {
+          consensusParameters: {
+            txParameters: { maxInputs },
```
petertonysmith94, line 1513: Have we questions the client team about using the `maxInputs` as the configurable for the maximum number of `excludedIds`?

This could become an issue when attempting to consolidate UTXOs.

arboleya: cc @nedsalk

packages/account/src/providers/utils/helpers.ts
Torres-ssf: https://github.com/FuelLabs/fuels-ts/pull/3697/commits/27b02765a77d6cabba134afe5ffe4f71fb29bb64

arboleya (commented):
[Lintfix](https://github.com/FuelLabs/fuels-ts/actions/runs/13397027733/job/37418533829?pr=3697) needed.

packages/account/src/providers/resource-cache.ts
```diff
@@ -26,54 +32,99 @@ export class ResourceCache {
...
+
+    activeData.utxos.reverse();
+    activeData.messages.reverse();
+
+    return activeData;
   }
 
-  // Check if a UTXO ID or message nonce is already cached and not expired
-  isCached(key: string): boolean {
+  isCached(owner: string, key: string): boolean {
```
arboleya, line 71: In the PR description, I thought this method accepted an object.

Perhaps removing the `:string` type notation can clear up this impression.


```ts
// before
provider.cache?.getActiveData();
provider.cache?.isCached(key);
```


```ts
// after
const owner = wallet.address.toB256();

provider.cache?.getActiveData(owner);
provider.cache?.isCached(key, owner);
```

arboleya (approved):

danielbate (approved):

PR #3704, opened 2025-02-12
fix: contract operations in transaction summary

### LINKED ISSUE #2530: Fix transaction summary with contract calls
The [current implementation](https://github.com/FuelLabs/fuels-ts/blob/e165e37307d65deda99f675723ea26470bf807d8/packages/account/src/providers/transaction-summary/call.ts#L14) of the transaction summary uses the `param1` property from a contract call receipt to build out the contract element of the summary.

As [recently deduced](https://github.com/FuelLabs/fuels-ts/discussions/2424), `param1` no longer represents the encoded function selector. Likewise, `param2` does not represent the encoded function arguments. They now represent pointers to their respective data.

Therefore currently the only way to deduce the function selector from a transaction is to decode it from the `scriptData` that was passed on the transaction, as documented [here](https://github.com/FuelLabs/fuels-ts/discussions/2424#discussioncomment-9766250). This is because `scriptData`, when used with contracts, represents the entire [contract call script](https://github.com/FuelLabs/fuels-ts/blob/e165e37307d65deda99f675723ea26470bf807d8/packages/program/src/contract-call-script.ts#L202).

Therefore we need to:
- Fix the transaction summary to properly display contract call info
- Add more test cases for contract calls in the transaction summary
- Check all references to `param1` in the SDK and confirm they are not being misused

> Reference:
> - https://github.com/FuelLabs/fuels-ts/discussions/2424

### LINKED ISSUE #2531: Implement contract call script decoding
Consumers of the SDK may want to extract the `scriptData` from a transaction to see function / arguments / asset / forward amount / gas values by decoding the contract call script. An advanced user that knows our [call script](https://github.com/FuelLabs/fuels-ts/blob/e165e37307d65deda99f675723ea26470bf807d8/packages/program/src/contract-call-script.ts#L202) and the encoding used could reverse engineer this themselves. However we should provide a utility to assist with this to make this information more easily available. It _could_ be used in the transaction summary, as expressed in #2530.

> Reference:
> - https://github.com/FuelLabs/fuels-ts/discussions/2424

### LINKED ISSUE #3703: `getTransactionSummary` throws when passed ABI map
Consider the following use case:

```ts
    const response = await contract.functions.assert_u8(100, 100).call();
    await response.waitForResult();

    const summary = await getTransactionSummary({
      id: response.transactionId,
      provider,
      abiMap: {
        [contract.id.toB256()]: CallTestContract.abi,
      },
    });
```

When passed an abi map, used to build out the contract call operations, this function throws:

```sh
FuelError: function 0x00000000000028f0 not found: undefined.
```

> This is related to the `param1` issue, as mentioned in #2530.

danielbate:
- Closes #3703 
- Closes #2531 
- Closes #2530

# Release notes

In this release, we:

- Introduced `decodeScriptData` to decode contract operations from a call receipt
- Fixed contract operations in transaction summaries

# Summary

- Uses the raw payload for a transaction to obtain it's script data, and uses that to decode the contract operations rather than using `param1`/`param2` in the call receipt
- Introduces the `decodeScriptData` function to directly decode the bytes from a contract call script

# Checklist


packages/abi-coder/src/utils/scriptData.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+import type { JsonAbi } from '../types/JsonAbiNew';
+
+import { decodeScriptData } from './scriptData';
+
+/**
+ * @group node
+ * @group browser
+ */
+describe('decodeScriptData', () => {
```
nedsalk, line 12: We'll have to change these tests whenever the `scriptData` encoding or abi structure change in a breaking manner, thus IMO it's better for maintenance to remove these tests and add more in the `fuel-gauge` [test](https://github.com/FuelLabs/fuels-ts/pull/3704/files#diff-9761807ab226a5228bc773ccfef74e9b6cf3acc0a1f44c23f395217b2b3829ddR305) if the existing tests don't cover it. From personal experience, it was always a pain to have to copy/paste new ABIs over old ABIs to get a test working while an equivalent higher-level test testing asserting the same results was succeeding.

Torres-ssf (approved):

packages/account/src/providers/transaction-summary/call.ts
```diff
@@ -17,21 +21,35 @@ export interface FunctionCall {
...
+  receipt,
+  rawPayload,
+}: GetFunctionCallProps): FunctionCall => {
+  const [transaction] = new TransactionCoder().decode(arrayify(rawPayload), 0);
+
+  if (!transaction.scriptData) {
+    throw new FuelError(
+      ErrorCode.NOT_SUPPORTED,
+      'Cannot get function calls for this transaction type.'
+    );
```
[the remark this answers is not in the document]
danielbate: I did think about this 🤔 

Decided to go with the `throw` because `getTransactionSummary` has wrapped `getFunctionCall` in a `isTypeScript`. So consumers should only be affected by this is they are calling `getFunctionCall` directly. Which I'd be surprised, neither the wallet or connectors do this.

packages/abi-coder/src/utils/scriptData.test.ts
danielbate: Turns out this function actually isn't exported from `fuels` which was my mistake, so I've removed the test. But I will add some more high level tests anyway.

packages/fuel-gauge/src/transaction-summary.test.ts
```diff
@@ -302,6 +302,175 @@ describe('TransactionSummary', () => {
...
+    expect(callOperation.calls?.[0].argumentsProvided).toStrictEqual({
+      recipient: { bits: recipient.address.toB256() },
+      asset_id: { bits: assetId },
+      amount: bn(1000).toHex(),
+    });
+  });
+
+  // Tx summary with multicall does not set contract operations correctly
+  // https://github.com/FuelLabs/fuels-ts/issues/3706
+  it.skip('should ensure getTransactionsSummaries executes just fine [w/ ABI & multicall]', async () => {
```
danielbate, line 401: [Follow up issue](https://github.com/FuelLabs/fuels-ts/issues/3706) added for multi call.

packages/fuel-gauge/src/transaction-summary.test.ts
nedsalk, line 1: All the added tests test the `TransactionResponse.getTransactionSummary` method.  Should we add tests for the [`getTransactionSummary`](https://github.com/FuelLabs/fuels-ts/blob/6d3cafbdfb3b7eb5e6fc4e663e4f1713a2109b8b/packages/account/src/providers/transaction-summary/get-transaction-summary.ts#L27) function? They both use the `assembleTransactionSummary` under the hood which, in the end, calls the `getFunctionCall` function that this PR touched. The difference is that `getTransactionSummary` calls the `provider.operations.getTransactionWithReceipts` method, while this one uses the subscription results.

arboleya (approved):

Torres-ssf (dismissed):

packages/fuel-gauge/src/transaction-summary.test.ts
danielbate: [Tests added](https://github.com/FuelLabs/fuels-ts/pull/3704/commits/a7a9b66160be8a8d739edbfe907585dff9a8f904), but I did find a problem with comparing the contract input, which will be addressed in https://github.com/FuelLabs/fuels-ts/issues/3708

Torres-ssf (approved):

nedsalk (approved):

Issue #3669, opened 2025-02-06
Improve SDK and Wallet Communication


danielbate:
## Problem

When a transaction is passed from an application to the wallet, the wallet performs estimation and funding network requests that have already been performed by the SDK. This can be visualised in the following diagram:

<img width="803" alt="Image" src="https://github.com/user-attachments/assets/1d524441-084b-44a1-8cce-f75384029b1f" />

This is because the wallet has no knowledge of the state of a transaction, so attempts to account for anything that *could* be missing.

## Solution

We should improve handoff from the SDK to connectors/wallets so that they can act accordingly, and not introduce any redundancy in the transaction flow.

Currently, the context that the SDK provides to the wallet includes:

```ts
interface SendTransactionJsonRpc {
  address: string,
  transaction: TransactionRequest,
  provider: {
    url: string,
  },
}
```

This should be extended to:

```ts
  interface SendTransactionJsonRpc {
    address: string,
    transaction: TransactionRequest,
    provider: {
      url: string,
      maxInputs: number, // required for estimatiom
      estimatedGasPrice: number, // required for estimation, so the wallet does not need to re-fetch this value
    },
    isEstimated: boolean, // mitigate any redundant dry-runs
    isFunded: boolean, // mitigate any redundant 
    isSigned: boolean, // tx should not be augmented, invalidating any signatures present i.e. gasless paymaster
    receipts: Receipt[] // reciepts to build out the final tx summary on the approval screen
  }
```

With the following additional information, the wallet can act accordingly and should not perform any redundant network requests.

> [!NOTE]  
> The specific work that needs to be done in the SDK, Connectors and Wallet is still being worked out but will be added to this issue

danielbate:
**Phase 1 - Estimation and Funding**

Phase 1 aims to give more context to the wallet as to whether it is needs to perform the estimation and funding stages.

This includes the addition of the `isEstimated` , `isFunded` and `isSigned` flags to the Json RPC object:

```ts
 interface SendTransactionJsonRpc {
    address: string,
    transaction: TransactionRequest,
    provider: {
      url: string,
    },
    isEstimated: boolean,
    isFunded: boolean,
    isSigned: boolean,
  }
```

For the implementation, we will pass an object throughout the transaction preparation flow:

`{ isEstimated: false, isFunded: false, isSigned: false }` 

When an operation is performed that sets one of these flags, we will store the transaction hash against the flag:

[...]

PR #3709, opened 2025-02-14
feat: added the state to the transaction request

petertonysmith94:
- Related to #3669 

# Summary

- Introduce the state field on a transaction request ('funded', 'signed') to enable underlying connectors to skip certain steps based upon this state to reduce HTTP requests.

```ts
 interface SendTransactionJsonRpc {
    state: undefined | 'funded' | 'immutable',
  }
```

| State       | Description                                                                                       |
| ----------- | ------------------------------------------------------------------------------------------------- |
| `undefined` | The transaction has not been estimated or funded.                                                 |
| `funded`  | The transaction is estimated and funded, the underlying wallet can skip this process.             |
| `signed`  | The transaction is signed (therefore, estimated and funded). This is the same as `skipCustomFee`. |

# Checklist


packages/account/src/providers/transaction-request/script-transaction-request.ts
```diff
@@ -87,6 +87,8 @@ export class ScriptTransactionRequest extends BaseTransactionRequest {
 
     await account.fund(this, txCost);
 
+    this.updateFlags({ isEstimated: true, isFunded: true });
```
arboleya, line 90: Is there a case where these two can exist in isolation, and if not, should we merge them into a single flag?

packages/account/src/providers/transaction-request/transaction-request.ts
```diff
@@ -706,4 +723,29 @@ export abstract class BaseTransactionRequest implements BaseTransactionRequestLi
...
+
+  /**
+   * @hidden
+   *
+   * Used internally to update the status flags of a transaction request.
+   *
+   * @param flags - The flags to update.
+   */
+  public updateFlags(flags: Partial<{ [keys in keyof TransactionStatusFlags]: boolean }>) {
+    const CHAIN_ID = 0;
```
arboleya, line 735: Is this supposed to be hardcoded?

packages/account/src/providers/transaction-request/script-transaction-request.ts
petertonysmith94: My initial solution was to have the `isEstimate` triggered using an `estimate` method, then the `isFunded` on the `fund` method.

@danielbate suggested that we only update in the `estimateAndFund` method for now as this is the primary suggested entry point for estimate + funding.

Given that users can estimate and fund with other avenues, I believe keeping these flags separate gives us more flexibility to enhance this feature later without external changes.

packages/account/src/providers/transaction-request/transaction-request.ts
petertonysmith94: Yes, these flags will only serve as a checksum hash to ensure the transaction body hasn't changed between steps.

By hardcoding the `CHAIN_ID`, we avoid having to either pass through a chain ID or making the async.

packages/account/src/providers/transaction-request/script-transaction-request.ts
arboleya: How do we expect the Wallet to behave in these two scenarios?

```ts
// 1
this.updateFlags({ isEstimated: true, isFunded: false });

// 2
this.updateFlags({ isEstimated: false, isFunded: true });
```

packages/account/src/providers/transaction-request/transaction-request.ts
arboleya: I may be missing the point or how a corner case like this could happen, but this seems like it could lead to future problems. Was this agreed upon with @FuelLabs/frontend?

packages/account/src/providers/transaction-request/script-transaction-request.ts
danielbate: `{ isEstimated: true, isFunded: false`} - wallet does not dry-run the transaction on page load. It also expects the `gasLimit` and `maxFee` to be set correctly. Based on this, the wallet will fetch resources and add them to the request.

`{ isEstimated: false, isFunded: true }` - wallet increases `gasLimit` by 20% and dry-runs the transaction. It assumes funding is correct but IF the prior change has increased the `maxFee`, we would need to re-fund.

As @nedsalk suggested in sync, I'm wondering whether a status enum would be more flexible. This would make it more resistant to any changes to dry-run that could allow it to also introduce funding.

Torres-ssf: I'm missing something here as well. Why is the `gasLimit` increased by 20% for a dry run when it was never accurately estimated in the first place?

I'm trying to understand if this is meant to address a specific corner case.

packages/account/src/providers/transaction-request/transaction-request.ts
Torres-ssf: I agree with @arboleya. Maybe I'm missing something, but since all methods calling `updateFlags` are async, what’s the benefit of making this method synchronous?

Also, `provider.getChain()` will usually resolve immediately since it relies on cached data.

danielbate: This is to produce a hash used internally in the SDK, which results in a boolean value that is passed to the FE. This hash is not transmitted via JSON RPC.

I can appreciate the confusion though of different tx hashes, internal or not. And it would be better to use the actual chain ID.

petertonysmith94: By storing the hash, it enables the detection of changes like the following:

```ts
const request = new ScriptTransactionRequest({ });
request.estimateAndFund(wallet);

// Sub-sequential changes to the request wo/ estimating or funding can be detected
request.addResources(...)
```

The consumer of these flags (the wallet), can choose whether they use these flags or not:

```ts
// Simple if statement
if (!request.flags.isEstimated)

// Or checking the request hash
const requestHash = request.getTransactionId(0);
if (request.flags.isEstimated !== requestHash)
```

Happy to change these flags to simple boolean values, but I believe having a means of detecting changes post estimate and fund would benefit this process.

There was no agreement with the @FuelLabs/frontend team, but it is in the proposed [solution](https://github.com/FuelLabs/fuels-ts/issues/3669#issuecomment-2657037422).

packages/account/src/providers/transaction-request/script-transaction-request.ts
danielbate: Apologies I wanted to think about this a little more - a transaction could not be funded and not estimated. We're going to rethink approach a little and will be making some changes to the spec.

packages/account/src/providers/transaction-request/transaction-request.ts
danielbate: I can appreciate the confusing with using a different chain ID, I think it makes sense to use the actual value.

packages/account/src/providers/transaction-request/script-transaction-request.ts
petertonysmith94: I've implemented the changes based on the [re-spec](https://github.com/FuelLabs/fuels-ts/issues/3669#issuecomment-2671670923), so am resolving this conversation.

packages/account/src/providers/transaction-request/transaction-request.ts
petertonysmith94: I've implemented the changes based on the [re-spec](https://github.com/FuelLabs/fuels-ts/issues/3669#issuecomment-2671670923), so am resolving this conversation.

Issue #3708, opened 2025-02-14
Transaction ID not correctly set on Contract input


danielbate:
When we create a new contract input against a transaction, we always zero out the transaction ID.

https://github.com/FuelLabs/fuels-ts/blob/02503a68f402f1984487f4bddc5886adc67c5c0d/packages/account/src/providers/transaction-request/input.ts#L117

This has not been a problem because this value is meant to be zeroed out for hashing/signing.

However when building out a transaction summary from a transaction response (`transactionResponse.getTransactionSummary()`) the summary is technically not correct as this input has not been populated correctly. Whereas a summary generated using a transaction ID (`getTransactionSummary(txId)`) fetches the full tx from the node, has this input set correctly.

Torres-ssf:
I believe this happens because we no longer [fetch the transaction](https://github.com/FuelLabs/fuels-ts/blob/70bff60371f70a165c7556e25454ef02eddf139b/packages/account/src/providers/transaction-response/transaction-response.ts#L219-L226) within `TransactionResponse` when a transaction request is [given](https://github.com/FuelLabs/fuels-ts/blob/70bff60371f70a165c7556e25454ef02eddf139b/packages/account/src/providers/transaction-response/transaction-response.ts#L157) during its instantiation.

Instead, we [update specific parts of it](https://github.com/FuelLabs/fuels-ts/blob/70bff60371f70a165c7556e25454ef02eddf139b/packages/account/src/providers/transaction-response/transaction-response.ts#L181) using data from subscription result.

> cc @nedsalk

danielbate:
Once this has been completed, we should enable the tests in #3710.

Torres-ssf:
@arboleya This issue was resolved in [#3747](https://github.com/FuelLabs/fuels-ts/pull/3747).

Issue #2481, opened 2024-06-07
Review exports from all packages


nedsalk:
Decide what to actually continue exporting and what to stop.

PR #3710, opened 2025-02-14
fix: multi-call contract operations in tx summary

### LINKED ISSUE #3706: Assembling a transaction summary using multicall incorrectly sets contract operations
Follow on from #2530, assembling a transaction summary from a multicall is not building the operations correctly. Whilst it no longer throws, the second operation uses the call from the first operation. 

I had proposed a solution using the dereferenced `param1` as mentioned [here](https://github.com/FuelLabs/fuels-ts/issues/2530#issuecomment-2656552871). But I can't work out how to correctly set the base offset. I don't think it's possible and instead we need the work from the on going discussions around repurposing `param1`.

danielbate:
- Closes #3706 

# Release notes

In this release, we:

- Populated transaction summaries with all contract operations

# Summary

Builds on the work in #3704 but makes correct usage of the `param1` and `param2` contract call receipt properties by using them in conjunction with a calculated script data size, enabling multi-call contract operations.

# Checklist

### DIFF
--- a/packages/account/src/providers/transaction-summary/operations.ts
+++ b/packages/account/src/providers/transaction-summary/operations.ts
@@ -428,27 +428,32 @@ export function getContractCallOperations({
       (receipt) => receipt.to === contractInput.contractID
     );
 
-    const callScriptBaseOffset =
-      SCRIPT_FIXED_SIZE +
-      calculateScriptVariableSize(callReceiptsForContract) +
-      // Potential weakness here for legacy transactions, if the script data was generated
-      // with a different max inputs value, the offset will be incorrect.
-      calculateVmTxMemory({ maxInputs: maxInputs.toNumber() });
+    // Potential weakness here for legacy transactions, if the script data was generated
+    // with a different max inputs value, the offset will be incorrect and therefore decoding
+    // will fail. In this scenario we won't return the operations.
+    try {
+      const callScriptBaseOffset =
+        SCRIPT_FIXED_SIZE +
+        calculateScriptVariableSize(callReceiptsForContract) +
+        calculateVmTxMemory({ maxInputs: maxInputs.toNumber() });
 
-    const operations = callReceiptsForContract.flatMap((receipt) =>
-      processCallReceipt(
-        receipt,
-        contractInput,
-        inputs,
-        abiMap,
-        rawPayload as string,
-        maxInputs,
-        baseAssetId,
-        callScriptBaseOffset
-      )
-    );
+      const operations = callReceiptsForContract.flatMap((receipt) =>
+        processCallReceipt(
+          receipt,
+          contractInput,
+          inputs,
+          abiMap,
+          rawPayload as string,
+          maxInputs,
+          baseAssetId,
+          callScriptBaseOffset
+        )
+      );
 
-    return operations;
+      return operations;
+    } catch (error) {
+      return [];
+    }
   });
 }
 
--- a/packages/account/src/providers/transaction-summary/operations.ts
+++ b/packages/account/src/providers/transaction-summary/operations.ts
@@ -361,32 +361,39 @@ function processCallReceipt(
 }
 
 /**
- * Calculates the size of the contract call script based off
+ * Calculates the size of the contract call script based off.
+ *
+ * This is a hardcoded implementation of the contract call script size calculation, validated
+ * by an ASM implemenation of the same logic in `operations.test.ts`.
  *
  * @param calls - The contract call receipts to calculate the size of.
  * @returns The size of the contract call script.
  */
-const calculateScriptVariableSize = (calls: ReceiptCall[]): number => {
+export const calculateScriptVariableSize = (calls: ReceiptCall[]): number => {
   // Calculate the length of the call script for each call and sum
   const offset = calls.reduce(
-    (total, call) => {
+    (total) => {
       let callOffset = total;
-
-      callOffset += asm.movi(0x10, 0).to_bytes().byteLength;
-      callOffset += asm.movi(0x11, 0).to_bytes().byteLength;
-      callOffset += asm.lw(0x11, 0x11, 0).to_bytes().byteLength;
-      callOffset += asm.movi(0x12, 0).to_bytes().byteLength;
-
-      const gasOffset = call.gas.toNumber()
-        ? asm.movi(0x13, call.gas.toNumber()).to_bytes().byteLength
-        : asm.call(0x10, 0x11, 0x12, asm.RegId.cgas().to_u8()).to_bytes().byteLength;
-
-      callOffset += gasOffset;
+      // Call Data Offset - asm.movi(0x10, 0)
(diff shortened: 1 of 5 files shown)


packages/fuel-gauge/src/transaction-summary.test.ts
```diff
@@ -340,6 +346,112 @@ describe('TransactionSummary', () => {
...
+    } = launched;
+
+    const baseAssetId = await provider.getBaseAssetId();
+    const contractId = contract.id.toB256();
+
+    const call = await contract.functions
+      .return_context_gas()
+      .callParams({
+        forward: [bn(100_000), baseAssetId],
+      })
```
petertonysmith94, line 433: See comment above:
-

packages/fuel-gauge/src/transaction-summary.test.ts
```diff
@@ -594,6 +704,72 @@ describe('TransactionSummary', () => {
...
+
+    const responseSummary = await res.transactionResponse.getTransactionSummary({
+      [contractId]: TokenContract.abi,
+    });
+
+    expect(summary.operations).toStrictEqual(responseSummary.operations);
+
+    // TODO: Contract txId not set correctly in`transactionResponse.getTransactionSummary`
+    // https://github.com/FuelLabs/fuels-ts/issues/3708
+    // expect(summary).toStrictEqual(responseSummary);
```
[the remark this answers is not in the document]
danielbate: https://github.com/FuelLabs/fuels-ts/issues/3708#issuecomment-2665114360

packages/account/src/providers/transaction-summary/operations.ts
```diff
@@ -342,6 +359,49 @@ function processCallReceipt(
...
+/**
+ * Calculates the size of the contract call script based off.
+ *
+ * This is a hardcoded implementation of the contract call script size calculation, validated
+ * by an ASM implemenation of the same logic in `operations.test.ts`.
+ *
+ * @param calls - The contract call receipts to calculate the size of.
+ * @returns The size of the contract call script.
+ */
+export const calculateScriptVariableSize = (calls: ReceiptCall[]): number => {
```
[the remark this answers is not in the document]
danielbate: I exported it for the test, we currently export everything from `operations`. IMO this we should re-open #2481 and include that here.

packages/account/src/providers/transaction-summary/operations.ts
```diff
@@ -342,6 +359,49 @@ function processCallReceipt(
...
+      callOffset += amountOffset.byteLength;
+      // Load Asset ID- asm.lw(0x11, 0x11, 0)
+      const assetIdOffset = new Uint8Array([93, 69, 16, 0]);
+      callOffset += assetIdOffset.byteLength;
+      // Asset ID - asm.movi(0x12, 0)
+      const loadBytes = new Uint8Array([114, 72, 0, 0]);
+      callOffset += loadBytes.byteLength;
+      // Gas Offset - asm.call(0x10, 0x11, 0x12, asm.RegId.cgas().to_u8())
+      const gasOffset = new Uint8Array([45, 65, 20, 138]);
+      callOffset += gasOffset.byteLength;
```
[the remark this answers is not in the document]
danielbate: Correct, so I for a single call it is working as expected, as that it was that test case is asserting. 

However, I've just checked multi-call and it is throwing. We need to deref the gas offset in the same way we do for `param1` etc.

Given it's failing gracefully and not throwing, I've added a [follow-up](https://github.com/FuelLabs/fuels-ts/issues/3716).

packages/fuel-gauge/src/transaction-summary.test.ts
danielbate: https://github.com/FuelLabs/fuels-ts/pull/3710#discussion_r1959448827

packages/account/src/providers/transaction-summary/operations.ts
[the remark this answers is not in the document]
petertonysmith94: I'm happy with this approach.

IMO, the operations should be a sub-package of the account package (`@fuel-ts/account/operations`) and not exported directly by `fuels`. Not sure how possible that is, though.

petertonysmith94 (approved):

packages/account/src/providers/transaction-summary/operations.ts
```diff
@@ -364,19 +424,36 @@ export function getContractCallOperations({
...
+    );
+
+    // Potential weakness here for legacy transactions, if the script data was generated
+    // with a different max inputs value, the offset will be incorrect and therefore decoding
+    // will fail. In this scenario we won't return the operations.
+    try {
+      const callScriptBaseOffset =
+        SCRIPT_FIXED_SIZE +
+        calculateScriptVariableSize(callReceiptsForContract) +
+        calculateVmTxMemory({ maxInputs: maxInputs.toNumber() });
```
arboleya, line 438: IIRC, judging by the answers to your questions, we can't rely on max inputs, can't we?

danielbate: No we can't. But I think there is value as for a tx summary created from a tx response, where it is likely that the max inputs is the same, we will get a more complete summary. It is also failing gracefully (empty array for operations).

For legacy transactions where it is possible that submitted max inputs !== current max inputs, I've got an idea whereby we could query the block header that the tx was included in, and get the consensus param version and check that against the one we want to use for the calculation. This would need another request though.

Additionally, there is the [tracer](https://github.com/FuelLabs/execution-trace/tree/master) but that will also require an additional request.

arboleya: Should we postpone this feature until it is 100% reliable (via the tracer or similar)?

IIRC, one of the use cases for this was to show stuff on the Explorer, correct?

Which, given the circumstances, will not happen anyway.

danielbate: Max inputs has not changed since we launched main-net. Right now, if we merged this feature, the explorer could display all contract operations for all transactions without making any additional requests.

Right now, we don't have any call operations in the tx summary. If max inputs changes, this feature mirrors that outcome for legacy txs. So IMO there is little downside vs where we are right now.

I'll leave it at that, I think there's value, but I'll leave it to you and the team if there isn't agreement and I'll convert to draft 👍🏻

arboleya: My point is that there's no guarantee it won't change, right? This is the same caveat from @Torres-ssf's presentation on why we can't hardcode the `maxInputs` value and save one extra request. The same rationale was applied to the [RPC Consistency](https://github.com/FuelLabs/fuels-ts/pull/3462#discussion_r1958726753) pull request, accounting for variations and "what if it changes" cases.

danielbate: There is no guarantee. 

I'll get an issue written up to investigate/introduce tracing as that'll likely be the most robust, and will close this PR with that issue.

danielbate:
Superseded by #3734

PR #3741, opened 2025-02-27
chore: support optimistic concurrency control

nedsalk:
# Release notes

In this release, we:

- Did this and that 

# Summary

# Breaking Changes

# Checklist


apps/demo-wallet-sdk-react/next-env.d.ts
```diff
@@ -2,4 +2,4 @@
 /// <reference types="next/image-types/global" />
 
 // NOTE: This file should not be edited
-// see https://nextjs.org/docs/app/building-your-application/configuring/typescript for more information.
+// see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
```
nedsalk, line 5: This got changed automatically.

packages/account/src/providers/provider.ts
```diff
@@ -495,10 +514,87 @@ export default class Provider {
         fullRequest = await options.requestMiddleware(fullRequest);
       }
 
-      return options.fetch ? options.fetch(url, fullRequest, options) : fetch(url, fullRequest);
+      const operationName = fullRequest.body
+        ?.toString()
+        .match(/"operationName":"(.+)"/)?.[1] as keyof SdkOperations;
```
Torres-ssf, line 519: Can we improve readability here by making this a helper?

```ts
  private static extractOperationName(body: string | null | undefined): keyof SdkOperations {
    return body?.match(/"operationName":"(.+)"/)?.[1] as keyof SdkOperations;
  }
```

packages/account/src/providers/provider.ts
```diff
@@ -495,10 +514,87 @@ export default class Provider {
...
+      let extensions: {
+        current_fuel_block_height?: number;
+        fuel_block_height_precondition_failed: boolean;
+      };
+
+      if (url.endsWith('-sub')) {
+        const reader = responseClone.body?.getReader() as ReadableStreamDefaultReader<Uint8Array>;
+        const { event } = await FuelGraphqlSubscriber.readEvent(reader);
+
+        extensions = event?.extensions as typeof extensions;
```
Torres-ssf, line 562: Here we are reading the first stream update? Just trying to understand this code.

packages/account/src/providers/provider.ts
```diff
@@ -495,10 +514,87 @@ export default class Provider {
...
+      } else {
+        extensions = (await responseClone.json()).extensions;
+      }
+
+      Provider.setCurrentBlockHeight(
+        url.replace(/-sub$/, ''),
+        extensions?.current_fuel_block_height
+      );
```
Torres-ssf, line 570: Can `Provider.setCurrentBlockHeight` handle the url replace?

packages/account/src/providers/provider.ts
```diff
@@ -495,10 +514,87 @@ export default class Provider {
...
+        ++retryAttempt;
+        const sleepTime = getWaitDelay(retryOptions, retryAttempt);
+        await sleep(sleepTime);
+      } else {
+        break;
+      }
+    } while (blockHeightPreconditionFailed);
+
+    return response;
+  }
```
Torres-ssf, line 584: Maybe I am missing something here, but why not use the variable to control the loop to consider `maxRetries` instead of explicitly breaking the loop?

packages/account/src/providers/provider.ts
nedsalk: The code here deals only with the initial event which also carries the `{ fuel_block_height_precondition_failed }` property. All subsequent events are read via [this `onEvent` callback](https://github.com/FuelLabs/fuels-ts/pull/3741/files#diff-69a6670d0426c74935ed2830ce552e7df2379c2d112a582306e57569fcd17029R827).

packages/account/src/providers/provider.ts
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/3741/commits/1d16d3ad5e156121f91581a7b4060876e54df41b

packages/account/src/providers/provider.ts
nedsalk: You're right that it could be improved. I have refactored the logic dealing with block height into a method in https://github.com/FuelLabs/fuels-ts/pull/3741/commits/4b7c01fc5a05dc4d6c5598df9064d91dd9667685. With this, IMO there's no need to refactor the `operationName` extraction further as the whole code block in the new method is cohesive and straightforward.

packages/account/src/providers/provider.ts
nedsalk: https://github.com/FuelLabs/fuels-ts/pull/3741/commits/cac250db6ca42f70e62435f98f919dca78bcddce

packages/fuels/test/features/deploy.test.ts
nedsalk, line 1: For some reason, the `Provider` static instance isn't the same between the one available in the tests that's initialized in `launchTestNode` and the one that's used for deployment via `runDeploy`; perhaps it has something to do with `comandeer`. 

The problem is that both of them live simultaneously across all tests in this file and what would happen is the first test would finish successfully, kill the node and clear the `currentBlockHeight` cache of its static instance... but it wouldn't clear the other `Provider`'s `currentBlockHeight` which would remain high (e.g. `15`). In the next test, `runDeploy` would try to deploy to a new `fuel-core` node initialized for that test, but that node would have a block height of `0` and it would stall as the `required_fuel_block_height` isn't reached. Our custom awaiting mechanism isn't working here because we're running the node with `--poa-instant true` so we're only creating blocks when submitting transactions, but we can't submit a transaction because the block height is too low.

In the end, I opted to rewrite the tests to not use the same `4000` port across all tests, which resolves us of the problem. Because `runDeploy` is our test utility, it won't affect the users. For users who want to run all their `launchTestNode` tests on the same port as was done in this test, [this](https://github.com/FuelLabs/fuels-ts/pull/3741/files#diff-3c0413287e9234f6662e455d68cdf5c2daf63406cd32c3b392f53c6fd5cd525fR98) change will suffice.

petertonysmith94 (commented):
First pass

packages/account/test/optimistic-concurrency-handling.test.ts
```diff
@@ -0,0 +1,438 @@
...
+    await expectToThrowFuelError(
+      () =>
+        provider.operations.getCoinsToSpend({
+          owner: wallet.address.toB256(),
+          queryPerAsset: { amount: '10', assetId: baseAssetId },
+        }),
+      new FuelError(
+        ErrorCode.INVALID_REQUEST,
+        `The required fuel block height is higher than the current block height. Required: ${expectedHeight}, Current: 0`
+      )
```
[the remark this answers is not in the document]
nedsalk: IMO it's straightforward and descriptive as-is. The request _is_ invalid, and the message explains it all.

packages/account/src/providers/provider.ts
```diff
@@ -729,6 +843,13 @@ export default class Provider {
           query,
           fetchFn: (url, requestInit) => fetchFn(url as string, requestInit, this.options),
           variables: vars,
+          operationName: opDefinition.name.value,
+          onEvent: (event) => {
+            Provider.setCurrentBlockHeight(
+              this.urlWithoutAuth,
+              event.extensions?.current_fuel_block_height as number
+            );
+          },
```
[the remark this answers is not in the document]
nedsalk: The code in `fetchAndProcessBlockHeight` deals only with the initial event which carries the `{ fuel_block_height_precondition_failed }` property that governs retry logic. The subsequent events also carry the `current_block_height` value in them, and we want to read that and update our state accordingly. We must do that from inside the `FuelGraphQLSubscriber`.

packages/account/src/providers/provider.ts
```diff
@@ -498,10 +528,93 @@ export default class Provider {
         fullRequest = await options.requestMiddleware(fullRequest);
       }
 
-      return options.fetch ? options.fetch(url, fullRequest, options) : fetch(url, fullRequest);
+      if (Provider.ENSURE_RPC_CONSISTENCY) {
+        Provider.applyBlockHeight(fullRequest, url);
+      }
+
+      return Provider.fetchAndProcessBlockHeight(url, fullRequest, options);
```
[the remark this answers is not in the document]
nedsalk: The `autoRetryFetch` function handles the retry strategy generally for _all_ requests and is governed via the `ProviderOptions.retryOptions` field. `fetchAndProcessBlockHeight` is specifically concerned with the RPC consistency problem. 

Because `fuel-core` nodes can be set up with a `--graphql-required-block-height-tolerance` of e.g. `0`, when we send a `required_fuel_block_height: 100` and the node's block height is `99`, the node will not hold our request until it gets the block `100` and will instead immediately return an error and have `extensions: { fuel_block_height_precondition_failed: true, current_block_height: 99 }` in the response. We will then await on our side based on our retry strategy and resend the request, hoping that the node has been synced with the network by then and has the block height we're requiring.

On the other hand, if the node was set up with the default `--graphql-required-block-height-tolerance 10`, the request above will be awaited on its side and the response it returns will have  `extensions: { fuel_block_height_precondition_failed: false, current_block_height: >= 100 }`. In this case, the node itself awaited and our retry logic wasn't run.

packages/account/test/optimistic-concurrency-handling.test.ts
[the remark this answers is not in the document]
arboleya: The message is clear, but the request is not necessarily invalid, right?

Referring to the `ErrorCode.INVALID_REQUEST` specifically — not sure if that's what @petertonysmith94 meant.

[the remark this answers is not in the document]
nedsalk: > The message is clear, but the request is not necessarily invalid, right?

That's true. I changed it to `ErrorCode.RPC_CONSISTENCY` in https://github.com/FuelLabs/fuels-ts/pull/3741/commits/a43306496412af09ae8d209b599847a135f1acce.

apps/docs/src/guide/provider/rpc-consistency.md
```diff
@@ -0,0 +1,8 @@
+# RPC Consistency
+
+Because the Fuel network consists of numerous nodes that at the same moment might be at different block heights, querying the state of the blockchain might yield different results from node to node. To fix that, every node appends its current block height to every response it returns. This enables the SDK to then append this block height to block-sensitive requests that it makes towards the network, thereby requiring the node it hits to be at that block height before processing the request.
+If the node is at a lower block height, the request will either be awaited on the node's side if the node is configured to do so, or it'll be retried from the SDK side for a couple of times with an exponential backoff. Once the maximum retry amount is reached, the SDK will throw an error.
+
+This functionality is turned on by default. If you want to turn it off to e.g. have faster response times, you can do the following:
```
danielbate, line 6:
```suggestion
A common issue with querying distributed networks is ensuring each node is consistent. At any moment each node could be at a different block height, therefore querying the state of the blockchain may yield different results from node to node.

To defend against this, the SDK appends the block height to block-sensitive requests. The node will then ensure it has met the block height criteria before it processes the request.

If the block height criteria has not been met by the node, the request will either be awaited on the node's side (if the node supports this), or it'll be retired from the SDK side. The SDK will attempt to retry the request until it hits a maximum amount of retries, and then will throw an error.

This functionality is enabled by default, but can be disabled as so:
```

Rewritten to break it up a bit and improve the flow

packages/account/src/providers/provider.ts
```diff
@@ -436,6 +436,13 @@ type SdkOperations = Omit<Operations, 'statusChange' | 'submitAndAwaitStatus'> &
   getBlobs: (variables: { blobIds: string[] }) => Promise<{ blob: { id: string } | null }[]>;
 };
 
+const BLOCK_HEIGHT_SENSITIVE_OPERATIONS: Array<keyof SdkOperations> = [
+  'submit',
+  'statusChange',
+  'getCoinsToSpend',
+  'submitAndAwaitStatus',
```
danielbate, line 443: Why aren't other queries block sensitive, like get transactions, coins, blocks? Surely these will all be potentialyl out of date?

arboleya: Discrepancies usually occur when a read operation is performed right after a write one.

apps/docs/src/guide/provider/rpc-consistency.md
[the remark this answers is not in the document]
nedsalk: I couldn't say for certain; it depends on the speed of block syncing and how much the node we hit lags behind.

arboleya (commented):
E2E tests are [failing](https://github.com/FuelLabs/fuels-ts/actions/runs/13942637373/job/39023267335?pr=3741#step:5:182). IIRC, pre-confs should not be in `0.42.0`.

Please, clarify what's going on, work around it, and report back.

We need this finalized, green, and out of the way ASAP.

packages/account/src/providers/provider.ts
Torres-ssf: Exactly. There's no need to consider every request, only those that could lead to discrepancies.

One that might be worth considering from that list is `getTransactionWithReceipts`, as it is used within `TransactionResponse`.

> cc @nedsalk

nedsalk: > only those that could lead to discrepancies

Almost every operation is block-sensitive. Users can manually call e.g. `provider.operations.getCoins` after a successful transaction execution and could come across this RPC consistency problem. Maybe instead of having a `BLOCK_HEIGHT_SENSITIVE_OPERATIONS` array, we should perhaps have a `BLOCK_HEIGHT_INSENSITIVE_OPERATIONS` instead? That list would be a lot smaller.

> One that might be worth considering from that list is `getTransactionWithReceipts`, as it is used within `TransactionResponse`.

I added it.

danielbate: > Discrepancies usually occur when a read operation is performed right after a write one.

So we should make all read operations consider block height? I think height insensitive operations makes more sense to me.

apps/docs/src/guide/provider/rpc-consistency.md
```diff
@@ -0,0 +1,11 @@
+# RPC Consistency
+
+A common issue with querying distributed networks is ensuring each node is consistent. At any moment each node could be at a different block height, therefore querying the state of the blockchain may yield different results from node to node.
```
arboleya, line 3: Nit.

```suggestion
A common issue with querying distributed networks is ensuring each node is consistent. At any moment, each node could be at a different block height; therefore, querying the state of the blockchain may yield different results from node to node.
```

apps/docs/src/guide/provider/rpc-consistency.md
```diff
@@ -0,0 +1,11 @@
...
+
+To defend against this, the SDK appends the block height to block-sensitive requests. The node will then ensure it has met the block height criteria before it processes the request.
```
arboleya, line 5: Nit.

```suggestion
To defend against this, the SDK appends the block height to block-sensitive requests so the node will verify that it meets the block height criteria before processing the request.
```

apps/docs/src/guide/provider/rpc-consistency.md
```diff
@@ -0,0 +1,11 @@
...
+
+If the block height criteria has not been met by the node, the request will either be awaited on the node's side (if the node supports this), or it'll be retried from the SDK side. The SDK will attempt to retry the request until it hits a maximum amount of retries, and then will throw an error.
```
arboleya, line 7: Nit.

```suggestion
If the node has not met the block height criteria, the request will either be awaited on the node's side (if the node supports this) or retried from the SDK side. The SDK will attempt to retry the request until it hits maximum retries and then throw an error.
```

packages/account/src/providers/fuel-graphql-subscriber.ts
```diff
@@ -41,24 +53,45 @@ export class FuelGraphqlSubscriber implements AsyncIterator<unknown> {
      */
     await new FuelGraphqlSubscriber(errorReader).next();
 
-    return new FuelGraphqlSubscriber(resultReader);
+    return new FuelGraphqlSubscriber(resultReader, onEvent);
   }
 
-  private events: Array<{ data: unknown; errors?: { message: string }[] }> = [];
-  private parsingLeftover = '';
+  public static async readEvent(
```
arboleya, line 59: This method is not very obvious, some basic comments could help understanding it.

packages/account/src/providers/fuel-graphql-subscriber.ts
```diff
@@ -73,27 +106,39 @@ export class FuelGraphqlSubscriber implements AsyncIterator<unknown> {
...
-        try {
-          this.events.push(JSON.parse(match.replace(/^data:/, '')));
-        } catch (e) {
-          throw new FuelError(
-            ErrorCode.STREAM_PARSING_ERROR,
-            `Error while parsing stream data response: ${text}`
-          );
-        }
-      });
+  async next(): Promise<IteratorResult<unknown, unknown>> {
```
arboleya, line 117: Same here - some comments could alleviate the lack of context.

petertonysmith94 (commented):
Tested this locally and functionally seems really solid.

packages/account/src/providers/provider.ts
danielbate: Okay following on from yesterday's conversations, I am now happy with this sentiment.

Although any get request could be fetching out of date information, this mechanism has been designed to defend against scenarios where a block is written to and follow-up request fetches out of date information.

Hence we have included `statusChange` and `getTransactionWithReceipts` as the SDK performs them as follow-up in building a response.

Why in that case, do we include `submit`, `submitAndAwaitStatus` and `getCoinsToSpend`?

packages/account/test/block-sensitive-operations.test.ts
```diff
@@ -0,0 +1,240 @@
...
+  await sleep(550);
+
+  const { encodedTransaction, transactionId } = await createTransfer(wallet);
+
+  const {
+    chain: {
+      latestBlock: { height: latestBlockHeight },
+    },
+  } = await provider.operations.getLatestBlockHeight();
+
```
Torres-ssf, line 49:
```suggestion
  // allow for block production
  await sleep(550);

  const { encodedTransaction, transactionId } = await createTransfer(wallet);

  const {
    chain: {
      latestBlock: { height: latestBlockHeight },
    },
  } = await provider.operations.getLatestBlockHeight();

  expect(latestBlockHeight).toBeGreaterThan(0);
```

packages/account/test/block-sensitive-operations.test.ts
```diff
@@ -0,0 +1,240 @@
...
+    using launched = await setupTest();
+
+    const { provider, encodedTransaction, latestBlockHeight } = launched;
+
+    const fetchSpy = vi.spyOn(global, 'fetch');
+
+    await provider.operations.submit({ encodedTransaction });
+
+    expectBlockHeight({ included: true, fetchSpy, latestBlockHeight });
+  });
```
Torres-ssf, line 96: Do you think we should add test cases here with Providers that were instantiated with custom fetch fn?

packages/account/test/optimistic-concurrency-handling.test.ts
```diff
@@ -0,0 +1,440 @@
...
+      fetchSpy.mockReset();
+
+      const body = JSON.parse(request?.body?.toString() as string);
+      body.extensions.required_fuel_block_height = expectedHeight;
+
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      request!.body = JSON.stringify(body);
+
+      return fetch(url, request);
+    });
```
Torres-ssf, line 77: For this test suite, would it be better, instead of mocking the returned value by the request, to ensure that the GraphQL request genuinely took longer than usual because it was waiting for the block to sync?

Maybe we could use:
```
--poa-interval-period 200ms
--graphql-required-block-height-tolerance 5
```

to simplify the validation.

packages/account/src/providers/provider.ts
nedsalk: > Why in that case, do we include submit, submitAndAwaitStatus and getCoinsToSpend?

Because they are block-sensitive as well. You might have built a transaction with an UTXO that was created in block N. If you submit it to a node that's on block N-1, it'll fail. The same goes for `getCoinsToSpend` - it goes over UTXOs to give you the appropriate coins.

In the end, as agreed yesterday, I'll be adding the expected block height to every request, but only _after_ the first `mutation` is done.

packages/account/src/providers/provider.ts
```diff
@@ -521,10 +553,93 @@ export default class Provider {
...
+  ): Promise<Response> {
+    const fetchFn = () =>
+      options.fetch ? options.fetch(url, request, options) : fetch(url, request);
+
+    let response: Response = await fetchFn();
+
+    const retryOptions: RetryOptions = {
+      maxRetries: 5,
+      baseDelay: 500,
+    };
```
Torres-ssf, line 594: So am I correct in assuming these are fixed default values that the user cannot modify?

apps/docs/src/guide/provider/rpc-consistency.md
Torres-ssf, line 7: Should we include information here about the values used by the TS SDK for retry attempts?

```ts
maxRetries: 5,
baseDelay: 500,
```

Also, it would be helpful to mention that `Provider.clearChainAndNodeCaches` resets the `currentBlockHeightCache` for all provider instances associated with the given URL or for all URLs when a specific URL is not informed.

On a related note, we should also clarify that all Provider instances created with the same URL share the same `currentBlockHeightCache`. Including code snippets with multiple provider instances could help make this behavior easier to understand.

Torres-ssf, line 7: Should we include information here about the values used by the TS SDK for retry attempts?

```ts
maxRetries: 5,
baseDelay: 500,
```

Also, it would be helpful to mention that `Provider.clearChainAndNodeCaches` resets the `currentBlockHeightCache` for all provider instances associated with the given URL or for all URLs when a specific URL is not informed.

On a related note, we should also clarify that all Provider instances created with the same URL share the same `currentBlockHeightCache`. Including code snippets with multiple provider instances could help make this behavior easier to understand.

Torres-ssf: Duplicated

packages/account/src/providers/provider.ts
nedsalk: Yes, you're right.

packages/account/test/optimistic-concurrency-handling.test.ts
nedsalk: That's exactly what the mock is doing. I have reset the implementation on line `68`  so that the _actual_ fetch implementation is called and the node is hit.

apps/docs/src/guide/provider/rpc-consistency.md
nedsalk: > Should we include information here about the values used by the TS SDK for retry attempts?

Yes, it might be useful for users to make a more informed decision on if they want the flag enabled or not.

> Also, it would be helpful to mention that `Provider.clearChainAndNodeCaches` resets the `currentBlockHeightCache` for all provider instances associated with the given URL or for all URLs when a specific URL is not informed.

IMO it's best to document that on the `clearChainAndNodeCaches` method itself.

> On a related note, we should also clarify that all Provider instances created with the same URL share the same `currentBlockHeightCache`. Including code snippets with multiple provider instances could help make this behavior easier to understand.

I didn't find any documentation generally around the `Provider` having a static instance of the cache that is accessed by all `Provider` instances. This should be documented not only for `currentBlockHeightCache` but also for the other caches. I think this would be better done in a subsequent PR that's focused on documenting it all, `currentBlockHeightCache` included.

packages/account/test/block-sensitive-operations.test.ts
nedsalk: The existing tests still cover the custom fetch fn functionality and would fail if [this line](https://github.com/FuelLabs/fuels-ts/blob/ns/feat/optimistic-concurrency-control/packages/account/src/providers/provider.ts#L586) ignored the `options.fetch` usage, so IMO there's no need to write tests around this with regards to RPC consistency.

packages/account/src/providers/provider.ts
```diff
@@ -463,10 +471,22 @@ export default class Provider {
   cache?: ResourceCache;
 
   /** @hidden */
```
nedsalk, line 473: This tag isn't shown on the method when hovering over it. Its functionality might have been overridden by the doc comments below.

petertonysmith94: 58635663f958d0387a37a4f110e5b217dc04eda0

packages/account/src/providers/fuel-graphql-subscriber.ts
petertonysmith94: 6f2c962698b3679a281ccacb8430eeb6752c5248

packages/account/src/providers/fuel-graphql-subscriber.ts
petertonysmith94: 6de928e416c0f1fd70c182cd590e1fc5a851ac62

packages/account/src/providers/provider.ts
arboleya: As discussed in the last sync, we can make all methods sensitive by default, but only after a written action happens during the program's lifecycle. Before that, we send no block height requirement.

Torres-ssf (approved):

.changeset/green-radios-taste.md
```diff
@@ -0,0 +1,6 @@
+---
+"@fuel-ts/account": patch
+"@fuel-ts/errors": patch
+---
+
+chore: support RPC consistency mechanism
```
petertonysmith94, line 6: > [!NOTE]
> This still requires change - we will only apply the required block height once a write operation has been performed.

packages/fuels/test/features/deploy.test.ts
Torres-ssf: @arboleya Yes, I vaguely remember this. You were the one who identified it while we were investigating an issue during a meeting.

If I recall correctly, the problem had to do with the same function/helper ( a test helper, if I am not mistaken ) being exported from multiple locations.

If I recall correct, this was the change that we made to fix [this](https://github.com/FuelLabs/fuels-ts/commit/45f4a86097ce249d79769451fc8e8ec0b5192717#diff-fdd9aa19e64f6e22d00f00a7610422e7b4cdd000952e6156c07a6e6f6477a229R106-R108)

arboleya: @Torres-ssf Fantastic memory! It was during the launch. 😅

@petertonysmith94 Could and should we apply the same here? 🤔 

Overall, this is a serious problem that should be addressed by design, _if feasible_. I would need to explore it further to gain a better understanding. That said, I think we can proceed with these temporary fixes as they are as well, but perhaps we could include some comments (unlike our past selves did [here](https://github.com/FuelLabs/fuels-ts/commit/45f4a86097ce249d79769451fc8e8ec0b5192717#diff-fdd9aa19e64f6e22d00f00a7610422e7b4cdd000952e6156c07a6e6f6477a229R106-R108)) and begin to leave detailed notes for these workarounds and the reasons why they are necessary.

petertonysmith94: I will give you my direction tomorrow if that is alright @arboleya?

arboleya: Of course! 😅

.changeset/green-radios-taste.md
arboleya: Was this done? cc @Torres-ssf

Torres-ssf: @arboleya Yes, this was addressed

arboleya (approved):

apps/docs/src/guide/provider/rpc-consistency.md
```diff
@@ -0,0 +1,11 @@
+# RPC Consistency
+
+A common issue with querying distributed networks is ensuring each node is consistent. At any moment, each node could be at a different block height; therefore, querying the state of the blockchain may yield different results from node to node.
```
arboleya, line 3:
```suggestion
A common issue with querying distributed networks is ensuring consistency between nodes. At any moment, each node could be at a different block height; therefore, querying the state of the blockchain may yield different results from node to node.
```

apps/docs/src/guide/provider/rpc-consistency.md
```diff
@@ -0,0 +1,11 @@
...
+
+To defend against this, the SDK appends the block height to block-sensitive requests so the node will verify that it meets the block height criteria before processing the request.
+
+If the node has not met the block height criteria, the request will either be awaited on the node's side (if the node supports this) or retried from the SDK side. The SDK will attempt to retry the request until it hits maximum retries and then throw an error.
```
arboleya, line 7:
```suggestion
If the node has not met the block height criteria, the request will either be awaited on the node's side (if the node supports this) or retried from the SDK side. The SDK will attempt to retry the request until it reaches the maximum number of retries and then throw an error.
```

apps/docs/src/guide/provider/rpc-consistency.md
```diff
@@ -0,0 +1,11 @@
...
+
+This functionality is enabled by default, but can be disabled as so:
```
arboleya, line 9:
```suggestion
This functionality is enabled by default but can be disabled as so:
```

arboleya (approved):
Optional doc nits.

arboleya (approved):

petertonysmith94 (approved):

packages/account/test/optimistic-concurrency-handling.test.ts
```diff
@@ -0,0 +1,404 @@
...
+    await provider.operations.getCoinsToSpend({
+      owner: wallet.address.toB256(),
+      queryPerAsset: { amount: '10', assetId: baseAssetId },
+    });
+
+    const [getCoinsToSpend] = getFetchOperationsByName(fetch, 'getCoinsToSpend');
+    expect(getCoinsToSpend.extensions.required_fuel_block_height).toEqual(expectedBlockHeight);
+  });
+
+  it('should update block height when new subscription events are received', async () => {
```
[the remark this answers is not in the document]
petertonysmith94: Not that I'm aware of, just thought it might be worthwhile having more tests around subscriptions for this functionality.

arboleya (approved):

Torres-ssf (approved):

Issue #3736, opened 2025-02-24
Connector Aware Estimation


danielbate:
## TLDR

Connector Aware Estimation is the process of giving the SDK more awareness of the account level semantics earlier in the transaction flow to reduce round trips to the network.

## Problem

Let's imagine the following transaction flow for non-native wallets:

<img width="789" alt="Image" src="https://github.com/user-attachments/assets/eba670da-abb8-41a0-8362-4bae715688f8" />

In the first stage, the SDK estimates the transaction as normal.

Once it hands off to the connector, and we introduce the non-native account, we must add the predicate to the transaction. Therefore meaning the transaction must be re-prepared.

This is obvious redundancy that could be mitigated by giving the SDK the context earlier in the flow:

## Potential Solution

1. Introduce a `prepareTransaction` function to the connector interface, that can alter a transaction request with any of it's own specifics, such as adding the non-native predicate.
```ts
  /**
   * Modifies a transaction request with anything connector specific.
   * 
   * @param transaction - The tx to prepare.
   *
   * @returns The prepared tx request
   */
  async prepareTransaction(transaction: TransactionRequestLike): Promise<TransactionRequest>;
```
2. Introduce `connector.prepareTransaction()` at the estimation and funding SDK entry points.
3. Remove connector level estimation and funding from within the `connector.sendTransaction()`  calls.

## Notes
- Must be additive, allowing for gradual upgrade by the connectors
- Must work with conjunction with `prepareForSend` - #3665

danielbate:
I have not done a proof of concept here yet, so this may evolve.

danielbate:
@petertonysmith94 I think I was overdoing it in our previous discussion, something simple like this would work by moving the connector level logic from submission -> preparation.

But have a look and see what you think, I will pick up any progress made next week :)

danielbate:
Blocked until #3669 and #3737 are complete.

danielbate:
Okay, I'm now blocked on this until `0.42.0` is available on testnet.

The [WIP PR](https://github.com/FuelLabs/fuels-ts/pull/3769) is looking pretty solid and includes an e2e test using a associated predicate account that requires an external signer (very similar to the EVM/SVM predicates).

[...]

Issue #3665, opened 2025-02-05
Inverted Submission Flow


danielbate:
## Problem

When a transaction is submitted from a connector and approved in a user wallet, it takes **3** requests to submit a tx and build it's summary inside the application. Versus it only taking **1** request when submitting a tx from the SDK when we have the private key readily accessible.

The main reason for this is that as the wallet submits the tx, the application loses context of both the submission and finalised transaction, and therefore it must fetch both the raw tx and tx status to build it's summary.

This issue aims to bring parity to these 2 flows and enable single request submission and summary inside apps using connectors, as visualised below:

<img width="853" alt="Image" src="https://github.com/user-attachments/assets/2f73d57e-db64-4cfd-8ddb-30cea99f0052" />

## Solution

To enable the above flow, the app must take ownership of tx submission. With the wallet passing back the finalised tx, propagated via the connectors.

This would involve the following work:

### SDK

1. Introduce a transaction preparation function to be utilised by connectors, via the [connector interface](https://github.com/FuelLabs/fuels-ts/blob/master/packages/account/src/connectors/fuel-connector.ts). The expected signature is as follows:
```ts
  /**
   * Should perform all necessary operations (i.e estimation,
   * funding, signing) to prepare a tx so it can be submitted
   * at the app level.
   *
   * @param address - The address to sign the tx
   * @param transaction - The tx to prepare
   *
   * @returns The prepared tx request
   */
  async prepareForSend(address: string, transaction: TransactionRequestLike): Promise<TransactionRequest>;
```
2. Utilise the preparation method inside `Account.sendTransaction` so that the process is still obfuscated to consumers.

https://github.com/FuelLabs/fuels-ts/blob/01375ce29bf5f18b5d8f5e97fec5d82f4fb6ae5f/packages/account/src/account.ts#L653-L659

### Non Native Wallets

For EVM and SVM wallets, all logic is inside the connector, so we need to do only the following

1. Introduce the new preparation method inside the relevant connector, identical to `sendTransaction` however passing back the prepared tx rather than submitting

### Native Wallets

Wallets such as Fuel Wallet and Bako, more of the logic is on the wallet side so would involve the following:

1. Introduce the preparation method inside the wallet itself, with a new event to trigger the flow
2. Introduce the method inside the connector, that submits the event

danielbate:
[...]

PR #3754, opened 2025-03-06
chore: revert inverted submission

danielbate:
- Part of #3665 

# Summary

Reverts #3689 so we can remove the `usePrepareForSend` flag in isolation. As the approach is also going to be similar to #3736, these should be done in conjunction.

# Checklist


.changeset/yummy-dragons-begin.md
```diff
@@ -0,0 +1,4 @@
+---
+---
+
+chore: revert inverted submission flow
```
danielbate, line 4: Conversation to ensure we update release notes.

arboleya: We can do that manually on the changeset PR _and_ GitHub release.

arboleya (approved):

Torres-ssf (approved):

.changeset/yummy-dragons-begin.md
arboleya: Next time, adding an empty changeset while removing the old one is better.

This way, changelogs wouldn't need to be manually edited afterward.

arboleya: @danielbate Actually, never mind.

I think we're good because this PR doesn't have a release note.

Please let me know if I misunderstood your point.

PR #3759, opened 2025-03-06
fix: correct revert error name

### LINKED ISSUE #3758: Fix Missing `OutputVariable` Name Error
When a Sway script attempts to execute a transfer, but the transaction lacks the corresponding `OutputVariable`, Sway reverts with the reason [FAILED_TRANSFER_TO_ADDRESS_SIGNAL](https://github.com/FuelLabs/sway/blob/8a3d8df198b5ecd7b66fdee6c1003e0c8ffb85a6/sway-lib-std/src/error_signals.sw#L16C11-L16C44).

However, in the TypeScript SDK, we [incorrectly indicate](https://github.com/FuelLabs/fuels-ts/blob/07f0177df95c4099f7819985350be173f80cdfe9/packages/account/src/providers/utils/extract-tx-error.ts#L98-L101) that the transaction is missing an `OutputChange` instead.

Torres-ssf:
- Closes #3758

# Checklist


nedsalk (approved):

packages/fuel-gauge/src/revert-error.test.ts
```diff
@@ -187,7 +187,7 @@ describe('Revert Error Testing', () => {
     );
   });
 
-  it('should throw for a missing OutputChange', async () => {
+  it('should throw for a missing OutputVariable', async () => {
```
danielbate, line 190: Should we extend this test to add the Output Variable and then assert the tx succeeds? As maybe we could have caught this prior.

danielbate (approved):

arboleya (approved):

packages/fuel-gauge/src/revert-error.test.ts
Torres-ssf: I don't believe this is needed because we have many tests covering this functionality. In fact, identifying missing `OutputVariable` based on the [FAILED_TRANSFER_TO_ADDRESS_SIGNAL](https://github.com/FuelLabs/fuels-ts/blob/4aa4ddb8bcc22b6142e04004ed81fe0ebcaad2eb/packages/account/src/providers/utils/receipts.ts#L28-L32) reason is one of the key behaviors of [Provider.estimateTxDependencies](https://github.com/FuelLabs/fuels-ts/blob/4aa4ddb8bcc22b6142e04004ed81fe0ebcaad2eb/packages/account/src/providers/provider.ts#L1106-L1107).

For example, in this [test suite](https://github.com/FuelLabs/fuels-ts/blob/4aa4ddb8bcc22b6142e04004ed81fe0ebcaad2eb/packages/fuel-gauge/src/dry-run-multiple-txs.test.ts#L23), we have multiple cases validating that `estimateTxDependencies` correctly detects this error and adds the missing `OutputVariable`, resulting in the TX succeeding.

The issue occurred when we were constructing an error message for the user. I mistakenly used `MissingOutputChange` instead of `MissingOutputVariable` while implementing this.

danielbate: Okay makes sense, as long as we have tests correctly rectifying 👍🏻

danielbate (approved):

Issue #3750, opened 2025-03-04
Stop cloning transaction request instances


arboleya:
Today, we often return a cloned `transactionRequest` (check below), which is not very flexible and can cause users to duplicate other code chunks if they want to customize and mutate their requests before submitting them.

https://github.com/FuelLabs/fuels-ts/blob/86b8e94c32b0f8329f1bca35551fe783e5206e05/packages/program/src/functions/base-invocation-scope.ts#L246

> cc @luizstacio

PR #3763, opened 2025-03-10
feat!: remove `BaseInvocationScope.getTransactionId()`

### LINKED ISSUE #3538: Remove `BaseInvocationScope.getTransactionId`
We have a [getTransactionId](https://github.com/FuelLabs/fuels-ts/blob/0fb6fa8e7d964e9c56eaa5e5554fde660f438bf4/packages/program/src/functions/base-invocation-scope.ts#L475) method on the `BaseInvocationScope`. It is very unlikely that the transaction ID that this method outputs is the same as the request that gets submitted as `call` will fund the request before [submission](https://github.com/FuelLabs/fuels-ts/blob/0fb6fa8e7d964e9c56eaa5e5554fde660f438bf4/packages/program/src/functions/base-invocation-scope.ts#L383), potentially causing unnecessary confusion.

danielbate:
- Closes #3538

# Release notes

In this release, we:

- Removed `BaseInvocationScope.getTransactionId()`

# Summary

Removed `getTransactionId()` from the `BaseInvocationScope` as this value is unlikely to be the actual submitted tx ID, as a contract call is funded on `.call()`.

# Breaking Changes

- `getTransactionId()` is no longer available on the `BaseInvocationScope`.

# Checklist


nedsalk (approved):

petertonysmith94 (dismissed):

apps/docs/src/guide/scripts/snippets/script-with-configurable.ts
```diff
@@ -39,9 +39,6 @@ tx.callParams({ gasLimit: 7500 });
 // Get the entire transaction request prior to
 const txRequest = await tx.getTransactionRequest();
 
-// Get the transaction ID
-const txId = await tx.getTransactionId();
```
arboleya, line 43: What is the alternative for this now?

Should we add a before/after example snippet to the breaking changes?

danielbate: The alternative is to pick it off the transaction request, but this is prone to the same issue of it being different on `call` so I'm hesitant to recommend it. For a contract call it should only really be picked off the submitted transaction. 

Thoughts on this for breaking change snippet:

```ts
// before
const transactionId = await await contract.functions.something().getTransactionRequest().getTransactionId();
```
```ts
// after
const { transactionId } = await contract.functions.something().call();
```

Torres-ssf: @danielbate I believe there's no way around this except to clarify on the docs that, in the context of contract/script calls (`BaseInvocationScope`), the transaction is estimated and funded right before submission and right after `.call()` is invoked.

While users can still retrieve the transaction ID, it's unlikely to be the final one, as the transaction will most likely be modified after calling `.call()`.

danielbate: Agreed @Torres-ssf, hence I think we should remove it from that class as it's unlikely to be correct.

What are you saying is the action to take here? More documentation? And no breaking change before/after?

nedsalk: I doubt that anybody was using this method because of this problem, so I'm in favor of removing it. 

We already have multiple other places in our docs that use the `ScriptTransactionRequest.getTransactionId` method (e.g. [signing](https://github.com/FuelLabs/fuels-ts/blob/817bf1389dfb942f4f7b0d809c198d6b2b6080c1/apps/docs/src/guide/wallets/snippets/signing/sign-transaction.ts#L25)), so the user has a way of knowing how to get transaction id. 

In the snippet changed in this PR, the `txId` wasn't used and, on top of that, the value isn't the same as the one they user would get from the `TransactionResult`; so even if we were to opt for showing this detail via `txRequest.getTransactionId`, it'd still be confusing to the user.

As far as the snippet goes, I'd update it to this without the `txId` part:
```ts
// You can get the underlying ScriptTransactionRequest for further modification
const txRequest = await tx.getTransactionRequest();
```

arboleya: @danielbate We must consider that people may want to get the TX ID before submitting it.

@Torres-ssf I'm curious to understand [your](https://github.com/FuelLabs/fuels-ts/pull/3763#discussion_r1987172761) POV as well.

danielbate: Makes sense. I still think we should remove this function from the base invocation scope as it is prone to error.

However, we could also improve our `TransactionRequest.getTransactionId()` [documentation](https://docs.fuel.network/docs/fuels-ts/transactions/modifying-the-request/#getting-the-transaction-id-for-a-transaction-request), to include an example that specifically references a contract call, and the dangers of doing so before the call has been funded.

arboleya: Yes, the problem with the snippet you linked is that it doesn't show how to get the `TransactionRequest`.

Could you pls slightly adjust it as part of this PR? I think something similar could be done with the before/after snippet for the breaking changes here as well.

Torres-ssf: In the context of `BaseInvocationScope,` there is no safe way to get the transaction ID before calling `.call`.

And calling `.call` will result in submitting the transaction.

The only way to get the TX ID would be manually submitting the TX.

```ts
const account = contract.account

const request = await contract.function.xyz().fundWithRequiredCoins()

const transactionId = request.getTransactionId(chainId)

account.sendTransaction(request)
```

However, with this approach, users will lose reference to the scope invocation, meaning that if the contract returns a `value` that they expect, this value will not be returned anymore.

arboleya: The plot thickens.

nedsalk: > However, with this approach, users will lose reference to the scope invocation, meaning that if the contract returns a value that they expect, this value will not be decoded anymore as we depend on the scope invocation to properly decode values returned by contracts.

If `fundWithRequiredCoins` was returning a `BaseInvocationScope`, then we could remain with the `BaseInvocationScope.getTransactionId` method and document this gotcha instead:

```ts
const fn = contract.functions.xyz();
console.log(fn.getTransactionId()); // one value logged

const fundedFn = await fn.fundWithRequiredCoins();
console.log(fundedFn.getTransactionId()); // another value logged which is the correct tx's id

const request = fundedFn.getTransactionRequest();

provider.sendTransaction(request);
```

arboleya: Ok, what about not breaking the API but doing something like:

```ts
// before
const fn = contract.functions.xyz();
const txId = fn.getTransactionId(); // throw error: tx must be funded
```

```ts
// after
const fn = contract.functions.xyz();

await fn.fundWithRequiredCoins(); // fund tx

const txId = fn.getTransactionId(); // all good

const { value } = await fn.call();
```

We should still document it, but it seems a more promising direction.

nedsalk: The problem with the current implementation is that `fn.fundWithRequiredCoins` currently doesn't modify the underlying transaction request:
```ts
test('fails', async () => {
  using contract = await launchTestContract(CoverageContractFactory);
  const fn = contract.functions.echo_u8(1);
  const txRequest = await fn.getTransactionRequest();
  await fn.fundWithRequiredCoins();
  const txRequest2 = await fn.getTransactionRequest();
  expect(txRequest).not.toEqual(txRequest2);
});
```

I'd say this is unexpected behavior as well.

arboleya: Which sends us back to this I guess:
 - https://github.com/FuelLabs/fuels-ts/issues/3750

Torres-ssf: @arboleya there is a problem with the suggested version.

`fundWithRequiredCoins` [will always clean](https://github.com/FuelLabs/fuels-ts/blob/94646e0d57beb0164198f5a0f87f09944a48fe8b/packages/program/src/functions/base-invocation-scope.ts#L252) the input coins before estimating the TX and fund it again.

The `.call` [will always](https://github.com/FuelLabs/fuels-ts/blob/94646e0d57beb0164198f5a0f87f09944a48fe8b/packages/program/src/functions/base-invocation-scope.ts#L374) execute `.fundWithRequiredCoins`.

 
This means that in the suggested approach,  `.fundWithRequiredCoins` will be executed 2 times. This will result in `getTransactionCost` being called 2 times as well, which will result in at least 2 dry-runs.

Also because `.fundWithRequiredCoins` will always remove all input coins, there is a slight chance that, on the second time that we are estimating and funding the transaction, different UTXOs are fetched, resulting in the submitted TX having a different TX ID from the one that it was get before.

With this change [here](https://github.com/FuelLabs/fuels-ts/pull/3773) we can avoid having `.fundWithRequiredCoins` cleaning all input coins.

But we would also need some way to skip the estimation and fund part when calling .`call`

arboleya: Would a simple flagging suffice?

- `fundWithRequiredCoins()` — sets a flag to `true`
- `call()` — checks the above flag and skip funding again

Something alongside these lines.

anandfresh (approved):

apps/docs/src/guide/scripts/snippets/script-with-configurable.ts
arboleya: @Torres-ssf Gentle ping - what do you think of the above suggestion?

Should we still move forward with this PR?

Torres-ssf: @arboleya This might resolve the issue, but it still needs to be validated.

Regarding whether we should move forward with this PR — give me a bit of time to validate everything. I’ll follow up shortly with an answer.

Torres-ssf: @arboleya After considering it, I think it makes more sense to address the suggested approach as part of resolving [this issue](https://github.com/FuelLabs/fuels-ts/issues/3815).

Also, since using `getTransactionId` from `BaseInvocationScope` isn’t safe, we should remove it. We need to update the docs accordingly. I'll take care of that here.

Torres-ssf:
Close in favor of https://github.com/FuelLabs/fuels-ts/pull/3864

PR #3747, opened 2025-03-04
feat: integrate with new `AssembleTx` GQL end-point

Torres-ssf:
- Closes #3737

# Release notes

In this release, we:

- Implemented the new `AssembleTx` GraphQL endpoint for `Provider`

# Summary

Introducing `assembleTx` a more robust and optimal way to estimate and fund transactions.

### Parameters
The `AssembleTxParams` interface includes the following parameters:

```ts
export type AssembleTxParams<T extends TransactionRequest = TransactionRequest> = {
  // The transaction request to assemble
  request: T;
  // Block horizon for gas price estimation (default: 10)
  blockHorizon?: number;
  // Coin quantities required for the transaction
  accountCoinQuantities: AccountCoinQuantity[];
  // Account that will pay for the transaction fees
  feePayerAccount: Account;
  // Resources to exclude from the transaction
  excludeInput?: ExcludeResourcesOption;
  // Whether to estimate predicates (default: true)
  estimatePredicates?: boolean;
  // Amount of gas to reserve (optional)
  reserveGas?: number;
};
```

### Return Value

The method returns an object containing:

- `assembledRequest`: The fully assembled transaction request with all necessary inputs, outputs, and policies
- `gasPrice`: The estimated gas price for the transaction
- `receipts`: The transaction receipts from the dry run
### Usage Example

```ts
const request = new ScriptTransactionRequest();

request.addCoinOutput(accountB.address, transferAmount, baseAssetId);

const accountCoinQuantities: AccountCoinQuantity[] = [
  {
    amount: transferAmount,
    assetId: baseAssetId, // Asset ID
    account: accountA,
    changeOutputAccount: accountA, // Optional
  },
];

// Assemble the transaction
const { assembledRequest, gasPrice, receipts } = await provider.assembleTx({
  request,
  accountCoinQuantities,
  feePayerAccount: accountA,
  blockHorizon: 10,
});

// The assembledRequest is now ready to be signed and sent
const submit = await accountA.sendTransaction(assembledRequest);
await submit.waitForResult();
```

## Best Practices

1. Always provide the correct `feePayerAccount` that has sufficient funds for the transaction fees
2. Pay special attention to change outputs when dealing with multiple accounts and the same asset ID:
   - In Fuel, only one change output is allowed per asset ID
   - If a transaction includes resources from the same asset ID for multiple accounts, only one account will receive the change from all spent resources
   - Make sure to coordinate with all parties involved to determine which account should receive the change output

## Notes

[...]

PR #3764, opened 2025-03-10
fix: update txId for contract input in tx summary

### LINKED ISSUE #3708: Transaction ID not correctly set on Contract input

danielbate:
- Closes #3708

# Release notes

In this release, we:

- Fixed some scenarios where a transaction summary contract input contained a zeroed transaction ID

# Summary

In a scenario where a transaction summary is built from a transaction request, we do not update the transaction ID for the input, meaning the summary is not valid as opposed to if the transaction was fetched from the node.

This PR updates that value when the summary is built from a passed transaction request object.

# Checklist

### DIFF
--- a/packages/account/src/providers/transaction-response/transaction-response.ts
+++ b/packages/account/src/providers/transaction-response/transaction-response.ts
@@ -24,7 +24,7 @@ import type {
   Output,
   TransactionType,
 } from '@fuel-ts/transactions';
-import { OutputType, TransactionCoder, TxPointerCoder } from '@fuel-ts/transactions';
+import { InputType, OutputType, TransactionCoder, TxPointerCoder } from '@fuel-ts/transactions';
 import { arrayify, assertUnreachable } from '@fuel-ts/utils';
 
 import type {
@@ -33,7 +33,12 @@ import type {
   GqlSubmitAndAwaitStatusSubscription,
 } from '../__generated__/operations';
 import type Provider from '../provider';
-import type { JsonAbisFromAllCalls, TransactionRequest } from '../transaction-request';
+import type {
+  ContractTransactionRequestInput,
+  JsonAbisFromAllCalls,
+  TransactionRequest,
+  TransactionRequestInput,
+} from '../transaction-request';
 import { assembleTransactionSummary } from '../transaction-summary/assemble-transaction-summary';
 import { processGqlReceipt } from '../transaction-summary/receipt';
 import { getTotalFeeFromStatus } from '../transaction-summary/status';
@@ -41,6 +46,7 @@ import type { TransactionSummary, GqlTransaction, AbiMap } from '../transaction-
 import { extractTxError } from '../utils';
 
 import { getDecodedLogs } from './getDecodedLogs';
+import { getInputsContract } from '../transaction-summary';
 
 /** @hidden */
 export type TransactionResultCallReceipt = ReceiptCall;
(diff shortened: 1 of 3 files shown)


packages/account/src/providers/transaction-response/transaction-response.ts
```diff
@@ -212,13 +212,34 @@ export class TransactionResponse {
...
     bytes: Uint8Array;
   }> {
     if (this.request) {
       const tx = this.request.toTransaction() as Transaction<TTransactionType>;
       this.applyMalleableSubscriptionFields(tx);
+      this.updateTxIds(tx);
+
       return {
         tx,
         bytes: this.request.toTransactionBytes(),
```
petertonysmith94, line 245:
```suggestion
        bytes: new TransactionCoder().encode(tx),
```

If we're mutating the transaction, surely the bytes from the request are incorrect?

packages/account/src/providers/transaction-response/transaction-response.ts
```diff
@@ -212,13 +212,34 @@ export class TransactionResponse {
...
+      TransactionType.Script | TransactionType.Create | TransactionType.Blob
+    >;
+
+    tx.inputs.forEach((input) => {
+      if (input.type === InputType.Contract) {
+        // eslint-disable-next-line no-param-reassign
+        input.txID = this.id;
+      }
+    });
+  }
```
Torres-ssf, line 232: Is this correct? It appears we are setting the transaction ID for the contract input to be the same as the script transaction that interacts with the contract.

Shouldn't it instead be the ID of the transaction that created/deployed the contract?

danielbate: So the spec [just says](https://docs.fuel.network/docs/specs/tx-format/input/#inputcontract) "hash of the transaction".

Also I have enabled some tests that validate both approaches. I.e. testing this approach vs populating the contract input TX id from the chain, and it gives the same value. So I am assuming this is correct.

Not sure how we could expect a user to know the tx ID that a contract was created in, for every contract input 🤔

Torres-ssf: But this seems to be the case. The following test case can validate this.

https://gist.github.com/Torres-ssf/6caff181d727d63264f31df55a2fccdb

Torres-ssf: According to the client team, this value corresponds to the last transaction that modified the contract state. The only reliable way to obtain it is by fetching the processed transaction.

We can use [assembleTx](https://github.com/FuelLabs/fuels-ts/pull/3747/commits/f460fe6a211673c4c3f359a12b2cc2d4b7f7a9fc) to update it accordingly.

danielbate: Ah okay well that defeats the point of this PR, nice find. I'll close this PR on the basis that we should update the input instead via `assembleTx`.

danielbate:
Closing based on [this comment](https://github.com/FuelLabs/fuels-ts/issues/3708#issuecomment-2715276508).

PR #3768, opened 2025-03-12
fix: `forc` and `fuel-core` installation after working on `git:` branch

nedsalk:
# Summary

When working with `git:` branches of `forc` and `fuel-core`, the e.g. `fuel-core --version` call on the built binary would return the latest version which would usually match the one our `master` branch currently supports via its `VERSION` file. Due to that, when switching from a branch that's working with `git:` binaries to a branch that's working with the released version, running `pnpm install` wouldn't install the released binary as the versions would match and the script would skip installation. This PR fixes it so that the script doesn't skip installation in this scenario.

# Checklist


internal/forc/lib/install.js
```diff
@@ -28,19 +28,29 @@ import {
   const binDir = join(__dirname, '../');
 
   const binPath = join(binDir, 'forc-binaries', 'forc');
+  const madeFromGitPath = join(binDir, 'MADE-FROM-GIT');
```
arboleya, line 31: Can you check the `VERSION` files directly?

internal/forc/lib/install.js
```diff
@@ -69,5 +79,6 @@ import {
 
     // Cleanup
     rmSync(pkgPath);
+    rmSync(madeFromGitPath, { force: true });
```
arboleya, line 82: You should keep the built repo in place so you can re-use it when you switch back to it.

internal/forc/lib/install.js
nedsalk: I've changed the implementation to copy the `VERSION` file into the bin directory after installing / running `cargo build --release`. There was no need for a new `MADE-FROM-GIT` file.

arboleya: My point remains: Why not read the `VERSION` files directly without copying them?

nedsalk: We are reading the `VERSION` files already. The problem is that we're comparing them against `fuel-core --version` which would return a release version (e.g `0.41.7`) even when it was built from a `git:` branch. Because of this, when we checkout an SDK branch that's working with a released version but our binary was built from a `git:` branch beforehand, the install script _wouldn't_ install the released version because `VERSION` == `fuel-core --version` even though the binary isn't the same as the released one.

What I did in this PR is always put the `VERSION` file alongside the binary and check against that instead of `fuel-core --version`. This way, if our binary was built from `git:master`, that `binDir/VERSION` file would contain `git:master`, and when we checkout a branch that has the `VERSION` file set to `0.41.7`, the script will compare the contents of these two files, see that they're different, and install the correct one.

internal/forc/lib/install.js
nedsalk: The `madeFromGitPath` naming might have been confusing. I refactored this. All in all, the binaries built by `cargo build --release` are untouched and will be reused in the `buildFromGitBranch` flow.

arboleya (approved):

petertonysmith94 (approved):

PR #3769, opened 2025-03-12
feat: front-end optimisation strategies

### LINKED ISSUE #3665: Inverted Submission Flow

### LINKED ISSUE #3736: Connector Aware Estimation

danielbate:
- Closes #3736
- Closes #3665 

# Release notes

In this release, we:

- Added the `onBeforeAssembleTx` hook to wallet connectors for quicker estimation
- Enabled a snappier transaction summary generation flow for connectors

# Summary

This PR introduces two approaches to enable more optimal transaction speeds from estimation to summary generation:

- [Connector Aware Estimation](https://github.com/FuelLabs/fuels-ts/issues/3736)
- [Inverted Submission Flow](https://github.com/FuelLabs/fuels-ts/issues/3665)

### Connector Aware Estimation

Adds the `onBeforeAssembleTx` function to the connector interface. This allows for connector specific modifications to a transaction request to be added before the SDK assembles the transaction. A specific use case for this is the SVM/EVM connectors that fund a transaction using an associated predicate account.

### Inverted Submission Flow

Utilises the change made to the `connector.sendTransaction()` signature in #3669 whereby we can now return a `string | TransactionResponse`.

https://github.com/FuelLabs/fuels-ts/blob/e8494fa488848ba215e339e2dd12364ab5b3a1a1/packages/account/src/connectors/fuel-connector.ts#L47-L51

This approach enables #3665 to allow connector transactions to submit and generate a transaction summary in apps in a single request.

## Proof Of Concept

A demo using the above approaches can be seen in the connectors repo [here](https://github.com/FuelLabs/fuel-connectors/pull/501). The result of which can be seen here:

https://github.com/user-attachments/assets/c5fab006-8c04-4c71-9902-aa0859dc716f

> Running against [Devnet](https://devnet.fuel.network/v1/playground) on `fuel-core@0.42.0`

# Checklist


danielbate:
@FuelLabs/sdk-ts both inverted submission flow and connector aware estimation are all merged here now. I'd created a mock solana connector using the actual SVM predicate. Had some troubles with signing correctly using that test. Was going to do the same for the EVM predicate then this should be ready to go. Anyone is welcome to pick up.

PR #3773, opened 2025-03-13
feat: support customizing TX request from `BaseInvocationScope`

### LINKED ISSUE #3750: Stop cloning transaction request instances

Torres-ssf:
- Closes #3750

# Release notes

In this release, we:

- Support customizing TX request from `BaseInvocationScope`

# Checklist


Torres-ssf:
@luizstacio I believe this change fixes the problem that you raised. Can you confirm that?

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -248,8 +248,6 @@ export class BaseInvocationScope<TReturn = any> {
     const txCost = await this.getTransactionCost();
     const { gasUsed, missingContractIds, outputVariables, maxFee } = txCost;
     this.setDefaultTxParams(transactionRequest, gasUsed, maxFee);
-    // Clean coin inputs before add new coins to the request
-    transactionRequest.inputs = transactionRequest.inputs.filter((i) => i.type !== InputType.Coin);
```
arboleya, line 252: To successfully close #3750, we also need to stop `clone()` in these places, no?

https://github.com/FuelLabs/fuels-ts/blob/f41887f754febbb4469d990e1cdf0d74fc9d9160/packages/program/src/functions/base-invocation-scope.ts#L230

https://github.com/FuelLabs/fuels-ts/blob/f41887f754febbb4469d990e1cdf0d74fc9d9160/packages/program/src/functions/base-invocation-scope.ts#L246

Torres-ssf: I believe cloning the TX request was not the source of the problem. The clone occurs only after `.call` is invoked. By that point, all custom modifications to the TX request have already been applied.

Additionally, there's another cloning step before calling `getTransactionCost`, but this is only to ensure that the request remains unmodified during estimation.

The real issue was that all input coins were being removed during `.call`, causing any manually added UTXOs to be discarded.

arboleya (approved):

danielbate (approved):

PR #3775, opened 2025-03-13
test: add withdraw to base layer test in network suite

Torres-ssf:
# Checklist

### DIFF
--- a/packages/fuel-gauge/src/network.test.ts
+++ b/packages/fuel-gauge/src/network.test.ts
@@ -2,9 +2,11 @@
 
 import type {
   BaseWalletUnlocked,
+  Block,
   ChangeTransactionRequestOutput,
   CoinTransactionRequestOutput,
   ContractTransferParams,
+  ReceiptMessageOut,
   ReceiptTransfer,
 } from 'fuels';
 import {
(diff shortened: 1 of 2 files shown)


packages/fuel-gauge/src/network.test.ts
```diff
@@ -99,6 +102,46 @@ describe('network', () => {
...
+
+    /**
+     * Waiting for the commit block, which means we need to wait at least one block
+     * after the block in which the withdrawal transaction was included.
+     */
+    while (waitForCommit) {
+      commitBlock = (await provider.getBlock('latest' as string)) as Block;
+
+      if (commitBlock.height.lte(txBlock.height)) {
+        await sleep(1000);
```
danielbate, line 126: Is there a risk that we miss the commit block due to the 1 second sleep? And therefore would query the incorrect block?

danielbate, line 126: Is there a risk that we miss the commit block due to the 1 second sleep? And therefore would query the incorrect block?

Torres-ssf: The `provider.getMessageProof` accepts any block later than the one that included the transaction. I verified this by increasing the sleep duration to 5 seconds, allowing more blocks to be created before retrieving the proof.

danielbate (approved):

arboleya (approved):

PR #3776, opened 2025-03-14
fix: ensure is possible to forward 0 amount to contract

### LINKED ISSUE #3777: Ensure Is Possible To Forward `0` Amount To Contract Calls
While Sway/VM accepts it, forwarding a `0` amount to a contract call when using `BaseInvocationScope` is not possible.

This limitation exists because [coinQuantityfy](https://github.com/FuelLabs/fuels-ts/blob/c7850d72fa5ee53011797b65b600864cb28bb715/packages/account/src/providers/coin-quantity.ts#L12) is used and it rounds any value below `1` up to `1`.

However, contract logic can still utilize a `0` amount for specific validations, as all payable Sway functions require a forwarded amount to be provided.

Torres-ssf:
- Closes #3777

# Checklist


packages/account/src/providers/coin-quantity.ts
```diff
@@ -23,10 +23,9 @@ export const coinQuantityfy = (coinQuantityLike: CoinQuantityLike): CoinQuantity
     max = coinQuantityLike.max ?? undefined;
   }
 
-  const bnAmount = bn(amount);
   return {
     assetId: hexlify(assetId),
-    amount: bnAmount.lt(1) ? bn(1) : bnAmount,
+    amount: bn(amount),
```
danielbate, line 28: This check seemed to be a [defence for fractions](https://github.com/FuelLabs/fuels-ts/pull/1603#issuecomment-1881210134), is this case still handled?

Torres-ssf: @danielbate Indeed, but this only applies to fractions less than `1`. With the previous check, we had:

`0.1` → `1`
However, for values greater than 1, the behavior was different:

`1.1` → `1`
`999.9` → `999`

This inconsistency doesn’t make sense the handling of values below 1 shouldn’t differ from those above `1`.

Additionally, from what I understood, the intent behind [this change in the past](https://github.com/FuelLabs/fuels-ts/pull/1603) was to ensure that when `getResourcesToSpend` is used with an amount of `0`, it properly accounts for those values.

To maintain this behavior, I made this change [here](https://github.com/FuelLabs/fuels-ts/blob/be35108c68bb710c454cb5d6efbe8607bd2132b1/packages/account/src/providers/provider.ts#L1580).

We already have a test for this [here](https://github.com/FuelLabs/fuels-ts/blob/6b3753dab33418d1e74880bb0a0d469ec456680c/packages/account/src/account.test.ts#L149)

danielbate (approved):

nedsalk (approved):

arboleya (approved):

PR #3788, opened 2025-03-18
chore: eliminate chain requests across concurrent provider instances

### LINKED ISSUE #3766: Eliminate concurrent provider chain requests
# Summary

Multiple providers can be instantiated simultaneously, which can lead to multiple redundant chain + node info requests. We should ensure that multiple providers only request the chain + node info once.

petertonysmith94:
- Closes #3766

# Release notes

In this release, we:

- Eliminated duplicate chain requests across multiple concurrent provider instances

# Summary

- Multiple instances of a provider could request the chain info simultaneously, causing the cached values to be missing. This ensures that only a single request is made across multiple provider instances.

# Checklist

### DIFF
--- a/packages/account/src/providers/provider.ts
+++ b/packages/account/src/providers/provider.ts
@@ -14,6 +14,8 @@ import type { GraphQLClientResponse, GraphQLResponse } from 'graphql-request/src
 import gql from 'graphql-tag';
 import { clone } from 'ramda';
 
+import { deferPromise } from '../connectors/utils/promises';
+
 import { getSdk as getOperationsSdk } from './__generated__/operations';
 import type {
   GqlReceiptFragment as TransactionReceiptJson,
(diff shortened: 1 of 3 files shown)


packages/account/src/providers/provider.ts
```diff
@@ -627,18 +634,32 @@ export default class Provider {
...
+      const inflightRequest: Promise<number> =
+        Provider.inflightFetchChainAndNodeInfoRequests[this.urlWithoutAuth];
+
+      // When an inflight is request is available, wait for it to complete
+      // Then fetch (which will hit the cached values)
+      if (inflightRequest) {
+        const now = await inflightRequest;
+        this.consensusParametersTimestamp = now;
+        return this.fetchChainAndNodeInfo();
+      }
```
danielbate, line 646: Really nice implementation, but should this have some kind of max retries on it before it attempts it's own fetch? Otherwise all concurrent requests will just infinitely loop awaiting the result the first.

danielbate: Ah makes sense, I'd missed the `await inflightRequest`. I think it's okay as is.

danielbate (approved):

Issue #3808, opened 2025-03-24
Implement auto-consolidation flow


arboleya:
Follow-up after:
- https://github.com/FuelLabs/fuels-ts/issues/3645

The PR must also be based on:
 - https://github.com/FuelLabs/fuels-ts/pull/3717

Tasks:

- [ ] Implement auto-detection to decide when UTXO consolidation is necessary. This must evaluate errors from `assembleTx`, but to distinguish between `insufficient balances` and `too many coins`  we may have to fetch user `balances` for baser asset ID, and if this is non-negotiable, we may just as well fetch balances for base asset with in the same HTTP/GQL call that we submit the `assembleTx` operation.

- [ ] Adding the consolidation transactions to the JSON RPC interface

arboleya:
Related:
- https://github.com/FuelLabs/fuels-ts/issues/3803

PR #3929, opened 2025-07-16
feat: added auto-consolidation of coins

petertonysmith94:
- Related to #3808 

# Release notes

In this release, we:

- Implemented the auto-consolidation flow to detect when the max UTXO limit has been reached.
- Added helper function for consolidating coins (`consolidateCoins`).

# Summary

With our previous implementation for consolidating coins, we encountered some inconsistent behaviour and limitations, namely, with the funding of the transaction. The new method aims to address those limitations.

We now fetch all of the base asset coins (and order them by their amount) for funding purposes. We will consolidate these down into a single UTXO on the initial consolidation transaction. The output from each step in the consolidation chain will be passed to the subsequent request in the _sequential_ transaction chain. 

# Checklist


Torres-ssf (approved):

packages/account/src/consolidate-coins.test.ts
```diff
@@ -752,4 +755,51 @@ describe('consolidate-coins', () => {
...
+
+      const { coins } = await sender.getCoins(baseAssetId);
+      expect(coins.length).toEqual(totalCoins);
+
+      const startConsolidationSpy = vi.spyOn(sender, 'startConsolidation');
+      const sendAmount = (maxInputs + 1) * 1_000;
+      await sender.getResourcesToSpend([{ amount: sendAmount, assetId: baseAssetId }]);
+      expect(startConsolidationSpy).toBeCalledTimes(1);
+    });
+  });
```
LuizAsFight, line 804: should we create an issue for it and put on project utxo consolidation?

packages/account/src/utils/consolidate-coins.ts
```diff
@@ -0,0 +1,241 @@
...
+  const all: Coin[] = [];
+  let coins: Coin[];
+  let hasNextPage = true;
+  let after: string | undefined;
+
+  while (hasNextPage) {
+    ({
+      coins,
+      pageInfo: { hasNextPage },
+    } = await account.getCoins(assetId, { after }));
```
LuizAsFight, line 73: can this be done with `const` instead?

packages/account/src/utils/consolidate-coins.ts
```diff
@@ -0,0 +1,241 @@
...
+
+      currentTx.maxFee = fee;
+      currentTx.gasLimit = bn(1000);
+
+      opts.onTransactionStart?.({
+        tx: currentTx,
+        step,
+        assetId,
+        transactionId: currentTx.getTransactionId(chainId),
+      });
```
LuizAsFight, line 215: was this `onTransactionStart` needed for connectors integration? if yes for what?

LuizAsFight (changes requested):

LuizAsFight:
looking good overall, just noticed some code improvements that could be done

packages/account/src/utils/consolidate-coins.ts
```diff
@@ -0,0 +1,241 @@
...
+      scriptData: '0x',
+    });
+
+    // Add our dust coins as inputs
+    request.addResources(batch);
+
+    return request;
+  });
+
+  const submitAll: SubmitAllCallback = async (opts = {}) => {
```
LuizAsFight, line 171: as the txs are being sent sequentially, why do we need the funding / dust logic? 

we could simplify by doing like this?
1- get all coins
2- sort bigger coins first
3- put 255 coins in the tx
4- put 1 output
5- wait tx success
6- repeat until user as only 1 coin

LuizAsFight (changes requested):

packages/account/src/account.ts
```diff
@@ -1153,59 +1265,6 @@ export class Account extends AbstractAccount implements WithAddress {
...
-    } else if (maxFee.gt(setMaxFee)) {
-      throw new FuelError(
-        ErrorCode.MAX_FEE_TOO_LOW,
-        `Max fee '${setMaxFee}' is lower than the required: '${maxFee}'.`
-      );
-    }
-
-    return request;
-  }
-
```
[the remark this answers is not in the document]
petertonysmith94: These were removed as they are private and no longer used within the class.

packages/program/src/functions/base-invocation-scope.ts
```diff
@@ -290,31 +294,48 @@ export class BaseInvocationScope<TReturn = any> {
...
+        skipAutoConsolidation,
+      });
 
-    return assembledRequest;
+      if (!shouldRetry) {
+        throw error;
+      }
+
+      return await assembleTx();
+    }
```
LuizAsFight, line 330: why we need retry here? is it to try again after consolidation? if that's the case I believe the user should do the action manually after consolidation

petertonysmith94: The retry will only be performed for wallets where we have the private key (i.e., `UnlockedWallet`). For connectors, we will not retry, and the user needs to do the action manually after consolidation.

LuizAsFight (approved):

